PR #521, opened 2021-10-01
add rerun-dependabot service

macrael:
## Summary

Attempt to adopt the rerun dependabot lambda shimona made for us

At first this is deploying just on this review-app but we’ll want to deploy it standalone in one of our environments once its working.


macrael:
I'm closing this, with the new GitHub Dependabot changes we should be able to do this in a simpler way.

PR #590, opened 2021-10-26
Add zip lambda

mojotalantikite:
## Summary
Adds a lambda to do bulk zipping. Each section on the submissions summary page should have a download button for all files in that section.

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-10581


services/app-api/handlers/bulk_download.ts
```diff
@@ -0,0 +1,164 @@
...
+            headers: {
+                'Access-Control-Allow-Origin': '*',
+                'Access-Control-Allow-Credentials': true,
+            },
+        }
+    }
+
+    const bulkDlRequest: S3BulkDownloadRequest = JSON.parse(event.body)
+    console.log('Bulk download request:', bulkDlRequest)
+    console.log('wat', bulkDlRequest.keys)
```
macrael, line 53: lol this is some of my nonsense logging. Probably can be removed now

services/app-web/src/components/DownloadButton/DownloadButton.tsx
```diff
@@ -0,0 +1,22 @@
...
+    zippedFilesURL: string
+}
+
+export const DownloadButton = ({
+    text,
+    zippedFilesURL,
+}: DownloadButtonProps): React.ReactElement => {
+    return (
+        <div>
+            <Link href={zippedFilesURL} target="_blank">
```
macrael, line 15: I don't think we want to nest a button inside a link, instead you can apply the button styles to the link like: 

`<Link className="use-button" variant="unsettled" >{text}</Link>`

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
```diff
@@ -79,9 +82,44 @@ export const ContractDetailsSummarySection = ({
...
+                console.log('ERROR: TODO: DISPLAY AN ERROR MESSAGE')
+                return
+            }
+
+            setZippedFilesURL(zippedURL)
+        }
+
+        void fetchZipUrl()
+    }, [getKey, getBulkDlURL, submission])
+    const [zippedFilesURL, setZippedFilesURL] = useState<string>('')
```
macrael, line 113: I think I prefer that all useState() calls be at the top of the function just to keep them near the props.

services/app-web/src/s3/s3Amplify.ts
```diff
@@ -123,6 +123,35 @@ function newAmplifyS3Client(bucketName: string): S3ClientT {
...
+            const prependedKeys = keys.map((key) => `allusers/${key}`)
+            const prependedFilename = `allusers/${filename}`
+
+            const zipRequestParams = {
+                keys: prependedKeys,
+                bucket: bucketName,
+                zipFileName: prependedFilename,
+            }
+
+            // the below is just a sketch, the URL needs to be fixed and the return value
```
macrael, line 142: old comments I think

macrael (approved):
Code looks great. I'm seeing a permissions error trying to download the zipped files in the review app, not sure what's going on there. We should probably add a cypress test to check that downloads download. 

```
<Error>
<Code>AccessDenied</Code>
<Message>Access Denied</Message>
<RequestId>1FK25NACBMTMFFWF</RequestId>
<HostId>
tLcBEzBnOoCXJdDysxYmSys89xAa9Zoq0MiaSwJ2Bt46lxmgxsDJpxtidp0RInqwaGrmsjnGIzk=
</HostId>
</Error>
```

mojotalantikite:
> Code looks great. I'm seeing a permissions error trying to download the zipped files in the review app, not sure what's going on there. We should probably add a cypress test to check that downloads download.
> 
> ```
> <Error>
> <Code>AccessDenied</Code>
> <Message>Access Denied</Message>
> <RequestId>1FK25NACBMTMFFWF</RequestId>
> <HostId>
> tLcBEzBnOoCXJdDysxYmSys89xAa9Zoq0MiaSwJ2Bt46lxmgxsDJpxtidp0RInqwaGrmsjnGIzk=
> </HostId>
> </Error>
> ```

Ah, yeah that error is from the virus scan. I was going to move the zip creation to the submit event so we don't have to worry about that 30 second-ish lag. In the future if we want to do sub-selection zip download on this page we can introduce a design element that indicates it's processing.

macrael:
> Ah, yeah that error is from the virus scan. I was going to move the zip creation to the submit event so we don't have to worry about that 30 second-ish lag. In the future if we want to do sub-selection zip download on this page we can introduce a design element that indicates it's processing.

Ahh, that makes sense. Part of me wants us to just not scan zips that we are making. What would that take?

mojotalantikite:
Oh, that's a good idea. Right now anything written to that bucket triggers a scan and a policy gets applied that only allows access to files that have gone through the scan. We could maybe create a `zips` bucket to write our zipped files to, that should be easy enough. CMS may have requirements around every file being scanned though, we should figure that out, although I agree it shouldn't pose a problem if we're in control of it and all the files already have been scanned.

services/app-web/src/components/DownloadButton/DownloadButton.tsx
mojotalantikite: I tried to apply the styling right onto the link like this and it wasn't styling as a button, it kept things as a link for whatever reason.

mojotalantikite:
Just as an update, we're writing zipped files to a `zips` bucket that is defined in the `uploads` service, which the `app-api` lambda then has permission to write to. We no longer have 30 second pauses/access denied errors when retrieving from this.

haworku:
I could not get the app running locally so made this change without being able to look at the rendered app 99ed79d but hopefully its working or almost there.

mojotalantikite:
Yeah I was getting a similar error -- my fault for having been only using `dev-hybrid` on this one! I fixed this in local by falling back onto a local env variable when extending a policy for the zip lambda.

Changes look good to me, thanks!

PR #673, opened 2021-12-06
Upgrade storybook deps

haworku:
## Summary
It's _a good idea™_  to upgrade storybook dependencies all together rather than incrementally.  There is a command `npx sb upgrade` to do this.  Since dependabot is letting us know we have some deps ready for upgrade, let's do that now.


macrael (approved):
all good

PR #675, opened 2021-12-08
User can retry if scanning fails

haworku:
#### Related issues
https://qmacbis.atlassian.net/browse/OY2-13896

#### Test cases covered
React component tests:
- `<FileItem />` - displays security scan failed message and both retry and remove buttons when status is SCANNING_ERROR
- `<FileItemList />` - here I added a updated snapshot that includes the Retry button UI
- `<FileUpload />` - calls `uploadFile` again when file scanning fails and is retried

## QA guidance
- Please test in review app. There is a universal bad file (EICAR anti-virus test file) that could be used


haworku:
When AWS came back up, I'm seeing this is a bit more complicated because the retry wasn't firing. Realized that we immediately delete scanning failed files, so retry was assuming the file wasn't present anymore. Have to adjust the logic a bit more to enable retry.

macrael (approved):
Looks great

PR #679, opened 2021-12-09
8040 form error summary part 1

jim:
## Summary

This is the first part of the implementation of a new error summary UI. It has been added to the first two pages of the form, and I'd like to get the team to review the general approach before going too much further.

Some comments pulled from commits:

* Some of the validations have been updated to conditionally apply in sync with the UI. This allows the Yup validations to be the source of truth for what error messages need to be displayed.
* For the document errors, the error determination was extracted into a variable and prepended to the Yup validation errors. This pattern will only work as long as additional errors are before or after the Yup errors and cannot support interleaved fields without additional changes.

Please also check for inline comments as you review 😄 

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-8040

#### Screenshots

Screen Shot 2021-12-09 at 2 23 03 PM

Screen Shot 2021-12-09 at 2 23 25 PM

#### Questions

* I haven't written unit tests for `ErrorSummary` yet. Any preference for how deep to go there? It doesn't do a lot but there is a little param juggling code.

#### Remaining tasks

- [ ] Replace buttons with links in error summary
- [ ] Don't autofocus after initial render


services/app-web/src/components/Form/ErrorSummary/ErrorSummary.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+function flattenMessage(message: string | string[]) {
+    if (typeof message === "string") {
+        return message;
+    }
+
+    return message.join(" and ");
+}
+
+export type ErrorSummaryProps = {
```
jim, line 48: This was heavily based on a similar component from EASi: https://github.com/CMSgov/easi-app/blob/master/src/components/shared/ErrorAlert/index.tsx

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.test.tsx
```diff
@@ -88,8 +88,8 @@ describe('ContractDetails', () => {
 
         // check that there are the errors we expect
         expect(
-            screen.queryByText('You must select at least one item')
-        ).toBeInTheDocument()
+            screen.queryAllByText('You must select at least one item')
```
jim, line 91: I can also see us having separate expectations for the field error message and the one in the summary above. WDYT?

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetailsSchema.ts
```diff
@@ -10,38 +10,50 @@ export const ContractDetailsFormSchema = Yup.object().shape({
...
-        .when(
-            // ContractDateEnd must be at minimum the day after Start
-            'contractDateStart',
-            (contractDateStart: Date, schema: Yup.DateSchema) => {
-                const startDate = dayjs(contractDateStart)
-                if (startDate.isValid()) {
-                    return schema.min(
-                        startDate.add(1, 'day'),
-                        'The end date must come after the start date'
+        .when("contractType", (contractType) => {
```
jim, line 24: As mentioned above, these changes are to keep the validations and UI visibility in sync.

services/app-web/src/components/Form/ErrorSummary/ErrorSummary.tsx
```diff
@@ -0,0 +1,91 @@
...
+}: ErrorSummaryProps): React.ReactElement | null => {
+    const numberOfErrors = Object.keys(errors).length;
+
+    const headingRef = useRef<HTMLHeadingElement>(null);
+
+    useEffect(() => {
+        if (numberOfErrors === 0) {
+            const { current } = headingRef;
+            if (current) {
+                current.focus();
```
jim, line 72: This will focus the heading when there are errors. I still need to verify this is working properly in screen readers.

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.test.tsx
haworku: Yeah either separate assertions or verify the length is 2. I would love to lean more heavily on jest here and be sure we have both inline errors and summaries up. That way we dont need to add much to Cypress.

services/app-web/src/components/Form/ErrorSummary/ErrorSummary.tsx
haworku: The  assertion confused me- based on what its doing seems like it would be number of !==0 to run the effect

jim: Yeah, that is absolutely a bug. It should be `numberOfErrors > 0`!

macrael:
@Jim looks like the build is broken because of:
```
src/pages/StateSubmission/Contacts/Contacts.tsx
  Line 30:10:  'ErrorSummary' is defined but never used  @typescript-eslint/no-unused-vars
1659
```

macrael:
We should figure out how to be that strict in the unit tests. We shouldn't have to get all the way to building for the deploy to hit this

services/app-web/src/components/Form/ErrorSummary/ErrorSummary.tsx
```diff
@@ -0,0 +1,91 @@
...
+type ErrorSummaryMessageProps = {
+    errorKey: string,
+    message: string
+}
+const ErrorSummaryMessage = ({
+    errorKey, message
+}: ErrorSummaryMessageProps): React.ReactElement => {
+    return (
+        <button
+            type="button"
```
macrael, line 15: I remember there being a long discussion about Button vs. Link in the A11y course we took. since this is navigating us to somewhere one the page would it make more sense as a link?

services/app-web/src/components/Form/ErrorSummary/ErrorSummary.tsx
macrael: Since this effect is being run every time the number of errors changes, is there a risk here that we're going to focus the error summary as we're just filling in form fields, after having clicked "continue" once? I'm waiting for a deploy to play with it live but I'm curious about this.

macrael (approved):
Looks great. We generally approve on first review unless there's something egregious, so feel free to merge this in whenever you think it's ready.

services/app-web/src/components/Form/ErrorSummary/ErrorSummary.tsx
jim: I think the link guidance is that you have to have a value for `href` for it to be a link. We have `id` attributes on the inputs, so I can change these to links and use those anchors.

services/app-web/src/components/Form/ErrorSummary/ErrorSummary.test.tsx
```diff
@@ -41,7 +41,24 @@ describe('ErrorSummary component', () => {
...
+    it('autofocuses the heading by default', () => {
+        render(
+            <ErrorSummary
+              errors={{title: "You must provide a title"}} />
+        )
+
+        expect(screen.getByText("There is 1 error on this page")).toHaveFocus()
+    })
+
+    it('does not autofocus when autofocus is false', () => {
```
haworku, line 55: nice test thank you

haworku (approved):
This is also going to make Cypress screenshots more useful sometimes we get into error states and cypress is scrolled to the bottom of the form so we cant see what went wrong.

PR #689, opened 2021-12-14
Add ADRs to the repo

macrael:
## Summary

Now that our repo is public, I’d like our projects Engineering ADRs to live here in this repo canonically. It will make them easier to share.


haworku (approved):

haworku:
If this is the source of truth should be linking to each for each entry on confluence then? We could still fill out confluence if its needed (or have that be the space for comments and discussions). But then when its finalized link to github.

macrael:
Once I merge this I was going to delete the confluence versions. It’s annoying to have it in two different formats so I think we should figure out how to do markdown the whole way. Maybe another use of https://hackmd.io for editing

PR #698, opened 2021-12-14
8040 form error summary part 2

jim:
## Summary

This is the second half of the work to complete Story 8040.

MacRae and I decided to opt for explicit focus management instead of having `ErrorSummary` autofocus in certain situations. The earlier uses of ErrorSummary have been updated to reflect this.

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-8040

#### Screenshots

Screen Shot 2021-12-14 at 3 58 15 PM


services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
```diff
@@ -343,12 +354,12 @@ export const SubmissionType = ({
                                 disabled={isSubmitting}
                                 onClick={() => {
                                     setShouldValidate(true)
+                                    setFocusErrorSummaryHeading(true)
                                 }}
                             >
                                 Continue
                             </Button>
                         </ButtonGroup>
-                        {/* <FormikFocusOnErrors /> */}
```
macrael, line 351: that one's been a long time coming!

services/app-web/src/pages/StateSubmission/Contacts/Contacts.tsx
```diff
@@ -129,6 +129,38 @@ const actuaryContactErrorHandling = (
...
 }
 
+// Convert the formik errors into a shape that can be passed to ErrorSummary
+const flattenErrors = (
+    errors: FormikErrors<ContactsFormValues>
+): { [field: string]: string } => {
+
+    const flattened: {[field: string]: string} = {}
+
+    if (errors.stateContacts && Array.isArray(errors.stateContacts)) {
```
macrael, line 139: There might be some clever way tot use [Array.flatMap()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) here but I am a big fan of this simplicity.

services/app-web/src/pages/StateSubmission/Contacts/Contacts.test.tsx
```diff
@@ -89,14 +89,14 @@ describe('Contacts', () => {
 
         await waitFor(() => {
             expect(
-                screen.getByText('You must provide a name')
```
macrael, line 92: Have you used [`within`](https://testing-library.com/docs/dom-testing-library/api-within/) before? I think checking to see that the error text is repeated is a fine way to do this but want to make sure you know about `within` b/c it's useful.

macrael (approved):
Fab.

services/app-web/src/pages/StateSubmission/Contacts/Contacts.test.tsx
jim: I have! I was torn about if it was worth having separate assertions against the error summary and the field error area.

services/app-web/src/pages/StateSubmission/Contacts/Contacts.tsx
jim: This code is right on the line where if it gets more complicated (or repeats itself a third time) it might be worth some additional refinement. I'm happy to do that sooner if we think it's worthwhile.

services/app-web/src/pages/StateSubmission/Contacts/Contacts.test.tsx
jim: I kinda want these tests to say "at least once" and then have the specific error display and behavior tested elsewhere. But hopefully this won't change again soon, so it's probably fine as-is.

services/app-web/src/pages/StateSubmission/Contacts/Contacts.tsx
macrael: nah it’s fine

tests/cypress/integration/stateWorkflow/stateSubmissionForm/documents.spec.ts
```diff
@@ -9,6 +9,17 @@ describe('documents', () => {
             const draftSubmissionID = pathname.split('/')[2]
             cy.visit(`/submissions/${draftSubmissionID}/documents`)
 
+            // Add two of the same document
```
jim, line 12: This is here until we can write a Jest test that covers this bug. So far, that has been elusive.

jim:
Closed in favor of #702

PR #699, opened 2021-12-15
New submission ID format

rswerve:
## Summary

[13267](https://qmacbis.atlassian.net/browse/OY2-13267)
[13200](https://qmacbis.atlassian.net/browse/OY2-13200)

We want the submission ID to begin with `MCR-` and we want it to contain the names of the associated programs.

#### Screenshots

image

#### Test cases covered

New cypress test to confirm that the submission ID starts with `MCR-` and that it contains at least one of the selected program names.

## QA guidance

Add a submission; check the ID.


macrael (approved):
Looks great. The order you select programs seems to matter right now. I think that long term that might be annoying because we want to make it easy for folks to compare submissions over time from the same set of programs.

How about for now we at least sort the IDs when constructing the ID, so that if we end up sorting things on the dashboard as well we at least won't have to change any IDs.

services/app-web/src/common-code/domain-models/submission.ts
```diff
@@ -80,9 +80,13 @@ const isDraftSubmission = (sub: unknown): sub is DraftSubmissionType => {
...
 }
 
+const naturalSort = (a: string, b: string): number => {
+    return a.localeCompare(b, "en", { numeric: true })
+}
+
 function submissionName(submission: SubmissionUnionType): string {
     const padNumber = submission.stateNumber.toString().padStart(4, '0')
-    return `${submission.stateCode.toUpperCase()}-${submission.programIDs[0].toUpperCase()}-${padNumber}`
+    return `MCR-${submission.stateCode.toUpperCase()}-${submission.programIDs.sort(naturalSort).join('-').toUpperCase()}-${padNumber}`
```
haworku, line 89: Could we make a shared helper for submission name. This is used also in `app-api/emailer/temlates.ts` for emails.  I am remembering there was still a TODO to share that code somehow.

macrael: I think this could be the shared helper since it’s in the domain model. I think I added that? I’m not sure if it’s being called in the email templates though

haworku (approved):
LGTM

rswerve:
> Looks great. The order you select programs seems to matter right now. I think that long term that might be annoying because we want to make it easy for folks to compare submissions over time from the same set of programs.
> 
> How about for now we at least sort the IDs when constructing the ID, so that if we end up sorting things on the dashboard as well we at least won't have to change any IDs.

I added sort to the id creation.

PR #701, opened 2021-12-15
strip down how we set env vars in CI

macrael:
## Summary
I was using this branch to test SES emails in DEV, but now it's completely unrelated to that. I went through our deploy scripts and simplified how we were handling env vars. Now jobs should mostly have env vars passed directly to them instead of relying on the build_vars script. This should make it more clear where env vars need to be set in the future. 

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-14289


haworku:
🗒️ Reminder to add the missing ENV variables (the mocked versions for local dev) to envrc

haworku:
Also re slack thread - Maybe we should also document something like  _" for local dev, .envrc is loaded and then it loads everything in .local"_  in the developer README.

Right now it just says _"For local dev, we use `.envrc` and `.envrc.local` to set the appropriate environment variables. Any local changes should be made to `.envrc.local`"._  I suppose the relationship isn't entirely clear just from that.

docs/Configuration.md
```diff
@@ -108,3 +108,51 @@ Sets the "from" address for all emails sent by the system. This address must hav
 Read by `app-api`
 
 Sets the "to" addresses for generic emails sent to the shared CMS inbox (there may be multiple). Value is comma separated string.
+
+### `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_DEFAULT_REGION`
```
mojotalantikite, line 112: oh awesome, thanks for adding more docs around these vars!

mojotalantikite (approved):
This is great, thanks for cleaning these up. It always felt bad to have the vars exported everywhere all the time and not specifically (and only) where they were needed.

PR #704, opened 2021-12-16
Add config to debug tests in vscode

rswerve:
## Summary

On past projects, we've shared a vscode config for doing things like running tests in the vscode debugger.  If folks would prefer to keep their config individually and locally, we can just close this.

@jim did the heavy lifting to get the debugger working on this project (for tests in /app-web).

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,36 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Debug all app-web tests",
+            "type": "node",
+            "request": "launch",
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,36 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
macrael, line 4: If VSCodes doens’t care then I don’t care but comments aren’t valid in JSON.

macrael (approved):
Looks good to me. Would it be helpful to specify any other config? Like linters or something?

.vscode/launch.json
rswerve: vscode does not care 🙂

rswerve:
> Looks good to me. Would it be helpful to specify any other config? Like linters or something?

The one that would be really helpful is attaching to our backend process, so we could debug.  But I think that will take a little more experimentation to get working.

PR #706, opened 2021-12-17
Restore offwhite background everywhere except the dashboard

rswerve:
## Summary

[14428](https://qmacbis.atlassian.net/browse/OY2-14428)

The submission table page should have a white background, and other pages should have an offwhite background.  Screenshots below.

#### Screenshots

image
image

#### Test cases covered

No tests.

## QA guidance

Probably best to defer to design on this one.


macrael (approved):
short and sweet

services/app-web/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -99,119 +99,127 @@ export const Dashboard = (): React.ReactElement => {
...
-                                    asCustom={NavLink}
-                                    className="usa-button"
-                                    variant="unstyled"
-                                    to={{
-                                        pathname: '/submissions/new',
-                                    }}
-                                >
-                                    Start new submission
-                                </Link>
+            <div className={styles.wrapper}>
```
haworku, line 102: Oh! This would be useful on other full width pages as well. 🤔  We don't have a pattern around this yet - an element that fills over the background of the app - and then you lay the grid on top. I feel like Landing page could use this too.

Would you mind pulling this out as a component?  `FullWidthPageWrapper` or something like that? It could takes all the props a div `JSX.IntrinsicElements['div']`  takes - that way anyone using it could override the styles

haworku: oh I'm too late nvm lol. We can do this in a future PR

PR #707, opened 2021-12-18
Change date rage separator to 'to'

macrael:
## Summary

Our date ranges on the summary pages get read strangely by Voiceover. This is a change to address that. 

We display date ranges (like contract effectiveness dates) on pages like this:

12/03/2021 - 08/14/2022

Voiceover has a real bug with this where it reads the 08 as “two thousand eight” for some reason. JAWS did not. But both of them read them very literally. “twelve slash zero three slash twenty twenty one dash…”

[This guidance](https://www.gov.uk/guidance/content-design/writing-for-gov-uk#date-ranges) from gov.uk specifies using “to” instead of a dash, and that was enough to get VO to behave. It also suggests writing out months explicitly. So then I tried this:

Dec 3 2021 to Aug 14 2022

and VO read it like a person would. “December third two thousand twenty one to…”
I’m definitely going to advocate we use “to” as separator, and am curious how our users would like slashless dates, they certainly seem nicer to listen to

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-11681

#### Screenshots
Screen Shot 2021-12-17 at 5 16 29 PM

### DIFF
--- a/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
+++ b/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
@@ -39,29 +39,11 @@ export const RateDetailsSummarySection = ({
                                     ? 'Rating period of original rate certification'
                                     : 'Rating period'
                             }
-                            data={
-                                <>
-                                    <time
-                                        dateTime={dayjs(
-                                            submission.rateDateStart
-                                        ).format('YYYY-MM-DD')}
-                                    >
-                                        {dayjs(submission.rateDateStart).format(
-                                            'MM/DD/YYYY'
-                                        )}
-                                    </time>{' '}
-                                    to{' '}
-                                    <time
-                                        dateTime={dayjs(
-                                            submission.rateDateEnd
-                                        ).format('YYYY-MM-DD')}
-                                    >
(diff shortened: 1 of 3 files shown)


rswerve (approved):

haworku (approved):

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
```diff
@@ -39,11 +39,29 @@ export const RateDetailsSummarySection = ({
...
+                                    <time
+                                        dateTime={dayjs(
+                                            submission.rateDateStart
+                                        ).format('YYYY-MM-DD')}
+                                    >
+                                        {dayjs(submission.rateDateStart).format(
+                                            'MM/DD/YYYY'
+                                        )}
+                                    </time>{' '}
+                                    to{' '}
```
haworku, line 53: i would put the to in a `<span>{' to '}</span>`

to me random spaces `'  '` outside an element are a problem in jsx - they are so easy to accidentally delete (or add more spaces accidental with prettier) and then change the way the UI looks. you can also use `&nbsp;` for spaces fyi

haworku:
could use two elements here -- one for reading one for screenreaders. It might ensure that the experience is the same between assistive devices if that is proving troublesome (wrote a bit more about this on jira off the thread there)

PR #709, opened 2021-12-20
Readme edits: add more detail about how to verify ctkey setup

haworku:
## Summary
Adding some notes on how to verify ctkey setup with serverless.  The `which serverless` expected result is tied to work in #590.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ which serverless # should return something like /managed-care-review/scripts/ser
 which sls # should return something like /managed-care-review/scripts/sls`
 ```
 
-These should both point to paths inside the codebase (not to paths in usr/local/bin).
+These should both point to paths inside the codebase (not to paths in /usr/local/bin).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -244,6 +244,20 @@ Currently, `ctkey-wrapper` requires the user to be running the openconnect-tinyp
...
+
+To verify serverless (and AWS access) is set up properly with ctkey, run:
+
+ ```shell
+which serverless # should return something like /managed-care-review/scripts/serverless`
+which sls # should return something like /managed-care-review/scripts/sls`
+```
+
+These should both point to paths inside the codebase (not to paths in usr/local/bin).
+
```
mojotalantikite, line 257: Cool, looks great, thanks for adding this!

Just a typo on this path, '/usr/local/bin'.

mojotalantikite (approved):

PR #710, opened 2021-12-20
Conditionally keep Aurora online

mojotalantikite:
## Summary

Because Aurora turns itself off if there is no traffic to it, we have long cold-start times for our application in val and prod. This creates a bad user experience as a fresh user to the application when the database is off has to wait 30 seconds or more before the application can handle requests.

We've decided to keep a minimal configuration online in val and prod so there are no more cold start times.

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-14082

### DIFF
--- a/services/postgres/serverless.yml
+++ b/services/postgres/serverless.yml
@@ -58,7 +58,6 @@ custom:
       - main
   pauseAurora:
     other: true
-    dev: true
     val: false
     prod: false
 
--- a/services/postgres/serverless.yml
+++ b/services/postgres/serverless.yml
@@ -19,17 +19,17 @@ provider:
       statements:
         - Effect: 'Allow'
           Action:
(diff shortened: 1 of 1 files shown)


services/postgres/serverless.yml
```diff
@@ -56,6 +56,11 @@ custom:
       - val
       - prod
       - main
+  pauseAurora:
+    other: true
+    dev: true
```
macrael, line 61: I think this should be "main" we don't have a "dev" stage

macrael (approved):
LGTM

services/postgres/serverless.yml
mojotalantikite: Ah yeah, of course. I actually just deleted `dev` altogether and let it fall back to the catchall `other`.

PR #713, opened 2021-12-21
Prepend npx to serverless local command

mojotalantikite:
## Summary

Now that ctkey is working for `serverless` commands, when running `./dev local` the command that is being run uses the ctkey wrapped command. This doesn't work when we don't want to actually connect to AWS. 

This fix runs `npx serverless ...` so that the command being run by `./dev local` uses the regular serverless environment and skips ct-key.


macrael (approved):
lgtm

PR #714, opened 2021-12-21
Fix upload "scan" failures

rswerve:
## Summary

[Ticket 13893](https://qmacbis.atlassian.net/browse/OY2-13893)

During a walkthrough, we saw multiple file upload scan failures and decided to investigate.

In pairing on this with @mojotalantikite and @haworku, we discovered two issues:
1. Large files were being uploaded in parts, and multipart upload *requires a different trigger* in the lambda
2. Scans of larger files were taking longer than our three polling retries allowed.

#### Screenshots

image

#### Test cases covered

No tests; just some config changes.

## QA guidance

Upload large files, and several files at the same time.  Should work.

### DIFF
--- a/services/app-web/src/s3/s3Amplify.ts
+++ b/services/app-web/src/s3/s3Amplify.ts
@@ -144,7 +144,7 @@ const waitFor = (delay = 1000) =>
 const retryWithBackoff = async (
     fn: () => Promise<void | S3Error>,
     retryCount = 0,
-    maxRetries = 3,
+    maxRetries = 6,
     err: null | S3Error = null
 ): Promise<void | S3Error> => {
     if (retryCount > maxRetries) {
--- a/services/uploads/serverless.yml
+++ b/services/uploads/serverless.yml
@@ -120,6 +120,8 @@ resources:
                 - DELETE
                 - HEAD
               MaxAge: 3000
+              ExposedHeaders:
+                - ETag
         NotificationConfiguration:
           LambdaConfigurations:
             - Event: s3:ObjectCreated:Put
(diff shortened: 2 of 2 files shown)


mojotalantikite (approved):

services/app-web/src/s3/s3Amplify.ts
```diff
@@ -144,7 +144,7 @@ const waitFor = (delay = 1000) =>
 const retryWithBackoff = async (
     fn: () => Promise<void | S3Error>,
     retryCount = 0,
-    maxRetries = 3,
+    maxRetries = 6,
```
haworku, line 147: If we retry 6 times how much time is that? I think this puts us over a minutes for uploading for one file if its failing. 

1. Can you please add those notes to the `scanFile` function in the code? I've been trying to document how long this can take at most per file in the code.  

 Also flagging this for design/product review- I think we should display some hint UI on our form uploads now to say "This could take up to X seconds' because we are pushing a minute.

rswerve: @haworku I see what you mean.  A known-bad file took 150 seconds to return an error.  That does seem too long.  But we found 4 retries wasn't enough, so 6 seemed safe, but it does take quite a while.

haworku: @rswerve yeah thats fair- I do  think we need to tell users what to expect in that case

rswerve: Yeah, @mojotalantikite were kicking around ideas, like accepting the file immediately and showing the status on the summary page, or...something else.

I'm not clear on what you'd like changed or added to the PR though.

haworku: Excuse me for adding to any confusion!  Thanks for pinging  on slack

PR #716, opened 2021-12-22
File input announces when a file has been selected

haworku:
## Summary
WIP 

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-10541


haworku:
This PR could use more work. I have the "1 file selected" announcing on the first selection but subsequent selections are not being announced as files are added. Probably the aria live element needs to be rerendered when new files are dropped to trigger the alert again? That was my guess...

This will take some reading around patterns for changing aria-live in React.

haworku:
closing in favor of @jim work last sprint

PR #718, opened 2021-12-27
Break out app-web/ui-auth/storybook into their own deploys

macrael:
## Summary

This removes deploy.sh in favor of explicitly calling serverless deploy for every service we have. Each service should now be configured with exactly the env vars they require.


mojotalantikite (approved):
Finally! Looks great.

PR #720, opened 2021-12-28
Collect CI Runtime Stats

macrael:
## Summary

This adds a script to our repo that crunches the numbers on how long individual pieces of our CI are taking. 

It outputs a list of all jobs and steps in order of how much time they have taken on average recently. 

https://qmacbis.atlassian.net/browse/OY2-14972

should we run this in CI for any reason?


mojotalantikite (approved):

mojotalantikite:
I'm not sure if we'd want to run this in CI at any point or not. I can see a use case in the future where we have an endpoint to send observability stats to, and sending in CI stats at every run would be an interesting use case. Right now I think it's fine to just have a method in which we can pull these stats from time to time like it is now.

PR #723, opened 2021-12-28
Fixing wrapping tag layout on dashboard

joshmfranklin:
Tags on the dashboard are wrapping strangely with longer program names. 

I've added `display: inline-block` to the tag layout to allow for the tags to wrap to new lines when they are longer.


macrael (approved):
👍🏻

PR #732, opened 2021-12-30
Thin out Setup Env script

macrael:
## Summary

The setup env script is used everywhere b/c it installs yarn. I’m trying to thin it out and not use it in places it’s not needed.


.github/workflows/deploy-app-to-env.yml
```diff
@@ -67,6 +64,12 @@ jobs:
       - name: Setup env
         uses: ./.github/actions/setup_env
 
+      - name: Generate Code
+        shell: bash
+        run: |
+          cd services/app-graphql && yarn gqlgen
+          cd ../app-proto && yarn protogen
```
mojotalantikite, line 71: Looks like `ubuntu-latest` has Lerna [installed by default now](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md), so we could define a task that is, say, `generate-code` in our `app-graphql` and `app-proto` `package.json` files and call it as `lerna run generate-code` and these two tasks will happen in parallel.

mojotalantikite (approved):

PR #742, opened 2022-01-06
Clean up email tests and make red text bold

haworku:
## Summary
- make it BOLD
- get tests back in there without all fighting all the whitespace. Test smaller chunks of text.

#### Related issues
follow up from #741 

#### Test cases covered

Added a new set of tests to make sure all state contacts are included in to addresses for state email.  Otherwise just added more assertions to existing tests to hopefully make it less breakable but still have a test that will fail if the wrong email template is prepared.

### DIFF
--- a/services/app-api/emailer/templates.ts
+++ b/services/app-api/emailer/templates.ts
@@ -39,11 +39,7 @@ const newPackageCMSEmailTemplate = (
 
             View submission: ${submissionURL}`,
         bodyHTML: `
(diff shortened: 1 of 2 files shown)


haworku:
@rswerve yeah you were right. what I thought was an issue with nesting the _expect...contains w_as just another whitespace fail I missed 😵‍💫

rswerve (approved):

PR #745, opened 2022-01-07
Announce selected file count (#10541)

jim:
## Summary

 This PR adds some text to the file upload component that describes how many files have been selected. This text updates automatically, and whenever a new file is added or one is removed, it is auto-focused. This will cause it to be announced by JAWS.

This PR also includes some additional markup to add ARIA roles to the progress display used during file uploads. That work will likely continue.

For more details, see commit messages 😄 

#### Related issues

This PR supersedes #716, which took a different approach to addressing this issue. 

#### Screenshots

https://user-images.githubusercontent.com/3331/148585769-5d6a1bcb-cbe0-492e-acb8-909091ca6fdb.mov

## QA guidance

As you make changes to the file list, the number of file should always be announced.

### DIFF
--- a/services/app-web/src/components/FileUpload/FileItem/FileItem.tsx
+++ b/services/app-web/src/components/FileUpload/FileItem/FileItem.tsx
@@ -120,16 +120,27 @@ export const FileItem = ({
         retryItem(item)
     }
 
+    let statusValue = "";
+    if (isLoading) {
+        statusValue = "uploading";
+    } else if (isScanning) {
+        statusValue = "scanning for viruses"
+    } else if (hasDuplicateNameError || hasScanningError || hasUploadError || hasUnexpectedError) {
+        statusValue = "error"
+    }
+
     return (
         <>
             <div className={styles.fileItemText}>
-                <img
-                    id={item.id}
-                    data-testid="file-input-preview-image"
-                    src={SPACER_GIF}
-                    alt=""
-                    className={imageClasses}
-                />
+                <div role="progressbar" aria-valuetext={statusValue} aria-label={`Status of file ${name}`}>
+                    <img
+                        id={item.id}
+                        data-testid="file-input-preview-image"
+                        src={SPACER_GIF}
+                        alt=""
+                        className={imageClasses}
+                    />
+                </div>
                 <span
                     style={{
                         display: 'flex',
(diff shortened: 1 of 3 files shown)


services/app-web/src/components/FileUpload/FileUpload.tsx
```diff
@@ -298,6 +310,8 @@ export const FileUpload = ({
         addFilesAndUpdateList(files)
     }
 
+    const summary = `${fileItems.length} file${fileItems.length !== 1 ? 's' : ''} selected`
```
rswerve, line 313: Not a blocker, and probably more of a pain than it's worth, given its location in app-api, but I recently added a `pluralizer` helper to the codebase that might have saved you one-third of a second.  🙂

kaleighsimmons:
We decided to change "selected" to "added" (so it should now read, "2 files added")

jim:
> We decided to change "selected" to "added" (so it should now read, "2 files added")

@kaleighsimmons Thanks for catching this! Here's what it looks like now:

image

services/app-web/src/components/FileUpload/FileUpload.tsx
jim: Do we have any rules around importing across package boundaries? I kinda think it's better to have a lib that both services import, but like you said, that's a lot of work for a couple lines of code.

haworku: Rules no. Limitations (at least up until a couple months ago) yes. At one point because of using CRA and _not_ having yarn workspaces set up at the beginning we simply weren't able to import from anywhere else in the codebase into app-web. 

That led to weird patterns like the `common-code` folder in app-web which is where we put code we want to share between. That should be the starting place for any lib. I think we are well positioned to that now with the yarn workspaces.  If ya want, Isuggest for now just put code you want to share between there.

services/app-web/src/components/FileUpload/FileItem/FileItem.tsx
```diff
@@ -120,16 +120,27 @@ export const FileItem = ({
...
         <>
             <div className={styles.fileItemText}>
-                <img
-                    id={item.id}
-                    data-testid="file-input-preview-image"
-                    src={SPACER_GIF}
-                    alt=""
-                    className={imageClasses}
-                />
+                <div role="progressbar" aria-valuetext={statusValue} aria-label={`Status of file ${name}`}>
```
haworku, line 135: this use of progressbar is 💅 I like how it maps to the visual UI we already have established

haworku (approved):

haworku:
Merging this to continue other a11y work. nice wokr

PR #746, opened 2022-01-07
OY2-13565 make step header h2

rswerve:
## Summary

Closes [13565](https://qmacbis.atlassian.net/browse/OY2-13565)

Small change to use react-uswds's new capability to specify the header level in a step indicator to make sure that our headers progress sequentially.

#### Test cases covered

Added one line to a unit test to check that the passed-in level is what's in the DOM.

## QA guidance

Should look and work exactly the same, but the DOM should have an h2 in the header, and screen readers should not be confused by the app headers.

### DIFF
--- a/services/app-web/package.json
+++ b/services/app-web/package.json
@@ -64,7 +64,7 @@
     },
     "dependencies": {
         "@apollo/client": "^3.4.15",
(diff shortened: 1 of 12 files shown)


haworku (approved):
LGTM! ⭐

PR #752, opened 2022-01-12
Fix bug in uploading files with commas in name

rswerve:
## Summary

Uploading files with commas in their names was failing in Chrome.  Some googling led me to the `content-disposition` header, and I'm just wrapping the file name in quotes, which resolves the issue.

Closes [15489](https://qmacbis.atlassian.net/browse/OY2-15489)

#### Screenshots

image

#### Test cases covered

No changes

## QA guidance

Meghan confirmed this works on the deployed branch, but feel free to try it out for yourself.

### DIFF
--- a/services/app-web/src/s3/s3Amplify.ts
+++ b/services/app-web/src/s3/s3Amplify.ts
@@ -42,7 +42,7 @@ function newAmplifyS3Client(bucketName: string): S3ClientT {
             try {
                 const stored = await Storage.put(`${uuid}.${ext}`, file, {
                     contentType: file.type,
-                    contentDisposition: `attachment; filename=${file.name}`,
+                    contentDisposition: `attachment; filename=${`file.name`}`,
                 })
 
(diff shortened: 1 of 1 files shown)


services/app-web/src/s3/s3Amplify.ts
```diff
@@ -42,7 +42,7 @@ function newAmplifyS3Client(bucketName: string): S3ClientT {
             try {
                 const stored = await Storage.put(`${uuid}.${ext}`, file, {
                     contentType: file.type,
-                    contentDisposition: `attachment; filename=${file.name}`,
```
haworku, line 45: huh i'm surprised the backticks  worked. I would think `attachment; filename="${file.name}"`

haworku (approved):

services/app-web/src/s3/s3Amplify.ts
rswerve: that might well have worked, too

PR #753, opened 2022-01-12
Add file list detail to the FileUpload summary

haworku:
## Summary
This adds more detail to the existing summary h5 including complete, error and pending files

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-13947

### Tests
list summary heading... 
- displays list count - X files added
- displays complete, errors, and pending in list

## how to verify 
- make the text changes as expect as file uploads complete.
- make sure it is only proactively announced in JAWS (interrupting the user's flow) when the user first adds a new file


services/app-web/src/components/FileUpload/FileUpload.tsx
```diff
@@ -329,7 +353,16 @@ export const FileUpload = ({
                 </span>
             )}
 
-            <h5 tabIndex={-1} ref={summaryRef} className="text-normal font-body-sm margin-0">{summary}</h5>
+            <div style={{ display: 'flex' }}>
+                <h5
+                    tabIndex={-1}
+                    ref={summaryRef}
+                    className="text-normal font-body-sm margin-0"
+                >
```
haworku, line 361: @jim This is my new idea. Keep the H5 only on "the X files added" text. Move the rest of the summary text (`summaryDetailText`) outside. That way its not `aria-hidden` but its also on the page. Thoughts?

services/app-web/src/common-code/formatters/index.ts
```diff
@@ -0,0 +1,3 @@
+import { pluralize } from './pluralizer'
+
+export { pluralize }
```
haworku, line 3: @rswerve fyi pluralize is now in in `common-code` in app-web so its shareable.

PR #755, opened 2022-01-13
Focus file list when there are file errors (#13946)

jim:
## Summary

Within forms that allow uploads:
* Clicking on the error messages for validation errors within the file list (duplicate files or upload/scan errors) will now focus the file list
* Errors related to files not being selected will continue to focus the file input when clicked

I attempted to implement this by passing `React.Ref`s down into the `ErrorSummary`, but in the end the behavior was flakey, there was a lot more code, and ultimately I fell back to the approach used here.

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-13946

#### Test cases covered

I didn't add unit tests for this functionality due to the limitations with JSDOM and Jest that we uncovered in an earlier story. I did add a new story to Storybook, though, that shows the functionality.

## QA guidance

* Trigger file errors and verify that the file list is focused when activating the error message in the error summary.
* Verify the file input is still focused when no files have been selected.

### DIFF
--- a/services/app-web/src/components/Form/ErrorSummary/ErrorSummary.tsx
+++ b/services/app-web/src/components/Form/ErrorSummary/ErrorSummary.tsx
@@ -29,9 +29,11 @@ const ErrorSummaryMessage = ({
             href={href}
             className={classnames(styles.message)}
             data-testid="error-summary-message"
+            onClick={(event) => {
                 const fieldElement: HTMLElement | null = document.querySelector(fieldSelector);
             
                 if (fieldElement) {
+                    event.preventDefault();
                     fieldElement.focus();
                 }
             }}
--- a/services/app-web/src/components/Form/ErrorSummary/ErrorSummary.tsx
+++ b/services/app-web/src/components/Form/ErrorSummary/ErrorSummary.tsx
@@ -10,15 +10,26 @@ type ErrorSummaryMessageProps = {
 const ErrorSummaryMessage = ({
     errorKey, message
 }: ErrorSummaryMessageProps): React.ReactElement => {
+    let fieldSelector : string;
+    let href: string;
+
+    // Treat keys that begin with # as ids
+    if (errorKey.startsWith("#")) {
+        fieldSelector = errorKey;
+        href = errorKey
+
+    // Otherwise, assume that keys correspond to name attributes and ids
+    } else {
+        href = "#" + errorKey
+        fieldSelector = `[name="${errorKey}"]`;
+    }
+
     return (
         <a
-            href={"#" + errorKey}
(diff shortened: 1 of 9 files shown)


services/app-web/src/components/Form/ErrorSummary/ErrorSummary.tsx
```diff
@@ -59,6 +72,11 @@ export type ErrorSummaryProps = {
  *
  * This component relies on the errors object returned by Formik, so it should generally
  * rendered inside of a Formik form context.
+ * 
+ * By default, keys in the errors property are assumed to correspond to the name attribute
+ * of the relevant form inputs. By passing a key that begins with a number sign ("#"),
+ * you can have it treated as an ID. This is useful when you want to focus something that
+ * doesn't have a name attribute when an error message is clicked.
```
jim, line 79: This is the main change: `errorKey`s that start with `#` are treated as IDs instead of values of `name` attributes.

There's some cleanup that would be nice to do- we have places where the names and `id`s don't match but probably could, and this means that some of our links to those elements don't go anywhere. It's not an issue since we handle the focus in JS, but adding more consistency over time would be nice.

haworku: Thanks for adding the comment here too I think that's clear.

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
```diff
@@ -101,6 +101,8 @@ export const RateDetails = ({
             : showFileUploadError && !hasValidFiles
             ? ' You must remove all documents with error messages before continuing'
             : undefined
+    const documentsErrorKey =
+        fileItems.length === 0 ? 'rateDocuments' : '#file-items-list'
```
haworku, line 105: oof seeing this I wish we had clearer ids `rate-documents-file-input`.  suspect that change could break tests but I can circle back here to look at this at the end of the sprint.

haworku:
This PR branch is having strange errors in CI. I have rerun a few times and getting different errors - about env variables not coming down for app-api, one was a linting error that looking at the codebase doesn't seem to exist for app-web. 🤔 Looking around a bit

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
jim: @haworku We might want to pass the list id in as a prop to `FileUpload`, that way the form could dictate the value.

haworku (approved):
Yay this deployed successfully and I was able to test.

PR #756, opened 2022-01-13
Make all error messages in state submission form use "polite"

haworku:
## Summary
@jim  saw some conflicts between our `<ErrorSummary />` and the inline `<ErrorMessage />` in screenreaders. This may be because both use `role=alert`, which uses assertive announcement by default and interrupts the users flow.

The ideal solution is still to remove `role="alert` all together from inline errors and prefer `aria-describedby` attribute to associate the errors (and hints) with their input. Looking around the codebase more, this seems liked a more significant code and tests change,  so getting this small PR in first so its in place for our upcoming 508 tests as an interim step. Further refactor will follow.

This PR adds a `<PoliteErrorMessage/>` that uses `aria-live="polite"` instead of assertive to make a slightly better user experience.

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-15518

#### Test cases covered
tests unchanged


haworku:
@mojotalantikite I force pushed over my branch, which only had one committ, and broke something I think with our service diff logic 😄. Gonna try pushing a brand new branch but FYI 
Screen Shot 2022-01-14 at 8 38 51 AM

jim (approved):
Looks good to me! I agree with the incremental approach used here.

PR #762, opened 2022-01-19
Destroy script in TypeScript

mojotalantikite:
## Summary

We've been having an issue where the old `destroy.sh` script will return successfully, but sometimes resources will not completely be removed from AWS. This causes us to sometimes hit resource limits due to dangling, unused AWS resources.

Taking this as an opportunity to rewrite the script in TypeScript and using the `aws-sdk` directly, rather than using `jq` on cli output where we're clearly missing things.

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-14896


.github/workflows/destroy.yml
```diff
@@ -13,21 +13,31 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
       - name: destroy all
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
           STAGE_PREFIX: ${{ secrets.STAGE_PREFIX }}
-        run: ./destroy.sh $STAGE_PREFIX$branch_name
```
haworku, line 30: should we also delete `destroy.sh`

haworku (approved):

.github/workflows/destroy.yml
mojotalantikite: yes! we should (I will)

scripts/destroy_stage.ts
```diff
@@ -0,0 +1,239 @@
...
+    const bucketParams = {
+        Bucket: bucket.PhysicalResourceId ?? '',
+    }
+
+    // get all versioned objects
+    let versionKeys: s3ObjectKey[] = []
+    let deleteMarkerKeys: s3ObjectKey[] = []
+
+    const objectVersions = await s3.listObjectVersions(bucketParams).promise()
+    // AWSError is unfortunately not backed by Error, so we can't just
```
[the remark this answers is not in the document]
mojotalantikite: ahh ok awesome! I had been thinking of writing a type guard around `AWSError` but figured it was too much for just this little destroy script. This looks like it would be helpful, thanks!

scripts/destroy_stage.ts
```diff
@@ -0,0 +1,234 @@
...
+async function main() {
+    stackPrefixes.map(async (prefix) => {
+        const sn = `${prefix}-${stage}`
+        const clearBucketOutput = await clearServerlessDeployBucket(sn)
+        if (clearBucketOutput instanceof Error) {
+            console.log(clearBucketOutput)
+        }
+        const deleteStackOutput = await deleteStack(sn)
+        if (deleteStackOutput instanceof Error) {
+            console.log(deleteStackOutput)
```
macrael, line 32: Should we be returning a non-zero exit code if something fails here?

scripts/destroy_stage.ts
```diff
@@ -0,0 +1,234 @@
...
+            `Could not get buckets in stack ${stackName}: ${buckets}`
+        )
+    }
+
+    // clean out each of those buckets
+    buckets.map(async (bucket) => {
+        // turn off versioning on the bucket
+        const versionResponse = await turnOffVersioningOnBucket(bucket)
+        if (versionResponse instanceof Error) {
+            return versionResponse
```
macrael, line 58: I think these errors inside of buckets.map are not getting propagated up, just stored in the unused result of .map(), we probably want to surface an error eventually, (though we probably still want to try and keep destroying buckets if just one fails, too)

scripts/destroy_stage.ts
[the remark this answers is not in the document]
macrael: what is returned when there is an error?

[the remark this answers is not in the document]
macrael: Oh, seeing this down below, maybe we want something like this here too?
```
if (stack.$response.error != null) {
         return new Error(
             'Error on describeStacks: ' + stack.$response.error.message
         )
     }
```

macrael (approved):
YES, kill bash

scripts/destroy_stage.ts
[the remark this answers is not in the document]
mojotalantikite: oh, yup, I can add a check there too -- thanks!

scripts/destroy_stage.ts
mojotalantikite: For sure, I'll add that.

scripts/destroy_stage.ts
mojotalantikite: Oh yup, forgot I was in a map here when sketching it. Good catch, I'll fix this.

PR #766, opened 2022-01-21
Add aria-describedby to submission description

joshmfranklin:
Add `aria-describedby` to submission description so that SR's will read the help text associated with the textarea.

I also made a copy change to the hint text based on the feedback and iteration we had after our Nov. PP test.


rswerve:
Looks good.  Works with VoiceOver.

rswerve (approved):

PR #767, opened 2022-01-21
restyle file input

joshmfranklin:
Update to file input styling. Lots of text stacked on top of the file input component, trying to even it out for less visual noise.

Before:
Screen Shot 2022-01-21 at 12 31 47 PM
Screen Shot 2022-01-21 at 12 31 35 PM

After:
Screen Shot 2022-01-21 at 12 30 54 PM
Screen Shot 2022-01-21 at 12 30 44 PM

**Extra fun**: I also made a small tweak to the step indicator styling so that it expanded correctly depending on browser size.


rswerve:
This looks good to me, but I'll leave the final approval to @haworku, who knows more about the app's styling than I do.

haworku (approved):
💅

haworku:
i think test failure is just a bad merge Screen Shot 2022-01-24 at 11 09 54 AM

PR #768, opened 2022-01-21
Show supporting documents in a table

rswerve:
## Summary

Closes [10539](https://qmacbis.atlassian.net/browse/OY2-10539)

Supporting documents (not rate or contract documents) will be displayed in a table (rather than a list).  Figma designs are linked in the associated ticket.

#### Screenshots

image

#### Test cases covered

Refactored some tests to look for rows, rather than the list test-id.

## QA guidance

Uploads and upload errors should work properly in the supporting docs table.


joshmfranklin (commented):
Looks good @rswerve ! the only thing I see is that the error row styling is a bit off. the background color just needs to match the bg color of the error rows on the other file input. I believe the class that gives it the proper bg color is `bg-secondary-lighter`

tests/cypress/integration/stateWorkflow/stateSubmissionForm/documents.spec.ts
```diff
@@ -18,7 +18,9 @@ describe('documents', () => {
                 'documents/trussel-guide.pdf'
             )
             cy.findByText(/0 complete, 1 error, 1 pending/).should('exist')
-            cy.waitForDocumentsToLoad()
```
macrael, line 21: why did you drop .waitForDocumentsToLoad() in favor of a wait()?

services/app-web/src/components/FileUpload/FileItemList/FileItemsList.tsx
```diff
@@ -23,26 +26,51 @@ export const FileItemsList = ({
...
-                    <FileItem
-                        deleteItem={deleteItem}
-                        retryItem={retryItem}
-                        item={item}
-                    />
-                </li>
-            ))}
-        </ul>
-    )
+    if (renderMode === 'table') {
```
macrael, line 29: I think that it might be cleaner here to use some composition instead of just a switch. I’m imagining new very small view-only components:

<FileItemsTableView>
<FileItemsListView>
<FileItemsTableRowView>
<FileItemsListRowView>

or something. And then you can pass the component in to <FileItemsList> ? 

If you want to pair on this and talk it through we could try and figure something out. I think we definitely still want to keep all the logic that is in FileItem.tsx solo there, so I don’t think it’s good to duplicate everything, but I think that we should be able to separate out these two views into their own things.

macrael: I do wonder what it’ll look like once we have the checkboxes in the table view. then we’ll have controls in one that aren’t in the other.

macrael (approved):
This looks great. LMK if you want to chat about refactoring the views out

joshmfranklin:
> > Looks good @rswerve ! the only thing I see is that the error row styling is a bit off. the background color just needs to match the bg color of the error rows on the other file input. I believe the class that gives it the proper bg color is `bg-secondary-lighter`
> 
> 👍 updated!

hmm looks like it is using the blue background instead of the red error background
Screen Shot 2022-01-21 at 3 54 12 PM

rswerve:
> > > Looks good @rswerve ! the only thing I see is that the error row styling is a bit off. the background color just needs to match the bg color of the error rows on the other file input. I believe the class that gives it the proper bg color is `bg-secondary-lighter`
> > 
> > 
> > 👍 updated!
> 
> hmm looks like it is using the blue background instead of the red error background Screen Shot 2022-01-21 at 3 54 12 PM

Oh weird, I see that too, now that I rebuilt locally.  Let me find the right color.  😆

tests/cypress/integration/stateWorkflow/stateSubmissionForm/documents.spec.ts
rswerve: The table, unlike the list, doesn't show an image next to the completed upload.  So, while loading, there's a spinner, and then, when loading is finished, nothing.  waitForDocumentsToLoad relies on the list behavior: check the image, see if it's loading.  In the table, you can't check the class on the image, because the image might not be there.  And you can't do a simple check that the image is _not_ there, because that will always be true before the page has fully rendered.  So I wait for the page to render, then make sure the spinner isn't still there.

services/app-web/src/components/FileUpload/FileItemList/FileItemsList.tsx
rswerve: Yeah, I didn't love this, and in one of the many commits attached to this PR, there's a time when I had this broken out into separate components (fileItem and fileRow), but I ultimately undid that and opted for keeping the fileItem logic in one place.  I was torn about it, and am happy to look again to see if the components can be separated without maintaining the upload logic in two places.

services/app-web/src/components/FileUpload/FileItem/FileItem.tsx
```diff
@@ -120,45 +122,127 @@ export const FileItem = ({
...
+                        }}
+                    >
+                        <DocumentError
+                            hasDuplicateNameError={hasDuplicateNameError}
+                            hasScanningError={hasScanningError}
+                            hasUploadError={hasUploadError}
+                            hasUnexpectedError={hasUnexpectedError}
+                        />
+                        <>
+                            {(isLoading || isScanning) && (
```
haworku, line 170: I would like to see some further refinement here to share the conditional logic between the two UI states more. 

I have a bunch of ideas I could share if you want to brainstorm together around this - but at a minimum want to see some kind of shared logic. Even if its just making variables for the conditional logic to reduce all the` ||` and `&& `in JSX.

services/app-web/src/components/FileUpload/FileItem/FileItem.tsx
```diff
@@ -120,45 +122,127 @@ export const FileItem = ({
...
+                        <img
+                            id={item.id}
+                            data-testid="file-input-preview-image"
+                            src={SPACER_GIF}
+                            alt=""
+                            className={imageClasses}
+                        />
+                    </div>
+                    <span
+                        style={{
```
haworku, line 157: Hey could you pull this out in the scss module and using classes? This might have been me, its a bad habit I have from rushing during MVP stage. We def don't want inline CSS going on.

services/app-web/src/components/FileUpload/FileItemList/FileItemsList.tsx
haworku: My take- the checkboxes will likely go inside the `FileItem` component since that is where the table rows and `<td>` are.  We need to contain form elements in a table inside a cell - that's an accessibility requirement.

I feel like that component (`FileItem`)  is a bit harder to read and has more essential logic than this one. `FileItemsList` is more of a  presentation component.

services/app-web/src/components/FileUpload/FileItem/FileItem.test.tsx
```diff
@@ -64,22 +64,44 @@ describe('FileItem component', () => {
 
     beforeEach(() => jest.clearAllMocks())
     it('renders without errors', () => {
```
haworku, line 66: The way the code is written right now, with the duplication in FileItem, I would want to double all the tests because we have double the conditionals.  Test the exact same thing `renderMode=table` because right now the code could easily diverge.  🤷  or just refactor

macrael (commented):
This looks great. I think it woud be nice to get some naming consistency for the FileItem/FileListItem/FileProcessor so we dont’ have things named different than the file they are in. 

Right now it looks like FileItemProps and FileRowProps are identical, could those be shared?

services/app-web/src/components/FileUpload/FileItem/FileItem.tsx
```diff
@@ -1,141 +1,42 @@
...
+}: FileListItemProps): React.ReactElement => {
+    const { name } = item
     return (
         <>
             <div className={styles.fileItemText}>
                 <div
                     role="progressbar"
                     aria-valuetext={statusValue}
-                    aria-label={`File status`}
+                    aria-label={`Status of file ${name}`}
```
haworku, line 39: I don't think this change is needed. JAWS will already announce the the file name immediately after the file status: loading because its written on the screen in `<span>{name}</span>`. We are trying not to over use `aria-label` - in this case it will re-announce something  (the file names, and file names are often very lengthy in our use case) that is already being announced.

The only reason we had to change things for the Remove and Retry buttons was because in "button" mode in JAWS it skips over anything that is not a button and it wasn't clear what button went with what file in button mode

services/app-web/src/components/FileUpload/FileItem/FileItem.tsx
```diff
@@ -1,141 +1,42 @@
...
-    hasUploadError: boolean
-    hasUnexpectedError: boolean
-}): React.ReactElement | null => {
-    if (hasDuplicateNameError)
-        return (
-            <>
-                <span className={styles.fileItemBoldMessage}>
-                    Duplicate file
-                </span>
-                <span className={styles.fileItemBoldMessage}>
```
haworku, line 42: Small stylistic things. Not blocking but we can make  follow on tickets to address this sprint

1. We missing conditional logic in documents error to adjust the UI for the the file list versus the table. It looks like in the list, we displayed error text in two spans, whereas in the table we display with a comma between. There could probably be a helper function that uses React.createElement to properly take error text and display it the appropriate way for the renderMode type

2. @joshmfranklin what did we decide about the difference colors of red are we using the same red in the table/list as the alert or no? I think the divergence has been there for awhile
Screen Shot 2022-01-27 at 9 46 08 AM

services/app-web/src/components/FileUpload/TableWrapper/TableWrapper.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const TableWrapper = ({
+    fileItems,
+    deleteItem,
+    retryItem,
+}: TableWrapperProps): React.ReactElement => {
+    return (
+        <Table fullWidth>
+            <thead>
+                <tr>
+                    <th>Document name</th>
```
haworku, line 20: Small stylistic thing. Not blocking. Do we want an empty state here?
Screen Shot 2022-01-27 at 9 41 07 AM

haworku (approved):
LGTM to me.  Thanks for adding more tests.  

We do can do polish work as follow on PRs (if you don't end up handling it yourself) just make tickets for us please so we don't forget.

services/app-web/src/components/FileUpload/FileItem/FileItem.tsx
joshmfranklin: @haworku good catch on the difference in error background color. I don't think I'd noticed them until put the error summary on the page! Lets consolidate them and use a single bg color. Lets match the background of the error message which is `$theme-color-error-lighter` from USWDS and uses the hex of ##f4e3db.

services/app-web/src/components/FileUpload/TableWrapper/TableWrapper.tsx
joshmfranklin: @haworku yes I think we have an empty state design here https://www.figma.com/file/a0PPSr65q1G798zV4bhzzO/State-submission-form?node-id=6432%3A59072

rswerve:
Fixed the renaming issues and merged.  Thanks all for the thorough review.

services/app-web/src/components/FileUpload/FileItem/FileItem.tsx
rswerve: I made these changes after I merged in main and had failing tests; so I think they came from one of the accessibility tickets?  We can revert in a future PR if it's a problem.

PR #769, opened 2022-01-21
Move ./scripts sourcing to .envrc.local

mojotalantikite:
## Summary
This moves the sourcing of our `./scripts` directory for `ctkey-wrapper` out of our regular `.envrc`  setup. This sourcing was causing issues for some team members that weren't wanting or needing access to AWS via CloudTamer.

Instead devs that wish to use it should source it in their `.envrc.local`. I've updated the docs.

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-15319

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -7,9 +7,6 @@
 # exporting them in .envrc.local         #
 ##########################################
 
-# Add local paths for binaries and scripts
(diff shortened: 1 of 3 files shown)


haworku (approved):
Yay docs much clearer too!

PR #772, opened 2022-01-22
Use pa11y to check app in CI

macrael:
## Summary

We want to be running pa11y in CI. Since we are already using cypress to login and setup our tests, we can use cypress-audit to be where we run our pa11y checks as well. 

Cypress Audit is a Cypress plugin that gives us a cy.pa11y() command that runs pa11y checks on the current page. This is a boon because it means that we can login and setup the form using Cypress, and then just use pa11y to run accessibility tests on that configured page. 

when you call cy.pa11y, it opens a new browser tab in the same cypress browser and loads the page fresh. As far as I can tell it's not possible to perform the check directly on the existing cypress browser tab, even though that would be more ideal.

The current app has dozens of errors reported in pa11y, so for this PR I've added cy.pa11y() calls with a ratchet. They ignore the current number of errors, but will cause CI errors if any new errors are added to the project. I've filed a number of JIRA tickets to track fixing the current issues. 

## Question

Right now I made minimal changes to get this running. the pa11y function takes an "action" parameter that allows you to specify an element to wait on before running. This is necessary because otherwise pa11y will run while the page is still loading form data. We don't actually have a lot of elements with ids for us to wait on so I'm waiting on some kind of random ones. Should we just modify our pages to have more stable ids for this purpose? 

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-15636


rswerve:
I feel for you having to count existing errors 😆.  This looks good, and I think we'll add ids where we need them...

rswerve (approved):

docs/ADRs/015-use-cypress-audit-for-pa11y.md
```diff
@@ -0,0 +1,36 @@
+# 015 — Use Cypress Audit to run pa11y tests
+
+This is an update to [ADR 005 — Define Frontend Toolset for Accessibility Standards](./005-frontend-a11y-toolset.md). When that ADR was written we were still using Nightwatch and not Cypress. Now that we are using Cypress, it makes sense to use [Cypress Audit](https://github.com/mfrachet/cypress-audit) as the tool for running pa11y tests against our app.
```
haworku, line 3: awesome! Can you remove the dependencies for pa11y-ci, the `test:a11y` command, and pa11y-ci config file and replace the readme documentation for `Technologies.md` around pa11y-ci

haworku:
> We don't actually have a lot of elements with ids for us to wait on so I'm waiting on some kind of random ones. Should we just modify our pages to have more stable ids for this purpose?

Sure if you want. It is a common pattern. personally, I like putting the `id` on the main page container (rather than on headers like I think I see ya doing - the headers are more subject to change/deletion/replacement whereas the container elements often never change even if the entire page is refactored).That's where the `data-testid="dashboard-page` is for dashboard page already

haworku (approved):
⭐

PR #776, opened 2022-01-24
Update CMS submission email to include key dates

haworku:
## Summary
Update CMS submission email to include key dates

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-15321

#### Test cases covered
- I added `templates.test.ts` which includes several new unit tests that cover all of the current logic in our email templates (including new dates work).
-  Simplified `submitDraftSubmission.test.ts` accordingly.
- Added `emailerHelpers.ts`
- Exported the `defaultContext` used in our gql handlers for tests - that way the tests can determine who is the "current user"

## QA guidance
Make a submission, then check shared email to ensure CMS email includes new dates fields

### DIFF
--- a/services/app-api/emailer/templates.test.ts
+++ b/services/app-api/emailer/templates.test.ts
@@ -56,8 +56,8 @@ describe('Email templates', () => {
         it('includes expected data summary for a contract only submission', () => { 
             const sub: StateSubmissionType  = {
                 ...mockContractOnlySubmission(),
-                contractDateStart: new Date('01-01-2021'),
-                contractDateEnd: new Date('01-01-2025')
+                contractDateStart: new Date('01/01/2021'),
+                contractDateEnd: new Date('01/01/2025')
             }
             const template = newPackageCMSEmailTemplate(
                 sub,
@@ -81,7 +81,7 @@ describe('Email templates', () => {
                expect(template).toEqual(
                    expect.objectContaining({
                        bodyText: expect.stringContaining(
-                           'Contract effective dates: 2021-01-01 to 2025-01-01'
+                           'Contract effective dates: 01/01/2021 to 01/01/2025'
                        ),
                    })
                )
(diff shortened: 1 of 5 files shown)


services/app-api/emailer/templates.test.ts
```diff
@@ -0,0 +1,285 @@
...
+                     ),
+                 })
+             )
+
+        })
+
+
+        it('includes expected data summary for a contract only submission', () => { 
+            const sub: StateSubmissionType  = {
+                ...mockContractOnlySubmission(),
```
macrael, line 58: oh I like this pattern!

services/app-api/resolvers/submitDraftSubmission.test.ts
```diff
@@ -252,59 +210,55 @@ describe('submitDraftSubmission', () => {
...
+        const config = testEmailConfig
+        const mockEmailer = testEmailer(config)
+        //mock invoke email submit lambda
+        const server = await constructTestPostgresServer({
+            emailer: mockEmailer,
+        })
+        const draft = await createAndUpdateTestDraftSubmission(server, {})
+        const draftID = draft.id
+
+        await new Promise((resolve) => setTimeout(resolve, 2000)) // TODO: why is this here in other tests??
```
macrael, line 226: Does the test fail without this? Seems pretty unnecessary

services/app-api/testHelpers/emailerHelpers.ts
```diff
@@ -0,0 +1,212 @@
...
+   const mockUser = (): CognitoStateUserType  => {
+        return {
+            role: 'STATE_USER',
+            email: 'test+state+user@example.com',
+            name: 'Test State User',
+            state_code: 'MN',
+            }
+   }
+   
+   const mockContractAndRatesSubmission = (
```
macrael, line 52: I bet at some point we'll want to break out these mock submissions somewhere else, but I think it makes sense for them to be here for now since they are only used in email tests rn.

macrael (approved):
This is _really_ nice testing work, thank you for doing this. For some reason I'm no longer happy with us calling those methods "newPackageCMSEmailTemplate()" etc. b/c it seems like what it returns is not a template but a filled out template, i.e. an actual email. It's probably fine to leave as is but it's started bugging me for some reason.

haworku:
@macrael thanks for the review! I agree we can rename the newNAME_OF_EMAILtemplate files. Will merge this for now just to move on but we have more email work coming up soon with unlock/lock.

PR #783, opened 2022-01-31
Add categorization checkboxes to the document summary table

rswerve:
## Summary

Closes [5476](https://qmacbis.atlassian.net/browse/OY2-5476)

Add checkboxes to the supporting document summary table, so that the user can indicate whether a given document is rate- or contract-supporting.  This is the UI piece; @haworku already wired it up on the backend.

#### Screenshots

image

#### Test cases covered

Added unit tests to make sure the right events were firing, and a cypress test to confirm that the checkboxes are working as intended.

Also tried it out with NVDA (AssistivLabs) and VoiceOver.


joshmfranklin (commented):
Looks like the checkboxes still show up if the user is on a contract only submission. A contract only submission shouldn't show the checkboxes and all documents should be categorized as contract only . 

When there are file errors I don't think we want to show the checkboxes since you shouldn't be able to categorize files that have errors.

rswerve:
> Looks like the checkboxes still show up if the user is on a contract only submission. A contract only submission shouldn't show the checkboxes and all documents should be categorized as contract only .
> 
> When there are file errors I don't think we want to show the checkboxes since you shouldn't be able to categorize files that have errors.

Oh, good call.  Let me fix those issues.

rswerve:
@joshmfranklin One more thing: the AC doesn't mention this, but in Figma, I see that we want to require that at least one checkbox be selected.  Is that a real-time check or just after the user tries to continue?  (And there's no corresponding error message above the table?)

macrael:
@rswerve That’s a different ticket rn https://qmacbis.atlassian.net/browse/OY2-10582

rswerve:
> @rswerve That’s a different ticket rn https://qmacbis.atlassian.net/browse/OY2-10582

Oh, great, thank you.

services/app-web/src/components/FileUpload/FileProcessor/FileProcessor.tsx
```diff
@@ -4,6 +4,7 @@ import { FileRow } from '../FileRow/FileRow'
 import { FileListItem } from '../FileListItem/FileListItem'
 
 import styles from '../FileUpload.module.scss'
+import { DocumentCategoryType } from '../../../common-code/domain-models'
```
macrael, line 7: Technically this might actually want to be imported from the generated GraphQL code instead of domain models, one of the reasons I'm excited for us to axe the GraphQL version of the DraftSubmission someday. It's fine as is I mostly just point this out in case at some point this breaks and knowing this might help.

macrael (approved):
Looks great and works in the review app!

rswerve:
Ok, updated!  I was banging my head against how to test that docs were being marked properly when there were no checkboxes, and then realized that this will be easy to test once we do the ticket for adding this info to the summary page.

PR #787, opened 2022-02-01
Import programs data and script

jim:
## Summary

This PR adds a more complete dataset of programs. A few things of note:

* The script is JavaScript instead of TypeScript because I couldn't find a neat way to run TS files without compiling them first.
* I had to add a state-prefix to the program's slugs/ids to prevent collisions. Without this, the script aborts early due to `id` collisions:
    ```console
    $ node scripts/import-programs.js
    Error processing row:
    
    {"State":"GA","Program":"Non-Emergency Medical Transportation","Nickname":"NEMT","Entity type (e.g., MCO, PIHP, PAHP, PCCM)":"","Population Served (Medicaid & CHIP, Medicaid only,CHIP only)":"","Federal Operating Authority":"","Please put your initials in this column, once the task is complete ":"TP"}
    
    ID 'non-emergency-medical-transportation' is already in use! The state containing the already existing ID is:
    
    {"name":"Washington, DC","programs":[{"id":"non-emergency-medical-transportation","fullName":"Non-Emergency Medical Transportation","name":"NEMT"},{"id":"dc-healthy-families-program","fullName":"DC Healthy Families Program","name":"DCHFP"},{"id":"child-and-adolescent-supplemental-security-income-program","fullName":"Child and Adolescent Supplemental Security Income Program","name":"CASSIP"}],"code":"DC"}
    ```
    
    The collision detection code is still in place, so if the script is run with different data in the future it should hopefully prevent duplicate ids from being encountered later in the pipeline.

* Program names are included in case they prove to be useful in the future.
* Each program has a v5 UUID generated from its `id`. They should be reasonable stable, but any change in the program's name could affect its `id` and therefor its `uuid`.

I know that exactly how the program identifier stuff works out is a little bit unknown, so hopefully this will provide a bunch of options.

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-12124

#### Screenshots

image


services/app-api/scripts/import-programs.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+function generateUUID(id) {
+    return uuidv5(id, uuidNamespace);
+}
+
+const states = {};
+const programIDs = [];
+
+fs.createReadStream('State programs, population, and nicknames.xlsx - Sheet1.csv')
```
macrael, line 82: I think it would be nice to have this as a command line argument instead of hard coded

services/app-api/scripts/import-programs.js
```diff
@@ -0,0 +1,123 @@
+const csv = require('csv-parser');
+const fs = require('fs');
+const slugify = require("slugify");
+const { v5: uuidv5 } = require('uuid');
+
+const stateNames = {
```
macrael, line 6: I’m going to add American Samoa to our google sheet b/c we use it for testing, AS will need to be added here.

macrael (approved):
This is great. 

I might move this to our top level /scripts directory where we already have a couple typescript scripts compiling. I know you need to start gearing up for your next project, these changes are small so I’m happy to take care of them if you are ready to be done.

jim:
@macrael I can make those updates today!

PR #795, opened 2022-02-08
Add revisions table

haworku:
## Summary
Add “revisions” table to postgres

## Related Issue
https://qmacbis.atlassian.net/browse/OY2-14193

## QA guidance
- All app-api tests pass unchanged
- Revisions table has the proper information in Postico
- Revisions table has proper relationship with StateSubmission table (used SQL query to confirm)

### DIFF
--- a/services/app-api/postgres/submissionWithRevisionsHelpers.ts
+++ b/services/app-api/postgres/submissionWithRevisionsHelpers.ts
@@ -1,0 +1,34 @@
+import { StateSubmission, StateSubmissionRevision } from '@prisma/client'
+import { StoreError } from './storeError'
+
+export type StateSubmissionWithRevisions = StateSubmission & {
+    revisions: StateSubmissionRevision[]
+}
+
+// Return first revision associated with a submission or return a StoreError if there is submission or revisions
+// used validate prisma results have useable submission
+const getCurrentRevision = (
+    submissionID: string,
+    submissionResult: StateSubmissionWithRevisions | null
+): StateSubmissionRevision | StoreError => {
+    if (!submissionResult)
+        return {
+            code: 'UNEXPECTED_EXCEPTION' as const,
+            message: `No submission found for id: ${submissionID}`,
+        }
+
+    if (!submissionResult.revisions || submissionResult.revisions.length < 1)
+        return {
+            code: 'UNEXPECTED_EXCEPTION' as const,
+            message: `No revisions found for submission id: ${submissionID}`,
+        }
+
+    // TODO FIGURE OUT HOW TO ENSURE PROPERLY ORDERED REVISIONS HERE
+    return submissionResult.revisions[0]
+}
+
+
+
+export {
+    getCurrentRevision
+}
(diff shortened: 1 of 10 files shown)


services/app-api/prisma/migrations/20220208221835_add_revisions_table/migration.sql
```diff
@@ -0,0 +1,19 @@
...
+    "id" TEXT NOT NULL,
+    "createdAt" TIMESTAMP(3) NOT NULL,
+    "submissionID" TEXT NOT NULL,
+    "submissionFormProto" BYTEA NOT NULL,
+
+    CONSTRAINT "StateSubmissionRevision_pkey" PRIMARY KEY ("id")
+);
+
+-- AddForeignKey
+ALTER TABLE "StateSubmissionRevision" ADD CONSTRAINT "StateSubmissionRevision_submissionID_fkey" FOREIGN KEY ("submissionID") REFERENCES "StateSubmission"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
```
haworku, line 14: what the best way to verify the relationship between the StateSubmission and StateSubmissionRevision is set up properly? I'm guessing by making a SQL query so trying that

macrael:
We need to change our postgres package to be aware of the revisions table. So, for instance, the insertDraftSubmission endpoint should create a new draft and also a new revision, findDraftSubmission should return the most recent revision for the submission with that ID, etc.

services/app-api/prisma/migrations/20220208221835_add_revisions_table/migration.sql
```diff
@@ -0,0 +1,19 @@
...
+
+-- CreateRevisionForEachExistingSubmission
+INSERT INTO "StateSubmissionRevision" 
+SELECT uuid_generate_v4() AS "id", now() AS "createdAt", "id" AS "submissionID","submissionFormProto"
+FROM "StateSubmission";
```
macrael, line 19: Once we’ve done some of the API work, we should drop the submissionFormProto column from the old StateSubmission table.

haworku:
@macrael I thought that was the next PR. This is only the first ticket. Are you saying its a bad idea to merge this without starting the api work ticket as well to test it? I could see that hehe.

macrael:
I think the next ticket has actual changing functionality. Adding the “unlock” endpoint. This ticket is about changing the DB and the current APIs to use the new table, all invisibly to the API.

services/app-api/postgres/submissionWithRevisionsHelpers.ts
```diff
@@ -0,0 +1,34 @@
...
+
+// Return first revision associated with a submission or return a StoreError if there is submission or revisions
+// used validate prisma results have useable submission
+const getCurrentRevision = (
+    submissionID: string,
+    submissionResult: StateSubmissionWithRevisions | null
+): StateSubmissionRevision | StoreError => {
+    if (!submissionResult)
+        return {
+            code: 'UNEXPECTED_EXCEPTION' as const,
```
macrael, line 16: Ha, this is kinda funny to me, like I’d expect checking that you got a StateSubmission happen before calling getCurrentRevision, but I think this is fine.

macrael (approved):
This looks great! I’d like us to make a transaction or use raw SQL for update, but if you’d rather me try that in a follow on PR that’s fine by me.

PR #798, opened 2022-02-09
Add OTEL Lambda Layer for JS

mojotalantikite:
## Summary

This adds a lambda layer to all of our functions in `app-api` to allow for instrumentation of our lambda functions. It configures the collector to export to our New Relic account.

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-15454


.github/workflows/deploy.yml
```diff
@@ -224,10 +224,11 @@ jobs:
       iam_path: ${{ secrets.IAM_PATH }}
       full_iam_permissions_boundary_policy: ${{ secrets.FULL_IAM_PERMISSIONS_BOUNDARY_POLICY }}
       react_app_auth_mode: AWS_COGNITO
+      nr_license_key: ${{ secrets.NR_LICENSE_KEY }}
 
   finishing-prep:
     name: Finishing Prep
-    needs: [deploy-infra, begin-deployment]
+    needs: [deploy-infra, begin-deployment, deploy-app]
```
mojotalantikite, line 231: I was hitting an issue where `Finishing Deploy` was running before unit tests and `app-api` completed deploying, as the infra had been previously deployed and skipped. It was then adding the cypress users, which seems like it was messing up the migration task. Adding this here to wait on app deployment before `Finishing Deploy` begins.

services/app-api/serverless.yml
```diff
@@ -61,6 +61,11 @@ provider:
   apiGateway:
     restApiId: ${self:custom.appApiGatewayId}
     restApiRootResourceId: ${self:custom.appApiGatewayRootResourceId}
+  layers:
+    - arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-nodejs-amd64-ver-1-0-1:1
+  tracing:
```
macrael, line 66: is this Amazon X Ray? I saw that referenced a bunch but don't understand what it is.

mojotalantikite: So yes and no -- it's Amazon X-Ray, which is their own tracing platform, but we're not using their X-Ray backend to send the data to. In order to get the auto-instrumentation we turn on active tracing on the lambda, but we configure it to be sent to NR via the OTEL collector. We shouldn't be getting any tracing into the X-Ray managed service itself.

PR #799, opened 2022-02-11
Error functionality for supporting document file upload

rswerve:
## Summary

Closes [10582](https://qmacbis.atlassian.net/browse/OY2-10582)

The major changes here are  
1) making error handling consistent with the file upload _list_; specifically, when errors are triggered and when they're cleared 
2) making the error messages consistent with other spots in the app, in that they now link to the offending row
3) adding a check in the api to disallow saving submissions with missing document categories

Also added the snippet of code to make `upload_error.pdf` fail locally, so we can simulate upload errors. 

#### Screenshots

image

#### Test cases covered

Added tests for the api check and the presence of links in the messages in the error list, and updated a few to work with the new functionality.

## QA guidance

There's a lot of fiddling with logic in this PR, so errors appearing/disappearing at times that don't seem right is what's most likely to be off.  The existing tests were super helpful, so 🤞.


services/app-web/src/pages/StateSubmission/Documents/Documents.tsx
```diff
@@ -44,20 +44,48 @@ export const Documents = ({
...
+                fileItems.some(
+                    (docs) =>
+                        !docs.documentCategories.includes('CONTRACT_RELATED')
+                )))
     const hasLoadingFiles =
         fileItems.some((item) => item.status === 'PENDING') ||
         fileItems.some((item) => item.status === 'SCANNING')
     const showFileUploadError = shouldValidate && !hasValidFiles
 
+    const documentsErrorMessages = () => {
```
macrael, line 62: I think we’re missing checking “showFileUploadError” or “shouldValidate” somwehere here. I was seeing the “You must select a document category” error message as soon as I added a document to the list. I expect to only see that error message if I try and click Continue before filling everything out correctly. 

It’s being checked in generating the errorSummary below, which I think I’m fine with being separate from this list, but it might be worth while collecting all those errors before we get to the return and checking if we should display them in one place.

macrael (approved):
Looks great. I saw the one bug where I was getting error messages before I’d clicked “Continue” but otherwise this is really solid. You got the API validation stuff just right, yay

haworku:
Are the errors between table and list supposed to be the exact same UI? I'm seeing difference between the two for the exact same files.

**Contract Details page:** 
Screen Shot 2022-02-14 at 8 45 29 AM

**Supporting documents page:**
Screen Shot 2022-02-14 at 8 53 13 AM

haworku (commented):
The work around adding document categories errors looks good to me except for the comment MacRae mentioned (we are showing the error summary too early - when the form page is not in validation state yet). I really like the backend changes too.

The work around changing how error summary works across (show errors specific to the line and also with a specific error) I think needs a bit more attention. Maybe we could take those changes out of this PR to get this through?  Specifically there are changes on the `Documents.tsx `that will need to also be added to Contract Details and Rate Details to unify how error handling works on all the pages. The same error on different pages should display the same .

services/app-web/src/pages/StateSubmission/Documents/Documents.tsx
```diff
@@ -44,20 +44,48 @@ export const Documents = ({
...
+                errorsObject[key] =
+                    'You must remove files that failed the security scan'
+            } else if (item.status === 'UPLOAD_ERROR') {
+                errorsObject[key] =
+                    'You must remove or retry files that failed to upload'
+            } else if (item.documentCategories.length === 0) {
+                errorsObject[key] = 'You must select a document category'
+            }
+        })
+        return errorsObject
```
haworku, line 78: I think this might be the place that is causing the differences seen here https://github.com/CMSgov/managed-care-review/pull/799#issuecomment-1039182993

services/app-web/src/pages/StateSubmission/Documents/Documents.test.tsx
```diff
@@ -625,12 +629,15 @@ describe('Documents', () => {
...
-                    )
-                ).toHaveLength(2)
+                const errorMessage = screen.getByText(
+                    'You must remove duplicate files'
+                )
+                const errorSummary = screen.getByText(
+                    'You must remove all documents with error messages before continuing'
+                )
+                expect(errorMessage).toBeInTheDocument()
+                expect(errorMessage).toHaveAttribute('href')
```
haworku, line 639: nice thanks for adding these to the tests 🥇

services/app-web/src/components/FileUpload/FileUpload.tsx
```diff
@@ -33,6 +33,8 @@ export type FileUploadProps = {
     deleteFile: (key: string) => Promise<void>
     onFileItemsUpdate: ({ fileItems }: { fileItems: FileItemT[] }) => void
     isContractOnly?: boolean
+    shouldValidate?: boolean
```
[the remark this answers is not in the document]
rswerve: This is one of the places I was thinking of when I said we need to consider our error handling.  Maybe React Context for error state?  We should chat.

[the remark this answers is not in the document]
macrael: I think that if we passed the file view component into the file uploads component that would go a long way to keeping the FileUpload component clean. 

I also think it’s a good point about “shouldValidate” not neccecarily belonging inside FileUpload. Let the Documents.tsx page manage when to display errors, if possible. the uploads componenet just reports the errors it sees upwards. Maybe with us having more detailed errors in the errors view, we can get rid of the error display that is inline in the documents component to simplify things?

I’m totally happy with us adding a refactor ticket, this is a tricky component and it could be fun for us all to collaberate on it.

services/app-web/src/pages/StateSubmission/Documents/Documents.tsx
rswerve: Yeah, missing a shouldValidate for actually showing the error summary.  🤦 

Can you say what you mean by "this list"?  Want to be sure I'm understanding you correctly.

macrael: we’ve got documentsErrorMessages and we’ve got errorSummary, those were the two lists I was talking about. Though errorSummary isn’t actually a list

services/app-web/src/components/FileUpload/FileUpload.tsx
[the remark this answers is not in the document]
rswerve: @haworku and I chatted about this (offline? online? in Zoom) and we're going to leave it as-is, because we both agree that a refactor of error handling across the various components that use fileUpload is necessary.

services/app-web/src/pages/StateSubmission/Documents/Documents.tsx
rswerve: Gotcha, thanks.

haworku (approved):

PR #806, opened 2022-02-14
Refactor DoubleColumnRow to display n children

JasonLin0991:
[OY2-10657](https://qmacbis.atlassian.net/browse/OY2-10657)

## Summary
Refactor `DoubleColumnRow` to accept `n` number of children and place them in double columns on multiple rows and changing component to `DoubleColumnRows`

#### Screenshots
The screens should still look the same.
Screen Shot 2022-02-14 at 1 38 52 PM

#### Test cases covered
Tests cover the function `pairedChildren` to correctly place pairs of children in an array and to render column and rows in the right order.

## QA guidance
Children passed into this component will render in left, right order in the array. Currently if you want to have a row display a single item on the right column, you need to pass in a react fragment `<></>`, let me know if there should be a better approach.


services/app-web/src/components/DoubleColumnRows/DoubleColumnRows.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const pairedChildren = (children: ChildrenType[]) => {
+    return children.reduce((pairedChildren: ChildrenPairType, child, index, array) => {
+        if (index % 2 === 0) {
+            pairedChildren.push(array.slice(index, index + 2))
+        }
+        return pairedChildren
+    }, [])
+}
+
+export const DoubleColumnRows = ({
```
macrael, line 21: I think we might be due for a rename here. Maybe DoubleColumnGrid?

services/app-web/src/components/DoubleColumnRows/DoubleColumnRows.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    })
+
+    it('Renders odd elements in correct row and column order with no errors', () => {
+        render(
+            <DoubleColumnRows>
+                {testDataOdd.map(child => child)}
+            </DoubleColumnRows>
+        )
+
+        expect(screen.getByTestId('grid-row-0')).toHaveTextContent('Row 1 Column 1Row 1 Column 2')
```
macrael, line 57: great tests! I think this is a fine use of getByTestId, if you wanted to just use getByText or something, you could probably pull one and check that it only has the one expected sibling with it, but I think that would be more finicky in the end.

services/app-web/src/components/DoubleColumnRow/DoubleColumnRow.tsx
```diff
@@ -1,19 +0,0 @@
-import { Grid } from '@trussworks/react-uswds'
```
macrael, line 1: Even though these components have changed a ton, I think it would be nice to have the git rename here instead of a delete and re-add

macrael (approved):
Looks great! ping me if you have any questions.

PR #807, opened 2022-02-14
Get FileUpload storybook working again and add table

haworku:
## Summary
Get storybook working again for file upload success case and add an example with render mode "table"

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-12519


rswerve:
Looks good.  Do we also want table failures?

haworku:
@rswerve yes good I think yes I'll add those

rswerve (approved):

PR #811, opened 2022-02-14
Fix global lint

haworku:
## Summary
@JasonLin0991 and @macrael found out that precommit lint is not working. I vaguely remember this being an issue, or else just lint taking really really long when we originally did the lerna work. @mojotalantikite  might remember more.

Here's a WIP branch with changes to get lint working on precommit again (although we may decide it takes to long and we only want to lint say on prepush).  Fixed the `yarn lint` in each package, now trying to get global precommit working without crashing on errors.


package.json
```diff
@@ -23,7 +23,7 @@
         "prettier": "prettier --ignore-path .gitignore \"**/*.+(ts|tsx|json)\"",
         "cypress:open": "cypress open --env AUTH_MODE=LOCAL",
         "cypress:run": "cypress run --env AUTH_MODE=LOCAL",
-        "precommit": "lint-staged"
+        "precommit": "lerna run --concurrency 1 --stream precommit --since HEAD --exclude-dependents"
```
haworku, line 26: This is... almost working? It works find if there are no lint errors. If there are actual lint errors though it crashes after finding the first error correctly. Crash looks like this

```
cripts: [FAILED] fatal: Unable to create '/Users/worku/Code/MCRRS/managed-care-review/.git/index.lock': File exists.
scripts: [FAILED]
scripts: [FAILED] Another git process seems to be running in this repository, e.g.
scripts: [FAILED] an editor opened by 'git commit'. Please make sure all processes
scripts: [FAILED] are terminated then try again. If it still fails, a git process
scripts: [FAILED] may have crashed in this repository earlier:
scripts: [FAILED] remove the file manually to continue.
scripts:   ✖ lint-staged failed due to a git error.
scripts: [STARTED] Cleaning up temporary files...
````

haworku:
@mojotalantikite  if you have any spare cycles this sprint could you take a look over here

mojotalantikite:
@haworku ok so I just pushed up some changes to get this working again. But you're right, we originally didn't enable it because of how long linting was taking. It seems much faster now than I remember it being though.

Anyway, I switched us over to just using `lint-staged` directly with yarn workspaces without leaning on `lerna`, as it seems like lerna causes it some issues. Want to test it out a bit and lmk how it looks?

haworku:
Just confirming I see linting and prettier (auto formatting) working again on both app-web and app-web. Also its properly running commands only on the services that have the changes in them. Thank you for mojo!

Its taking less time yes but still up to 20 seconds per commit I think I saw. @macrael @rswerve @JasonLin0991 tagging you all for visibility. If you want to see the precommit hook timing in action all you have to do is  introduce a linting error (for example, mispelling a method or import is an easy one) and then attempt to git commit it.

Personally I feel okay about this for now - we can see how painful it is. I did peek around the internetz and supposedly there could be a way to just run on [prepush](https://laurieontech.com/posts/husky/#pre-push) by diffing from main but not sure we want to go that hacky path.

haworku:
someone else will need to approve this eventually so it can merge.

macrael (approved):
Yay

haworku:
@mojotalantikite  I was hoping to merge this today since its the beginning of the new sprint and number of WIP PRs has gone a lot.

But getting this error after merging main and I'm confused where its coming from. Does this mean anything to you? Screen Shot 2022-03-18 at 8 53 23 AM

PR #814, opened 2022-02-15
Add Unlock submission functionality

macrael:
## Summary

This PR covers adding the ability for a CMS user to unlock a submission and no more. CMS users can unlock a submission from the submission summary page, which creates a new revision in the DB. A state user can then edit that new revision using the same old fetch and update DraftSubmission GQL operations. 

Doing this work has helped me develop a design for the next chunk of work. I don’t think that status info like who unlocked or resubmitted a submission belongs in the submission form data models, so I think we need a new graphql endpoint to get that info, and that work will be in a follow on PR. 

NOTE: the text field in the unlock confirmation dialog is not connected to anything yet.

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-12184

#### Screenshots

https://user-images.githubusercontent.com/55759/154122497-32bc5b71-4a40-436e-895c-2baf3c6b16bb.mov

#### Test cases covered
* added API tests for unlock and re-submit
* added FE tests for clicking the unlock button and getting success or errors

## QA guidance
* It’s a known issue that after clicking unlock there is no visual indication that anything worked and if you refresh the page it will error. We need to add some new API to allow CMS users to see unlocked submissions (in the draft state) 
* You should be able to log in as a state user and submit, then as a CMS user and unlock, then as a state user and submit again.


services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -46,6 +64,39 @@ export const SubmissionSummary = (): React.ReactElement => {
...
+            }
+
+            if (result.data?.unlockStateSubmission.draftSubmission) {
+                console.log('Submission Unlocked')
+                modalRef.current?.toggleModal(undefined, false)
+            } else {
+                console.error('ERROR: Got nothing back from unlock')
+                modalRef.current?.toggleModal(undefined, false)
+                setUserVisibleUnlockError('Error attempting to unlock. Please try again.')
+            }
```
haworku, line 90: I'm confused is this possible? I would hope unlock request would fail overall or return `.errors` if there is not a valid draft submission coming back. Can we push this error handling to backend

haworku (approved):
LGTM. Going to run through deployed app now but everything in code looks exactly as your described in the PR.

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
macrael: OK, I wrote a wrapper for this controller. It would make sense to do the same on submit, but I’m going to leave that for another day.

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -1,26 +1,70 @@
...
+            const result = await unlockStateSubmission({
+                variables: {
+                    input: {
+                        submissionID: id
+                    }
+                }
+            })
+
+            if (result.errors) {
+                return result.errors
```
haworku, line 44: Yes we need something like this. And  me this should log the errors and then either throw a `Error` or return a error type. `GraphQLError`  may be something we need looks like. 

An array is not a clear error type, especially on the frontend where often the result is an array of data we iterate over.

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -46,6 +92,22 @@ export const SubmissionSummary = (): React.ReactElement => {
 
     if (error || !submission) return <GenericError />
 
+    const onUnlock = async () => {
+        const result = await unlockMutationWrapper(unlockStateSubmission, submission.id)
+
+        if (result instanceof Error) {
+            console.error('ERROR: got an Apollo Client Error attempting to unlock', result)
+            setUserVisibleUnlockError('Error attempting to unlock. Please try again.')
+        } else if (Array.isArray(result)) {
```
haworku, line 101: `isGraphQLError (result)` would be nice here.

haworku (approved):

services/app-api/resolvers/unlockStateSubmission.test.ts
```diff
@@ -0,0 +1,196 @@
...
+        })
+
+        expect(unlockResult.errors).toBeDefined()
+        const err = (unlockResult.errors as GraphQLError[])[0]
+
+        expect(err.extensions['code']).toEqual('BAD_USER_INPUT')
+        expect(err.message).toEqual('A submission must exist to be unlocked: foo-bar')
+    })
+
+    it('returns an error if the DB errors', async () => {
```
haworku, line 159: Yeah I like these!

PR #816, opened 2022-02-16
Add empty state for supporting documents  in ReviewSubmit

haworku:
## Summary
Add an empty state that displays for supporting documents in ReviewSubmit and SubmissionSummary so that when there are no supporting contract or rate documents the state and cms user explicitly can see that.

#### Related issues
Follow on from https://qmacbis.atlassian.net/browse/OY2-10583 product and design review

#### Screenshots [UPDATED, thanks @joshmfranklin ]
Screen Shot 2022-02-16 at 10 07 38 AM

#### Test cases covered
No tests added just adjusted the tests for no supporting docs case

### How to verify
Navigate to review and submit page with no supporting documents for contract and rates


joshmfranklin:
@haworku looks good! I just did some quick restyling and copy update.

haworku:
@joshmfranklin just made a small update to copy as well (just in case through some weird turn of events someone ends up on ReviewSubmit without contract or rate documents). We will show "No documents" for that case "No supporting documents" for no supporting documents cases. Can you give a 👍 if things look good- I can merge!

services/app-web/src/components/SubmissionSummarySection/UploadedDocumentsTable/UploadedDocumentsTable.tsx
```diff
@@ -61,27 +83,23 @@ export const UploadedDocumentsTable = ({
 
         void refreshDocuments()
     }, [documents, getKey, getURL])
-    
+
+    // Empty State
+    if (refreshedDocs.length < 1) {
```
macrael, line 88: is `.length < 1` a common js-ism? It seems more clear to me to do `.length === 0`

macrael (approved):
Looks good!

services/app-web/src/components/SubmissionSummarySection/UploadedDocumentsTable/UploadedDocumentsTable.tsx
haworku: no its not. I can try to write things that way, agree its clear. If I think hard maybe its  cuz I have a tendency to just avoid `=== 0` cuz of typoed `== 0` being falsey? but with linting that doesn't happen.

PR #817, opened 2022-02-16
New Relic Infrastructure Monitoring

mojotalantikite:
## Summary

Adds the resources needed to enable a stream of infra data to NewRelic. This will allow us to monitor resource usage.

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-15459


.github/workflows/deploy.yml
```diff
@@ -229,7 +230,7 @@ jobs:
   finishing-prep:
     name: Finishing Prep
     needs: [deploy-infra, begin-deployment, deploy-app]
-    if: always()
+    if: always() && needs.deploy-infra.result == 'success' && needs.deploy-app.result == 'success'
     runs-on: ubuntu-latest
```
mojotalantikite, line 234: This isn't related to the feature, but I was running into issues where `finishing-prep` was running before other things and causing issues. I even saw the container hang for a good 30 minutes for a reason I'm still unsure of.

services/infra-api/serverless.yml
```diff
@@ -23,8 +27,32 @@ provider:
...
+      permissionsBoundary: ${self:custom.iamPermissionsBoundaryPolicy}
+      statements:
+        - Effect: 'Allow'
+          Action:
+            - '*'
+          Resource: !GetAtt FirehoseStreamToNewRelic.Arn
 
 resources:
+  Conditions:
+    CreateNRInfraMonitoring:
```
mojotalantikite, line 42: We only want this infra to exist once per environment. It doesn't make sense to have these roles in our PR branches since NR is just reading from the entire `dev` environment anyway.

.github/workflows/deploy.yml
mojotalantikite: It was also happening with Cypress, so I added some conditions there as well.

mojotalantikite:
Just wanted to call out that I put this in infra-api, but it doesn't have to live in that service. I don't think it makes sense to create a whole new service just for this though, as that will add to CI time and whatnot.

I think there is something to be said for having a shared infra place, but that's probably a separate ticket to consolidate those things in a single service. 

Open to moving this elsewhere if people have strong opinions though.

services/infra-api/serverless.yml
```diff
@@ -48,6 +76,140 @@ resources:
...
+
+    CloudWatchMetricStream:
+      Type: AWS::CloudWatch::MetricStream
+      Condition: CreateNRInfraMonitoring
+      Properties:
+        Name: ${self:custom.nrMetricStreamName}
+        FirehoseArn: !GetAtt FirehoseStreamToNewRelic.Arn
+        RoleArn: !GetAtt MetricStreamRole.Arn
+        OutputFormat: 'opentelemetry0.7'
+
```
mojotalantikite, line 212: CloudWatch lets us transform this data, but they have OpenTelemetry support built in which is what NewRelic will use.

PR #818, opened 2022-02-17
fix: Complete and error summary in FileUpload list view

haworku:
## Summary
Two bugs 
- error count is wrong
- complete count is also wrong

**6f2e54adfffde8c4713b46fa6405b28259278b52 is the actual bugfix**

Also made a change to rely just on one prop (`shouldDisplayMissingCategoriesError`) when FileUpload determines if/when categories are relevant.  The prop is just set to false in list view or when contract only submission.

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-16435

#### Test cases covered

- Add tests to the "list summary" describe block in `FileUpload.test.tsx` for error count, complete count, and pending count.
- Add tests for error state and checkboxes still appearing when expected on Documents page after I broke that (accidentally removed the checkboxes when they were supposed to be there in error state) but tests still passed.

## QA guidance

Place both file upload list and file upload table in the different error states and check that count is correct. Can use storybook for quickly viewing error states for failed scan and failed upload.


rswerve:
I'm curious if the cypress tests still passed when you broke checkboxes.

haworku:
@rswerve ooo good point probably we have cypress for this ❤️ . I was developing locally and wasnt checking cypress. I can check that

rswerve:
> @rswerve ooo good point probably we have cypress for this ❤️ . I was developing locally and wasnt checking cypress. I can check that 

I was just curious; it should have been caught, but if you have unit tests now, all the better.

rswerve:
🤔  this isn't behaving the way I expect.  Files with errors other than missing categories shouldn't display checkboxes.  This is what happens when I hit continue.  Note also that the first row, which has a checked box, is showing an error state.

image

haworku:
@rswerve Oo thanks. I'm on it, will add tests. I'm just gonna back out  my tiny refactor commit all together and stick to the bugfix.

haworku:
@rswerve lol yup! I really did tell myself I was renaming variables and not refactoring 😆 . But anyway good thing is I had isolated that in a commit so its just a revert

haworku:
Screen Shot 2022-02-17 at 12 36 11 PM

I think this issue is resolved now @rswerve - thanks for bringing it up

services/app-web/src/components/FileUpload/FileProcessor/FileProcessor.test.tsx
```diff
@@ -357,6 +357,12 @@ describe('FileProcessor component', () => {
                 {...categoryCheckboxProps}
             />
         )
+        expect(
```
rswerve, line 360: 👍

rswerve:
Looks great.

rswerve (approved):

PR #820, opened 2022-02-18
Handle user switching between contract/rates and contract-only

rswerve:
## Summary

@MegMurphy found a bug, detailed [in this ticket](https://qmacbis.atlassian.net/browse/OY2-10582?focusedCommentId=106447)

This one is tricky, because to fix the bug Meghan found, which was an actual showstopper, we'll be introducing a little strangeness: if the user

1) selects contract-only
2) uploads a supporting document
3) goes back to select contract and rates
4) returns to the supporting documents page

then they'll see the checkbox for `Contract related` already checked.  I can't think of a non-horrible way to avoid this, while fixing the original bug, but I'm all ears.

Would this state of affairs be acceptable @MegMurphy @joshmfranklin ?

### DIFF
--- a/services/app-web/src/components/FileUpload/FileUpload.tsx
+++ b/services/app-web/src/components/FileUpload/FileUpload.tsx
@@ -101,7 +101,7 @@ export const FileUpload = ({
         setFileItems((prevItems) => {
             const newItems = [...prevItems]
             return newItems.map((item) => {
-                if (item.documentCategories.length === 0 || isContractOnly) {
+                if (renderMode === 'table' && isContractOnly) {
                     return {
                         ...item,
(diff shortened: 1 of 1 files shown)


MegMurphy:
@rswerve I am cool with that personally. It isn't unreasonable to think when a user uploads their supporting docs originally  under the contract-only submission, that the docs they uploaded at the time were related to a contract submission anyway perhaps...and at least it's not a show stopper.

macrael:
I don't hate this. Imagine someone submitted a contract only submission and then it was unlocked and they added rates. We would already have shown these documents as contract supporting documents. I think it's right to go ahead and associate them that way.

macrael (approved):
This makes sense to me, great use for useEffect

PR #825, opened 2022-02-22
Remove double alerts on promote fail

mojotalantikite:
## Summary

We just got two alerts on a promote failure, and it looks like it's because in #777 we added the condition to alert on a cypress val or prod 'skipped' state. This will happen whenever cypress-dev fails, in addition to other cases of earlier failure.

This removes the redundant case of when cypress alerts on failure in dev or val.


macrael (approved):
cool

PR #827, opened 2022-02-24
Add fetchSubmission2 Query

macrael:
## Summary

This PR adds a new GraphQL Query: `fetchSubmission2` that returns a submission along with its revisions.

This is a big change! It's part of my overall vision of the database/protobuf refactor. For now, I've simplified the graphQL api to just return the protobuf directly for each revision instead of it having its own full version of the submission form type. This is something we can change, but I think it would be nice for the form data for us to be able to rely on the protobuf and the domain models alone, cutting out one type representation in our app. 

You can see how this is used in the submissionSummary page where we now are able to view the submission after it's been unlocked without crashing, and we make the unlock button disabled. 

### Things to consider
I would love some help with naming. I think `updateStateSubmission` in the postgres module is really terrible, and I named this fetchSubmission2 mostly so it would be easy to find and replace it everywhere when we deprecate the old API or rename things one

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-16430
https://qmacbis.atlassian.net/browse/OY2-12184 : This PR fixes most of the review issues

#### Test cases covered

Added a new cypress test that unlocks and resubmits, did a fair amount of backend testing. Still working on some FE stuff

## QA guidance

Def try and break this, it's a big change!


tests/cypress/integration/cmsWorkflow/unlockResubmit.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        cy.fillOutSupportingDocuments()
+        cy.navigateForm('Continue')
+        cy.findByRole('heading', {name: /Review and submit/}).should('exist')
+
+        // Store submission url for reference later
+        cy.location().then((fullUrl) => {
+
+            const reviewURL = fullUrl.toString()
+            fullUrl.pathname = path.dirname(fullUrl)
```
haworku, line 28: cypress has a helper for this. would prefer we use that

services/app-api/resolvers/configureResolvers.ts
```diff
@@ -22,6 +24,7 @@ export function configureResolvers(store: Store, emailer: Emailer): Resolvers {
             fetchCurrentUser: fetchCurrentUserResolver(),
             fetchDraftSubmission: fetchDraftSubmissionResolver(store),
             fetchStateSubmission: fetchStateSubmissionResolver(store),
+            fetchSubmission2: fetchSubmission2Resolver(store),
```
haworku, line 27: just checking my understanding - is the goal is that `fetchSubmission2`  (or whatever we name it) would replace `fetchDatchSubmission` and `fetchStateSubmission` both?

services/app-api/resolvers/fetchSubmission2.test.ts
```diff
@@ -0,0 +1,281 @@
...
+        expect(subData.contractDocuments).toEqual([
+            {
+                name: 'contractDocument.pdf',
+                s3URL: 'fakeS3URL',
+                documentCategories: ['CONTRACT'],
+            },
+        ])
+    })
+
+    it('returns multiple submissions payload with multiple revisions', async () => {
```
haworku, line 58:
```suggestion
    it('returns a submitted and then unlocked submission with multiple revisions', async () => {
```

services/app-api/resolvers/fetchSubmission2.test.ts
```diff
@@ -0,0 +1,281 @@
...
+            variables: { input },
+        })
+
+        expect(result.errors).toBeUndefined()
+
+        expect(result.data?.fetchSubmission2.submission).toBeDefined()
+        expect(result.data?.fetchSubmission2.submission).not.toBeNull()
+    })
+
+    it('returns an error if you are requesting for a different state (403)', async () => {
```
haworku, line 198:
```suggestion
    it('returns an error if a state user requests a submission from a different state (403)', async () => {
```

services/app-api/resolvers/fetchSubmission2.test.ts
```diff
@@ -0,0 +1,281 @@
...
+        }
+        expect(result.errors?.length).toEqual(1)
+        const resultErr = result.errors[0]
+
+        expect(resultErr?.message).toEqual(
+            'user not authorized to fetch data from a different state'
+        )
+        expect(resultErr?.extensions?.code).toEqual('FORBIDDEN')
+    })
+
```
haworku, line 240: how about a test for what should happen if a submission at that ID does not exist? Or is that handled somewhere else.

services/app-api/prisma/schema.prisma
```diff
@@ -20,18 +20,20 @@ model TestUser {
...
 
 model StateSubmissionRevision {
   id                  String          @id
-  createdAt           DateTime        
+  createdAt           DateTime
   submission          StateSubmission @relation(fields: [submissionID], references: [id])
   submissionID        String
   submissionFormProto Bytes
+
+  submittedAt DateTime?
```
haworku, line 36: Is this `submittedAt` just the initial submission date? If so, I think we should called it that.  Or is it the date for the last time a submission was submitted by a state user and  thus re-locked?

services/app-web/src/common-code/domain-models/submission2.ts
```diff
@@ -0,0 +1,49 @@
+import { RevisionType, Submission2Status, Submission2Type } from './Submission2Type'
+
+
+// submissionStatus computes the current status of the submission based on 
+// the submit/unlock info on its revisions.
+// These methods ASSUME that revisions are returned most-recent-first.
```
haworku, line 6: I would like to see some basic unit tests here because now we have business logic going on in these functions. Also I wonder if we should add some safe guard here to throw an error if revisions are obviously not properly ordered ?

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -98,25 +105,63 @@ export const SubmissionSummary = (): React.ReactElement => {
...
             console.log('Submission Unlocked', unlockedSub)
         }
 
         modalRef.current?.toggleModal(undefined, false)
     }
 
+    const currentRevision = submissionAndRevisions.revisions[0].revision
+
+    const formBinData = Buffer.from(currentRevision.submissionData, 'base64')
+    const submissionResult = toDomain(formBinData)
```
haworku, line 134: _Maybe this is what you are working on now_ - should be in a helper and pushed as close to the client api as possible. No buffers in the JSX please. Looks like we need some kind of library that wraps our api (apollo) functions. Having that data transformation layer separate from the UI  (and with unit tests) is v important.

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
```diff
@@ -82,6 +80,8 @@ export const ContractDetailsSummarySection = ({
     const [zippedFilesURL, setZippedFilesURL] = useState<string>('')
     const contractSupportingDocuments = submission.documents.filter(doc => doc.documentCategories.includes('CONTRACT_RELATED' as const))
     const isSubmitted = submission.__typename === 'StateSubmission'
+    console.log("IS SUB", isSubmitted, submission.__typename)
```
haworku, line 83: eventually what will be the logic  on the frontend for `isEditing`?  Is it status is DRAFT | | UNLOCKED ?

services/app-api/postgres/updateStateSubmission.ts
```diff
@@ -1,27 +1,92 @@
...
+                            },
+                            data: {
+                                submissionFormProto: proto,
+                                submittedAt,
+                            }
+                        }
+                    }
+                },
+                    include: {
+                    revisions: true
```
macrael, line 55: add OrderBy

haworku:
Naming - `SubmissionWithRevisions` or maybe even go for the `HealthPlanPackageWithRevisions`

I have been anticipating there might also be EQRO submissions, EBRK submissions, and maybe pre-print submissions as entirely separate forms in our app. In which case "submission" is very general.

tests/cypress/integration/cmsWorkflow/unlockResubmit.spec.ts
macrael: I’m not sure I follow. I’m using dirname to pop off the end of the URL (‘review-and-submit’). It is importing werid so I’d probably rather do something else.

services/app-api/resolvers/fetchSubmission2.test.ts
macrael: This test caught a bug!

services/app-api/prisma/schema.prisma
macrael: This is the submittedAt for a revision. I’ll rename the initiallySubmittedAt for the overall package.

services/app-web/src/common-code/domain-models/submission2.ts
macrael: I did include unit tests for this. I’m going back and forth on checking createdAt ordering ourselves. It’s not like we’re going to have tons of revisions, so effeciancy isn’t a super argument, but it does seem like we should be able to order them correctly if we want.

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
macrael: I suspect so. The trick I took advantage of here is that we only set navigateTo on the reivew and submit page, when we are editing. It’s not even set on submission summary so we are already communicating edit status to the component.

services/app-api/prisma/schema.prisma
haworku: should we just call this `unlockedAt` then? It sounds like submission is really only relevant the first time and that is stored on the overall package.  after that we are just locking and unlocking revisions.

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -89,23 +90,38 @@ export const SubmissionSummary = (): React.ReactElement => {
...
     const environmentName = process.env.REACT_APP_STAGE_NAME || ''
-    const isProdEnvironment = ['prod', 'val', 'nope'].includes(environmentName)
+    const isProdEnvironment = ['prod', 'val'].includes(environmentName)
 
     const displayUnlockButton = !isProdEnvironment && loggedInUser?.role === 'CMS_USER'
 
     useEffect(() => {
-        updateHeading(pathname, 'TEMP NAME')
-    }, [updateHeading, pathname])
+        if (submissionAndRevisions) {
```
haworku, line 99: Looks good this is better. Still would like to see us abstract the api request + data transformation  farther from the React UI  code  as we migrate the  other apis.

services/app-web/src/common-code/domain-models/submission2.ts
haworku: I think we should at least check  first and last  revision submitted at dates  and throw an specific error  if things are wrong. personally don't think we should be reordering in this function though that seems like it could be a separate function that we could call optionally if that error type is thrown.

services/app-api/resolvers/fetchSubmission2.test.ts
```diff
@@ -0,0 +1,299 @@
...
+        // then see if we can fetch that same submission
+        const input = {
+            submissionID: createdID,
+        }
+
+        const result = await server.executeOperation({
+            query: FETCH_SUBMISSION_2,
+            variables: { input },
+        })
+
```
haworku, line 105: How about adding a third revision here as well and then we can also quick check that the orderBy is right across each position as well.

haworku:
Is is finding a valid error with rate dates being incorrect. I confirmed locally. The dates are changing between when the state user is on review and submit and when the state user looks at submission summary

BEFORE SUBMIT
Screen Shot 2022-02-25 at 9 58 54 AM
AFTER SUBMIT
Screen Shot 2022-02-25 at 9 59 07 AM

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
macrael: I’d love to see your thoughts. Should we wrap up the use…Query() call in something that handles that for us? In general I’m a fan of having a package that wraps up a dependency.

haworku (approved):
I don't see any other bugs with existing functionality besides the dates. 

I do  see some weirdness with when the banner appears for unlock error but that seems to me we can adjust it when we get to that ticket. Since this ticket is just about rewriting the api I think we should leave that for later. 

@macrael once this deploys can you double check zip downloads is fine? That's the only thing I could think of that I couldn't verify locally .

package.json
```diff
@@ -47,7 +47,8 @@
     },
     "dependencies": {
         "aws-sdk": "^2.1018.0",
-        "cypress-file-upload": "^5.0.8"
+        "cypress-file-upload": "^5.0.8",
+        "path-browserify": "^1.0.1"
```
haworku, line 51: do you mean to add this dependency?

macrael: It’s used in the cypress test to pop the end of the URL off, we could probably do something else, my editor was upset about the import. You were also talking about there being a Cypress fucntion to do this but I couldn’t find what you were talking about.

tests/cypress/integration/cmsWorkflow/unlockResubmit.spec.ts
haworku: Missed this. `cy.location('pathname')` I thought would help but I see what you are doing now and I'm wrong.  if this was in app-web theres a bunch of url helpers in react-router but at this level i guess we do need a dependency

PR #832, opened 2022-03-01
Add tracing to app-api

rswerve:
## Summary

For [16351](https://qmacbis.atlassian.net/browse/OY2-16351)

This adds the necessary packages (and proper versions) for OpenTelemetry in app-api, as well as some examples of instrumentation.  Information is being logged to the console (CloudWatch when deployed) and is also showing up in New Relic (thanks @mojotalantikite!)

Things that aren't working yet:  
No auto instrumentation.  This remains mysterious.
No associating each span with a full session trace across app-web, app-api, graphql, and AWS.  Mojo and I chatted about this a bit and we're both still noodling about the magic commands (we think headers) to do this.

#### Screenshots

image


services/app-api/handlers/otel_handler.ts
```diff
@@ -0,0 +1,22 @@
...
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+
+const provider = new NodeTracerProvider({
+    resource: Resource.default().merge(new Resource({
+        "service.name": "mc-review",
+      })),
+})
+
+// log to console and send to New Relic
+provider.addSpanProcessor(new SimpleSpanProcessor(new ConsoleSpanExporter()))
```
mojotalantikite, line 16: We might not totally care about the console exporter for the merge, since I feel like it's mainly good for debugging.

Or maybe we just comment it out with some instructions that say this line can be enabled for debugging?

macrael: I think we absolutely want it enabled in local dev by default, I think this would be a good env var toggled setting

mojotalantikite (approved):

services/app-api/handlers/otel_handler.ts
mojotalantikite: I think that's cool for now, but I'd vote for local dev that we setup a container that is running [jaeger](https://www.jaegertracing.io/) and we point traces there in local. That way we can lean on the OTEL backend to start correlating the spans to a trace and not have to parse through console output.

macrael: oh yes please. I don’t know what Jaeger is but i’d love for us to have local monitoring running

rswerve:
Closing this in favor of #844

PR #833, opened 2022-03-01
Split apart prisma lambda layer

mojotalantikite:
## Summary

We were running into size limits in the AWS Lambda environment, where the total bundled size of some lambdas + prisma layer + otel layer were exceeding the 250MB-ish limit. This moves us to using a layer for the migrations function, and using the webpack supplied bins for prisma in the lambdas that need to interact with the database (mainly graphql). The bundle for the graphql lambda function including prisma bins is now around 80MB, which should give us plenty of headroom.

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-16750


services/app-api/prisma/schema.prisma
```diff
@@ -20,15 +20,15 @@ model TestUser {
 
 // The StateSubmission Form Data
 model StateSubmission {
-  id                  String                    @id
-  state               State                     @relation(fields: [stateCode], references: [stateCode])
-  stateCode           String
-  revisions           StateSubmissionRevision[]
```
mojotalantikite, line 26: my autosave did this

.github/workflows/deploy.yml
```diff
@@ -230,7 +230,10 @@ jobs:
   finishing-prep:
     name: Finishing Prep
     needs: [deploy-infra, begin-deployment, deploy-app]
-    if: always() && needs.deploy-infra.result == 'success' && needs.deploy-app.result == 'success'
+    if: |
+      always() 
+      && (needs.deploy-infra.result == 'success' || needs.deploy-infra.result == 'skipped') 
+      && (needs.deploy-app.result == 'success' || needs.deploy-app.result == 'skipped')
```
mojotalantikite, line 236: I was seeing that Cypress would get skipped, because this step would sometimes get skipped, and a skip is not considered a success in the same way. Anyway, this makes it so we always run finishing prep and then get a Cypress run.

services/app-api/serverless.yml
```diff
@@ -169,21 +166,6 @@ functions:
...
-          method: get
-          cors: true
-    vpc:
-      securityGroupIds:
-        - ${self:custom.sgId}
-      subnetIds: ${self:custom.privateSubnets}
-    layers:
-      - !Ref PrismaClientLambdaLayer
-      - arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-nodejs-amd64-ver-1-0-1:1
-
```
mojotalantikite, line 186: we never use this function. I had made it when we were just getting prisma up and running, but by now we have coverage of postgres vis prisma with the graphql func.

.github/workflows/deploy.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v2
         with:
-          fetch-depth: 10
+          fetch-depth: 20
```
macrael, line 24: what’s up with fetch-depth?

services/app-api/prisma/schema.prisma
macrael: I gotta figure out all the linters. I hope this is the result of `prisma format`

macrael (approved):
Looks good, thanks for this

.github/workflows/deploy.yml
mojotalantikite: ohhh, yeah so when I was hitting failure after failure with each commit in my branch I hit an issue with the CI skipping script where lerna wouldn't be able to find the commit in it's history, because the GHA only fetches the past 10 commits. 

Anyway, this was to give more git history to GHA so lerna could do it's thing. I have a bug ticket open to clean up the skipping script to deal with this a bit better.

PR #837, opened 2022-03-02
Do not depend on Code Climate publishing success in CI

mojotalantikite:
## Summary
Code Climate is currently having an outage, which is causing CI to fail. This makes it so Code Climate stats publishing doesn't block CI runs from completing.

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-16793


rswerve (approved):

macrael (approved):
Yup.

PR #839, opened 2022-03-03
State user sees submissions with unlocked status on state dashboard

haworku:
## Summary
As a state user, when I am on my state dashboard I see if a previously submitted submission package has been unlocked, so that I know which submissions still need to be edited and resubmitted. 

DRAFT PR BECAUSE I AM FIXING UP WITH BOTH #827 and main so the diff is a little off until that merges

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-13844

#### Test cases covered
[api tests] `indexSubmission2` query
- returns a list of submissions that includes newly created entries
- a different user from the same state can index submissions'
- returns no submissions for a different state's user

[web tests] dashboard page
- adjusted table test to include an unlocked submission

## QA guidance
- open up the app in two tabs (on cognito). log into app as cms user on one side
- create submission, submit submission as a state user
- unlock submission as a cms user
- submission page appears unlocked as cms user
- go to edit unlocked submission as state user (should show unlocked tag in dashboard and navigate to review and submit page)
- submit unlocked submission again as state user
- go to dashboard page as state user, the just resubmitted submissions  should have the same treatment as a submitted submission 
- cms user should see the submission they just unlocked as locked


services/app-api/postgres/findAllSubmissionsWithRevisions.ts
```diff
@@ -0,0 +1,47 @@
...
+    const submissions: Submission2Type[] = []
+    const errors: Error | StoreError[] = []
+    findResult.forEach((submissionWithRevisions) => {
+        // check for current revision, if it doesn't exist, log an error
+        const currentRevisionOrError = getCurrentRevision(
+            submissionWithRevisions.id,
+            submissionWithRevisions
+        )
+        if (isStoreError(currentRevisionOrError)) {
+            // TODO: what should I do with these
```
haworku, line 38: ^ question for @macrael

services/app-api/resolvers/indexSubmissions2.test.ts
```diff
@@ -0,0 +1,200 @@
...
+                     name: 'Aang',
+                     state_code: 'FL',
+                     role: 'STATE_USER',
+                     email: 'aang@mn.gov',
+                 },
+             },
+         })
+
+         // TODO should we type these results
+         const result = await otherUserServer.executeOperation({
```
haworku, line 160: q for @macrael  - I noticed (or maybe re-discovered) that `result` isn't typed. I was trying to figure out a way to have it infer the type of the GraphQLReponse for a specific query but I couldn't. Any ideas here?

services/app-web/src/pages/Dashboard/Dashboard.test.tsx
```diff
@@ -66,18 +67,24 @@ describe('Dashboard', () => {
...
+        
+        const submissions = [draft, submitted, unlocked]
 
-        const submissions = [mockDraft(), mockStateSubmission(), mockDraft()]
-        submissions[2].id = 'test-abc-122'
-        submissions[2].name = 'MN-MSHO-0002' // the names collide otherwise
-        // set middle row to latest updatedAt to test sorting (it should be sorted to top)
-        submissions[1].updatedAt = '2100-01-01T00:00:00.000Z'
+        // TODO - Figure this test mocking out
+
```
haworku, line 77: Reminder for myself  - come back here and figure this out. This is the one failing app-web test because of the way things are mocked

services/app-graphql/src/queries/indexSubmissions2.graphql
```diff
@@ -0,0 +1,25 @@
...
+        edges {
+            node {
+                id
+                stateCode
+                status
+                intiallySubmittedAt
+                revisions {
+                    revision {
+                        id
+                        unlockInfo {
```
haworku, line 13: note for myself - need to add handling for `createdAt`

services/app-api/postgres/findAllSubmissionsWithRevisions.ts
macrael: It doesn’t look like you are actually using the currentRevisionOrError if it’s not an error, so this may be moot, but it’s a good question, especially for queries that return multiple objects. When getCurrentRevision errors it’s a real programming error, submissions should always have at least one revision associated with them. I think it would be reasonable to log it and return an error in this case, ultimately throwing an ApolloError in the handler.

services/app-api/resolvers/indexSubmissions2.test.ts
macrael: I’m not sure. So, this type is the return type we get from the client code, which we generate in app-web. Maybe we could generate those types for app-api and set the types right

services/app-api/resolvers/indexSubmissions2.ts
```diff
@@ -0,0 +1,59 @@
...
+                'user not authorized to fetch state data'
+            )
+            throw new ForbiddenError('user not authorized to fetch state data')
+        }
+
+        // fetch from the store
+        const results = await store.findAllSubmissionsWithRevisions(context.user.state_code)
+
+        if (isStoreError(results)) {
+            if (results.code === 'WRONG_STATUS') {
```
macrael, line 27: I think this isn’t relevant anymore. Submission2 postgres functions don’t specifiy draft/state submisssions.

services/app-web/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -1,55 +1,79 @@
...
     submittedAt?: string
     updatedAt: string
+    status: Submission2Status
     submissionType: GQLSubmissionType
 }
 
+const isSubmitted = (status: Submission2Status) => status === 'SUBMITTED' || status === 'RESUBMITTED'
+
+function submissionURL(
+        id: SubmissionInDashboard['id'],
```
macrael, line 33: Nice, I always forget about indexing into other types

services/app-web/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -58,37 +82,43 @@ export const Dashboard = (): React.ReactElement => {
...
-                    updatedAt: row.updatedAt,
-                    submissionType: row.submissionType,
-                }
+        .forEach((sub) => {
+            const currentRevision = sub.revisions[0]
+            const currentSubmissionData = base64ToDomain(
+                currentRevision.revision.submissionData
+            )
+            console.log('IN HERE')
+            if (currentSubmissionData instanceof Error) {
```
macrael, line 100: This is a good move, we don’t want to break the dashboard just b/c a row has old submissionData in it. We’ll solve this better one day but for now just hiding those rows seems good to me.

macrael (approved):
looks great 🥳

PR #842, opened 2022-03-07
Update copy on homepage to include unexecuted contracts

joshmfranklin:
Changes made to homepage:
- Added "Contracts may be fully executed or unexecuted by some or all parties" to the before you begin section
- Added "CHIP-only submissions" to list of what we don't accept yet
- Reworded "draft documents" to "documents for pre-approval"

Screen Shot 2022-03-07 at 3 11 35 PM


haworku (approved):
nice

PR #843, opened 2022-03-08
Add cms and state unlock email template

joshmfranklin:
## Summary
Adding template for CMS and State unlock email.

### DIFF
--- a/services/app-api/emailer/emailer.ts
+++ b/services/app-api/emailer/emailer.ts
@@ -4,6 +4,7 @@ import {
     newPackageCMSEmail,
     newPackageStateEmail,
     unlockPackageCMSEmail,
+    unlockPackageStateEmail
 } from './'
 import {
     StateSubmissionType,
@@ -42,6 +43,9 @@ type Emailer = {
     sendUnlockPackageCMSEmail: (
         unlockEmailData: UnlockEmailData
     ) => Promise<void | Error>
+    sendUnlockPackageStateEmail: (
+        unlockEmailData: UnlockEmailData
+    ) => Promise<void | Error>
 }
 
 function newSESEmailer(config: EmailConfiguration): Emailer {
@@ -78,6 +82,10 @@ function newSESEmailer(config: EmailConfiguration): Emailer {
         },
         sendUnlockPackageCMSEmail: async function (unlockEmailData){
             const emailData = unlockPackageCMSEmail(unlockEmailData, config)
+            return await this.sendEmail(emailData)
+        },
+        sendUnlockPackageStateEmail: async function (unlockEmailData){
+            const emailData = unlockPackageStateEmail(unlockEmailData, config)
             return await this.sendEmail(emailData)
         }
     }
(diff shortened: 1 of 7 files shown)


services/app-api/emailer/templates.ts
```diff
@@ -149,5 +149,33 @@ const newPackageStateEmail = (
     }
 }
 
-export { newPackageCMSEmail, newPackageStateEmail }
+const unlockPackageCMSEmail = (
+    submission: StateSubmissionType,
```
haworku, line 153: For this parameter instead of `submission: StateSubmissionType` I think it will be `revision: RevisionType`. You can go ahead and add that and you can find that type (so far key, fields are missing that @JasonLin0991  is adding) in domain-models/Submission2Type

@JasonLin0991 could you tell us what the `unlockInfo` and `submitInfo` will look like after your changes. That could unblock josh to keep working here.


If we find out more about what the revisions type will look like then the next step with this PR could be writing tests.

services/app-api/emailer/templates.ts
```diff
@@ -149,5 +149,33 @@ const newPackageStateEmail = (
...
+    config: EmailConfiguration
+): EmailData => {
+    const isTestEnvironment = config.stage !== 'prod'
+    const reviewerEmails = config.cmsReviewSharedEmails
+    const bodyHTML = `
+        <h1>Submission ${submissionName(submission)} was unlocked</h1>
```
haworku, line 159: Don't forget the `<span style="color:#FF0000;font-weight:bold;">Note: This submission is part of the MC-Review testing process. This is NOT an official submission and will only be used for testing purposes.</span>` at the top. Maybe we should be sharing this across templates and prepending everything.

services/app-api/emailer/emailer.ts
```diff
@@ -112,6 +121,18 @@ function newLocalEmailer(config: EmailConfiguration): Emailer {
             ${'(¯`·.¸¸.·´¯`·.¸¸.·´¯·.¸¸.·´¯`·.¸¸.·´¯`·.¸¸.·´¯`·.¸¸.·´)'}
         `)
         },
+        sendUnlockPackageCMSEmail: async (
+            unlockEmailData: UnlockEmailData
+        ) => {
+            const emailData = unlockPackageCMSEmail(unlockEmailData, config)
+            const emailRequestParams = getSESEmailParams(emailData)
```
macrael, line 128: I think it would be nice to extract this bit out into its own function since we are repeating it 4 times in the localEmailer. getSeSRemailParams and the console.log could be in their own function

services/app-api/emailer/templates.ts
```diff
@@ -149,5 +153,73 @@ const newPackageStateEmail = (
     }
 }
 
-export { newPackageCMSEmail, newPackageStateEmail }
+type UnlockEmailData = {
+    submissionName: string
```
macrael, line 157: Looks like our other emails have taken a whole submission here, instead of just the bits they need. Might be nice to keep that consistent but I can see it either way. This way we’re really clear what data is going into the email which is always nice. If we pass in the whole submission, then it makes it easier us to change this email to require different submision data over time.

macrael (approved):
Looking great!

PR #844, opened 2022-03-10
Instrumenting Resolvers

rswerve:
Closes [16351](https://qmacbis.atlassian.net/browse/OY2-16351)

Adds open telemetry instrumentation to our resolvers, sending context, error, and success details to New Relic.

There will need to be work to attach the same ID to requests all the way from the front end through to our lambdas.


services/app-api/resolvers/indexSubmissions.ts
```diff
@@ -6,17 +6,22 @@ import {
...
 export function indexSubmissionsResolver(
     store: Store
 ): QueryResolvers['indexSubmissions'] {
     return async (_parent, _args, context) => {
+        const { user, span } = context
+        setContextualAttributes('indexSubmissions', user, span)
+        
         // This resolver is only callable by state users
         if (!isStateUser(context.user)) {
             logError(
```
haworku, line 20: do we still need `logError` now?  Can we just remove logger or make logError be what calls `setErrorAttributes`

services/app-api/resolvers/attributeHelper.ts
```diff
@@ -0,0 +1,25 @@
+import { Context } from "../handlers/apollo_gql"
+import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
+
+export function setContextualAttributes(name: string, user: Context["user"], span: Context["span"]): void {
```
haworku, line 4: Can you add a quick comment on this function describing what kinds of things might be the preferred use case is for this? Feels like the success and error use case is easy to assume but this one gives me pause

services/app-api/resolvers/indexSubmissions.ts
rswerve: I think logError/logSuccess will go away once we have instrumentation.  I guess it's a question for @macrael: does anything in any of our scripts depend on the error/success logs?  If not, we can remove them as we add instrumentation...

services/app-api/resolvers/attributeHelper.ts
rswerve: good idea

rswerve: in fact, looking at this again, I think I can set these attributes just once in the middleware, and get rid of this helper

services/app-api/handlers/otel_handler.ts
```diff
@@ -0,0 +1,23 @@
+import opentelemetry from "@opentelemetry/api";
+import { NodeTracerProvider } from "@opentelemetry/sdk-trace-node";
+import { Resource } from "@opentelemetry/resources";
+import { OTLPTraceExporter} from "@opentelemetry/exporter-trace-otlp-grpc"
+import { SimpleSpanProcessor, ConsoleSpanExporter } from "@opentelemetry/sdk-trace-base";
+
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
macrael, line 7: what is setting this env var for?

services/app-api/handlers/otel_handler.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const provider = new NodeTracerProvider({
```
macrael, line 9: I think this file belongs somewhere else. We should probably create a new module called “otel” that exports this tracer, rather than having it live with our lambda handlers

services/app-api/resolvers/attributeHelper.ts
macrael: yeah, maybe a name that makes this more clearly tied to this being info about the specific resolver being invoked would help?

services/app-api/resolvers/attributeHelper.ts
```diff
@@ -0,0 +1,27 @@
...
+    span.setAttributes({
+        [SemanticAttributes.CODE_FUNCTION]: name,
+        [SemanticAttributes.ENDUSER_ID]: user.email,
+        [SemanticAttributes.ENDUSER_ROLE]: user.role,
+    })
+}
+
+export function setErrorAttributes(message: string, span: Context["span"]): void {
+    if (!span) return
+    span.setAttributes({
```
macrael, line 17: I think there’s also `client.error` for the actual error object? I’m not sure.

macrael: should we set resolver.success = false to match setSuccessAttributes?

services/app-api/resolvers/indexSubmissions.ts
macrael: let’s leave them in for now, we can remove them once we feel like we’ve gotten used to instrumentation. 

I don’t want to give up being able to look at the CloudWatch and see what’s up. If our otel logs are showing up there now, then we can probably ditch the logError stuff since we’ll still have something in cloudwatch if we look at it.

macrael (approved):
This looks great, we can bikeshed the names of these three functions a little but in general I think it’s great. 

To me “set*Attributes” feels like it’s lacking something about the fact that it’s operating on the current trace/span.

services/app-api/handlers/otel_handler.ts
rswerve: @mojotalantikite I think we put this in while debugging the New Relic piece.  Given your proxy work, will we still need this for logging to NR to work?

PR #845, opened 2022-03-11
[Draft] CMS user can input unlock reason and see unlock banner with correct data when unlocked.

JasonLin0991:
## Summary
As a CMS user when I click `Unlock submission` a modal opens where I can input a required reason for unlocking submission. Once a submission is unlocked, I will see an unlocked banner with unlock information.

Having issues passing cypress test 
Screen Shot 2022-03-11 at 11 15 27 AM

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-12724

#### Test cases covered
- unlockStateSubmissions.test.ts
  - unlocked submission revision returns correct `unlockInfo`
  - returns errors if unlocked reason is undefined

- indexSubmission2.test.ts
  - Fixed `unlockTestDraftSubmission` to include `unlockedReason`

- fetchSubmission2.test.ts
  - Fixed `unlockTestDraftSubmission` to include `unlockedReason`

- ReviewAndSubmitModal.test.tsx
  - Renders element with modal hidden
  - Renders element with modal visible with modal title
  - Renders element with modal visible without title

- SubmissionSummary.test.tsx
  - displays unlock banner with correct data for an unlocked submission
  - displays error when submitting without a unlock reason
  - Fixed other tests to input text in unlock modal unlock reason text input.

- unlockResubmit.spec.ts
  - Add input text in unlock modal unlock reason text input.
  - Add test for visible unlock banner with correct data after submission is unlocked.
  - Add test for hidden unlock banner when resubmitted

## QA guidance
- Create and submit a state submission as a state user.
- As a CMS User click `Unlock submission` and type in reason for unlocking. Then submit.
- As a CMS user you will now see a unlock banner on the top of the page with CMS User email, date of unlock and unlock reason
- As a CMS user unlocking submission without inputing an unlock reason should not allow you to submit on the unlock modal.


JasonLin0991:
Having trouble with this test. 
Screen Shot 2022-03-11 at 11 21 11 AM

Locally it passes on cypress. It looks like this is because the test is looking for the first record on the table and it should be `UNLOCKED`, but during the test it isn't the first record.
Screen Shot 2022-03-11 at 11 24 39 AM

Locally the newly unlocked submission does show up as the first record on the table.

services/app-api/postgres/submissionWithRevisionsHelpers.ts
```diff
@@ -47,12 +47,24 @@ function convertToSubmission2Type(dbSub: StateSubmissionWithRevisions): Submissi
             let submitInfo: UpdateInfoType | undefined = undefined
             if (r.submittedAt) {
                 submitInfo = {
-                    updatedAt: r.submittedAt
+                    updatedAt: r.submittedAt,
```
haworku, line 50: Can we do the same handling for `submitInfo` as we are doing for `unlockInfo` (make sure we use the previous values if they exist(.  I know thats not entirely  in this ticket but with this change we would start rewriting submitInfo with an empty string for certain values whenever this function is called.

services/app-web/src/components/Modal/ReviewAndSubmitModal/ReviewAndSubmitModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+interface ReviewAndSubmitModalProps {
+    modalTitle?: string,
+    onSubmit?: () => void,
+    onCancel?: () => void,
+    showModal: boolean,
+}
+
+export type ModalProps = ReviewAndSubmitModalProps & UswdsModalProps
+
+export const ReviewAndSubmitModal = ({
```
haworku, line 22: The goal with the /components folder is to make things as generalizable as possible. This is super close but can you also pass in props for the className and modal heading id? From what I can tell the only things left that make this modal related to just the review and submit page are those two things otherwise its generalizable and could be thrown anywhere in the app.  Let me know if you want to pair.

Basically I think this is close to just being "Modal" not ReviewAndSubmitModal". And a reusable Modal would be great!

rswerve:
> Having trouble with this test. Screen Shot 2022-03-11 at 11 21 11 AM
> 
> Locally it passes on cypress. It looks like this is because the test is looking for the first record on the table and it should be `UNLOCKED`, but during the test it isn't the first record. Screen Shot 2022-03-11 at 11 24 39 AM
> 
> Locally the newly unlocked submission does show up as the first record on the table.

Yeah, in CI, there are going to be submissions in the DB from other tests.  Can you do a findByText for "Unlocked" within the table?

JasonLin0991:
> 

I can. This wasn't a test I wrote so I wasn't sure how important it was for this unlocked submission to be the first one.

JasonLin0991:
> Yeah, in CI, there are going to be submissions in the DB from other tests. Can you do a findByText for "Unlocked" within the table?

I can. This wasn't a test I wrote so I wasn't sure how important it was for this unlocked submission to be the first one.

haworku:
Yeah... since in CI Cypress is running several tests simultaneously the test is too brittle that I wrote, like @rswerve  said. Realizing now should instead Is should have been checking the submission rows are are in the right order _nn relation to eachother_, not in relation to the table.

@JasonLin0991  Whatever you want to do to get this passing while checking for unlocked is fine with me. I think we will need to revisit the table tests in the dashboard page anyway.i

services/app-api/postgres/submissionWithRevisionsHelpers.ts
JasonLin0991: I wanted to get some clarification. Did you mean to add two fields like `submittedBy` and `submittedChange` in postgress that correspond with `updatedBy` and `updatedReason` ? Also I'm confused about what the previous values are.

haworku: I meant do something like this for submitInfo
```
submitInfo = {
                    updatedAt: r.submittedt,
                    updatedBy: r.submittedBy,
                    updatedReason: r.submittedReason,
                }
                }
```

instead of writing empty strings there. This isn't blocking though I could be misunderstanding something. It just caught my eye the use of empty strings here rewriting over other data.

JasonLin0991: Ah ok I understand.

I believe in order for me to do that I would need to add `submittedBy` and `submittedReason` to postgres since the `StateSubmissionRevision` (auto generated from prisma.schema?) does not include those two.  Does that sound correct @macrael 

I initially set empty strings there because `UpdateInfoType` in `Submission2Type.d.ts` required all three properties. I have no problems with adding those two fields into postgres and updating everything else to accommodate those changes.

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -119,6 +134,25 @@ export const SubmissionSummary = (): React.ReactElement => {
                 setPageLevelAlert('Error fetching the submission. Please try again.')
                 return
             }
+
+            if (submissionAndRevisions.status === 'UNLOCKED') {
+                const unlockedRevision = submissionAndRevisions.revisions.find(rev => rev.revision.unlockInfo)
+                const unlockInfo = unlockedRevision?.revision.unlockInfo
+
+                if (unlockInfo?.updatedBy && unlockInfo?.updatedAt && unlockInfo?.updatedReason) {
+                    //For some reason the type for unlockInfo from graphql is <UnlockedInfo | null | undefined>, not sure where the null came from
```
macrael, line 143: I dont’ know if you still get notified about comments on a closed PR, but pretty much everything in the graphql has null | undefined when things are optional. I don’t remember why

PR #846, opened 2022-03-11
State user sees unlock banner

haworku:
## Summary
As a state user 
when I am viewing a submission 
I want to know if the submission was unlocked by CMS and on what date
so that it is clear to me that the submission still needs to be modified and re-submitted. 

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-12725

#### Test cases covered
- loads unlocked submission and displays unlock banner with correct data for an unlocked submission (jest)
- state user sees banner when submission is unlocked and does not see banner when submission is locked (cypress)

## QA guidance
- unlock a submission and login as state user

### DIFF
--- a/services/app-web/src/gqlHelpers/submissionWithRevisions.ts
+++ b/services/app-web/src/gqlHelpers/submissionWithRevisions.ts
@@ -1,0 +1,91 @@
+import { base64ToDomain } from '../common-code/proto/stateSubmission'
+import { submissionName, SubmissionUnionType } from '../common-code/domain-models'
+import {  Submission2, DraftSubmission, StateSubmission} from '../gen/gqlClient'
+
+type GQLSubmissionUnionType = DraftSubmission | StateSubmission 
+
+const isGQLDraftSubmission = (sub: GQLSubmissionUnionType): boolean => {
+    return sub.__typename === 'DraftSubmission'
+}
+
+const getCurrentRevisionFromSubmission2 = (submissionAndRevisions?: Submission2 | null): SubmissionUnionType | Error => {
+    // check that package and valid revisions exist
+    if (submissionAndRevisions) {
+        if (
+            !submissionAndRevisions.revisions ||
+            submissionAndRevisions.revisions.length < 1
+        ) {
+            console.error(
+                'ERROR: submission in summary has no submitted revision',
+                submissionAndRevisions.revisions
+            )
+            return new Error(
+                'Error fetching the latest revision. Please try again.'
+            )
+        }
+
+        const newestRev = submissionAndRevisions.revisions.reduce(
+            (acc, rev) => {
+                if (rev.revision.createdAt > acc.revision.createdAt) {
+                    return rev
+                } else {
+                    return acc
+                }
+            }
+        ).revision
+
+        // Decode form data submitted by the state
+        const healthPlanPackageFormDataResult = base64ToDomain(
+            newestRev.submissionData
+        )
+        if (healthPlanPackageFormDataResult instanceof Error) {
+            console.error(
+                'ERROR: got a proto decoding error',
+                healthPlanPackageFormDataResult
+            )
+            return new Error('Error decoding the submission. Please try again.')
+        } else {
+            return healthPlanPackageFormDataResult
+        }
+    } else {
+        console.error('ERROR: no submission exists')
+        return new Error('Error fetching the submission. Please try again.')
+    }
+
+    }
+    
+const convertDomainModelFormDataToGQLSubmission = (submissionDomainModel: SubmissionUnionType): GQLSubmissionUnionType => {
+    // convert from domain model back into GQL types
+    const GQLSubmission: GQLSubmissionUnionType =
+        submissionDomainModel.status === 'DRAFT'
+            ? {
+                  ...submissionDomainModel,
+
+                  __typename: 'DraftSubmission' as const,
+                  name: submissionName(submissionDomainModel),
+                  program: {
+                      id: 'bogs-id',
+                      name: 'bogus-program',
+                  },
+              }
+            : {
+                  ...submissionDomainModel,
+                  __typename: 'StateSubmission' as const,
+                  name: submissionName(submissionDomainModel),
+                  program: {
+                      id: 'bogs-id',
+                      name: 'bogus-program',
+                  },
+                  submittedAt:submissionDomainModel.submittedAt
+              }
+
+    return GQLSubmission
+}
+
+export {
+    convertDomainModelFormDataToGQLSubmission,
+    getCurrentRevisionFromSubmission2,
+    isGQLDraftSubmission, 
+}
+
+export type {GQLSubmissionUnionType}
--- a/services/app-web/src/gqlHelpers/submissionWithRevisions.ts
+++ b/services/app-web/src/gqlHelpers/submissionWithRevisions.ts
@@ -1,8 +1,7 @@
 import { base64ToDomain } from '../common-code/proto/stateSubmission'
 import { submissionName, SubmissionUnionType } from '../common-code/domain-models'
-import {  Submission2, DraftSubmission, StateSubmission} from '../gen/gqlClient'
+import {  Submission2, Submission as GQLSubmissionUnionType} from '../gen/gqlClient'
 
-type GQLSubmissionUnionType = DraftSubmission | StateSubmission 
 
 const isGQLDraftSubmission = (sub: GQLSubmissionUnionType): boolean => {
     return sub.__typename === 'DraftSubmission'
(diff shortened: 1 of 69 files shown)


haworku:
Screen Shot 2022-03-11 at 10 42 26 AM

Now I have broken the submission form update and getting this error. Still trying to unravel whats wrong.

services/app-web/src/pages/StateSubmission/updateSubmissionTransform.ts
```diff
@@ -37,34 +38,66 @@ function cleanDraftSubmission(
 // There's probably some Typescript Cleverness™ we could do for this mapping function
 // but for now the compiler complains if you forget anything so ¯\_(ツ)_/¯
 function updatesFromSubmission(draft: DraftSubmission): DraftSubmissionUpdates {
+    console.log('draft', draft)
     return {
         programIDs: draft.programIDs,
         submissionType: draft.submissionType,
         submissionDescription: draft.submissionDescription,
-        documents: stripTypename(draft.documents),
```
haworku, line 44: 🤔  I think this might be the cause of my issues. I need to manually put back the type name everywhere its needed (and remove this code here) to get GQL recognizing my faked domain model > GQL submission

services/app-web/src/gqlHelpers/submissionWithRevisions.ts
```diff
@@ -0,0 +1,91 @@
+import { base64ToDomain } from '../common-code/proto/stateSubmission'
+import { submissionName, SubmissionUnionType } from '../common-code/domain-models'
+import {  Submission2, DraftSubmission, StateSubmission} from '../gen/gqlClient'
+
+type GQLSubmissionUnionType = DraftSubmission | StateSubmission
```
macrael, line 5: I think this is a type in the generated code

services/app-web/src/pages/StateSubmission/StateSubmissionForm.test.tsx
```diff
@@ -27,7 +27,7 @@ describe('StateSubmissionForm', () => {
                     apolloProvider: {
                         mocks: [
```
macrael, line 28: I wonder if we should configure some default mocks for the apolloProvider in renderWithProviders?

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
```diff
@@ -43,58 +48,126 @@ const activeFormPages = (draft: DraftSubmission): RouteT[] => {
...
+
+            // set unlock info
+            if (submissionAndRevisions.status === 'UNLOCKED') {
+                const unlockedRevision = submissionAndRevisions.revisions.find(
+                    (rev) => rev.revision.unlockInfo
+                )
+                const unlockInfo = unlockedRevision?.revision.unlockInfo
+
+                if (
+                    unlockInfo?.updatedBy &&
```
macrael, line 151: I’m surprised you can’t just do `if (unlockInfo)` here

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
```diff
@@ -127,63 +197,81 @@ export const StateSubmissionForm = (): React.ReactElement => {
...
                 />
+                {/* <PageBannerAlerts /> */}
             </div>
             <StateSubmissionContainer>
                 <Switch>
                     <Route path={RoutesRecord.SUBMISSIONS_TYPE}>
                         <SubmissionType
                             draftSubmission={draft}
                             updateDraft={updateDraft}
-                            formAlert={
```
macrael, line 148: Why did we used to pass formAlert down?

macrael (commented):
This is looking great!

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
haworku: It used to be displayed inside the form container. I've moved all alerts outside now since the info alert designs show it outside makes sense  (to me) to keep alerts at the same level so they are easier to grok. Now only error summary errors will be in the form. what do you think?

tests/cypress/integration/cmsWorkflow/unlockResubmit.spec.ts
```diff
@@ -91,8 +89,19 @@ describe('dashboard', () => {
...
+                // State user can resubmit and see resubmitted package in dashboard
                 cy.intercept('POST', '*/graphql').as('gqlRequest2')
+
+                //Unlock banner for state user to be present with correct data.
+                cy.findByTestId('unlockedBanner')
+                    .should('exist')
+                    .and('contain.text', 'zuko@example.com')
+                    .and('contain.text', 'Unlock submission reason.')
+                    .contains(
+                        /Unlocked on: (0?[1-9]|[12][0-9]|3[01])\/[0-9]+\/[0-9]+\s[0-9]+:[0-9]+[a-zA-Z]+\s[a-zA-Z]+/i
```
macrael, line 101: What is this regex? Are we just trying to match a mm/dd/yyyy timestamp?

macrael (approved):
I meant to approve this before, looks great.

tests/cypress/integration/cmsWorkflow/unlockResubmit.spec.ts
JasonLin0991: Yes, this regex is just checking the timestamp is in the correct format. I was having issues getting the correct current date time in cypress testing.

haworku: I think so, I grabbed this from Jason's work. I think he was seeing the date bug here where the date was not correct  (off by one day I believe) so instead added regex which I thought was creative workaround. @JasonLin0991 am I describing the bug you saw right? Did it sound similar to this https://qmacbis.atlassian.net/browse/OY2-12629

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
haworku: This is what the types look like in GQL

```
type UpdateInformation {
    updatedAt: Date
    updatedBy: String
    updatedReason: String
}
```

Is this what we intend 🔍 . Maybe those should all be required (with the `!`). Heres the way it looks in my editor on hover
Screen Shot 2022-03-14 at 4 52 39 PM

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
haworku: ^ I looked more into this and I'm gonna push a commit just making a new shared generic alert banner  in `/components`. Thats what we display when the api fails with 500, we show this red alert. Realized we could be sharing that between pages including outside the state submission form. I will post a photo of what it looks like when you have a unlocked submission and you have api error both displayed and tag design when the work is up on this PR

tests/cypress/integration/cmsWorkflow/unlockResubmit.spec.ts
JasonLin0991: > I think he was seeing the date bug here where the date was not correct (off by one day I believe)

Just checking some past cypress tests done today and I think it's similar, except it seems like in cypress the time will always be `12:00am UTC`.

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
macrael: sounds great

JasonLin0991 (approved):
Removed my comments leftover from other PR.

tests/cypress/integration/cmsWorkflow/unlockResubmit.spec.ts
macrael: oh I see it’s a whole timestamp, I thought we were just going to show the date here. I’m a little worried about the bug appearing here, I thought that was just with dates that had no time associated with them. Here we should be getting a bonafied Date() object which should do timezone adjustments automatically.

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
macrael: Ahh, yeah, I think those should all be required. If we have an update info, it should have all the info in it.

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
haworku: Screen Shot 2022-03-14 at 5 28 38 PM

Here we go. This is what it would look like if an unlocked submission failed to update on "Continue" click for example. The  alerts will stack, with the error on top. Right now using the generic error message for almost everything in state submission form that can go wrong, but we can add more custom error messaging as we need it. cc: @joshmfranklin @MegMurphy

PR #850, opened 2022-03-15
Update diagrams to include Revisions and use native github mermaid display

haworku:
## Summary
Updated diagrams. Welcome all comments.

### Issue
https://qmacbis.atlassian.net/browse/OY2-16465

### DIFF
--- a/docs/submissionDiagram.md
+++ b/docs/submissionDiagram.md
@@ -1,39 +1,64 @@
 # Submission Data Model
 
-This diagram is drawn using [Mermaid](https://mermaid-js.github.io/mermaid/#/entityRelationshipDiagram) to better visualize how data could be stored in the database if tables were used. We handle submission data as a blob. Thus, domain models for parts of a submission that are contained to the api (such as contract/rate amendment info) are not included.
+Diagrams are drawn using [Mermaid](https://mermaid-js.github.io/mermaid/#/entityRelationshipDiagram) to better visualize the in the app. You can edit and save new diagrams using the [Mermaid Live Editor](https://mermaid-js.github.io/mermaid-live-editor).
 
-You can edit and save new diagrams using the [Mermaid Live Editor](https://mermaid-js.github.io/mermaid-live-editor).
+## Database Models (birds-eye view)
 
-[![Mermaid diagram][1]][1]
+For more detail, see [prisma schema](../services/app-api/prisma/schema.prisma).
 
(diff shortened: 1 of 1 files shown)


rswerve (approved):
Looks good!

macrael (approved):
I think it’s a little confusing that we have SubmitInfo and UnlockInfo as their own standalone types even though they are actually UnlockInfo in the real typesystem. Let me look at some other exmaples of these kinds of diagrams for how that could be structured. I would expect either we’d have the references in the type that refrences them by name (like submitInfo as a field in HealthPlanPackageRevision) or we’d draw multiple lines with that name on it to the one UpdateInfo model.

haworku:
@macrael that's fair. I think its getting fuzzy here for me regarding are we making something more like  ER diagram (even though we aren't actually mapping the db)  or are we making a type diagram. I pulled them out for that reason - similar to  reasons why rateInfo and other nested objects are pulled out.

macrael:
Yeah, here are some examples I found snooping around. Like you said, they mostly seem like they are mapping 1:1 with database tables, so they have entries for foo_id that they can connect to the Foo table. (and intersstingly, none of the lines are labeled)

7_ERD_Subjectwise_Performance
_new_Entity-Relationship-Diagram-Example
er-diagram-example-simple-order-system

We need to break out the separate types like rateInfo in part because mermaid doesn’t to a nested datatypes, and it makes sense to me b/c RateInfoType is a separate type in our code as well. Since our domain models are all nested, I think it might be clearer if those relationships instead of just being labeled “contains” are labeled with the accessor on the parent type, like `rateInfo`. Then I think it would flow naturally that we’d need two separate accessors for UpdateInfo, one for submit and one for unlock. 

I think if you just put two relationships with the same start and end and a different label in mermaid it’ll draw two lines.

haworku:
Screen Shot 2022-03-21 at 10 39 26 AM

Here it is with all the nested types pulled out as keys on the parent. I pulled out contract info because thats what we are doing in protobuf not sure if we want that here as well if we are pulling out sub types...

Not sure now what to do with the documents and contacts so they are floating. If they were associated with all the types that have documents or contacts it adds a lot of lines. 

Feel free to go in and [make your own change ](https://mermaid-js.github.io/mermaid-live-editor/edit/#pako:eNqVVk1v2zAM_SuGz81h2Gm5tUmHFRiGwml3yoWV6VSoLQWSXCxI8t9HyZIt-aPreiic9yiSenySfc6ZLDFf56i2HA4Kmr34gVCb18caxCOwNzhgdt6LjP5KMJhxwQ2Huj7t2peGG4PlreloFG2TaQOm1R2gjeLi4CDcUJkOBaXglCl855pLQaHXuZKXy2p1uWQTovDr1hmTwgC3CWbWh7CkdaYQon6fjxZ9EJXMWlFL9mYfJ5R22-yo68e1lnr-LlWzBQOf7dnniZpYZ_u8a2Sf___Kbnd2ZUQ4YZwsrQO9LH5mHrs7TbECQUvhxZhpJuw2SO9X83JiiuAT8s0Lqs4nv9zzyFCRe8JyK4a2e96iZoofDbc9Dat6_ul0TIx3VNLa_GGrI2fosZfn_DIVK-T0PdJsmdExUUrWNigCaEMUxbgBODOEXx1fhOEo_9CrbNmtTxaEjeQQ0OAE3H19Ln4OaNrRhiocpOKoQ_qokt_KZwsZbmosZD1lsAFeR-mXzUKuXa3s_0Sj-MQkxDmaddBxmHQC3_9B1lp77CLLdfP1EdQAEqnMAncvynioDQjqvNyAIob27jQc6ApLVFDftuaV9B2zfd55Z9wSVlrc7ZLNod4TiRz-3M9niUVcLNWPumuTG2z0HdIMXRiWgfaDleYV1QMFzcW8SPICCLrja3tWnuRGvvPyy7dF_jcwxgUMZ5i6hCM3DrGHQvsKnSqLVOS0hW0GoZazj9VajjyPNFH-YpwKVUTMP88BNXgeboJRO0VyfXun29tiML-zr4XmbB3w3tIJukFleMUxsTtp2II6pdebq9szTdMKzpxKjworVChY-ra3BVLH252ks1FjJLmX-p1HCn3o82lA_C2AVYXM8PdZlRKyl4oayW_yBhVdaiV9Mrl0-5wGTHdibl-0JVbQ1u4dfaXQ7l1xX3IjVb6uoNZ4k0Nr5O4kWA90Uf7ry6PXv8AXZSs) @macrael

PR #854, opened 2022-03-17
Address a11y issues in new unlock modal

haworku:
## Summary
- Address accessibility issues in unlock modal with ESC key and focus on error. 
- Use shared Modal component on both ReviewSubmit and SubmissionSummary
- Use formik in the unlock modal form to make it easier to display dynamic and changing error messages
- Use the react-uswds Modal ref to handle state instead of adding our own state to the component
- Small bugfix for a page flicker I saw when the user creates a new submission

FYI for anyone interested-  root cause of issues here is 
1. not using an actual <form> semantic element  and a form label  (I used aria-label since we are using a visible label) even though we are submitting form data . That makes the field not be recognized by screenreader
2.   using `forceAction` on react-uswds Modal  - that overrides built in handling for a11y stuff and isn't needed

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-17215

#### Test cases covered

Forthcoming... I will add a couple tests for the focus behavior and also the Modal component.

#### QA guidance
Please give this a runthrough with voiceover and let me know what you find.

### DIFF
--- a/services/app-web/src/components/Modal/Modal.stories.tsx
+++ b/services/app-web/src/components/Modal/Modal.stories.tsx
@@ -11,9 +11,8 @@ const Template: Story<ModalProps> = (args) => <Modal {...args} />
 
 export const ModalShow = Template.bind({})
 ModalShow.args = {
-    showModal: true,
     modalHeading: 'Test Modal Title',
-    modalHeadingId: 'testModalTitle',
+    id: 'test-modal',
     children: [
         <div style={{ flexDirection: 'column', display: 'flex', padding: 10}}>
             <label htmlFor={'textarea1'}>Text Box 1</label>
--- a/services/app-web/src/components/Modal/Modal.test.tsx
+++ b/services/app-web/src/components/Modal/Modal.test.tsx
@@ -1,15 +1,17 @@
-import React from 'react'
-import { screen, render } from '@testing-library/react'
+import React , {createRef} from 'react'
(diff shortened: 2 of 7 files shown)


joshmfranklin (commented):
Modal works as expected with VO. The only thing I saw was that the trigger button for the modal is now using the primary button. The unlock button should be the secondary button style. 
Screen Shot 2022-03-18 at 11 50 39 AM

haworku:
@joshmfranklin oops thanks josh, I did change the underlying component there I will fix the styles.

haworku:
Josh found some issues in JAWS I will work on next:
- It seems like its double announcing content in the modal
- aria-label on character counter could be cause going to look at this

rswerve (approved):
This looks good and I was able to keyboard navigate everything without issue.  The one thing I noticed, and this should definitely not hold up the PR, is that there's no indication that `Submit` has been clicked in the modal, and it takes a second or two for the modal to go away and the summary to load, during which I hit enter again, which caused me to see the "already unlocked" banner when the summary finally loaded.  Doesn't affect usability at all, but it would be nice for folks not to wonder if their click was registered.

haworku:
@rswerve Good catch, it's easy enough to close the modal faster but also noticed we aren't doing the same UI handling   (displaying that spinner with the  "Loading" heading which will announce in screenreader as well) for the unlock call as we are for other API calls. Let me add that in also

PR #855, opened 2022-03-18
State and CMS receives email that submission has been resubmitted

JasonLin0991:
## Summary

When resubmitting an unlocked submission State and CMS contacts should receive an email with resubmission information.

**Placeholder Code Note:** 
On the ReviewAndSubmit page, there is placeholder code that will default resubmission reason. This should be changed or removed on [OY2-12183](https://qmacbis.atlassian.net/browse/OY2-12183)

#### Related issues
[OY2-12182](https://qmacbis.atlassian.net/browse/OY2-12182)
[OY2-12186](https://qmacbis.atlassian.net/browse/OY2-12186)


services/app-api/postgres/submissionWithRevisionsHelpers.ts
```diff
@@ -45,11 +45,11 @@ function convertToSubmission2Type(dbSub: StateSubmissionWithRevisions): Submissi
         stateCode: dbSub.stateCode,
         revisions: dbSub.revisions.map(r => { 
             let submitInfo: UpdateInfoType | undefined = undefined
-            if (r.submittedAt) {
+            if (r.submittedAt && r.submittedReason && r.submittedBy) {
```
haworku, line 48: @macrael Isn't this where the default values should go that we discussed? We said for a new revision on creation that default `submittedReason` should be `Initial submission` and for `updatedBy` should be the current user

macrael: I think we want to set those default values when we create the revision so that those default values are set in the database itself. This code would be right as is, converting the db fields to our domain submitInto type.

services/app-api/resolvers/submitDraftSubmission.ts
```diff
@@ -148,8 +162,47 @@ export function submitDraftSubmissionResolver(
...
+        //Set submitInfo default to initial submission
+        const submitInfo: UpdateInfoType = {
+            updatedAt: new Date(),
+            updatedBy: context.user.email,
+            updatedReason: 'Initial submission'
+        }
+
+        //If submission is a resubmission set submitInfo updated reason to input.
+        if (planPackageStatus === 'UNLOCKED' && submittedReason ) {
+            submitInfo.updatedReason = submittedReason
```
macrael, line 188: @haworku This looks good to me, setting the default value is happening at submit time and handling either setting the default in the initial submission case, or setting the value passed in on re-submission is handled all by the resolver code.

haworku: Ohhhhh. Thank you for calling it out I was looking for this code. 🔍

services/app-api/emailer/emailer.ts
```diff
@@ -81,40 +92,49 @@ function newSESEmailer(config: EmailConfiguration): Emailer {
...
     return {
         sendEmail: async (emailData: EmailData): Promise<void | Error> => {
-            const emailRequestParams = getSESEmailParams(emailData)
-            console.log(`
-            EMAIL SENT
-            ${'(¯`·.¸¸.·´¯`·.¸¸.·´¯·.¸¸.·´¯`·.¸¸.·´¯`·.¸¸.·´¯`·.¸¸.·´)'}
-            ${JSON.stringify(emailRequestParams)}
-            ${'(¯`·.¸¸.·´¯`·.¸¸.·´¯·.¸¸.·´¯`·.¸¸.·´¯`·.¸¸.·´¯`·.¸¸.·´)'}
-        `)
+            localEmailerLogger(getSESEmailParams(emailData))
```
macrael, line 133: You could put `getSEESEmailParams` into `localEmailerLogger` as well since we call it in every instance here.

services/app-api/emailer/templates.test.ts
```diff
@@ -314,19 +326,18 @@ describe('Email templates', () => {
         })
 
      })
-     describe('CMS unlock email', () =>{
+    describe('CMS unlock email', () =>{
+        const unlockData = {
+            submissionName: 'MCR-VA-CCCPLUS-0001',
```
macrael, line 331: this is very nice. I don’t think we’ve done stuff like this anywhere and I bet it would help. The only risk here is that that unlock data is shared across all these test runs, if any test were to modify it, that might make a different test fail which could be confusing. We’ve solved that problem elsewhere by writing functions that return a new object every time they are called, you can see that pattern in apolloHelpers.tsx. In this case, this is probably fine since we don’t expect any of these functions to do anything but read this, but it’s something to be aware of.

services/app-api/postgres/submissionWithRevisionsHelpers.ts
macrael, line 48: Just to note, this is somewhere where our db representation and our domain model representation differ, and it means that we can get the two into an inconsistent state. This code means that if we ever manage to write incomplete submit* fields to the database, we’ll end up with no submitInfo coming out. I think that’s correct, but I want to point it out because it’s a bit of a risk. If we were really worried about that happening, we could raise an error here instead

services/app-api/postgres/updateStateSubmission.ts
```diff
@@ -34,7 +31,7 @@ async function submitStateSubmissionWrapper(
 
         try {
             const currentRevision = currentRevisionOrError
-            const updateResult = await client.stateSubmission.update( {
+            const submission = await client.stateSubmission.update({
```
macrael, line 34: I would name this submissionResult to indiciate that it can in fact be an error

services/app-web/src/pages/StateSubmission/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -48,14 +50,23 @@ export const ReviewSubmit = ({
...
-    const handleFormSubmit = async (): Promise<void> => {
+    const handleFormSubmit = async (e: React.FormEvent): Promise<void> => {
+        e.preventDefault()
+
+        const input = { submissionID: draftSubmission.id }
+
+        if (unlocked) {
+            Object.assign(input, {
+                //This is placeholder text until we get the resubmission reason input modal added in.
+                submittedReason: 'Placeholder resubmission reason'
```
macrael, line 61: Nice! i like it

macrael (approved):
This is terrific work. You really spanned the whole stack with this one.

haworku (approved):

PR #858, opened 2022-03-21
Fix dashboard page error display

haworku:
## Summary
See bug description, this was a styling regression over the last couple months. Fixed it a bit with #854 but realized the container element was not quite right.

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-17030 

#### Screenshots
This is the new updated error display

Screen Shot 2022-03-21 at 10 09 42 AM


JasonLin0991 (approved):
Looks good to me

PR #859, opened 2022-03-21
Fix rate amendment dates so we can continue past the Contacts page

macrael:
## Summary

Showstopper bug found by Josh, we missed a spot in converting dates to work correctly with our weird handoff between our domain models and the old updateDraftSubmission inputs.

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-17270

## QA guidance

We should be able to submit a rate amendment (with rate start/end dates) on this branch.


JasonLin0991 (approved):
Looks good, I'm able to submit a rate amendment.

PR #861, opened 2022-03-21
Add test for rate amendment dates bug

macrael:
## Summary

This is a test that I confirmed will catch the showstopper bug we fixed today. 

#### Related issues

https://qmacbis.atlassian.net/browse/OY2-17270


haworku:
whew I've rerun tests a couple times here. cypress and gh not having a good day

haworku:
looks like valid test fail

macrael:
@haworku this is passing now

PR #867, opened 2022-03-23
[OY2-12183] State can list what changes were made when they resubmit

JasonLin0991:
## Summary
State user can input a summary of changes made in the submit modal when resubmitting and unlocked submission.

**NOTE:** I added `"defaultCommandTimeout": 10000,` to check if that fixes of the flaky tests we had. Many of them are having issues finding elements, buttons , ..etc and are sometimes not consistent on where it fails. I think it might be something with API requests taking longer than before and not loading the content before looking for elements. Not entirely sure. This hopefully gives commands more time to try and find things?
#### Related issues
[OY2-12183](https://qmacbis.atlassian.net/browse/OY2-12183)

#### Test cases covered
- Testing Library
   - opens submission modal with summary input when submit button is clicked on unlocked plan package
   - redirects if submission succeeds on unlocked plan package
   - displays an error if submission fails on unlocked plan package
   - displays form validation error when summary for submission is over 300 characters
   - displays form validation error when submitting without an submission summary
   - draws focus to submitted summary textarea when form validation errors exist

- Cypress
   - can unlock and resubmit
      - Modified to be able to resubmit with submittedReason (Summary of changes)


services/app-web/src/pages/StateSubmission/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -156,14 +204,58 @@ export const ReviewSubmit = ({
...
+                onSubmit={() => {
+                    setFocusErrorsInModal(true)
+                    formik.handleSubmit()
+                }}
             >
-                <p>
-                    Submitting this package will send it to CMS to begin their
-                    review.
-                </p>
+                <form>
```
haworku, line 215: The form element should be inside the `unlock ? (this JSX): (that JSX)` logic. We don't have a form element on the page at all if we aren't looking at an unlocked submission that will be resubmitted. That will lead to funky behaviors in assistive devices.

services/app-web/src/pages/StateSubmission/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -156,14 +204,58 @@ export const ReviewSubmit = ({
...
                 modalRef={modalRef}
                 id="review-and-submit"
-                modalHeading="Ready to submit?"
+                modalHeading={unlocked ? 'Summarize changes' : 'Ready to submit?'}
                 submitButtonProps={{ className: styles.submitButton }}
-                onSubmit={handleFormSubmit}
+                onSubmitText={unlocked ? 'Resubmit' : undefined}
+                onSubmit={() => {
+                    setFocusErrorsInModal(true)
+                    formik.handleSubmit()
```
haworku, line 212: I wonder if we can just call the `onSubmit` directly here when we are in a draft submission rather than going through formik even though we don't have a form in that modal.

services/app-web/src/pages/StateSubmission/ReviewSubmit/ReviewSubmit.test.tsx
```diff
@@ -238,4 +240,221 @@ describe('ReviewSubmit', () => {
 
         expect(errorText).toBeInTheDocument()
     })
+
+    it('opens submission modal with summary input when submit button is clicked on unlocked plan package', async () => {
```
haworku, line 244: These tests look really good 🏆 can you wrap them in a jest `describe` block because I feel like they are all related and also anticipating we are going to keep adding more and more complex logic to this page

services/app-web/src/components/Modal/Modal.tsx
```diff
@@ -66,7 +66,7 @@ export const Modal = ({
                         onClick={onSubmit}
                         {...submitButtonProps}
                     >
-                        Submit
+                        {onSubmitText || 'Submit'}
```
haworku, line 69: Can you add a test to `Modal.test.tsx` for this new prop that was added? That way we have test coverage for this new logic

services/app-web/src/pages/StateSubmission/ReviewSubmit/ReviewSubmit.tsx
JasonLin0991: Another approach I was thinking about doing was to check for unlocked right after `setFocusErrorsInModal(true)` then calling the appropriate function
```
    const submitHandler = async() => {
        setFocusErrorsInModal(true)
        if (unlocked) {
            formik.handleSubmit()
        } else {
            await onSubmit(undefined)
        }
    }
```  

```
      onSubmitText={unlocked ? 'Resubmit' : undefined}
      onSubmit={submitHandler}
```

cypress.json
```diff
@@ -7,6 +7,7 @@
     "videosFolder": "tests/cypress/videos",
     "supportFile": "tests/cypress/support",
     "projectId": "tt5hbz",
+    "defaultCommandTimeout": 10000,
```
haworku, line 10: I also was testing this on my branch. I don't think it is addressing the flakes unfortunately but its good to know about

tests/cypress/support/stateSubmissionFormCommands.ts
```diff
@@ -209,6 +209,9 @@ Cypress.Commands.add('submitStateSubmissionForm', (success = true) => {
     cy.findAllByTestId('modalWindow')
         .should('exist')
         .within(($modal) => {
+            if (resubmission) {
+                cy.get('#submittedReasonCharacterCount').type('Resubmission summary')
+            }
             cy.findByTestId('review-and-submit-modal-submit').click()
```
JasonLin0991, line 215: I added this part for resubmissions. Let me know if it looks ok.

haworku: yes I saw that! looks good

haworku (approved):

PR #871, opened 2022-03-24
OY2-17338: Reveal unlock button on all environments

haworku:
## Summary
We are done with this epic. The button should be available on all environment (only for CMS users)

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-17338


rswerve (approved):
I see it!

PR #872, opened 2022-03-24
OY2-16868: Add Jira Issue Collector Feedback Tab to Dev Environment

haworku:
## Summary
Add Jira Issue Collector Feedback Tab to Dev Environment

Notes: 
- added a useScript hook to handle this after trying a couple approaches that didn't work
- More about the ticket collector: https://support.atlassian.com/jira-cloud-administration/docs/use-the-issue-collector/
- Also saw this: _The JavaScript exposed by the issue collector is not considered a stable API and may change with new Jira releases._

### Screenshots
Screen Shot 2022-03-29 at 4 06 22 PM

#### Related issues
https://qmacbis.atlassian.net/browse/OY2-16868


haworku:
Got a new link and everything is working now!

services/app-web/src/pages/App/App.tsx
```diff
@@ -35,9 +36,14 @@ function App({
...
 }): React.ReactElement {
-    useEffect(() => {
-        logEvent('on_load', { success: true })
-    }, [])
+
+    // This is a hacky way to fake feature flags before we have feature flags.
+    // please avoid reading env vars outside of index.tsx in general.
+    const environmentName = process.env.REACT_APP_STAGE_NAME || ''
+    const isProdEnvironment = ['prod', 'val'].includes(environmentName)
+    const jiraTicketCollectorURL = `https://meghantest.atlassian.net/s/d41d8cd98f00b204e9800998ecf8427e-T/-9zew5j/b/7/c95134bc67d3a521bb3f4331beb9b804/_/download/batch/com.atlassian.jira.collector.plugin.jira-issue-collector-plugin:issuecollector/com.atlassian.jira.collector.plugin.jira-issue-collector-plugin:issuecollector.js?locale=en-US&collectorId=e59b8faf`
```
macrael, line 44: Is this a secret? I guess it just ends up in our app so there’s no way to hide it really.

services/app-web/src/pages/App/App.tsx
```diff
@@ -35,9 +36,14 @@ function App({
...
+    
+    useScript(jiraTicketCollectorURL, !isProdEnvironment)
```
macrael, line 46: Could we wrap the useScript call inside `if (!isProdEnvironment)` or does it need to be called every time?

services/app-web/src/pages/App/App.tsx
```diff
@@ -35,9 +36,14 @@ function App({
     apolloClient: ApolloClient<NormalizedCacheObject>
     s3Client: S3ClientT
 }): React.ReactElement {
-    useEffect(() => {
-        logEvent('on_load', { success: true })
```
macrael, line 39: Can you check with Mojo to make sure we don’t need this anymore? I’m not sure if we are logging an event on load through otel yet.

macrael (approved):
Yay, nice and easy.

services/app-web/src/pages/App/App.tsx
haworku: I have no idea

services/app-web/src/pages/App/App.tsx
haworku: According to [rule of hooks](https://reactjs.org/docs/hooks-rules.html#explanation) can't conditionally run a hook

services/app-web/src/pages/App/App.tsx
haworku: Let me just add that line back, I think we can clear out logEvent later I was getting hasty

haworku:
I will need to open a new branch to get this work through. getting api fails similar to our bug earlier last week. think my AWS env on this review app is not set up properly

PR #873, opened 2022-03-25
OY2-13875: Add change history to submission summary page

rswerve:
## Summary

Closes [#13875](https://qmacbis.atlassian.net/browse/OY2-13875)

Adds a history of unlocks and submissions to the bottom of the submission summary page, formatted as a list of accordions.

#### Screenshots

image

#### Test cases covered

Accordions are there, open and close, have correct content.
Also had to add a test to Section Header, because I had to tweak it to allow rending without a bottom border on the title header.

## QA guidance

Color contrast error on the accordion rows.  Not sure how we can override the uswds default, and googling isn't turning up any satisfying answers.  @haworku ?

How do _you_ expect VoiceOver to work here?  I'd like more guidance from VO about the fact that I'm on something like an accordion and can click down into the expanded content, but we're following the [guidance from USWDS](https://designsystem.digital.gov/components/accordion/).

Note that if you want to test this from scratch, you need to log in as a state user, submit, save the submission URL, then log in as a CMS user and unlock.

### DIFF
--- a/services/app-web/src/components/ChangeHistory/ChangeHistory.test.tsx
+++ b/services/app-web/src/components/ChangeHistory/ChangeHistory.test.tsx
@@ -92,4 +92,20 @@ describe('Change History', () => {
             screen.getByText('Placeholder resubmission reason')
         ).not.toBeVisible()
     })
+    it('should list the submission events in reverse chronological order', () => {
+        render(<ChangeHistory submission={submissionData} />)
+        expect(
+            screen.getByText('Placeholder resubmission reason')
+        ).not.toBeVisible()
+        const accordionRows = screen.getAllByRole('button')
+        expect(accordionRows[0]).toHaveTextContent(
+            '03/23/22 9:19pm ET - Submission'
+        )
+        expect(accordionRows[1]).toHaveTextContent(
+            '03/23/22 9:18pm ET - Unlock'
+        )
+        expect(accordionRows[2]).toHaveTextContent(
+            '03/22/22 10:08pm ET - Submission'
+        )
+    })
 })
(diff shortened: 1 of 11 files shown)


JasonLin0991 (commented):
When one accordion history item is expanded, it closes any other expanded items. Looking at Figma, it shows that all the history can be expanded. @joshmfranklin Should the user be able to have all the history expanded?

rswerve:
> 

Good call.  This is a simple prop change; I'll make it match Figma and we can change it later if Josh objects, but I'd bet Figma is a reliable guide here.

services/app-web/src/components/ChangeHistory/ChangeHistory.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+        const accordionRows = screen.getAllByRole('button')
+        userEvent.click(accordionRows[0])
+        expect(
+            screen.getByText('Placeholder resubmission reason')
+        ).toBeVisible()
+        userEvent.click(accordionRows[0])
+        expect(
+            screen.getByText('Placeholder resubmission reason')
+        ).not.toBeVisible()
+    })
```
haworku, line 94: Could you add some jest testing around the ordering of the history - I see  a lot of logic in that component around  that aspect.

rswerve:
> > 
> 
> Good call. This is a simple prop change; I'll make it match Figma and we can change it later if Josh objects, but I'd bet Figma is a reliable guide here.

done

haworku (approved):
Looks really good! Very exciting to see this in the app.

@rswerve I don't see a valid color contrast issue when I look at this with the [WAVE browser extension](https://wave.webaim.org/extension/). I _think_ the pa11y color contrast warning is false positive - similar to what happened to Jason (https://github.com/CMSgov/managed-care-review/pull/834#issuecomment-1058320246) - so it may need to be ignored the same way

PR #877, opened 2022-03-26
use all state programs

macrael:
## Summary

Moved programs to have UUIDs and short names. Before we were using the id in the title of the package, but now we have to coordinate with the list of programs for the state. 

Gotta make tests pass but this is a stake in the ground.


scripts/import-programs.ts
```diff
@@ -8,10 +8,8 @@
 //
 // Additional columns aren't used and should be ignored.
 
-import csv from 'csv-parser'
-import fs from 'fs'
-import slugify from 'slugify'
```
haworku, line 13: Remove the unused dependency from `package.json` as well

scripts/import-programs.ts
```diff
@@ -130,33 +110,16 @@ fs.createReadStream(file)
             };
         }
 
-        const id = generateSlug(code, data.Program)
-
         // Check to see if the id is unique. If it isn't abort and print some useful information.
```
haworku, line 113:
```suggestion

```

services/app-web/src/common-code/domain-models/submission.ts
```diff
@@ -104,6 +105,24 @@ function submissionName(submission: SubmissionUnionType): string {
...
+    })
+}
+
+function submissionNameWithPrograms(submission: SubmissionUnionType, statePrograms: ProgramT[]): string {
+    const padNumber = submission.stateNumber.toString().padStart(4, '0')
+    const pNames = programNames(statePrograms, submission.programIDs)
+    const formattedProgramNames = pNames.sort(naturalSort).map(n => n.replace(/\s/g, '-').replace(/[^a-zA-Z0-9+]/g, '').toUpperCase()).join('-')
+    return `MCR-${submission.stateCode.toUpperCase()}-${formattedProgramNames}-${padNumber}`
+}
+
```
haworku, line 125: Can you write some quick unit tests for this second function if it doesn't exist. We have some more involved logic here that would be good for testing

services/app-web/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -60,6 +59,17 @@ const StatusTag = ({status} : {status: Submission2Status}): React.ReactElement =
...
+// Pull out the programs names for display from the program IDs
+function programNames(programs: Program[], programIDs: string[]) {
+    return programIDs.map(id => {
+        const program = programs.find(p => p.id === id)
+        if (!program) {
+            return "Unknown Program"
+        }
+        return program.name
+    })
+}
```
haworku, line 71: This looks like the same function that's in the common code `submission.ts`

services/app-web/src/pages/StateSubmission/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -29,6 +30,10 @@ export const ReviewSubmit = ({
     >(undefined)
     const history = useHistory()
     const modalRef = useRef<ModalRef>(null)
+    const { loggedInUser } = useAuth()
+
+    // pull the programs off the user
+    const statePrograms = loggedInUser && 'state' in loggedInUser && loggedInUser.state.programs || []
```
haworku, line 36: Is there a case where a user at this page does not have a `state`?  If that happened, what error handling do we need?

haworku (approved):
Nice little cleanup here. I didn't run the app because I know you are still adding tests but let me know when and if you want someone to go through some submissions in a deployed app with different state users.

services/app-web/src/common-code/domain-models/submission.test.ts
```diff
@@ -290,4 +322,41 @@ describe('submission type assertions', () => {
...
+        ],
+        [
+            [
+                'abbdf9b0-c49e-4c4c-bb6f-040cb7b51cce',
+                'd95394e5-44d1-45df-8151-1cc1ee66f100',
+                'ea16a6c0-5fc6-4df8-adac-c627e76660ab',
+            ],
+            'MCR-MN-MSC+-PMAP-SNBC-0005',
+        ],
+    ])('submission name is correct', (programIDs, expectedName) => {
```
haworku, line 355: 🌮 thank you!

services/app-web/src/pages/StateSubmission/ReviewSubmit/ReviewSubmit.tsx
macrael: This page is only shown to StateUsers so I think we’re fine. But since this page and SubmissionSummary are both calling fetchSubmission2 we probably could do the same thing for both.

haworku: To clarify don't all `role=STATE_USER` already for sure have a `state`? Or is there some case we need error handling around where a state user does not have a state?

macrael: no they all should. I probably should have used an isStateUser() function. I think I’m going to put up a little cleanup PR

PR #880, opened 2022-03-29
[WIP][OY2-12726] Users can view and visits links to previous submission versions

JasonLin0991:
## Summary

WIP still need to add cypress tests.

#### Related issues
[OY2-12726](https://qmacbis.atlassian.net/browse/OY2-12726)

### DIFF
--- a/services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
+++ b/services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
@@ -1,4 +1,5 @@
 import { useEffect, useState } from 'react'
+import { useRouteMatch } from 'react-router-dom'
 import { DataDetail } from '../../../components/DataDetail'
 import { SectionHeader } from '../../../components/SectionHeader'
 import { UploadedDocumentsTable } from '../../../components/SubmissionSummarySection'
@@ -15,12 +16,12 @@ import { formatCalendarDate } from '../../../dateHelpers'
 import { DraftSubmission, StateSubmission } from '../../../gen/gqlClient'
 import { DoubleColumnGrid } from '../../DoubleColumnGrid'
 import { DownloadButton } from '../../DownloadButton'
+import { RoutesRecord } from '../../../constants/routes'
 import styles from '../SubmissionSummarySection.module.scss'
 
 export type ContractDetailsSummarySectionProps = {
     submission: DraftSubmission | StateSubmission
     navigateTo?: string
-    disableDownload?: boolean
 }
 
 const createCheckboxList = ({
@@ -52,8 +53,10 @@ const createCheckboxList = ({
 export const ContractDetailsSummarySection = ({
     submission,
     navigateTo,
-    disableDownload,
 }: ContractDetailsSummarySectionProps): React.ReactElement => {
+    //Checks if submission is a previous submission
+    const { path } = useRouteMatch()
+    const isPreviousSubmission = path === RoutesRecord.SUBMISSIONS_REVISION
     // Get the zip file for the contract
     const { getKey, getBulkDlURL } = useS3()
     useEffect(() => {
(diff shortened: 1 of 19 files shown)


services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummary.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+                <SubmissionTypeSummarySection
+                    submission={submission}
+                    showLastUpdated={false}
+                    headerChildComponent={
+                        submitInfo && (
+                            <p id="submissionVersion">{`${dayjs
+                                .utc(submitInfo?.updatedAt)
+                                .tz('America/New_York')
+                                .format('MM/DD/YY h:mma z')} ET version`}</p>
```
JasonLin0991, line 172: Do we want the timestamp to always be in ET like on the change history section?

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
```diff
@@ -47,6 +52,7 @@ const createCheckboxList = ({
 export const ContractDetailsSummarySection = ({
     submission,
     navigateTo,
+    disableDownload,
```
haworku, line 55: Rather than surface a new prop why not use the submission data  itself to determine whether or not to display the button?

I worry about dependencies between the React props and the submission data in our UI in general. Whenever possible its good to rely on the submission (or the URL) to figure out what UI to display. I suggest only add a React prop for truly _new_ data that cannot be calculated otherswise (such as the "navigateTo` link href here).

In this case I think we could make cont similar to `isSubmitted` and `isEditing` for `isPreviousSubmission`.

haworku: ^^ if we can't figure that out, I suggest naming the `disableDownload` button prop to `isPreviousSubmission` since I believe that is the context it is really capturing that is new.

services/app-web/src/pages/App/AppRoutes.tsx
```diff
@@ -68,6 +78,12 @@ const CMSUserRoutes = (): React.ReactElement => {
             <Route
                 path={RoutesRecord.SUBMISSIONS_FORM}
                 component={SubmissionSummary}
+                exact
```
haworku, line 81: yes that exact will be needed now ⭐

services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummary.module.scss
```diff
@@ -0,0 +1,32 @@
+@import '../../styles/uswdsImports.scss';
+@import '../../styles/custom.scss';
+
+.background {
+    background-color: $cms-color-white;
+    width: 100%;
+}
+
+.container {
+    max-width: 50rem; // should match ReviewSubmit max-width
```
haworku, line 10: Since we are adding a second instance of this and it relates to a third variable in ReviewSubmit this is a good chance to pull out into global css variable in `custom.scss`.  Something like `width-max-contained`  or whatever you like to name. We aren't consisted with our naming or casing in scss css fyi.

services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummary.tsx
```diff
@@ -0,0 +1,200 @@
+import React, { useEffect, useState } from 'react'
```
haworku, line 1: 🤔 💭 I'm wondering why make a separate component instead of reusing `SubmissionSummary`  tsx and referencing it at another route? Was that considered?

JasonLin0991: To me `SubmissionSummary` seemed already a bit long and complex, adding more logic plus another banner (in another ticket) would add to that complexity. Although I have no issues if we wanted to reuse `SubmissionSummary`.

haworku:
🐛 Looks like "View past submission" link is still appearing on the submission summary page.  This is good place to add tests to Cypress specifically, we definitely don't want this on the initial submission. When I click the link it tries to navigate to `https://d3oqvrrw4ih62n.cloudfront.net/submissions/0df475ed-1197-4715-b2a0-6e01a0b4629f/revisions/undefined`

PR #883, opened 2022-03-30
OY2-13581: Show date when document was submitted

rswerve:
## Summary

Closes [13581](https://qmacbis.atlassian.net/browse/OY2-13581)

We want to show the date that a document was submitted.  Had some discussion with @macrael about the best approach here, and we decided that rather than trying to wrangle types between the old submission format and the new submission format, or coming up with a third type, we can use the information that we have in the proto buffer and calculate the date for display on the fly.  We might decide later that we want the date set on the document object.

#### Screenshots

image

#### Test cases covered

Added what amounts to an integration test for SubmissionSummary, and a smaller test for the UploadedDocumentsTable, which is doing the actual date display.

## QA guidance

Note that the displayed date should be the date of submission, not upload.

And as I opened a few test files, I saw linter errors in my editor, so I addressed them, which is why you see a bunch of apparently unrelated tests in this PR.


services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -46,6 +46,10 @@ import { GenericErrorPage } from '../Errors/GenericErrorPage'
 import styles from './SubmissionSummary.module.scss'
```
haworku, line 46: This new column only supposed to appear on the main submission summary right? Or is supposed to appear on past revision summaries as well?  If so, that is separate file right now `SubmissionRevisionSummary`

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -127,6 +131,39 @@ export const SubmissionSummary = (): React.ReactElement => {
...
+                )
+                if (revisionData instanceof Error) {
+                    console.error(
+                        'failed to read submission data; unable to display document dates'
+                    )
+                    return
+                }
+                docBuckets.forEach((bucket) => {
+                    revisionData[bucket].forEach((doc) => {
+                        lookupTable[doc.name] = revisionData.updatedAt
```
haworku, line 160: Do we need to protect against unique keys at all here? I'm hoping not. What if the document with same name is added/remove in different revisions - we just keep the most recent update date right?

haworku (approved):
I see this working. Both on ReviewSubmit and SubmissionSummary. Thanks for updating tests and any lint errors

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
rswerve: I made sure our upload dupe check was relying on name before I used it here, so I think this should be fine.

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -130,6 +135,39 @@ export const SubmissionSummary = (): React.ReactElement => {
...
+        DocumentDateLookupTable | undefined
+    >({})
+
+    // const makeDateTable = (submissions: Submission2) => {
+    //     const docBuckets = [
+    //         'contractDocuments',
+    //         'rateDocuments',
+    //         'documents',
+    //     ] as const
+    //     const lookupTable = {} as DocumentDateLookupTable
```
haworku, line 150: reminder: remove this commented code

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.test.tsx
```diff
@@ -187,6 +189,46 @@ describe('SubmissionSummary', () => {
             ).toBeNull()
         })
 
+        it('extracts the correct dates from the submission and displays them in tables', async () => {
```
haworku, line 192: please add this test to the SubmissionRevisionSummary.test.tsx too cuz now we need to test both places

haworku (approved):
looks really good

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.test.tsx
rswerve: done

PR #885, opened 2022-03-30
OY2-12726: Users can view and visits links to previous submission

JasonLin0991:
## Summary
Users can view previous submissions by clicking link in change history accordion.

#### Related issues
[OY2-12726](https://qmacbis.atlassian.net/browse/OY2-12726)
[WIP PR with more context](https://github.com/CMSgov/managed-care-review/pull/880)

#### Test cases covered
- ChangeHistory.test.tsx
   - has correct href values for previous submission links
   - should list accordion items with links when appropriate

- ContractDetailsSummarySection.test.tsx
   - does not render download all button when on previous submission

- RateDetailsSummarySection.test.tsx
   - does not render download all button when on previous submission

- SubmissionTypeSummarySection.test.tsx
   - does not render Last Updated field when viewing previous submission
   - renders headerChildComponent component
   - can render all state submission type fields

- SubmissionRevisionSummary.test.tsx
   - renders correctly without errors
   
Cypress
- unlockResubmit.spec.ts
   - Can expand all change history accordion items
   - Can click and view previous submission


services/app-web/src/components/ChangeHistory/ChangeHistory.test.tsx
```diff
@@ -104,13 +149,60 @@ describe('Change History', () => {
...
+        //Previous submission should contain a link
+        expect(
+            screen.getByTestId('accordionItem_2022-03-23T02:08:52.259Z')
+        ).toHaveTextContent('View past submission version')
+    })
+    it('should not list links for initial submission without revisions', () => {
+        render(<ChangeHistory submission={submissionDataInitialSubmission} />)
+        //Initial submission should not have a link
+        expect(
+            screen.getByTestId('accordionItem_2022-03-23T02:08:52.259Z')
```
haworku, line 205: Thank you for these tests they are great 🌮  also going to help us stabilize these changes.

haworku (approved):
Thanks for the diligence here. Things look good to me for merge.  I am also going to comment on the Jira ticket about a design question I have, just FYI.

haworku:
@JasonLin0991  I noticed that we are enumerating the revision version in the urls from zero. e.g. https://d1uhxoxbnqp0yf.cloudfront.net/submissions/22e4f3d6-d32d-4a54-86a1-abb536564c33/revisions/0 is an initial submission. And /1 is the first re submission after that. Was that an intentional design update or just accidental? 

 In the comment on the ticket from @MegMurphy I understood we were starting with 1.

JasonLin0991:
> @JasonLin0991 I noticed that we are enumerating the revision version in the urls from zero. e.g. https://d1uhxoxbnqp0yf.cloudfront.net/submissions/22e4f3d6-d32d-4a54-86a1-abb536564c33/revisions/0 is an initial submission. And /1 is the first re submission after that. Was that an intentional design update or just accidental?
> 
> In the comment on the ticket from @MegMurphy I understood we were starting with 1.

This was intentional, but after reading the ticket again I see initial submission should be 1.

MegMurphy:
@JasonLin0991 @haworku that's a good question. I hadn't really considered if the original submission should start at 0 or 1. I think either way would work. I don't think users will realistically catch on and change the URL directly to navigate from version to version, but if they did, I think 1 is more intuitive as the beginning of a series from an end user perspective. But like I said, I don't really expect users to navigate in that way, so I think if there is a preference on the engineering side to do one way over the other, that would be fine.

JasonLin0991:
> @JasonLin0991 @haworku that's a good question. I hadn't really considered if the original submission should start at 0 or 1. I think either way would work. I don't think users will realistically catch on and change the URL directly to navigate from version to version, but if they did, I think 1 is more intuitive as the beginning of a series from an end user perspective. But like I said, I don't really expect users to navigate in that way, so I think if there is a preference on the engineering side to do one way over the other, that would be fine.

@MegMurphy @haworku I'll just leave as is with original submission starting at 0. Thanks for clarifying!

PR #887, opened 2022-03-30
 Fix delay when modal closes to redirect when resubmitting

JasonLin0991:
## Summary
Fix to address delay after modal closes and redirect when resubmitting a submission. 

**Note: **Also add code to close Unlock Modal on error. Does not seem to make sense to keep modal open and not see the error banner, but let me know otherwise.

#### Related issues
[OY2-12183](https://qmacbis.atlassian.net/browse/OY2-12183?focusedCommentId=108213)

### DIFF
--- a/services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
+++ b/services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
@@ -40,7 +40,10 @@ import {
     useFetchSubmission2Query,
     useUnlockStateSubmissionMutation,
 } from '../../gen/gqlClient'
-import { convertDomainModelFormDataToGQLSubmission, isGraphQLErrors } from '../../gqlHelpers'
+import {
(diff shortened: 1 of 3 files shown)


services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -255,7 +256,6 @@ export const SubmissionSummary = (): React.ReactElement => {
 
     const onModalSubmit = async (values: typeof modalFormInitialValues) => {
         const { unlockReason } = values
-        modalRef.current?.toggleModal(undefined, false)
```
haworku, line 258: This was a bug fix for something @rswerve found which was if you dont close the modal immediately its possible for someone to unlock twice and just interact with the form more while submit is in progress. 

Rather than keep the modal open longer here to fix delay to the redirect we could do something like show the loading page instead while things are delayed for an async request

haworku (approved):

PR #888, opened 2022-03-31
OY2-1477: User is alerted when they are viewing a previous submission version

JasonLin0991:
## Summary
[OY2-14777](https://qmacbis.atlassian.net/browse/OY2-14777)

Adds a new banner to SubmissionRevisionSummary page that alerts the user they are on a previous submission. An inlcuded link in the banner will take the user back to the current submission.


tests/cypress/support/submissionCommands.ts
```diff
@@ -0,0 +1,26 @@
+import { aliasQuery } from '../utils/graphql-test-utils'
+
+Cypress.Commands.add(
+    'navigateBackToCurrentSubmission',
+    () => {
```
haworku, line 5: This is a good helper 🌮 we will def reuse this

haworku (approved):
Looks good based on code. Good use of testID to ensure the right revision link is being clicked. Once deploy completes and tests pass good to merge

PR #890, opened 2022-03-31
Submission revisions start from 1 instead of 0

JasonLin0991:
## Summary
Change versions of submissions to start at 1


macrael (approved):
Easy

PR #895, opened 2022-04-02
MR-392: Add a NEW tag to newly uploaded documents

rswerve:
## Summary

Closes[392](https://qmacbis.atlassian.net/browse/MR-392)

Adds a NEW tag to newly uploaded docs, provided the user is a CMS user.

#### Screenshots

image

#### Test cases covered

Should see NEW when a doc is newer than the second submission (the first submission is the current one; the doc will be part of the current submission).
Should NOT see NEW if it's not a CMS user.
Updated the lookup table, which now has the date of the second submission.

## QA guidance

Not much.  Should see that tag if you're logged in as a CMS user and a doc has been uploaded since the last submission, and not if not.

### DIFF
--- a/services/app-web/src/components/SubmissionSummarySection/UploadedDocumentsTable/UploadedDocumentsTable.module.scss
+++ b/services/app-web/src/components/SubmissionSummarySection/UploadedDocumentsTable/UploadedDocumentsTable.module.scss
@@ -38,3 +38,14 @@
     display: flex;
     justify-content: space-between;
 }
+
+.newDocTag {
+    background-color: #009ec1;
+    display: inline-flex;
+    flex-direction: row;
+    align-items: center;
+    padding: 1px 8px;
+    border-radius: 2px;
+    height: 18px;
+    margin-right: 5px !important;
+}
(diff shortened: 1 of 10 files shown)


services/app-web/src/components/SubmissionSummarySection/UploadedDocumentsTable/UploadedDocumentsTable.module.scss
```diff
@@ -38,3 +38,14 @@
     display: flex;
     justify-content: space-between;
 }
+
+.newDocTag {
+    background-color: #009ec1;
```
rswerve, line 43: @haworku I couldn't get `$theme-color-info-dark` to work, even when I uncommented the entry in the master colors file.  Not sure what the problem is there.

haworku (approved):
Looks good.

services/app-web/src/components/SubmissionSummarySection/UploadedDocumentsTable/UploadedDocumentsTable.module.scss
haworku: Here is the new issue https://qmacbis.atlassian.net/browse/MR-2067

PR #899, opened 2022-04-04
MR-332: Users can see rate name on submission summary

JasonLin0991:
## Summary
[MR-332](https://qmacbis.atlassian.net/browse/MR-332)

Users can see rate name on submission summary

#### Test cases covered

- RateDetailSummarySection.tsx
   - <strike>can render all rate details fields for amendment to prior rate certification submission **(Updated)**</strike>
   - <strike>can render all rate details fields for new rate certification submission **(Updated)**</strike>
   - can render correct rate name for new rate submission
   - can render correct rate name for AMENDMENT rate submission

### DIFF
--- a/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.test.tsx
+++ b/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.test.tsx
@@ -8,7 +8,7 @@ import { RateDetailsSummarySection } from './RateDetailsSummarySection'
 import { createMemoryHistory } from 'history'
 import { Route } from 'react-router'
 import { RoutesRecord } from '../../../constants/routes'
-import { rateDateString } from './RateDetailsSummarySection'
+import { formatRateNameDate } from '../../../dateHelpers'
 
 describe('RateDetailsSummarySection', () => {
     const draftSubmission = mockContractAndRatesDraft()
@@ -91,9 +91,13 @@ describe('RateDetailsSummarySection', () => {
         )
         rateID.toBeInTheDocument()
         rateID.toHaveTextContent(
-            `MN-MSHO-0003-RATE-${rateDateString(new Date())}-${rateDateString(
-                new Date()
-            )}-CERTIFICATION-${rateDateString(new Date())}`
+            `MN-MSHO-0003-RATE-${formatRateNameDate(
+                submission.rateDateStart
+            )}-${formatRateNameDate(
+                submission.rateDateEnd
+            )}-CERTIFICATION-${formatRateNameDate(
+                submission.rateDateCertified
+            )}`
         )
     })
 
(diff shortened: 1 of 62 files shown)


services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
```diff
@@ -75,6 +109,12 @@ export const RateDetailsSummarySection = ({
                     )}
                 </SectionHeader>
 
+                <div aria-label="Rate ID" id="rateID" role="group">
```
haworku, line 112: A couple things to check here
- What is the intended meaning here of the rate id? Is it information meant to be a [heading](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) or not? 
- IF it is, the only permitted aria roles for that element from the linked guide are a "heading", [tab](https://w3c.github.io/aria/#tab), [presentation](https://w3c.github.io/aria/#presentation) or [none](https://w3c.github.io/aria/#none). `role=definition` is not allowed.
- If you are using heading here mainly to get the heading-like styles, use CSS for that instead, not HTML

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.test.tsx
```diff
@@ -75,6 +76,65 @@ describe('RateDetailsSummarySection', () => {
...
+            )}-${formatRateNameDate(
+                submission.rateDateEnd
+            )}-CERTIFICATION-${formatRateNameDate(
+                submission.rateDateCertified
+            )}`
+        )
+    })
+
+    it('can render correct rate name for AMENDMENT rate submission', () => {
+        const submission = {
```
haworku, line 105: good test!

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
JasonLin0991: I saw this mock up and a h4, though we wanted a h4 element. I'll change this to use a div with CSS.

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
```diff
@@ -35,6 +35,40 @@ export const RateDetailsSummarySection = ({
         doc.documentCategories.includes('RATES_RELATED')
     )
 
+    let rateName: string
+
+    const generateRateName = (
```
macrael, line 40: We’re you talkinga about putting this in common-code? Maybe that’s a follow on PR with the emails I’m remembering?

JasonLin0991: Yes, this will be done in the rate name in email PR.

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
haworku: just writing here what we discussed IRL for documentations - if a heading is being used it should be h3 since we want to only use headings in order (otherwise its an a11y fail). The Rate Details above is an h2 so if the rate name is a heading (I think we decided yes that makes sense, especially since we forsee the multi rate world coming) it should be an h3.  If the heading isnt clear via screenreader as read, an `aria-label` can be used.

haworku (approved):

PR #902, opened 2022-04-08
MR-330 & MR-334: Cms and state sees rate name on new submission email

JasonLin0991:


services/app-web/src/common-code/domain-models/submission.ts
```diff
@@ -135,6 +136,64 @@ function submissionName(
     return `MCR-${submission.stateCode.toUpperCase()}-${formattedProgramNames}-${padNumber}`
 }
 
+export type RateDataType = {
```
macrael, line 139: Exactly what I was talking about. This is one of the best parts of Typescript is that it's easy to make a little protocol. If we ever change our other types to not fulfill this we'll get errors.

services/app-web/src/common-code/domain-models/submission.ts
```diff
@@ -135,6 +136,64 @@ function submissionName(
...
+    rateDateEnd?: Date
+    rateDateCertified?: Date
+    rateAmendmentInfo?: {
+        effectiveDateEnd?: Date
+        effectiveDateStart?: Date
+    } | null
+}
+
+export const generateRateName = (
+    rateData: Partial<RateDataType>,
```
macrael, line 151: Why is this `Partial`? It seems like everything can already be null above. The question marks should make things optional.

services/app-web/src/common-code/domain-models/submission.ts
```diff
@@ -135,6 +136,64 @@ function submissionName(
...
+    rateDateStart?: Date
+    rateDateEnd?: Date
+    rateDateCertified?: Date
+    rateAmendmentInfo?: {
+        effectiveDateEnd?: Date
+        effectiveDateStart?: Date
+    } | null
+}
+
+export const generateRateName = (
```
macrael, line 150: This would be a good function to write some table tests for. Take a look at the "Submission name is correct" test in submission.test.ts

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
```diff
@@ -4,11 +4,12 @@ import { SectionHeader } from '../../../components/SectionHeader'
 import { UploadedDocumentsTable } from '../../../components/SubmissionSummarySection'
 import { DocumentDateLookupTable } from '../../../pages/SubmissionSummary/SubmissionSummary'
 import { useS3 } from '../../../contexts/S3Context'
-import { formatCalendarDate, formatRateNameDate } from '../../../dateHelpers'
+import { formatCalendarDate } from '../../../dateHelpers'
 import { DraftSubmission, StateSubmission } from '../../../gen/gqlClient'
 import { DoubleColumnGrid } from '../../DoubleColumnGrid'
 import { DownloadButton } from '../../DownloadButton'
 import { usePreviousSubmission } from '../../../hooks/usePreviousSubmission'
+import { generateRateName } from '../../../common-code/domain-models/submission'
```
macrael, line 12: This should be exported from common-code/domain-models, we're trying not to import stuff directly from the files themselves. So this needs to be added to the domain-models/index.ts

services/app-api/emailer/index.ts
```diff
@@ -8,6 +8,7 @@ export {
     resubmittedStateEmail,
     resubmittedCMSEmail,
     UpdatedEmailData,
+    generateRateName,
```
macrael, line 11: This should be exported by the domain-models package, not the emailer.

services/app-api/emailer/templates.ts
```diff
@@ -27,24 +29,17 @@ const stripHTMLFromTemplate = (template: string) => {
...
-    )
+    formatted = formatted.replace(/<a.*href="(.*?)".*>(.*?)<\/a>/gi, ' $2 ($1)')
 
     // everything else
-   return formatted.replace(/(<([^>]+)>)/gi, '')
+    return formatted.replace(/(<([^>]+)>)/gi, '')
 }
 
-const newPackageCMSEmail = (
+const generateNewSubmissionData = (
```
macrael, line 38: "Data" is a big generic of a name here. What is actually being generated? would "Body" be correct?

macrael (approved):
Great work!

I think that the templates.ts file is starting to get a bit big, at some point we should split that up into separate files, but I don't think that needs to happen for this PR.

services/app-web/src/common-code/domain-models/submission.ts
JasonLin0991: Yes, totally forgot to remove this when switching to `RateDataType`.

services/app-web/src/common-code/domain-models/submission.test.ts
```diff
@@ -359,4 +361,45 @@ describe('submission type assertions', () => {
...
+            },
+            submissionName: 'OH-TEST-NAME',
+            expectedName:
+                'OH-TEST-NAME-RATE-20210422-20220329-CERTIFICATION-20210523',
+        },
+    ]
+    test.each(rateNameTestArray)(
+        'submission rate name is correct',
+        ({ rateData, submissionName, expectedName }) => {
+            expect(generateRateName(rateData, submissionName)).toMatch(
```
JasonLin0991, line 400: I noticed that when testing the "table", if the first row passes, the entire test passes regardless if subsequent rows do not match. This happens with `submission name is correct`, change the last table row `expectedName` to something that will never match the function, the entire test will still pass.

Not sure whats going on. Any insights @macrael @haworku @rswerve ?

macrael: Hm. locally I was able to change the last test in the list for “submission name is correct” and get an error.
Screen Shot 2022-04-11 at 2 32 31 PM

macrael: is there a reason you are using `toMatch()` instead of `toBe()` here?

JasonLin0991: Oops, it was me the whole time! False alarm. I need to run these tests with `yarn test` and not using the built in Webstorm tool for jest.

services/app-api/emailer/templates.ts
haworku: This is not making the body, its making subsection of the body string. This smells very much like a nested template - an email template calling another template inside of it. We can easily create un-forseen dependencies between templates, which I want to avoid. 

We want to be able to control and change each template as a unit without dependencies between the markdown.  To me repetition in a template is okay and even expected.

The code sharing of markup here (and not only logic for individual variables) suggests to me that we def have way too much complexity in the email HTML strings to handle with javascript alone. Its feeling verbose and we want to refactor.

 How about we move towards mustache templates with shared variables ASAP rather than created shared template functions? That would be my suggestion here. I would back out this change, keep the verbose code, and then work on adding mustache templates instead. I don't see a backlog ticket about this I will create. I think we should probably prioritize we have gone too long without a templater tool

macrael: I love this idea. We can also break out the templates into their own files to make testing/comprehension easier. Do you want to block this PR on this or have that be a refactor Jason can start on after this merges?

haworku: or we can just merge since looks this is almost done ✅ . That's fine too. Just want to warn against us hacking javascript much further to do something that really is for templating tools.

tests/cypress/integration/cmsWorkflow/unlockResubmit.spec.ts
```diff
@@ -129,9 +129,6 @@ describe('dashboard', () => {
                 cy.wait('@fetchSubmission2Query', { timeout: 50000 })
                 cy.findByTestId('updatedSubmissionBanner').should('exist')
 
-                cy.findByText('Back to state dashboard').click()
```
haworku, line 132: was this flaking or what was happening here?

haworku (approved):

tests/cypress/integration/cmsWorkflow/unlockResubmit.spec.ts
JasonLin0991: This was something was testing before @macrael fixed the issue with signing out.

PR #903, opened 2022-04-08
Add LaunchDarkly to /health API endpoint

mojotalantikite:
## Summary
This adds LD to the `/health` api endpoint. It includes the appropriate secrets for each environment and is mainly for us to test enabling/disabling something via LD.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-402

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,5 @@
 {
-    "eslint.workingDirectories": ["services/app-web", "services/app-api"]
+    "eslint.workingDirectories": ["services/app-web", "services/app-api"],
+    "launchdarkly.project": "default",
+    "launchdarkly.env": "dev"
 }
--- a/services/app-api/handlers/health_check.ts
+++ b/services/app-api/handlers/health_check.ts
@@ -1,13 +1,30 @@
 import { APIGatewayProxyHandler } from 'aws-lambda'
+import LaunchDarkly from 'launchdarkly-node-server-sdk'
+
+const ldClientKey = process.env.LD_SDK_KEY ?? ''
+if (ldClientKey === '') {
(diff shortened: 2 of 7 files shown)


.vscode/settings.json
```diff
@@ -1,3 +1,5 @@
 {
-    "eslint.workingDirectories": ["services/app-web", "services/app-api"]
+    "eslint.workingDirectories": ["services/app-web", "services/app-api"],
+    "launchdarkly.project": "default",
```
macrael, line 3: Is there a VSCode plugin related to LD?

macrael (approved):
Cool!

.vscode/settings.json
mojotalantikite: There is, it let's you do things like see flag toggle state and whatnot from VSCode: https://docs.launchdarkly.com/integrations/vscode#installing-the-vscode-extension

services/app-api/handlers/health_check.ts
```diff
@@ -1,14 +1,38 @@
...
+import LaunchDarkly from 'launchdarkly-node-server-sdk'
+
+const ldClientKey = process.env.LD_SDK_KEY ?? ''
+if (ldClientKey === '') {
+    throw new Error('LD_SDK_KEY environment variable is not set')
+}
 
 export const main: APIGatewayProxyHandler = async () => {
+    const ldClient = LaunchDarkly.init(ldClientKey)
+    await ldClient.waitForInitialization()
```
rswerve, line 11: Do we know what happens here if LD doesn't respond?

mojotalantikite: Ah yeah, it'll promise reject -- I just pushed a change to try/catch it.

PR #912, opened 2022-04-11
Add more encryption policies

mojotalantikite:
## Summary
This is a follow on for the s3 bucket security hub findings we have in our environment. This adds a rule that denies unencrypted traffic from our buckets (not that we have any anyways).

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2066

### DIFF
--- a/services/storybook/serverless.yml
+++ b/services/storybook/serverless.yml
@@ -68,6 +68,16 @@ resources:
                   Fn::GetAtt:
                     - CloudFrontOriginAccessIdentity
                     - S3CanonicalUserId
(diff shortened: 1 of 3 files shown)


macrael (approved):
Which buckets are having their policies changed by the s3 bucket helper?

mojotalantikite:
> Which buckets are having their policies changed by the s3 bucket helper?

They all should be, it's on in all of our serverless configs. It only enforces versioning and blocking public access though.

Probably an opportunity for it to enforce the encryption policies.

PR #913, opened 2022-04-11
Add LD_SDK_KEY env variable to app-api serverless

mojotalantikite:
## Summary

This threads in the LD SDK key into one last place. There are separate ones for dev/val/prod added to CI already.


macrael (approved):
yup. so many places to thread

macrael:
Did this work in the review app?

PR #918, opened 2022-04-15
MR-2081: State and CMS have no character limit on their unlock or resubmission reason

JasonLin0991:
## Summary
[MR-2081](https://qmacbis.atlassian.net/browse/MR-2081)

- Unlock reason and resubmission summary should no longer have a maximum character limit.

- Unlock and Updated banner should have the functionality to expand and contract the `Reason for unlock:` and 
`Changes made:` text.

**NOTES**
- I'm pretty sure the ability to pass in ReactElement for `clamp` will lead to strange things when passing in complicated elements. The reason for this was to add the bold text to `Reason for unlock:` and `Changes made:` text. There is probably a better way.

#### Screenshots
Screen Shot 2022-04-15 at 5 37 24 AM
Screen Shot 2022-04-15 at 5 37 45 AM

Screen Shot 2022-04-15 at 5 38 02 AM
Screen Shot 2022-04-15 at 5 38 18 AM

#### Test cases covered

- ExpandableText.test.tsx
   - renders without errors and clamps long text
   - can expand and contract clamped text to show more or less text
   - renders short text without errors and not clamped
   - can render react elements and clamp long text correctly
   - can expand and contract clamped text inside react element to show more or less text

- Remove testing of textarea character limits in unlock and resubmission modals.
   - ReviewSubmit.test.tsx
   - SubmissionSummary.test.tsx

#### QA Guidence

- VO should ignore the `Show More` and `Show Less` buttons and read the entire text even if clamped.
- Positioning of the `Show More` and `Show Less` buttons are positioned like the mock up on Jira, but was given the ok. Attempting to position the buttons like the mock up along with clamping at whole sentences, for me, would have required a lot of code.


services/app-web/src/components/ExpandableText/ExpandableText.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { useLayoutEffect, useState } from 'react'
+import styles from './ExpandableText.module.scss'
+
+export type ExpandableTextProps = {
+    clamp: React.ReactElement | string
+    clampedLines?: string
+}
```
JasonLin0991, line 7: I'm pretty sure the ability to pass in ReactElement for clamp will lead to strange things when passing in complicated elements. The reason for this was to add the bold text to Reason for unlock: and Changes made: text. There is probably a better way.

haworku:
I have a small change I'm going to push up to add aria labels for read more, adjust types for `clampedLines`, and maybe change the type of `clamp` back to string. Going to check out the bold text issue you mentioned

haworku:
@JasonLin0991  I see the Firefox issue you mentioned. I don't see a quick workaround  Reading a bit more it might be that the `-webkit` values are not fully supported https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp . But when I used `-moz` values for firefox I wasn't seeing the issue resolved.

To me, this should be raised in a comment on the jira ticket to discuss with product and design. Do they feel comfortable reviewing and merging this work as is... with the firefox bug filed in our backlog?  If not, I think we need pursue a lineclamp solution that works for firefox next week to close out this work.

JasonLin0991:
> @JasonLin0991 I see the Firefox issue you mentioned on storybook. I don't see a quick workaround. Reading a bit more it might be that the `-webkit` values are not fully supported https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp . But when I used `-moz` values for firefox I wasn't seeing the issue resolved. Also the css `line-clamp` doesn't work for me although I would like to explore more...
> 
> ~To me, this should be raised in a comment on the jira ticket to discuss with product and design. Do they feel comfortable reviewing and merging this work as is... with the firefox bug filed in our backlog? If not, I think we need pursue a lineclamp solution that works for firefox next week to close out this work.~
> 
> @JasonLin0991 😮‍💨 just realized that this isn't happening in the deployed version in the app which you said. I think this is fine if its storybook only. Thanks for putting a comment in there

Awesome. I probably should have mentioned this in the PR description and my wording could have been better 😅.

haworku:
In regards to the how to Actions tab on storybook is flooded with actions. I've noticed storybook is very slow, almost crashing, when I view this story. I wonder if we have a memory leak in the React state. Going to look more at things here.

services/app-web/src/components/ExpandableText/ExpandableText.tsx
```diff
@@ -0,0 +1,61 @@
...
+                    showMore ? styles.textExpanded : styles.textContracted
+                }`}
+            >
+                {typeof clamp === 'string' ? <span>{clamp}</span> : clamp}
+            </span>
+            {showMoreButton && (
+                <Button
+                    aria-hidden // hide from screenreaders
+                    type="button"
+                    unstyled
```
JasonLin0991, line 51: @haworku I saw we changed this to the React USWDS Button component. The outline styling for this button partially covers up the last line of text. Do we want to add top `margin` to the button or reduce the `outline-offset` to 0?

services/app-web/src/components/ExpandableText/ExpandableText.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Button } from '@trussworks/react-uswds'
+import styles from './ExpandableText.module.scss'
+
+export type ExpandableTextProps = {
+    clamp: React.ReactElement | string // you can pass in an element with nested <span> or simple string
+    clampedLines?: number
+}
+
+export const ExpandableText = ({
+    clamp,
```
macrael, line 11: Since this component wraps other components, could we use children here instead of passing react elements in as props?

services/app-web/src/components/ExpandableText/ExpandableText.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+    useLayoutEffect(() => {
+        const ref = textRef?.current
+        if (
+            ref &&
+            (ref.offsetHeight < ref.scrollHeight ||
+                ref.offsetWidth < ref.scrollWidth)
+        ) {
+            if (clampedLines) {
+                ref.style.webkitLineClamp = clampedLines.toString()
```
macrael, line 29: neat! I didn't know this was a browser built-in

macrael (approved):

services/app-web/src/components/ExpandableText/ExpandableText.tsx
JasonLin0991: We could, this was where I'm unsure if we want to pass elements at all ( only strings ). I got stuck on how to add bold text to the `Reason for unlock:` and `Changes made:` text with out passing in elements. Being able to pass any element(s) causes unexpected/unintentional clamping results.

services/app-web/src/components/ExpandableText/ExpandableText.tsx
```diff
@@ -0,0 +1,61 @@
...
+            }
+            setShowMoreButton(true)
+        }
+    }, [showMoreButton, clampedLines, setShowMoreButton])
+
+    return (
+        <p {...props}>
+            <span
+                ref={textRef}
+                data-testid="clampElement"
```
JasonLin0991, line 39: > In regards to the how to Actions tab on storybook is flooded with actions. I've noticed storybook is very slow, almost crashing, when I view this story. I wonder if we have a memory leak in the React state. Going to look more at things here on Monday. I don't want to merge this work if there's some kind of performance issue going on that will make Storybook hard to use.

It's the spreading of props `{...props}` that's causing this issue. I don't think it's needed here or in the Updated and Unlocked banners. If there is no objections I can remove that

services/app-web/src/components/ExpandableText/ExpandableText.tsx
haworku: fwiw I did look into only using a string and I agree we may have to pass elements to match the current design, like you noticed @JasonLin0991 .  I also  added a comment about only using <span>s in my commit. You are right that it could have unintentional results if complex nested HTML is passed, but I think having a comment is the best we can do while still making this component flexible

How do you feel the proposal to render children instead of using a prop @JasonLin0991 ?

services/app-web/src/components/ExpandableText/ExpandableText.tsx
haworku: Huh, yeah that's fine to remove we aren't even passing anything in there lol

services/app-web/src/components/ExpandableText/ExpandableText.tsx
haworku: For chrome/safari its been there for awhile, firefox added support a lot later and it seems like its not fully there it seems Jason saw weirdness with this even in Storybook.  Eventually CSS will have `line-clamp` officially (without need for browser specific polyfill) but its out of [draft](https://www.w3.org/TR/css-overflow-3/#line-clamp) yet.

services/app-web/src/components/ExpandableText/ExpandableText.tsx
JasonLin0991: `How do you feel the proposal to render children instead of using a prop @JasonLin0991 ?`

Works for me!

haworku (approved):

PR #923, opened 2022-04-18
Fix apollo cache usage

macrael:
## Summary

After renaming everything, I have cleaned up some of our apollo cache usage to be consistent.

Apollo uses its cache by default to respond to requests if there is data in the cache that matches the query. So, for instance, in our app, moving between different pages of the app doesn’t cause new requests to fetch the HealthPlanPackage every time because useFetchHPPQuery() knows that it is asking for a single record with the given ID and that already exists in the cache. 

That’s quite straightforward for queries, if we’ve queried it before its in the cache and won’t be re-queried, but mutations can modify the actual state of things, so how does that work?

By default, if your mutation return all of the mutated objects, they are inserted into the cache automatically. So, now, with our update, submit, and unlock mutations, we don’t need to configure the calls at all for the cache. those endpoints return the object they modify in full so Apollo puts the new data in the cache. 

The only part that doesn’t work automatically for us is updating the indexHPPQuery. When we create a new package, that individual package is correctly inserted into the cache (though even so, we do re-fetch when the app navigates to the ContractDetails page, I suspect since that query itself hasn’t happened before it always fetches but I don’t know). But even though that single object is in the cache, apollo doesn’t try to be clever enough to update other queries that might reference that new object, so we have to explicitly update the cached indexHPPQuery to include the newly create package.

### DIFF
--- a/services/app-web/src/components/ChangeHistory/ChangeHistory.test.tsx
+++ b/services/app-web/src/components/ChangeHistory/ChangeHistory.test.tsx
@@ -30,7 +30,7 @@ const submissionData: HealthPlanPackage = {
                     __typename: 'UpdateInformation',
                 },
                 createdAt: '2022-03-25T01:18:44.665Z',
-                submissionData: 'qpoiuenad',
+                formDataProto: 'qpoiuenad',
                 __typename: 'HealthPlanRevision',
             },
             __typename: 'HealthPlanRevisionEdge',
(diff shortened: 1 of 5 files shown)


haworku:
Are there any places we intentionally don't want to cache? For example, if an action occurs on a submission by another user (such as an unlock or resubmit) while the current user is in the dashboard, at what point will that user get the updated submission. Is it when they fetch that submission by clicking into it? Do we have an cases with the multiple user experience we want to test here, I'm happy to jump in.

haworku (approved):

haworku:
@macrael can you write that ticket out or shall I?

macrael:
@haworku I can write it

PR #928, opened 2022-04-20
MR-2096: Documents removed from edited submissions not downloadable on previous submission

JasonLin0991:
## Summary
Only delete files from s3 when file does not exists in previous submission revisions. This ensures that previous submission documents will still be available for download/view while still being able to delete newly uploaded files on a resubmission from s3.

#### Test cases covered

- makeDocumentKeyLookupList.test.ts
   - should make two lists with document s3 keys
   - should return empty arrays for no documents in submission

- ContractDetails.test.tsx
   - Updated tests with new prop `handleDeleteFile`
- RateDetails.test.tsx
   - Updated tests with new prop `handleDeleteFile`
- Documents.tests.tsx
   - Updated tests with new prop `handleDeleteFile`
   - Merged delete file test from MacRae's [PR](https://github.com/CMSgov/managed-care-review/pull/932)
   - Added `does delete files not in previous revisions` test.


services/app-web/src/s3/s3Amplify.ts
```diff
@@ -62,22 +64,38 @@ function newAmplifyS3Client(bucketName: string): S3ClientT {
...
-                await Storage.remove(filename)
-                return
-            } catch (err) {
-                assertIsS3PutError(err)
-                if (err.name === 'Error' && err.message === 'Network Error') {
-                    console.log('Error deleting file', err)
-                    return {
-                        code: 'NETWORK_ERROR',
-                        message: 'Error deleting file from the cloud.',
+        deleteFile: async (
```
macrael, line 67: I don't think this checking logic belongs here. One reason this sticks out to me is that you needed to duplicate the logic in our Amplify and our Local deleteFile implementations. Another reason is that this s3 package is a wrapper around a dependency, we want this code to expose the functionality of the dependency and not much more. Logic like when to delete a file belongs in our apps "business logic" layer rather than a dependency wrapper.

services/app-web/src/s3/s3Amplify.ts
```diff
@@ -62,22 +64,38 @@ function newAmplifyS3Client(bucketName: string): S3ClientT {
...
+            filename: string,
+            planPackageStatus?: HealthPlanPackageStatusType,
+            fileItems?: FileItemT[]
+        ): Promise<void | S3Error> => {
+            // Only delete file from S3 bucket if submission status is DRAFT or if file has NOT been submitted before.
+            // Meaning it was uploaded, but health plan package has not been submitted/resubmitted yet.
+            const isSubmittedFile =
```
macrael, line 74: So, I don't think this solves our problem exactly. We need to be checking if a file has _ever_ been uploaded for this HealthPlanPackage, not just the most recent revision.

macrael (commented):
I don't think this PR is solving the bug exactly. I think the correct logic for determining if a file can be deleted is "if this file exists in any previous revisions of this package then it must not be deleted from S3" 

What this PR really needs is some tests where we can throw different HealthPlanPackages at the page and see what happens when we click delete. LMK if you want to pair on building that.

services/app-web/src/common-code/data-helpers/makeDocumentKeyLookupList.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (submissions) {
+        const revisions = submissions.revisions
+        if (revisions.length > 1) {
+            revisions.forEach((revision, index) => {
+                const revisionData = base64ToDomain(revision.node.formDataProto)
+                if (revisionData instanceof Error) {
+                    console.error(
+                        'failed to read submission data; unable to display document dates'
+                    )
+                    return
```
macrael, line 36: There are a couple different options of what to do if we fail to unbox the revision proto. 
1. abort and return an error. If we want a user to see an error if any of the previous revisions don’t decode, this function should return an error directly so that calling code can decide how to surface that. 
2. Ignore the documents from that revision and continue, returning all the documents we are able to decode. This is a little attractive to me b/c as is we know we have data that can become invalid and we don’t really want to lose this whole feature just because an old revision can’t be read anymore, that said, it would mean that a user might get some funky unexpected file-deletion behavior, which seems bad

I think we should probably go with option 1 for now, and treat these errors as very unexpected. If that bites us we can always relax later.

services/app-web/src/common-code/data-helpers/makeDocumentKeyLookupList.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const getKey = (s3URL: string) => {
+    const key = parseKey(s3URL)
+    return key instanceof Error ? null : key
+}
+
+export const makeDocumentList = (
+    submissions: HealthPlanPackage
+): LookupListType | undefined => {
```
macrael, line 17: Is it possible for this function to return undefied? Seems like it would always return an empty pair of arrays

services/app-web/src/common-code/data-helpers/makeDocumentKeyLookupList.ts
```diff
@@ -0,0 +1,52 @@
...
+    const docBuckets = [
+        'contractDocuments',
+        'rateDocuments',
+        'documents',
+    ] as const
+    const lookupList: LookupListType = {
+        currentDocuments: [],
+        previousDocuments: [],
+    }
+    if (submissions) {
```
macrael, line 27: It looks like submissions is not optional so we probably dont’ need this if

services/app-web/src/common-code/data-helpers/makeDocumentKeyLookupList.ts
```diff
@@ -0,0 +1,52 @@
...
+        const revisions = submissions.revisions
+        if (revisions.length > 1) {
```
macrael, line 29: I’m curious about this if, too, there’s no harm in calling .forEach() on an empty array. 

In general I’ve been prefering the `for (const revision of revisions) {}` syntax, but I don’t know if you can get an index out of that. I think it might be a little more elegant to make a separate function that takes a revision and an array to dump its documents into and call that differently for the first and then the rest of the reivisions.

services/app-web/src/pages/StateSubmission/Documents/Documents.test.tsx
```diff
@@ -357,6 +385,85 @@ describe('Documents', () => {
         })
     })
 
+    describe('the delete button', () => {
+        // For this test, we want to mock the call to deleteFile to see when it gets called
+        const mockS3 = testS3Client()
+        const deleteCallKeys: string[] = []
+        mockS3.deleteFile = async (key) => {
+            console.log('MOCK DELETE CALLED', key)
```
macrael, line 393: probably dont need this console.log anymore.

services/app-web/src/pages/StateSubmission/Documents/Documents.test.tsx
```diff
@@ -357,6 +385,85 @@ describe('Documents', () => {
         })
     })
 
+    describe('the delete button', () => {
```
macrael, line 388: Hana said she’d prefer this test live in StateSubmissionForm.test.tsx instead of here since that’s the actual component under test.

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
```diff
@@ -165,7 +170,24 @@ export const StateSubmissionForm = (): React.ReactElement => {
...
+        const isSubmittedFile =
+            previousDocuments &&
+            Boolean(
+                previousDocuments.some((previousKey) => previousKey === key)
+            )
+
+        if (!isSubmittedFile) {
+            const result = await deleteFile(key)
+            if (isS3Error(result)) {
+                throw new Error(`Error in S3 key: ${key}`)
```
macrael, line 184: what happens if we throw here? I think we’d want to display an error allowing for the deletion to be re-tried or something?

services/app-web/src/pages/StateSubmission/Documents/Documents.test.tsx
```diff
@@ -357,6 +385,85 @@ describe('Documents', () => {
...
+            )
+
+            // PERFORM
+
+            // we should be able to find delete buttons for each of the three recent files.
+            // the aria label for each button is a lifesaver here.
+            const removeOneTwo = await screen.findByLabelText(
+                'Remove one two document'
+            )
+            userEvent.click(removeOneTwo)
```
macrael, line 427: I’d add a couple more clicks here to make sure that it works across different revisions. Part of the reason I built the mock to write to an array is you could delete all the files and then assert that only the ones you expected got deleted.

services/app-web/src/common-code/data-helpers/makeDocumentKeyLookupList.test.ts
```diff
@@ -0,0 +1,44 @@
+import { makeDocumentList } from './makeDocumentKeyLookupList'
```
macrael, line 1: In my refactor I moved Maz’s document table code out of /common-code/data-helpers and just put it in pages/SubmissionSummary directly. 

This for sure doesn’t belon inside of /common-code b/c I’m trying to only put code in there that is imported by the front end and the back end, and this code operates on the GQL types so it wouldn’t make sense on the backend anyway. 

Maybe something like “documentsHelpers” could be a package in /src that could have your code and Maz’s in it?

macrael (approved):
Aweseome changes. I had most of my comments around the error handling, and I think that probably deserves some tests to ensure that if there is an error in deleting that we’re displaying the right message to the user. 

Passing the handle delete function down into the individual components from StateSubmissionForm feels a little weird to me because it moves our s3 calling code up into a component that didn’t previously have to care about it. I think you could accomplish the same thing by passing the list of previous files down into those components instead, but it’s not a blocker for me if you like this way best. I suspect we need to do more consolidation at some point of the share documents component to get that logic into one place. @haworku is that something that’s on our backlog? i forget.

services/app-web/src/common-code/data-helpers/makeDocumentKeyLookupList.ts
JasonLin0991: I don't believe you can get an index using `for..of` by itself. Usually you can do something like 
`for ( const [index, revision] of revisions.entries()) {...}`
but typescrypt does't like `revisions.entries()`

I could use  the standard `for` loop with an `iterator` or map the revision to an index like:
`const mappedRevisions = revisions.map((revision, index)=> ({ index, revision}))` then you could for `for...of` but that doesn't seem efficient.

JasonLin0991:
>  I suspect we need to do more consolidation at some point of the share documents component to get that logic into one place. @haworku is that something that’s on our backlog? i forget.

I remember @Hana mentioning consolidation of these functions. I'll move it back into the components for now.

haworku:
Yes, there are two existing tickets in the tech maintenance backlog about consolidating logic. They are intentionally separated to try and make the work more manageable and focused.  We want to avoid doing a huge refactor-to-end-all-refactor that risks touching area we have had a lot of bugs and breakages in.
[MR-888](https://qmacbis.atlassian.net/browse/MR-888)
[MR-768](https://qmacbis.atlassian.net/browse/MR-768)

macrael:
@haworku that first link i think is wrong, its about dependabot secrets, what’s the right one?

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
```diff
@@ -199,6 +201,15 @@ export const StateSubmissionForm = (): React.ReactElement => {
 
             setFormDataFromLatestRevision(planFormData)
 
+            //set previous submitted files
+            const documentList = makeDocumentList(submissionAndRevisions)
+            if (documentList instanceof Error) {
+                //Maybe a different error message here.
+                setFormDataError('MALFORMATTED_DATA')
```
haworku, line 208: Error state looks fine to me for now, I don't anticipate it will stay in this component since we are refactoring how this works.

FYI @rswerve this will be another thing to handle in our api wrapper that will isolate the data transformation.

haworku:
@macrael oops! Updated. FYI I'm finding these at the top of the tech maintaince backlog

haworku (approved):
Things look good to me +1 to MacRae's comments.

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
haworku: @JasonLin0991 if you have capacity, could you add a quick storybook story for this case ^ to FileUpload. We have a bunch of examples already in there to make it easier to verify these kinds of edge cases about FileUpload. We already have helpers to mock deleting scanning and uploading. Looks like we are missing a delete cases. 

If you can't get to it ping me, I can add on to one of my PRs this week.

services/app-web/src/documentHelpers/makeDocumentKeyLookupList.ts
```diff
@@ -0,0 +1,50 @@
...
+        currentDocuments: [],
+        previousDocuments: [],
+    }
+
+    const revisions = submissions.revisions
+
+    for (let index = 0; index < revisions.length; index++) {
+        const revisionData = base64ToDomain(revisions[index].node.formDataProto)
+        if (revisionData instanceof Error) {
+            return new Error(
```
JasonLin0991, line 33: >I’m curious about this if, too, there’s no harm in calling .forEach() on an empty array.

>In general I’ve been prefering the for (const revision of revisions) {} syntax, but I don’t know if you can get an index out of that. I think it might be a little more elegant to make a separate function that takes a revision and an array to dump its documents into and call that differently for the first and then the rest of the reivisions.

@macrael I switched to the vanilla for loop, returning errors now breaks out of the loop! I also added a few test, one which tests when the functions returns an error.

PR #929, opened 2022-04-20
Split up domain-models from common-code

macrael:
## Summary

Our domain models originally were intended to be used across the app, but with how GraphQL works, it ended up making more sense for us to use GraphQL types in app-web and only use the domain models in app-api. WIth the resrt of our refactor in, this splits out those types so that only things that are used across app-web and app-api are in common-code, things only used by one or the other are elsewhere. 

I also put app-api into /src to be consistent

```
Services/
  App-api/
    src/
      Domain-models
  app-web/
    src/
      common-code/
        config/
        healthPlanFormDataType/
        healthPlanFormDataMocks/
        dateHelpers/
        formatters/
        proto/
          healthPlanFormDataProto/
```


rswerve (approved):
Let's do it.

tests/cypress/integration/stateWorkflow/stateSubmissionForm/submissionType.spec.ts
```diff
@@ -9,6 +9,7 @@ describe('submission type', () => {
             const pathnameArray = pathname.split('/')
             const draftSubmissionId = pathnameArray[2]
             cy.visit(`/submissions/${draftSubmissionId}/type`)
+            cy.findByRole('heading', { level: 2, name: /Submission type/, timeout: 10000 })
```
haworku, line 12: Why not wait for the query to return here also? The problem with cy.visit to specific form page is it bypasses those wait function which do address flakes. I bet we could have flakes here in other tests as well that use cy.visit to jump ahead in the form

PR #931, opened 2022-04-21
Update unlock button copy

joshmfranklin:
## Summary
Update unlock button copy from "submit" to "unlock"
Screen Shot 2022-04-21 at 1 56 08 PM

### DIFF
--- a/services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
+++ b/services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
@@ -427,6 +427,7 @@ export const SubmissionSummary = (): React.ReactElement => {
                         formik.handleSubmit()
                     }}
                     modalRef={modalRef}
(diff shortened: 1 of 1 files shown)


macrael (approved):
I'm surprised this didn't break any tests

rswerve:
> I'm surprised this didn't break any tests

Snapshot tests are underrated.  🙂

haworku:
@macrael peeked at test and looks like we have a `aria-label` hard coded to "Submit" on our Modal.tsx  submit button. Since the aria label determines the accessible name, the test  did not fail because the name is unchanged. 

Probably that Modal component should have an `aria-label of `{onSubmitText || 'Submit'}`

macrael:
@haworku Ah ha, that would explain it. I bet you are right we want to aria-label to match, or if we want the button title to be what a screen reader reads do we even need to set aria-label?

PR #935, opened 2022-04-22
[MR-2112][MR-103]Address scrolling bugs

haworku:
## Summary

- Scroll to top whenever navigating to new page.
- Don't jump down on for pages when rendering new conditional fields

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2112
https://qmacbis.atlassian.net/browse/MR-103

#### Test cases covered
- added test for user is scrolled to top of new page

## QA guidance
- Please check this on a few browsers. For the conditional fields issue, that was a browser specific bug. Test in Chrome.


services/app-web/src/components/Form/FieldPreserveScrollPosition/FieldPreserveScrollPosition.tsx
```diff
@@ -0,0 +1,24 @@
...
+export function FieldPreserveScrollPosition<T>({
+    fieldName,
+}: {
+    fieldName: keyof T
+}): null {
+    const { values } = useFormikContext<T>()
+    const fieldValue = values[fieldName]
+    useEffect(() => {
+        document.querySelector(`#${fieldName}`)?.scrollIntoView()
+    }, [fieldValue, fieldName])
```
JasonLin0991, line 22: Should we use useLayoutEffect here also?

services/app-web/src/hooks/useScrollToPageTop.ts
```diff
@@ -0,0 +1,11 @@
...
+import { useLocation } from 'react-router'
+
+// Scroll to top of page on load, used on every page change
+export const useScrollToPageTop = (): void => {
+    const { pathname } = useLocation()
+
+    useLayoutEffect(() => {
+        window.scrollTo(0, 0)
+    }, [pathname])
+}
```
JasonLin0991, line 11: I like how you made this into a hook! Worked well when I tested this.

I'm interested to see how we tackle this scroll and the `FieldPreserveScrollPosition`, since they both would be firing.

services/app-web/src/components/Form/FieldPreserveScrollPosition/FieldPreserveScrollPosition.tsx
haworku: Yes, good call. I added it in [185b630](https://github.com/CMSgov/managed-care-review/pull/935/commits/185b63036ca33f267763e1e1ee1419c07d9c5b71) let me know if you see any different in this commit.

services/app-web/src/hooks/useScrollToPageTop.ts
haworku: Yes! I thought about this a bit. Let write out some thoughts here.
- The useScrollToPageTop effect will only fire _on initial mount_ with the route change. 
- The effect in FieldPreserveScrollPosition fires _if the formik field value changes after initial mount._ I  believe the only way that formik values could change after the initial mount is from some user interaction like clicking or typing in the field.

I don't think these should conflict. I tested this with and already filled out and saved contract details page (e.g. there is data that is not undefined/empty in there for the contract type field) and slow throttling to slow the api request. It still seems to work fine (only goes scrolls to the top on load - it doesn't try to jump down at all when the api returns the data for contract type). 

What do you think? Do you feel comfortable merging this?

services/app-web/src/components/Form/FieldPreserveScrollPosition/FieldPreserveScrollPosition.tsx
JasonLin0991: Looks great! There was an issue the first round of testing, where going back to the `Contract Details` page would scroll to the fields, this is no longer happening. I wonder if changing this to `useLayoutEffect` from `useEffect` fix that because, I think `useEffect` fires after `useLayoutEffect`.

services/app-web/src/hooks/useScrollToPageTop.ts
JasonLin0991: >I'm interested to see how we tackle this scroll and the FieldPreserveScrollPosition, since they both would be firing.

Initially I mentioned this because of what I saw going back to `Contract Details` page.
>Looks great! There was an issue the first round of testing, where going back to the `Contract Details` page would scroll to the fields, this is no longer happening. I wonder if changing this to `useLayoutEffect` from `useEffect` fix that because, I think `useEffect` fires after `useLayoutEffect`.

Otherwise everything looks awesome!

JasonLin0991 (approved):

JasonLin0991 (approved):

services/app-web/src/components/Form/FieldPreserveScrollPosition/FieldPreserveScrollPosition.tsx
haworku: I bet you are right!

PR #940, opened 2022-04-25
Add LD to app-web

mojotalantikite:
## Summary

This adds LD to `app-web`. It adds a testing "announcements" banner to the dashboard to make sure our React application communicates properly with LD.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-651

### Testing guidance

Open the review app and login. On the dashboard there should be no banner message displayed:
Screen Shot 2022-04-28 at 2 03 12 PM

Login to Launch Darkly and make sure you're in the "dev" environment. Look for the "test-frontend-banner" flag:

Screen Shot 2022-04-28 at 2 03 41 PM

Enable the flag and you should see the following test banner:

Screen Shot 2022-04-28 at 2 05 46 PM

Disable it and it should go away.


macrael:
What was the trick, in the end?

mojotalantikite:
@macrael getting a regex into ignoreURLs in OTEL and not trying to do anything with the LD client instantiation. There was some sort of precedence going on when I was trying to work with it from both the LD client and OTEL, doing all the ignoring from OTEL is what worked. Seems obvious now lol.

services/app-web/src/pages/App/App.tsx
```diff
@@ -70,4 +73,6 @@ function App({
     )
 }
 
-export default App
+export default withLDProvider<AppProps>({
```
macrael, line 76: my vote would be to move this call into index.tsx. That way we’d be checking for the config in there and immediately using it in wrapping the App component. It also decouples LD from the app so if we needed to host App in a test or something we still could eaily.

macrael (approved):
Looks great! seems like the ignore URLS thing finally started working?

services/app-web/src/pages/App/App.tsx
mojotalantikite: Ok so moving it over into index.tsx seems to have broken things, I'll see what's going on with it.

mojotalantikite: Ok I changed this over to using the async provider LD has -- seems to be working again from `index.html` via the async init.

PR #948, opened 2022-04-27
MR-43: Change submission type page buttons when editing

JasonLin0991:
## Summary
[MR-43](https://qmacbis.atlassian.net/browse/MR-43)

When editing an existing submission, the `Submission Type` page will now show the `Save as draft` button. I brought up keeping the `Cancel` button, it would be a clear way for the user to exit the editing process, to @joshmfranklin and will be keeping that in unless there are objections.

#### Screenshots
Screen Shot 2022-04-27 at 4 52 04 PM

#### Test cases covered
- PageActions.test.tsx
   - displays Save as draft, Cancel and Continue buttons for the first page when editing
   - calls continueOnClick when Continue button is clicked on the first page when editing
   - calls saveAsDraftOnClick when Save as draft button is clicked on the first page when editing
   - calls backOnClick when Cancel button is clicked on the first page when editing

- SubmissionType.test.tsx
   - displays a save as draft link when editing a submission
   - displays a cancel link on new submission
   - [removed] displays a cancel link
   - shows validation error messages when save as draft button is clicked and form fields are invalid

- submissionType.spec.ts
   - can navigate to and from type page
      - Added test for going to `dashboard` page form `type` and back using `Save as draft` button
   - can save submission edits using Save as draft button

### DIFF
--- a/services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
+++ b/services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
@@ -159,9 +159,7 @@ export const SubmissionType = ({
             FormikHelpers<SubmissionTypeFormValues>,
             'setSubmitting'
         >,
-        options?: {
-            redirectPath?: string
-        }
+        redirectPath?: string
     ) => {
         if (isNewSubmission) {
             try {
@@ -221,8 +219,8 @@ export const SubmissionType = ({
 
             try {
                 await updateDraft(draftSubmission)
-                if (options?.redirectPath) {
-                    history.push(options.redirectPath)
+                if (redirectPath) {
+                    history.push(redirectPath)
                 } else {
                     history.push(
                         `/submissions/${draftSubmission.id}/contract-details`
@@ -247,7 +245,6 @@ export const SubmissionType = ({
                 handleSubmit,
                 isSubmitting,
                 setSubmitting,
-                validateForm,
             }) => (
                 <>
                     <UswdsForm
@@ -417,20 +414,11 @@ export const SubmissionType = ({
                                 setFocusErrorSummaryHeading(true)
                             }}
                             saveAsDraftOnClick={async () => {
-                                setShouldValidate(true)
-                                //Validate form before trying to save a draft
-                                const validationErrors = await validateForm(
-                                    values
+                                await handleFormSubmit(
+                                    values,
+                                    { setSubmitting },
+                                    '/dashboard'
                                 )
-                                if (
-                                    Object.keys(validationErrors).length === 0
-                                ) {
-                                    await handleFormSubmit(
-                                        values,
-                                        { setSubmitting },
-                                        { redirectPath: '/dashboard' }
-                                    )
-                                }
                             }}
                             continueDisabled={isSubmitting}
                         />


JasonLin0991 (commented):
-

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
```diff
@@ -155,7 +155,13 @@ export const SubmissionType = ({
 
     const handleFormSubmit = async (
         values: SubmissionTypeFormValues,
-        formikHelpers: FormikHelpers<SubmissionTypeFormValues>
+        formikHelpers: Pick<
+            FormikHelpers<SubmissionTypeFormValues>,
+            'setSubmitting'
+        >,
```
JasonLin0991, line 161: Changed the `formikHelpers` type to pick out `setSubmitting` used in the function. I couldn't figure out a way, if there is one, to pass the helpers from `formik` to match `FromikHelpers` type.

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
```diff
@@ -155,7 +155,13 @@ export const SubmissionType = ({
...
+        options?: {
+            redirectPath?: string
+        }
```
JasonLin0991, line 164: Added the `options.redirectPath` to push to custom path ( '/dashboard' ) after  `updateDraft` mutation

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
```diff
@@ -392,13 +408,30 @@ export const SubmissionType = ({
...
                                 setFocusErrorSummaryHeading(true)
                             }}
-                            saveAsDraftOnClick={() => setShouldValidate(true)}
+                            saveAsDraftOnClick={async () => {
+                                setShouldValidate(true)
+                                //Validate form before trying to save a draft
+                                const validationErrors = await validateForm(
+                                    values
+                                )
+                                if (
```
JasonLin0991, line 425: Checking for errors at the time of clicking save as draft using `validateForm`. Using `error` from formik did not have up to date errors and would not display them.

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
```diff
@@ -392,13 +408,30 @@ export const SubmissionType = ({
...
+                                    Object.keys(validationErrors).length === 0
+                                ) {
+                                    await handleFormSubmit(
+                                        values,
+                                        { setSubmitting },
+                                        { redirectPath: '/dashboard' }
```
JasonLin0991, line 431: Passing relevant arguments for the `handleFormSubmit`

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
macrael: I don't get what FormikHelpers are, what are we using them for elsewhere?

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
macrael: This all seems solid, following the pattern of the other pages. ContractDetails just passes setSubmitting directly instead of wrapping it in an object

macrael (approved):
This looks good. We definitely could refactor this component a bit to keep from using the path to determine if we're at the start here or not, but that can wait for another day.

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
haworku: They are a way to access formik state. Read more here https://formik.org/docs/api/withFormik#the-formikbag  - whenever we submit a form with formik theres a bunch of state we have access to.

 In terms of how we are using it looks like in Contacts an Submission Type forms only -- which suggests it might be able to be refactored out if we unify the forms.

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
haworku: That looks good can you complete the refactor then -  only use `redirectPath`. You can remove the if/else and pass in the contract-details path for cases when we are redirecting there.

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
haworku: @JasonLin0991 Hmmmm... that sounds a bit off 🔍 likely its a bug you found. The use case for `validateForm` is pretty specific, its when you want to validate on the spot and bypass the usual flow. But formik values and errors both should be up to date here, if its not correct it means something might be off with our `yup` schema or the way we hook our inputs into formik

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
haworku: +1 We can pass `setSubmitting` directly here as well just like ContractDetails. No need to pass all the formik state, we aren't using it.

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
haworku: @JasonLin0991 FYI the type of `setSubmitting` is literally just `setSubmitting: (isSubmitting: boolean) => void`.  The way you are doing things here also works, that is the proper way to pick out a type from a set of typed properties. I think what you are doing works well.

> I couldn't figure out a way, if there is one, to pass the helpers from formik to match FromikHelpers type.

I _believe_ the only way you could get that type is directly from the parameter within the` <Formik onSubmit />`
Screen Shot 2022-04-28 at 8 44 41 AM


Hope that answers your type question. If not, let's look together when we pair.

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
JasonLin0991: @haworku Actually, the other pages don't seem to be checking for errors on `Save as draft`. I think this might be entirely unnecessary because the user may not have all the information for the edit but want to save what they have. Checking for errors would require them to fill out everything.

haworku: Oh sure. Makes sense. I didn't realize this was save as draft only. Yes yes do not validate on save as draft unless there are documents that are messed up.

haworku (approved):

PR #953, opened 2022-04-28
[MR-108][MR-42] Add loading button states and prevent duplicate clicks in button group when a request in progress

haworku:
## Summary
- Add loading state to buttons
- Address bug where in modals actions can be double-submitted. 
- Add `actionInProgress` prop to `<PageActions />` so that all buttons can be disabled easily while async request completes
- Prefer `aria-disabled` for handling disabled buttons. Adjust tests.
- Refactor out UnlockModal from SubmissionSummary
- Re-add work from #931 that was lost in bad merge
- Storybook: Create `<ActionButton />` and add to storybook. Update storybook config based on latest docs.
- Other misc: Address 404 bug in state submission form _(still present in submission summary - fixing that was a bit more involved so I left it_). 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-108
https://qmacbis.atlassian.net/browse/MR-42

## Test cases added
- Everything in ActionButton.test.tsx is new. Adjusted tests in PageActions to check for new prop.
- Had to change how we check for disabled buttons throughout the rest of our test


services/app-web/src/pages/SubmissionSummary/UnlockModal.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+    return (
+        <Modal
+            modalHeading="Reason for unlocking submission"
+            id="unlockReason"
+            onSubmit={() => {
+                setFocusErrorsInModal(true)
+                formik.handleSubmit()
+            }}
+            isSubmitting={formik.isSubmitting}
```
haworku, line 139: This is the only related code change, everything else is copy paste from submission summary

haworku:
I have a bunch more tests to fix because the second we don't use HTML `disabled` for buttons (for [accessibility](https://www.smashingmagazine.com/2021/08/frustrating-design-patterns-disabled-buttons/)) then we need a different matcher for `.toBeDisabled` in jest and cypress.

services/app-web/src/testHelpers/s3Helpers.ts
```diff
@@ -17,7 +17,7 @@ export const testS3Client: () => S3ClientT = () => {
             return key instanceof Error ? null : key
         },
         getS3URL: async (s3key: string, fileName: string): Promise<string> => {
-            return `fakes3://fake-bucket/${s3key}/${fileName}`
+            return `s3://fake-bucket/${s3key}/${fileName}`
```
haworku, line 20: Had to change this due to test failures for improperly formatted keys. We have a function `isValidS3URLFormat` that was throwing an error from the mocked key that didn't follow the expected format and then broke a test

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
```diff
@@ -333,6 +341,7 @@ export const StateSubmissionForm = (): React.ReactElement => {
                         <Contacts
                             draftSubmission={formDataFromLatestRevision}
                             updateDraft={updateDraftHealthPlanPackage}
+                            previousDocuments={previousDocuments}
                         />
```
JasonLin0991, line 345: I think we are passing `previousDocuments` here to satisfy `HealthPlanFormPageProps` type in the `Contacts` component. I'm wondering the reason for passing a prop that is never used vs specifying the prop types in `Contacts`, like 
```
export const Contacts = ({
    draftSubmission,
    showValidations = false,
    updateDraft,
}: {
    draftSubmission: HealthPlanFormPageProps['draftSubmission']
    showValidations?: boolean,
    updateDraft: HealthPlanFormPageProps['updateDraft'],
}): React.ReactElement => {
```

JasonLin0991 (commented):
Overall looks good to me. Just the one question about `previousDocuments` prop on `Contacts` component.

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
haworku: I can do that! No specific reason why its written the way it is.

services/app-web/src/pages/SubmissionSummary/UnlockModal.tsx
```diff
@@ -0,0 +1,173 @@
...
+                result
+            )
+            setModalAlert('Error attempting to unlock. Please try again.')
+            modalRef.current?.toggleModal(undefined, false)
+        } else if (isGraphQLErrors(result)) {
+            console.error('ERROR: got a GraphQL error response', result)
+            if (result[0].extensions.code === 'BAD_USER_INPUT') {
+                setModalAlert(
+                    'Submission is already unlocked. Please refresh and try again.'
+                )
```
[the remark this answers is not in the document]
haworku: Good idea. Will make this change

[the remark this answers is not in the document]
haworku: Edit: I realized I've seen this error a few times when someone already unlocked the submission. I think its a pretty common issue with concurrent editing to hit this error in which case the error message is instructive. So I still made the error message a bit more generic (to reflect what you said) but kept some of that messaging as well.

JasonLin0991 (approved):
✅

PR #954, opened 2022-04-28
[MR-2147] Speculative fix for app-api test timeouts

macrael:
## Summary

We’ve had a number of timeouts in the app-api tests.

Both of the cases I could find from recently were for tests that had a 2 second delay built in to the test themselves. That was introduced to check that updatedAt was changing on the submit action. Several other tests ended up with that line copied into them even though they didn’t need a 2 second delay. I’ve deleted the unnessecary delays and added a longer timeout for the test that is still waiting 2 seconds. 

#### Related issues

https://qmacbis.atlassian.net/browse/MR-2147

## QA guidance

This has been an itermittant failure, often seen on promote only. So I think we’re going to have to merge this and then wait and see if we see that error again to know if it fixed anything.

### DIFF
--- a/services/app-api/src/resolvers/submitHealthPlanPackage.test.ts
+++ b/services/app-api/src/resolvers/submitHealthPlanPackage.test.ts
@@ -26,8 +26,11 @@ describe('submitHealthPlanPackage', () => {
         const draft = latestFormData(initialPkg)
         const draftID = draft.id
 
-        // submit
+        // We delay here to enable checking that updatedAt changes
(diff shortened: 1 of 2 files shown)


mojotalantikite (approved):

haworku:
Just saw this error three times on Friday. I reran that many times and gave up. Today I re-ran and it seems to be working but I think the issue is persisting. 
https://github.com/CMSgov/managed-care-review/commit/9045d0fe1f8ff82e98fd49ac753cad9be7f50b69/checks

PR #955, opened 2022-04-29
Add logged in user to LD client on frontend

mojotalantikite:
## Summary
Adds the current user to the LD client. This allows us to target specific users with feature flags.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-403

#### Testing guidance
Add a user to the test role. In this case we're adding the test user Aang to the 'false' category on the 'test-frontend-banner' feature flag:

Screen Shot 2022-05-04 at 5 33 33 PM

Check that the user sees the correct flag state for their user. In this case Aang won't see the test banner, even though the flag is 'on':

Screen Shot 2022-05-04 at 5 33 17 PM

Verify that other users see the other flag state. In this case, Toph still sees the flag since it is 'on' and the user hasn't been added to the 'false' state:

Screen Shot 2022-05-04 at 5 38 10 PM


services/app-web/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -70,7 +71,10 @@ const StatusTag = ({
 export const Dashboard = (): React.ReactElement => {
     const { loginStatus, loggedInUser } = useAuth()
     const location = useLocation()
-    const { testFrontendBanner } = useFlags()
+
```
haworku, line 74: Curious what happened with `useFlags` and why that didn't work?  That looks more similar to what I have used before

services/app-web/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -70,7 +71,10 @@ const StatusTag = ({
...
+    // Get the LD feature flag
+    const ldClient = useLDClient()
+    const testFrontendBanner = ldClient?.variation('test-frontend-banner')
```
haworku, line 77: Hmmm, how can we have some type control over flags? Should the current flag names should be hard coded constants in a file?

Maybe the flags could be stored as a constant in common-code so both api and web have the same source of truth for the flags ist. When we add a new flag to the codebase, which requires a code change anyway, we add it to the list 

Want to make sure someone can't typo this randomly in the code.

services/app-web/src/pages/Dashboard/Dashboard.tsx
mojotalantikite: I was having some issues with registering users into LD and moved over to using `ldClient.variation`, which is what their docs had been referencing when talking about user segmentation. I can test out `useFlags` again though and see if we can get the same behavior again!

services/app-web/src/pages/Dashboard/Dashboard.tsx
mojotalantikite: Yeah I like this idea, it'd force us to keep a running list of flags somewhere, and it could be helpful for us when we get to the point where we have flags we want to remove.

I'll give it a shot and push it up to this PR. Thanks!

haworku: Nice! Also on the topic of typescript check out this `getFlagValue` wrapper. 

Let's create a wrapper like this also instead of directly calling ldClient in the React code. Usually this makes things easier to refactor and improve this code as we move along. Even if all our version of `getFlagValue` does is declare the type of the flags as `LaunchDarkly.LDFlagValue` instead of `any` at this points (since I know we are initializing the user a bit differently).
https://launchdarkly.com/blog/using-launchdarkly-with-typescript/

haworku (approved):
Everything else looks good and tested it in LD dashboard

services/app-web/src/pages/Dashboard/Dashboard.tsx
mojotalantikite: So I was writing a wrapper like this, but it turns out that [`useLDClient`](https://docs.launchdarkly.com/sdk/client-side/react/react-web#hooks) is a React hook and needs to be called in the component. I started writing something where we'd have to pass the client in to the wrapper function, but since I added the constants file where we pass a known safe flag to `ldClient.variation()` it didn't feel like it was saving us much. 

You think it's still worth it? Happy to add it if so.

haworku (approved):

PR #956, opened 2022-05-02
MR-2094: Date missing from documents table on unlocked review and submit page

JasonLin0991:
## Summary
[MR-2094](https://qmacbis.atlassian.net/browse/MR-2094)

Date missing from documents from Review/Submit page. 

Repeated the code from `SubmissionSummary` that generates the document dates in `StateSubmissionForm`. This is passed to `ReviewSubmit` which is then passed to `ContractDetailsSummarySection` and `RateDetailsSummarySection`. 

#### Test cases covered
- makeDocumentDateLookupTable.test
   - should use earliest document added date


haworku (approved):
Thanks for including tests.

PR #965, opened 2022-05-16
[MR-2147] Set timout correctly on our api test that keeps timing out

macrael:
## Summary

My initial fix for this issue didn’t take. It looks like I might have set the timeout incorrectly. I’ve also added some debug logging to see how far we are making it in the test before it times out. 

It’s still very possible that this timeout won’t fix things if the issue is about the sidecar db not being available instead of the test having a 2 second delay in it. Hopefully the debug printing will help narrow that down. 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2147


rswerve (approved):
good luck!

PR #967, opened 2022-05-20
MR-380: Show a modal when the user's session is about to expire

rswerve:
## Summary

We want the user to know that their session is about to expire, and to be able to continue it.  In this PR, we show a modal with a countdown timer, according to parameters that we can set in LaunchDarkly.  I also made a wrapper for our authenticated routes as an anchor point for the modal.

#### Screenshots

image

#### Test cases covered

Added some tests for new functionality in our Modal component (new cancel button text, and the ability to pass in a custom cancel method).

Added tests for the route wrapper, and whether the modal shows when it should.

## QA guidance

In Local and Dev, the current LaunchDarkly times are set to a two-minute session with a one-minute warning period.


services/app-web/src/contexts/AuthContext.tsx
```diff
@@ -39,11 +47,32 @@ function AuthProvider({
     >(undefined)
     const [loginStatus, setLoginStatus] =
         React.useState<LoginStatusType>('LOGGED_OUT')
-
+    const [isSessionExpiring, setisSessionExpiring] =
+        React.useState<boolean>(false)
+    const ldClient = useLDClient()
+    const countdownDuration =
+        ldClient?.variation(featureFlags.MODAL_COUNTDOWN_DURATION, 2) * 60
```
haworku, line 54: Are we expecting this feature flag to be evergreen? I was expecting we would use the `timeUntilLogout` value as the default here, overridden by the feature flag only if it exists.

services/app-web/src/pages/App/AppBody.tsx
```diff
@@ -33,7 +33,7 @@ export function AppBody({
                 {loginStatus === 'LOADING' ? (
                     <Loading />
                 ) : (
-                    <AppRoutes authMode={authMode} />
+                    <AppRoutes authMode={authMode} setAlert={setAlert} />
```
haworku, line 36: Can we name this prop `setGlobalAlert` or `setAppAlert` just to clarify what level this alert is occurring at? 

Also question - are you envisioning that is how this prop would be used in the future- by any app wide alert that' s not in a specific form ? I could also see us using a context for this - to store an app wide alert.

services/app-web/src/contexts/AuthContext.tsx
rswerve: I set it as a permanent flag, yeah, and timeUntilLogout gets its default value from this flag.  It gives us some flexibility to respond to user feedback around this feature, but maybe after a few months we'll want to refactor this to be hardcoded, or conditional, like you suggest.

services/app-web/src/pages/Wrapper/AuthenticatedRouteWrapper.tsx
```diff
@@ -0,0 +1,113 @@
...
+                    modalHeading="Session Expiring"
+                    onSubmitText="Continue Session"
+                    onCancelText="Logout"
+                    onCancel={logoutSession}
+                    submitButtonProps={{ className: styles.submitButton }}
+                    onSubmit={resetSessionTimeout}
+                    className={styles.expireModal}
+                    forceAction={true}
+                >
+                    <p>
```
haworku, line 94: I think we need to you some kind of `aria-live` here - I'm going off Chris's OTT from the easi project where he showed (after user testing sessions) that with aria-live users with screenreaders get properly notified about whats going on on with the countdown. 

Here's the code that adds this in their app https://github.com/CMSgov/easi-app/pull/304/files#diff-ec7778003924faeebbb2bd218f1005e66c8d085c44365afde23614aef3e7a693R124-R128

services/app-web/src/pages/App/AppBody.tsx
rswerve: I copied the pattern for catching errors in logout that we find in Header for the AuthenticatedRouteWrapper, so I needed to pass setAlert down to the wrapper.  This solution implies no vision whatsoever 🙂 .

services/app-web/src/pages/Wrapper/AuthenticatedRouteWrapper.tsx
haworku: Also regarding screenreader testing - when I tested this I saw something interesting where when the modal first opens it also re-announces the page heading.  It might be because the React context is remounting he entire tree here and that is triggering `usePage` again? 🔍 I can poke around more if its helpful

rswerve: Good call.  👀

services/app-web/src/pages/Wrapper/AuthenticatedRouteWrapper.tsx
```diff
@@ -0,0 +1,113 @@
...
+        modalRef.current?.toggleModal(undefined, isSessionExpiring)
+    }, [isSessionExpiring, modalRef])
+
+    useEffect(() => {
+        if (timeUntilLogout < 1) {
+            logoutSession()
+        }
+    }, [timeUntilLogout, logoutSession])
+    return (
+        <div className={styles.wrapper}>
```
haworku, line 79: This wrapper div is a bit problematic. Looks like its messing up our full screen styles for content inside `<main />`. See images below.

 Could you just use a React fragment here instead? Not sure we need the nested div here or its styles at all...
Screen Shot 2022-05-20 at 2 40 25 PM
Screen Shot 2022-05-20 at 4 14 35 PM

services/app-web/src/pages/Wrapper/AuthenticatedRouteWrapper.tsx
rswerve: I pushed a fix; let me know how it works for you.  And I'm not hearing the page heading thing.  We can chat Monday...

services/app-web/src/pages/Wrapper/AuthenticatedRouteWrapper.tsx
rswerve: Yup, fragment works fine.

services/app-web/src/pages/StateSubmission/ReviewSubmit/ReviewSubmit.tsx
```diff
@@ -215,56 +215,58 @@ export const ReviewSubmit = ({
...
-                    unlocked ? 'Summarize changes' : 'Ready to submit?'
-                }
-                submitButtonProps={{ className: styles.submitButton }}
-                onSubmitText={unlocked ? 'Resubmit' : undefined}
-                onSubmit={submitHandler}
-                isSubmitting={formik.isSubmitting || submitMutationLoading}
-            >
-                {unlocked ? (
-                    <form>
+            {!isSessionExpiring && ( // if the session is expiring, close this modal so the countdown modal can appear
```
haworku, line 218: 🤔 is there anyway to add this into shared `Modal` component so we always give priority to our system alerts? Otherwise we have to remember to add this conditional to every Modal we add to the app. 

One idea- we get the `isSessionExpiring`  value from within the modal itself, and if  that value is true and we are not in the session priority modal (check by component `id`), we self close the non-priority modal.

services/app-web/src/pages/Wrapper/AuthenticatedRouteWrapper.tsx
```diff
@@ -0,0 +1,124 @@
...
+            void extendSession()
+        }
+    }
+
+    useEffect(() => {
+        modalRef.current?.toggleModal(undefined, isSessionExpiring)
+    }, [isSessionExpiring, modalRef])
+
+    useEffect(() => {
+        if (timeUntilLogout < 1) {
```
haworku, line 79: Would you mind sneaking in a `setAlert` here to address https://qmacbis.atlassian.net/browse/MR-53 as well? I think that might address the bug - rather than waiting for the AuthContext refactor I have on another branch

services/app-web/src/pages/Wrapper/AuthenticatedRouteWrapper.tsx
haworku: 🌮 I am hearing the announcement now 🌮 . 

nit (not blocking): The  initial announcement is a bit  - right now it says "Your session is going to expire two o'clock"  which sounds like a time.  For everything else it says one fourty, one thirty, etc so it sounds fine. Its only the 2:00 and 1:00
that have this issue.

I think this code might be causing it 
```
 Your session is going to expire in{' '}
                        {dayjs
                            .duration(timeUntilLogout, 'seconds')
                            .format('mm:ss')}{' '}
```
 fwiw I'm not seeing the double announce of the page heading when the modal opens on JAWS today (i had used voiceover on Friday).  I don't think its an issue

haworku (approved):
Things look good to me, I haven't found any further bugs. It might be worth doing product/design review from the review app for this type of change to avoid messing with these flags on an a higher environment to test.

rswerve:
> Things look good to me, I haven't found any further bugs. It might be worth doing product/design review from the review app for this type of change to avoid messing with these flags on an a higher environment to test.

Sure.  One thing I was going to mention about flags, is that we can target them per user, in addition to per environment, so we could have a flag on only for josh and meghan in prod, or whatever.  (I know everyone knows this, but I think we were getting caught up in on/off per env.)

services/app-web/src/pages/StateSubmission/ReviewSubmit/ReviewSubmit.tsx
rswerve: Good idea.  Done.

services/app-web/src/pages/Wrapper/AuthenticatedRouteWrapper.tsx
rswerve: I don't think the quick fix is going to work (and didn't work locally) because everything gets wiped out on logout.  We'll have to persist the "just got logged out" state, probably in local storage.  (But open to other ideas.)

services/app-web/src/contexts/AuthContext.tsx
```diff
@@ -139,6 +174,9 @@ function AuthProvider({
                 loginStatus,
                 logout,
                 checkAuth,
+                isSessionExpiring,
```
macrael, line 177: We’ve got enough stuff going on in here that I think we should do a little documentation. Could you write up what these three new varables do and when they are set? For instance, it looks like timeUntilLogout isn’t set until we are displaying the alert. 

nit: I think sessionIsExpiring would be slightly more clear, isSessionExpiring sounds a little like a question to me for some reason.

macrael (approved):
Nice work. Keeping a reference to all the timers seems like a good pattern for this kind of stuff. 

I like seeing how this all came together, I think that there might be room for us to box it up a little tighter though. We’ve got a decent amount of logic that’s all related to us showing this alert spread across three different files: AuthContext, AppRoutes, and AuthenticatedRouteWrapper. I think it could stand to centralized a bit more, happy to pair with you to think about how. I would expect AppRoutes to be calling into a module to reset the timer, not setting timers and hitting LocalStorage on its own. 

I’m totally fine with merging this as is since it’s tested and working, but i’d like us to refactor it next if so.

services/app-web/src/pages/Wrapper/AuthenticatedRouteWrapper.tsx
haworku: that makes sense.

haworku:
>I would expect AppRoutes to be calling into a module to reset the timer, not setting timers and hitting LocalStorage on its own.
@macrael  I have that work going on my branch about cleaning up how we do local storage and handling as a hook. I tried to share with Maz but it wasn't tested yet and... so wasn't working 😄. I can continue working on this as part of my branch with the login and logout error handling.

haworku (approved):
Things look good still on deployed app.

services/app-web/src/components/Modal/Modal.tsx
```diff
@@ -30,13 +33,32 @@ export const Modal = ({
...
     submitButtonProps,
     onSubmitText,
+    onCancelText,
     isSubmitting = false,
     ...divProps
 }: ModalProps): React.ReactElement => {
+    const { sessionIsExpiring } = useAuth()
+
+    /* unless it's the session expiring modal, close it if the session is expiring, so the user can interact
+    with the session expiring modal */
```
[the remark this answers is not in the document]
rswerve: There is a test in AuthenticedRouteWrapper.test that would break if this were changed.

PR #970, opened 2022-05-25
[MR-2079] Switch “Items being amended” for “Modified provisions” in Contract details

macrael:
## Summary
This adds the "Modified Provisions" section to our Contract Amendment details pages, replacing the "items being amended" section. I made a new FieldYesNo component to try and keep the ContractDetails page succinct, though of course then I just wrapped that in a loop so that whole thing is only saving me like 5 lines.

Open Questions:
1. Should we nest the modified provisions inside their own object inside of ContractAmendmentInfo instead of having those keys be top level? Might be more future proof if we add more amended contract questions in the future
2. Should we have a separate UnlockedContractAmendementInfo that is Partial<ContractAmendmentInfo>? I had that at first but we only save complete amendment infos so it's not strictly necessary. (I think that save as draft is working as expected even though that breaks this type a little
3. Is the Yes/No component accessible correctly? I'm going to check that next but we're not using a label so it might be missing something. Am I using the right parent component there (Fieldset)?

#### Related issues

https://qmacbis.atlassian.net/browse/MR-2079

#### Screenshots
Screen Shot 2022-06-06 at 4 30 23 PM

Screen Shot 2022-06-06 at 4 31 28 PM

#### Test cases covered

I’m testing the error handling in ContractDetails, and that they are being shown in the review page

## QA guidance

This is a data model change so make sure we're getting everything everywhere.


services/app-web/src/common-code/proto/healthPlanFormDataProto/toDomain.ts
```diff
@@ -186,46 +185,40 @@ function parseProtoDocuments(
...
-function parseCapitationRates(
-    capRates: CapitationRateI | null | undefined
-): ContractAmendmentInfo['capitationRatesAmendedInfo'] | undefined {
-    if (!capRates) {
-        return undefined
-    }
-
-    return {
-        reason: enumToDomain(
-            mcreviewproto.CapitationRateAmendmentReason,
```
haworku, line 202: I missed this in the AC are we removing capitation rates amended reason entirely as well? I thought this was just about the items being amended list only so a bit confused.

services/app-web/src/components/Form/FieldYesNo/FieldYesNo.module.scss
```diff
@@ -0,0 +1,19 @@
...
+.optionsContainer label {
+    margin-top: .8em;
+}
+
+.yesnofield {
+    margin-top: .8em;
+    margin-bottom: 1.5em;
+}
+
+.yes {
```
haworku, line 14: nit: don't add empty scss classes

services/app-web/src/components/Form/FieldYesNo/FieldYesNo.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * inside of a Formik form context.
+ *
+ * If you want to use these components outside a Formik form, you can use the
+ * ReactUSWDS components directly.
+ */
+
+export type FieldYesNoProps = {
+    name: string
+    label: string
+    showError: boolean
```
haworku, line 22: suggestion - make `showError?: boolean`and default to false rather than making this a required prop. i try to keep required props to things that would break the markdown or accessibility for these kinds of re-useable elements

services/app-web/src/components/Form/FieldYesNo/FieldYesNo.tsx
```diff
@@ -0,0 +1,68 @@
...
+    showError,
+    id,
+    ...inputProps
+}: FieldYesNoProps): React.ReactElement => {
+    const [_field, meta] = useField({ name })
+
+    const isRequired =
+        inputProps['aria-required'] !== false && inputProps.required !== false // consumer must explicitly say this field is not required, otherwise we assume aria-required
+    return (
+        <Fieldset
```
haworku, line 38: Yes, this looks like the right use of this element

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
```diff
@@ -53,6 +55,32 @@ const createCheckboxList = ({
     )
 }
 
+// This function takes a ContractAmendmentInfo and returns two lists of keys sorted by whether they are set true/false
+export function sortAmendedItems(
```
haworku, line 59: An abstracted version would be a good resuseable helper to include in the yesno component folder.

services/app-web/src/common-code/proto/healthPlanFormDataProto/toDomain.ts
macrael: Yep. I confirmed this in slack, capitation rates reason is part of the Items Being Amended we're dropping. 

https://cmsgov.slack.com/archives/C014CRU197U/p1652994695062299

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
```diff
@@ -53,6 +55,32 @@ const createCheckboxList = ({
...
+    const modifiedProvisions = []
+    const unmodifiedProvisions = []
+
+    if (amendmentInfo) {
+        // We type cast this to be the list of keys in the ContractAmendmentInfo
+        const provisions = Object.keys(amendmentInfo) as Array<
+            keyof ContractAmendmentInfo
+        >
+
+        for (const provisionKey of provisions) {
```
haworku, line 71: An abstracted version would be a good re- useable helper to include somewhere with unit tests - maybe in `formHelpers'

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetailsSchema.ts
```diff
@@ -57,22 +62,21 @@ export const ContractDetailsFormSchema = Yup.object().shape({
...
-        then: Yup.string()
-            .nullable()
-            .defined('You must select a reason for capitation rate change'),
-    }),
-    capitationRatesOther: Yup.string().when('capitationRates', {
-        is: 'OTHER',
-        then: Yup.string().defined('You must enter a description'),
-    }),
+
+    modifiedBenefitsProvided: yesNoError,
```
haworku, line 66: Nice. We should probably do this more often, creating re-useable yup validation types

haworku (commented):
Nice additions of re-useable helpers  in this PR and interesting testing approaches. This is an initial review and everything looks good. I just want to run through the app live tomorrow am and test a bit before approving

services/app-web/src/components/Form/FieldYesNo/FieldYesNo.tsx
haworku: 🔴  alright, lets remove the fieldset at this level.  

From webaim
> Fieldset and legend should only be used when a higher-level label is necessary. Single checkboxes or basic radio buttons that make sense from their labels alone do not require fieldset and legend. Nested fieldsets can cause odd screen reader behavior and should be avoided

We aren't using a higher level label here anyway and these are just basic two option radios.

services/app-web/src/components/Form/FieldYesNo/FieldYesNo.tsx
```diff
@@ -0,0 +1,68 @@
...
+export type FieldYesNoProps = {
+    name: string
+    label: string
+    showError: boolean
+    id: string
+} & JSX.IntrinsicElements['input']
+
+export const FieldYesNo = ({
+    name,
+    label,
```
haworku, line 28: We should be using the label prop in contract details. It will solve a screenreader issue where when you go through the list the radios are announce with their values but the label text is not read.  We need to pass the label into the FieldYesNo > FieldRadio to address this bug I believe

haworku (approved):
The main accessibility issue I saw is the radio button group label name is not announcing when we tab through. We should fix that. If you tab through the form in JAWS you can see the bug where the radio button group name for the things like contract execution status are announced but nothing in the FieldYesNo is announced

Otherwise when the form is in error state even though its not clear what the error is for int he summary when you click the link it takes you to the right place and announce the full  name, so I think we are good there.

Screen Shot 2022-06-07 at 9 53 39 AM

services/app-web/src/components/Form/FieldYesNo/FieldYesNo.tsx
macrael: I added the legend back and it's working. The docs do say it's not ideal to nest field sets but our screen reader is working right now

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
macrael: I left this as is for now. I think it's mostly going to be useful for further yes no's on the summary page so I bet we can adapt it when we have more.

PR #973, opened 2022-05-31
[MR-2235] Add test environment banner to val and dev environments

haworku:
## Summary
Add banner.

Bonus:  strip out the jira collector from the frontend (cc: @MegMurphy, my understanding is this is no longer needed )
#### Related issues
https://qmacbis.atlassian.net/browse/MR-2235

#### Screenshots
Screen Shot 2022-05-31 at 5 16 17 PM

#### Test cases covered
- shows test environment banner in val
- does not show test environment banner in prod

### DIFF
--- a/services/app-web/src/pages/App/App.tsx
+++ b/services/app-web/src/pages/App/App.tsx
@@ -15,7 +15,6 @@ import TraceProvider from '../../contexts/TraceContext'
 import { GenericErrorPage } from '../Errors/GenericErrorPage'
 import { AuthModeType } from '../../common-code/config'
 import { S3Provider } from '../../contexts/S3Context'
-import { useScript } from '../../hooks/useScript'
 import type { S3ClientT } from '../../s3'
 
 function ErrorFallback({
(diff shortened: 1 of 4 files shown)


macrael:
I was out for this design, but my thought for this banner was that it would actually tell you what environment it is, not just "this is a test env". That way if you had local, review app, dev, or Val open in different tabs it's immediately obvious which one is which. @joshmfranklin what do you think of that? We could have different colors for different envs, or just text. I think it would be nice for it to have the branch name for review apps too, ideally.

macrael (approved):

joshmfranklin:
> I was out for this design, but my thought for this banner was that it would actually tell you what environment it is, not just "this is a test env". That way if you had local, review app, dev, or Val open in different tabs it's immediately obvious which one is which. @joshmfranklin what do you think of that? We could have different colors for different envs, or just text. I think it would be nice for it to have the branch name for review apps too, ideally.

@macrael I had been thinking of this ticket from the perspective of a state or CMS user who might make a parallel path submission on VAL and or get shared an incorrect link. I don't think that VAL or DEV are meaningful to those users so I just went with "test environment". Perhaps we take another spin at this down the road where we could show both "this is a test environment" and a subheader with the exact type of environment.

joshmfranklin (approved):

PR #991, opened 2022-06-10
MR-2179 Email copy changes

rswerve:
## Summary

Updates to the emails we send.  Required some new environment variables.  

@mojotalantikite @macrael As I write this, the build is failing with an environment variable failure that's not one of the ones I changed, so I'm not sure what's going on.  Also, I added env variables in the following places: .envrc, Github Actions Secrets, the Dev main graphql Lambda.  I didn't want to mess with the lambdas in Val and Prod, given the weirdness with the build here.  Maybe there's a lambda for this branch that I need to modify?  I wasn't able to find one.

#### Test cases covered

Updated old tests for removed bits around these being test submissions, and added new tests for the new language.

## QA guidance

This will need to build successfully...all tests were passing locally.


haworku:
@macrael ping me when this is ready for re-review

macrael:
@haworku this should be read to go now, looks like it's deploying.

I'm loving our emailer templates tests! they are covering a lot of cases, and I think as we start to add more logic there might be some ways for us to repeat some of those tests for some of the different config options. 

You are right that it's past time for us to introduce a templating library for this though. I think we're doing OK for these 3 events (submit, unlock, resubmit) X 2 user groups (state, CMS) X 2 overall states (CONTRACT_ONLY and CONTRACT_AND_RATES) but it's clear that there's a lot of repeated logic now going between our different emails. It's gotten to the point that I think we should try and tackle this sooner rather than later. I know that we'll have more things to notify about later on in the app, and I think that whenever we get to another one (sending emails to states? final letter? completeness review?) I think we need to be sure we've refactored this to use shared templates before we add another into this.

.envrc
```diff
@@ -23,9 +23,14 @@ export REACT_APP_S3_LOCAL_URL='http://localhost:4569'
 export REACT_APP_S3_DOCUMENTS_BUCKET='local-uploads'
 export DATABASE_URL='postgresql://postgres:shhhsecret@localhost:5432/postgres?schema=public&connection_limit=5' # pragma: allowlist secret
 export EMAILER_MODE='LOCAL'
+
 export SES_SOURCE_EMAIL_ADDRESS='dev@localhost'
 export SES_REVIEW_TEAM_EMAIL_ADDRESSES='mc-review-qa@truss.works'
 export SES_RATES_EMAIL_ADDRESSES='mc-rates-review-qa@truss.works'
+export SES_REVIEW_HELP_EMAIL_ADDRESS='MCOGDMCOActions@cms.hhs.gov'
```
haworku, line 30: Shouldn't these be mc-review-qa-help-email@truss.works ? I thought we were only using values here that are overridden in higher environments

haworku (approved):

.envrc
macrael: I'm down to change it, since we don't send emails to any of these addresses I thought it was OK to have them but I don't love that they are checked in even, really. I'm going to merge and we can revisit this after the submission tomorrow.

PR #1001, opened 2022-06-15
[MR-2321] reorder submission id to have number before programs

macrael:
## Summary

Reorder the submssion name to put the number before programs. 

#### Related issues

https://qmacbis.atlassian.net/browse/MR-2321

#### Screenshots

Screen Shot 2022-06-15 at 2 06 40 PM

#### Test cases covered

shockingly the only tests that broke were the ones related to the name generation code.

### DIFF
--- a/services/app-api/src/resolvers/indexHealthPlanPackages.test.ts
+++ b/services/app-api/src/resolvers/indexHealthPlanPackages.test.ts
@@ -203,9 +203,13 @@ describe('indexHealthPlanPackages', () => {
             query: INDEX_HEALTH_PLAN_PACKAGES,
         })
 
-        expect(result.errors).toBeUndefined() // Is this really what we want? I thought this would be 403 unauthorized
+        expect(result.errors).toBeUndefined()
 
         const indexHealthPlanPackages = result.data?.indexHealthPlanPackages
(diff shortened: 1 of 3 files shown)


services/app-api/src/resolvers/indexHealthPlanPackages.test.ts
```diff
@@ -203,9 +203,13 @@ describe('indexHealthPlanPackages', () => {
             query: INDEX_HEALTH_PLAN_PACKAGES,
         })
 
-        expect(result.errors).toBeUndefined() // Is this really what we want? I thought this would be 403 unauthorized
+        expect(result.errors).toBeUndefined()
 
         const indexHealthPlanPackages = result.data?.indexHealthPlanPackages
-        expect(indexHealthPlanPackages).toEqual({ edges: [], totalCount: 0 })
+        const otherStatePackages = indexHealthPlanPackages.edges.filter(
+            (pkg: HealthPlanPackageEdge) => pkg.node.stateCode !== 'VA'
```
haworku, line 210: Why do we have state specific logic here?

haworku (approved):
That's cool - maybe cuz we refer to submission name in tests in cypress via the justSubmitted url query parameter rather than hard coding? We also use the actual id quite a bit in the other tests and in our urls. really like that this was such an isolated change

PR #1005, opened 2022-06-16
[MR-2206] Add useOTEL hook, tracingHelper to app-web

haworku:
## Summary
This work add OTEL spans for significant JS exceptions that can happen in the frontend code, including a React crash (ErrorBoundary). Also added a span for each page view (approximately, I'm using route changes to determine this) after pairing  @JasonLin0991.

**Note:** 
The original scope of this ticket was "Add lib utility in app-web to add error states to traces". I don't think I have achieved that yet. My changes do not add the app-web errors to an existing trace, mainly because I couldn't see how to  identify  any existing traces) @mojotalantikite might have some insight here? Maybe there's a quick fix. Otherwise we could make that a follow on ticket. Personally, I think standalone spans are fine for now for monitoring purposes and they should be query-able.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2206


services/app-web/src/otelHelpers/tracingHelper.ts
```diff
@@ -0,0 +1,61 @@
...
+            span.recordException(err)
+        }
+    }
+
+    span.end()
+
+    return
+}
+
+function recordJSException(error: string | Error): void {
```
haworku, line 44: I tried to figure out if there was a fast way to pass in the ReactComponent name automatically to this function but I didn't see one. I think it requires adding `displayName` manually to each component which seemed like a bigger change.

I noted that  one of the [React auto-instrumentation tools](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/examples/react-load/react) for OTEL (which we aren't using because of the class components) uses a class components based way to add the name but there is an open ticket about creating a hook for use with functional components.

JasonLin0991 (commented):
Just leaving a comment on something I saw. To me everything looks good and works on Jaeger, no login for New Relic yet, so I'll let someone else approve.

services/app-web/src/otelHelpers/tracingHelper.ts
mojotalantikite: Yeah, when I first went about putting this together I was hoping we could use that lib, but that issue asking for functional component support has been open for two years 😭

services/app-web/src/hooks/useOTEL.ts
```diff
@@ -0,0 +1,29 @@
...
+// Adds a new OTEL span for each route change with some custom attributes. Proxy to users time on a given page.
+export const useOTEL = (title?: string): void => {
+    const { currentRoute } = useCurrentRoute()
+    const { loggedInUser } = useAuth()
+
+    useEffect(() => {
+        const tracer = getTracer()
+
+        const span: Span = tracer.startSpan('React: Page View')
+        span.setAttribute('current_route', currentRoute.toString())
```
mojotalantikite, line 17: I think we might want to use the `SemanticAttributes` to record the keys on these attributes. It's a spec of constants that OTEL is standardizing upon to make it easier to surface common keys. [Here's a list](https://open-telemetry.github.io/opentelemetry-js/modules/_opentelemetry_semantic_conventions.html) from the JS lib.

So for `user_id` we'd use `[SemanticAttributes.ENDUSER_ID]: loggedInUser.email`. They don't exist for everything, but where there is a standardized key I think we should adopt it.

mojotalantikite (approved):

docs/Technologies.md
```diff
@@ -43,6 +43,8 @@ To adjust warning levels, ignore certain types of warnings, or create actions (s
 
 We are running pa11y on every page of our form inside of our Cypress tests using the cypress-audit library. This allows us to sign in and fill out the form using Cypress, and only rely on pa11y to scan those pages. This is also one place in our app where we have implemented a ratchet in CI. The issues present when we started testing with pa11y have been ignored, allowing CI to ensure that we aren't adding new errors in the future. Tickets have been filed to address those existing issues.
 
+### Jaeger
+Jaeger is used to emulate Open Telemetry behavior for local dev. Read more in the [Jaeger docs](https://www.jaegertracing.io/docs/1.35/getting-started/) or visit port [16686](http://localhost:16686)) on local dev.
```
macrael, line 47: nit: I think I'd say that Jaeger is used for locally observing OTEL data? 

Not happy with that wording either. But I don't think it's emulating anything, it's a local replacement for NR

macrael (approved):
OK, wow, this is so cool. I love the idea of having a span start every time you visit a URL and have it end whenever you get to another one. 

I think what would take it to the next level would be if useOTEL() was something you could call in a component and you'd get the current span out. That would let us pass the span into `recordJSException`, and into our API calls(!!) tying all of those networking spans to the url the user was on when they initiated them. 

Tying recordJSException to specific span would also have some knock on benefits. Some of the stuff you are putting into the error message like ID and Error message could be added as span metadata instead of just hanging out inside of an error string. 

I think that connecting our API calls to the current span is definitely something that can happen in a follow on PR (maybe along with our refactor pulling the apollo stuff into our own lib?) but it would be neat if we could get the span into the current component, and use that with recordJSException in this PR.

@mojotalantikite per usual I'm a bit out of my depth when it comes to monitoring, does this sound like a good plan?

mojotalantikite:
> Awesome. Just a general comment/question, and maybe @mojotalantikite has an opinion as well: do we lose the connection between the front and back end traces by starting a new trace for various routes/actions on the front-end?

I'd have to test this out, but yes I'd think these wouldn't have a connection to a back end trace, unless it creates a fetch request. I think this fine for now and we can work towards finding what works for us in terms of tying particular spans together under a parent trace. Right now I think the most important thing is that we have a way to gather application errors and where those errors were triggered.

PR #1007, opened 2022-06-17
Add useLocalStorage hook

haworku:
## Summary
Add new `useLocalStorage` hook and a constants file for local storage key values. 

#### Related issues
Prep for https://qmacbis.atlassian.net/browse/MR-897 I want to be able to quickly add a new keys to local storage and have error handling if local storage is already full or JSON.parse fails.  In the case of this bug want to track if a user is coming back from a failed login 

#### Test cases covered
- This hook can be tested within the components where it is used. For example, we already have CMS unlock and resubmit tests that test that the hook is working when a CMS user tries to visit a submission url directly (and is redirected through Auth flow and then login) 

## QA guidance

Test that IDM local is working as expected for CMS user visiting a specific submission URL

### DIFF
--- a/services/app-web/src/hooks/useLocalStorage.ts
+++ b/services/app-web/src/hooks/useLocalStorage.ts
@@ -2,16 +2,19 @@ import { useState, useEffect } from 'react'
 import { LocalStorageKeyType } from '../constants/localStorage'
 type LocalStorage = {
     key: LocalStorageKeyType
-    value: string | null
+    value?: string | string[] | boolean | object
 }
 
-type UseLocalStorage = [LocalStorage['value'], (value: string | null) => void]
-// Get and set keys in local storage. If key is set to a value of null, clear and remove from local storage, return default fallback value
+type UseLocalStorage = [
+    LocalStorage['value'],
+    (value: string | undefined) => void
+]
+// Get and set keys in local storage. If key is set to a value of undefined, clear and remove from local storage, return default fallback value
 function useLocalStorage(
     key: LocalStorage['key'],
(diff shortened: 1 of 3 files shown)


services/app-web/src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,57 @@
+import { useState, useEffect } from 'react'
+import { LocalStorageKeyType } from '../constants/localStorage'
+type LocalStorage = {
+    key: LocalStorageKeyType
+    value: string | null
```
macrael, line 5: can we use undefined instead of null here?

services/app-web/src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,57 @@
...
+): UseLocalStorage {
+    const [storedValue, setStoredValue] = useState<string | null>(() => {
+        if (typeof window === 'undefined') {
+            console.error('Unable to find local storage. window is undefined')
+            return defaultValue
+        }
+
+        let value
+        try {
+            value = JSON.parse(
```
macrael, line 22: since values can only be strings or null, why are we JSON parsing / stringifying this?

macrael (approved):
I don't totally understand why we are combining useState with the local storage calls here, but I really like what it looks like using useLocalStoarge in a component. Nice.

haworku:
> why we are combining useState with the local storage calls 
just to avoid unnecessary synchronous reads into local storage, since we are already in a custom hook its not much to add state management there

PR #1022, opened 2022-06-24
[MR-2131] State analyst emails

JasonLin0991:

### DIFF
--- a/services/app-api/src/emailer/emailer.ts
+++ b/services/app-api/src/emailer/emailer.ts
@@ -26,6 +26,9 @@ type EmailConfiguration = {
     cmsRateHelpEmailAddress: string // email address for rates help
     cmsDevTeamHelpEmailAddress: string // email address for all other help
 }
+
+type StateAnalystsEmails = string[]
+
 type EmailData = {
     bodyText: string
     sourceEmail: string
@@ -49,7 +52,8 @@ type Emailer = {
     sendEmail: (emailData: EmailData) => Promise<void | Error>
     sendCMSNewPackage: (
         submission: LockedHealthPlanFormDataType,
-        submissionName: string
+        submissionName: string,
+        stateAnalystsEmails: StateAnalystsEmails
     ) => Promise<void | Error>
     sendStateNewPackage: (
         submission: LockedHealthPlanFormDataType,
@@ -59,7 +63,8 @@ type Emailer = {
     sendUnlockPackageCMSEmail: (
         submission: UnlockedHealthPlanFormDataType,
         updatedEmailData: UpdatedEmailData,
-        submissionName: string
+        submissionName: string,
+        stateAnalystsEmails: StateAnalystsEmails
     ) => Promise<void | Error>
     sendUnlockPackageStateEmail: (
         submission: UnlockedHealthPlanFormDataType,
@@ -73,7 +78,8 @@ type Emailer = {
     ) => Promise<void | Error>
     sendResubmittedCMSEmail: (
         submission: LockedHealthPlanFormDataType,
-        updatedEmailData: UpdatedEmailData
+        updatedEmailData: UpdatedEmailData,
+        stateAnalystsEmails: StateAnalystsEmails
     ) => Promise<void | Error>
 }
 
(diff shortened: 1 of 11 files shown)


services/app-api/src/emailer/emailer.ts
```diff
@@ -39,11 +42,18 @@ type EmailData = {
     bodyHTML?: string
 }
 
+// TODO: Next steps for state specific logic
```
macrael, line 45: are these TODOs done?

services/app-api/src/emailer/templates.test.ts
```diff
@@ -19,14 +19,20 @@ import {
 import { formatRateNameDate } from '../../../app-web/src/common-code/dateHelpers'
 import { unlockedWithFullContracts } from '../../../app-web/src/common-code/healthPlanFormDataMocks'
 
+const stateAnalystEmails = [
```
macrael, line 22: For this kind of test data, I strongly recommend making this a function that you call that returns the array rather than just a const array b/c that way if any individual test ends up modifying the array it won’t affect any other tests b/c they will get a fresh array when they call the function.

services/app-api/src/parameterStore/getStateAnalystEmailsStore.ts
```diff
@@ -0,0 +1,25 @@
+import { Context } from '../handlers/apollo_gql'
+import { logError } from '../logger'
+import { setErrorAttributesOnActiveSpan } from '../resolvers/attributeHelper'
+import { getParameterStore } from './parameterStore'
+
+export const getStateAnalystEmailsStore = async (
```
macrael, line 6: yay, I love making this its own package. Way to wrap a dependecny.

services/app-api/src/parameterStore/getStateAnalystEmailsStore.ts
```diff
@@ -0,0 +1,25 @@
...
+    stateCode: string,
+    span?: Context['span'],
+    operation?: string
+): Promise<string[]> => {
+    const analystsParameterStore = await getParameterStore(
+        `/configuration/${stateCode}/stateanalysts/email`
+    )
+    if (analystsParameterStore instanceof Error) {
+        logError(
```
macrael, line 15: I think this is fine, but it’s interesting to think about where we want to handle this error. The other option would be for this wrapper to reutrn an `| Error` and have the calling code be responsible for deciding to log the error. In that case we’d simplify this wrapper b/c it wouldn’t need the span or the error reporting logic in it. 

Not a case where there’s a clear “right” answer to me but worth considering the tradeoffs.

services/app-api/src/parameterStore/parameterStore.ts
```diff
@@ -0,0 +1,33 @@
+import { SSM } from 'aws-sdk'
+
+type GetParameterResult = SSM.GetParameterResult
+
+const ssm = new SSM({ region: 'us-east-1' })
+
+const getParameterStore = async (name: string): Promise<string | Error> => {
```
macrael, line 7: Excellent.

services/app-api/src/resolvers/submitHealthPlanPackage.test.ts
```diff
@@ -225,7 +225,7 @@ describe('submitHealthPlanPackage', () => {
 
         expect(submitResult.errors).toBeUndefined()
         expect(mockEmailer.sendEmail).toHaveBeenCalledTimes(2)
-    })
+    }, 20000)
```
macrael, line 228: were you seeing timeout failures here? I’m a little surprised, shouldn’t this all be real fast?

services/app-api/src/parameterStore/getStateAnalystEmailsStore.ts
macrael, line 6: nit: I’m not sure “store” is adding anything to this function name. Seems like since we pass in the state code we are just getting a list of analysts out, so I’d name this function “getStateAnalystEmails” and leave the package to have “store” in the name.

macrael (approved):
Very solid work. I really like how you wrapped up this dependency, it’s going to make it straightforward for us to replace with the db table later on.

PR #1032, opened 2022-06-28
[MR-2121] 404 Not Found page flickers when going from Dashboard to unlocked submission review and submit

JasonLin0991:
## Summary
[MR-2121](https://qmacbis.atlassian.net/browse/MR-2121)

Refactored error handling in `StateSubmissionForm.tsx` that caused the 404 flickering.

#### Test cases covered

- `StateSubmissionForm.test.tsx`
   - 'shows a generic error when updating submission fails'


services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
```diff
@@ -192,13 +191,19 @@ export const StateSubmissionForm = (): React.ReactElement => {
...
             }
 
             return updatedSubmission
         } catch (serverError) {
             setShowPageErrorMessage(true)
-            return serverError
+            recordJSException(
+                `StateSubmissionForm: Apollo error reported. Error message: ${serverError.message}`
+            )
+            throw new Error(serverError)
```
JasonLin0991, line 206: Throwing this error because `updateDraftHealthPlanPackage` is called in child components that has a `catch` block.

JasonLin0991 (commented):
There seems to be a duplicate `GenericApiErrorBanner` banner in `StateSubmissionForm` and `SubmissionType` both get triggered on errors with `updateDraftHealthPlanPackage`, but `SubmissionType` also triggers it with errors in `createHealthPlanPackage`. Should I tackle this on this PR?

services/app-web/src/pages/StateSubmission/StateSubmissionForm.test.tsx
```diff
@@ -445,6 +446,69 @@ describe('StateSubmissionForm', () => {
...
+
+        it('shows a generic error when updating submission fails', async () => {
+            const mockSubmission = mockDraftHealthPlanPackage({
+                submissionDescription:
+                    'A real submission but updated something',
+            })
+            const formData = base64ToDomain(
+                mockSubmission.revisions[0].node.formDataProto
+            )
+            if (formData instanceof Error) throw Error
```
haworku, line 458: I could see this unwrap proto code being its own test helper

> const formData = base64ToDomain(mockSubmission.revisions[0].node.formDataProto)
if (formData instanceof Error) throw Error

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
```diff
@@ -192,13 +191,19 @@ export const StateSubmissionForm = (): React.ReactElement => {
             if (!updatedSubmission) {
                 setShowPageErrorMessage(true)
                 console.log('Failed to update form data', updateResult)
+                recordJSException(
+                    `StateSubmissionForm: Apollo error reported. Error message: Failed to update form data ${updateResult}`
+                )
                 return new Error('Failed to update form data')
             }
```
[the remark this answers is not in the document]
haworku: Thanks for catching this, we talked about handling in this ⭐

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
macrael: I’m a little confused. We are returning a new error in one place but throwing in another in this same function? I think we should treat this function like one of our dependency wrappers, we are calling the apollo API but we should be able to return an | Error like we do everywhere else. 

That can wait until we extract all our apollo calls into a wrapper module, I know we have a ticket for doing that for the fetch API so we can do the proto unzipping behind the same facade.

macrael (approved):
Great new test there. 

What exactly is making this fix work here? What happens `if (!formDataFromLatestRevision)` now?

JasonLin0991:
@haworku @macrael Fixed the missing 404 page and added a test for making sure it shows up. Additionally I removed the generic banner on update draft errors in the `SubmissionType` page since `StateSubmissionForm` handles showing error message for updating draft on all the other form pages. It should now show the banner, on `SubmissionType` page, only on errors in `useCreateHealthPlanPackageMutation` and I added a test for this too.

services/app-web/src/testHelpers/apolloHelpers.tsx
```diff
@@ -836,25 +852,94 @@ const mockSubmittedHealthPlanPackageWithRevision = (): HealthPlanPackage => {
...
+                            pkgID: id,
+                            healthPlanFormData: updatedFormData,
+                        },
+                    },
+                },
+                result: { data: { updateHealthPlanFormData: { pkg } } },
+            }
+        case 403:
+            return {
+                request: { query: UpdateHealthPlanFormDataDocument },
```
macrael, line 884: why does this not have any input passed in? Isn’t that required for it to match?

services/app-web/src/testHelpers/apolloHelpers.tsx
```diff
@@ -836,25 +852,94 @@ const mockSubmittedHealthPlanPackageWithRevision = (): HealthPlanPackage => {
...
+                request: {
+                    query: UpdateHealthPlanFormDataDocument,
+                    variables: {
+                        input: {
+                            pkgID: id,
+                            healthPlanFormData: updatedFormData,
+                        },
+                    },
+                },
+                result: { data: undefined },
```
macrael, line 898: I don’t think this is what is returned on a 404 for UpdateHPFD. 404 returns something that looks like this for the fetch query:

```
{
    "data": {
        "fetchHealthPlanPackage": {
            "pkg": null,
            "__typename": "FetchHealthPlanPackagePayload"
        }
    }
}
```

But, while fetching a non existant package is treated as not an error, and we just return undefined, attempting to update a package that does not exist is treated as an error:

from the resolver:
```
            throw new UserInputError(errMessage, {
                argumentName: 'pgkID',
            })
```

services/app-web/src/testHelpers/apolloHelpers.tsx
```diff
@@ -836,25 +852,94 @@ const mockSubmittedHealthPlanPackageWithRevision = (): HealthPlanPackage => {
...
-type updateHealthPlanFormDataMockSuccessProps = {
+type updateHealthPlanFormDataMockProps = {
     pkg?: HealthPlanPackage
     updatedFormData: string
     id: string
+    statusCode?: 200 | 403 | 404 | 500
 }
 
-const updateHealthPlanFormDataMockSuccess = ({
+const updateHealthPlanFormDataMock = ({
```
macrael, line 862: I’ve been moving away from these switch based helpers to individual helpers. “updateHPFDMockSuccesss”, “updateHPFDMockAuthFailure”, “updateHPFDMockNotFound”

it’s slightly more repeated code at the top of each function but I think is a bit more clear. I might need to go through and consolidate that.

macrael (approved):
Some small test formatting comments

services/app-web/src/testHelpers/apolloHelpers.tsx
macrael: maybe, or maybe we aren’t actually calling it. Another good reason to break this out from the switch statement, it will be obvious if that function is being called or not

JasonLin0991: I split the helpers I consolidated into into individual ones.

PR #1040, opened 2022-06-30
MR-2285 Upgrade to React 18 and latest Testing Library

rswerve:
## Summary

Upgrading React from 17 to 18
Upgrading the various Testing Library packages to their latest versions

Here's the [official write-up](https://reactjs.org/blog/2022/03/29/react-v18.html) of what's new in React 18.

One part to pay attention to right away is [new strict mode behaviors](https://reactjs.org/blog/2022/03/29/react-v18.html#new-strict-mode-behaviors).  In short, when developing in strict mode (which we use), React will mount every component twice.  React components are ideally idempotent functions, and this double-mounting helps find those components that aren't.  I didn't come across any problems with this behavior and it didn't seem to break any tests.

There are also some hooks and UI transition improvements we might want to take advantage of in the future.

The one code change React required is switching from ReactDOM.render in app-web/index.tsx to creating a root and doing root.render.

The big change for Testing Library is that all "userEvent" methods are now async.  In the few cases we still need synchronous tests (like a transitory loading state) we can use "fireEvent" methods.  One test for Cognito login was removed because it wouldn't pass and @haworku and I decided it wasn't essential.

#### Test cases covered

Most tests needed syntax updates to `await` user interactions.  No new tests.

## QA guidance

Everything should work exactly as before.


rswerve:
Closing in favor of https://github.com/CMSgov/managed-care-review/pull/1041 to let serverless deploy start fresh

PR #1041, opened 2022-06-30
MR-2285: React 18 and Testing Library major version upgrades

rswerve:
## Summary

Upgrading React from 17 to 18
Upgrading the various Testing Library packages to their latest versions

Here's the [official write-up](https://reactjs.org/blog/2022/03/29/react-v18.html) of what's new in React 18.

One part to pay attention to right away is [new strict mode behaviors](https://reactjs.org/blog/2022/03/29/react-v18.html#new-strict-mode-behaviors).  In short, when developing in strict mode (which we use), React will mount every component twice.  React components are ideally idempotent functions, and this double-mounting helps find those components that aren't.  I didn't come across any problems with this behavior and it didn't seem to break any tests.

There are also some hooks and UI transition improvements we might want to take advantage of in the future.

The one code change React required is switching from ReactDOM.render in app-web/index.tsx to creating a root and doing root.render.

The big change for Testing Library is that all "userEvent" methods are now async.  In the few cases we still need synchronous tests (like a transitory loading state) we can use "fireEvent" methods.  One test for Cognito login was removed because it wouldn't pass and @haworku and I decided it wasn't essential.

#### Test cases covered

Most tests needed syntax updates to `await` user interactions.  No new tests.

## QA guidance

Everything should work exactly as before.


services/app-web/src/components/Header/Header.test.tsx
```diff
@@ -53,7 +53,7 @@ describe('Header', () => {
                 const signInButton = screen.getByRole('link', {
                     name: /Sign In/i,
                 })
-                userEvent.click(signInButton)
+                void userEvent.click(signInButton)
```
macrael, line 56: what does `void` do here?

services/app-web/src/components/Header/PageHeadingRow/PageHeadingRow.tsx
```diff
@@ -82,6 +82,6 @@ export const PageHeadingRow = ({
     } else if (loggedInUser.__typename === 'StateUser') {
         return <StateUserRow user={loggedInUser} heading={heading} />
     } else {
-        return <h1>Programming Error: Unkown User Type: {loggedInUser}</h1>
+        return <h1>{`Programming Error: Unkown User Type: ${loggedInUser}`}</h1>
```
macrael, line 85: why do we have to interpolate this? I thought JSX would work with the old version. Is this a new React change?

services/app-web/src/components/Modal/Modal.test.tsx
```diff
@@ -34,18 +34,19 @@ describe('Modal', () => {
...
-        expect(
-            screen.getByRole('heading', { name: 'Test Modal Title' })
-        ).toBeInTheDocument()
-        expect(
-            screen.getByRole('button', { name: 'Cancel' })
-        ).toBeInTheDocument()
-        expect(
-            screen.getByRole('button', { name: 'Submit' })
-        ).toBeInTheDocument()
+        handleOpen()
```
macrael, line 37: I don’t understand why handleOpen is this way. seems strange that we wrapped it it waitFor before, is `toggleModal` async?

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.test.tsx
```diff
@@ -541,7 +541,7 @@ describe('RateDetails', () => {
             expect(imageElFile2).toHaveClass('is-loading')
 
             // click continue while file 2 still loading
-            continueButton.click()
+            fireEvent.click(continueButton)
```
macrael, line 544: when do we want to use fireEvent vs. userEvent?

tests/cypress/support/stateSubmissionFormCommands.ts
```diff
@@ -82,56 +78,103 @@ Cypress.Commands.add('fillOutAmendmentToBaseContractDetails', () => {
...
-    cy.findByText('Non-risk payment arrangements').parent().within(() => {
-        cy.findByText('Yes').click()
-    })
+    cy.findByText('Benefits provided by the managed care plans')
+        .parent()
+        .within(() => {
+            cy.findByText('Yes').click()
+        })
 
+    cy.findByText('Geographic areas served by the managed care plans')
```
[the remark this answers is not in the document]
macrael: made it! https://qmacbis.atlassian.net/browse/MR-2379

macrael (approved):
await approved()

I asked a bunch of questions but that’s mostly me hoping you’ll tell me about what changed in testing library. Feel free to tell me to RTFM for any of them.

services/app-web/src/components/Header/Header.test.tsx
rswerve: Silencing a no-floating-promises linter error.

services/app-web/src/components/Header/PageHeadingRow/PageHeadingRow.tsx
rswerve: This does look like new behavior, though I haven't tracked down the root cause.  The new error is `This JSX tag's 'children' prop expects a single child of type 'ReactNode', but multiple children were provided.`  So I turned it into a single thing.

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.test.tsx
rswerve: I buried this too far in the summary, but fireEvent is still synchronous, so in a case like this, where we're trying to test a transitional state, it can be useful.

services/app-web/src/components/Modal/Modal.test.tsx
rswerve: > seems strange that we wrapped it it waitFor before  

This also seemed strange to me, so I removed the waitFor and tests still passed, so I kept it out.  toggleModal [is not](https://github.com/trussworks/react-uswds/blob/main/src/components/Modal/utils.ts#L29) async.

rswerve:
> 

@macrael The changes in testing library are not super well-described.  For [React](https://github.com/testing-library/react-testing-library/releases/tag/v13.0.0) and for [UserEvent](https://github.com/testing-library/user-event/releases/tag/v14.0.0).  The main change we've discussed, and that's the default to async actions for all the userEvent methods (fireEvent remains synchronous).

I can also imagine some scenarios where the [userEvent.setup()](https://testing-library.com/docs/user-event/setup/#starting-a-session-per-setup) is useful, but I didn't run into any in the upgrade.

PR #1045, opened 2022-07-01
[MR-2376] Dependency injection for parameter store

JasonLin0991:
## Summary
[MR-2376](https://qmacbis.atlassian.net/browse/MR-2376)

- Dependency injection for our email configuration parameterStore module to allow local testing. `PARAMETER_STORE_MODE` configuration variable is set using matching variable name in `.envrc` when running locally, otherwise this variable is found in parameter store `/configuration/parameterStoreMode`

- Fixed flaky tests in `RateDetails.test.tsx` and cypress.
  - The flaky cypress tests, especially when locally testing, seems to be our `safeClick` work around breaking. Maybe due to changes in React 18? I read the linked [github issue ](https://github.com/cypress-io/cypress/issues/7306), seems everyone is still having issues with this. Good news is that looks like cypress activity working on this from [April post](https://github.com/cypress-io/cypress/issues/7306#issuecomment-1098106831) and this more [recent post](https://github.com/cypress-io/cypress/issues/7306#issuecomment-1154362254).
  - Adding a 500ms wait to `safeClick` did fix all the flaky tests, especially in local. Definitely not a great fix, but a bandaid for now until fix from cypress is released or maybe we could add a ticket to look into this further and possibly adopt one of the many attempts at a fix from that thread. Maybe this [one](https://github.com/cypress-io/cypress/issues/7306#issuecomment-1076451070)

#### Test cases covered

- `submitHeathPlanPackage.test.ts`
   - 'does send email when request for state analysts emails fails'
   - 'does log error when request for state specific analysts emails failed'
- `unlockHealthPlanPackage.test.ts`
   - 'does send unlock email when request for state analysts emails fails'
   - 'does log error when request for state specific analysts emails failed'


services/app-api/src/resolvers/submitHealthPlanPackage.test.ts
```diff
@@ -256,6 +260,12 @@ describe('submitHealthPlanPackage', () => {
 
         const programs = [defaultFloridaProgram()]
         const name = packageName(sub, programs)
+        const stateAnalystsEmails = getTestStateAnalystsEmails(sub)
+
+        const cmsEmails = [
+            ...config.cmsReviewSharedEmails,
+            ...stateAnalystsEmails,
```
haworku, line 267: can you also add a line here removing any non unique emails? That way we ensure we dont email people twice or attach their name twice.

haworku (approved):
Things look good here. Let's get these cypress/jest flakes fixes in as well. I like the tests around error handling,logging, and fallback (we should still send some kind of email even if state analysts lookup fails) .

tests/cypress/support/commands.ts
```diff
@@ -38,6 +38,6 @@ Cypress.Commands.add('restoreLocalStorage', () => {
 
 Cypress.Commands.add('safeClick', { prevSubject: 'element' }, ($element) => {
     const click = ($el) => $el.click()
-    return cy.wrap($element).should('exist').should('be.visible').pipe(click)
+    return cy.wrap($element).should('exist').should('be.visible').wait(500).pipe(click)
```
haworku, line 41: This is a fine solution for now. We have tried several other things before getting to this point and it seems like the hit to our cypress tests time is acceptable to the team. Let's keep an eye on Cypress dashboard after this merges to see how much slower things get and watch out if we keep needing to bump this number higher and higher in the future to get same results.

mojotalantikite (approved):

services/app-api/src/resolvers/submitHealthPlanPackage.test.ts
JasonLin0991: That sounds great. I'll merge this in this PR and tack on making this helper function to [MR-2361](https://qmacbis.atlassian.net/browse/MR-2361)

PR #1050, opened 2022-07-11
Move GHA secrets to SSM Param Store

mojotalantikite:
## Summary

This changes moves values that we need for deployment into AWS parameter store. Previously we were setting these in the environment, read from GHA secrets. After this change, serverless will be reading them directly from parameter store.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-1182


services/app-api/serverless.yml
```diff
@@ -23,23 +23,14 @@ custom:
   applicationEndpoint: ${env:APPLICATION_ENDPOINT, cf:ui-${sls:stage}.CloudFrontEndpointUrl}
   appApiGatewayId: ${env:CF_CONFIG_IGNORED_LOCALLY, cf:infra-api-${sls:stage}.ApiGatewayRestApiId}
   appApiGatewayRootResourceId: ${env:CF_CONFIG_IGNORED_LOCALLY, cf:infra-api-${sls:stage}.AppApiGatewayRootResourceId}
-  emailSource: ${env:SES_SOURCE_EMAIL_ADDRESS}
   emailerMode: ${env:EMAILER_MODE}
-  parameterStoreMode: ${env:PARAMETER_STORE_MODE, ssm:/configuration/parameterStoreMode}
```
JasonLin0991, line 28: `parameterStoreMode` is being removed, are we getting it form somewhere else now? This was added for my work with configuring parameter store mode in `app-api/src/handlers/apollo_gql.ts`.

mojotalantikite: Good catch! I think this got blown away when pulling in `main` and fixing up the merge conflict. Adding it back in.

mojotalantikite:
Ok, I think I've got all the places sorted out for local/dev/val/prod on all the ssm params, but there's a lot of them and I very well could have overlooked some places. If people could test the branch local that'd be helpful before we merge this.

services/app-api/serverless.yml
```diff
@@ -64,6 +56,13 @@ custom:
       '/configuration/default/vpc/subnets/private/a/id': 'offline'
       '/configuration/default/vpc/subnets/private/b/id': 'offline'
       '/configuration/default/vpc/subnets/private/c/id': 'offline'
+      '/configuration/iam/full_permissions_boundary_policy': 'arn:aws:iam::local:policy/local/developer-boundary-policy'
+      '/configuration/email/sourceAddress': ${env:SES_SOURCE_EMAIL_ADDRESS, 'offline'}
```
macrael, line 60: Is there a practical difference between setting the serverless-offline-ssm with an env fallback like sourceEmailAddress and setting a variable like reactAppOtelCollectorUrl above with an env var with a ssm fallback?

services/app-api/serverless.yml
```diff
@@ -128,12 +127,12 @@ provider:
     SECRETS_MANAGER_SECRET: ${self:custom.secretsManagerSecret}
     EMAILER_MODE: ${self:custom.emailerMode}
     PARAMETER_STORE_MODE: ${self:custom.parameterStoreMode}
-    SES_SOURCE_EMAIL_ADDRESS: ${self:custom.emailSource}
-    SES_REVIEW_TEAM_EMAIL_ADDRESSES: ${self:custom.emailReviewTeamAddresses}
-    SES_REVIEW_HELP_EMAIL_ADDRESS: ${self:custom.emailReviewHelpAddress}
-    SES_RATE_HELP_EMAIL_ADDRESS: ${self:custom.emailRateHelpAddress}
-    SES_DEV_TEAM_HELP_EMAIL_ADDRESS: ${self:custom.emailDevTeamHelpAddress}
-    SES_RATES_EMAIL_ADDRESSES: ${self:custom.emailRatesTeamAddresses}
+    SES_SOURCE_EMAIL_ADDRESS: ${ssm:/configuration/email/sourceAddress}
```
macrael, line 130: I’m glad we are still configuring the app using env vars, constraining the SSM stuff to serverless.yml is great.

macrael (approved):
working for me locally

services/app-api/serverless.yml
mojotalantikite: They work the same really. I started with lookup for an env var since that's what we had been doing, and then a lookup to ssm if nothing was in the env var. We could just as easily do it the other way around if we want, I don't have a strong opinion here.

For the email config things I wanted to make it so we wouldn't need to change anything for local dev. We always pull the value from ssm, but locally that ssm value is set from `.envrc`.

macrael: Ok, tickles my One Way To Do Things meter that we have two different ways to do this, but I think we need the serverless-offline-ssm plugin for the per-state-email-config for now anyway so it’s not like changing this would save us from this dependecny. I don’t think we’ll run into trouble with confusion over how to do this.

PR #1052, opened 2022-07-12
[MR-2361] Refactor unlock/resubmit email to use Eta templates

JasonLin0991:
## Summary
[MR-2361](https://qmacbis.atlassian.net/browse/MR-2361)

- Refactor unlock/resubmit emailer functions to use Eta template
- Clean up some parameters passed to unlock/resubmit functions
- Refactor generating state receiver email into helper function and pruning duplicate emails
- Prune duplicate emails from generated CMS reviewer emails.

#### Test cases covered

- `newPackageCMSEmail.test.ts`
   - `'to addresses list does not include duplicate review email addresses'`
- `newPackageCMSEmail.test.ts`
   - `'to addresses list not include duplicate state contacts on submission'`


haworku (commented):
Things are looking good to me here so far. 

@macrael this would be a good, straightforward PR to add any emailer config refactors. You had mentioned wanting to see further emailer refactors last sprint - did you have specific changes in mind? My understanding was you wanted to simplify the parameters that the various email functions get called with. Looks like @JasonLin0991 has done a bit of this already here (removing `rateName` and `submissionName`  from being parameters on email functions and generating them closer to the template). _Anything else you want to see here?_

services/app-api/src/emailer/templateHelpers.ts
```diff
@@ -100,12 +105,29 @@ const generateReviewerEmails = (
...
+}
+
+const generateStateReceiverEmails = (
+    pkg: UnlockedHealthPlanFormDataType | LockedHealthPlanFormDataType,
+    user?: UserType
+): string[] => {
+    const stateReceiverEmails: string[] = []
+    //Only add current user email if user is passed in a role is state user
+    if (user?.email && user?.role === 'STATE_USER') {
+        stateReceiverEmails.push(user.email)
```
haworku, line 123: Is this new logic you are adding?  I thought we wanted to always add as a receiver the person triggering the email (the current user) and not restrict this to state users.

JasonLin0991: I scoped this function to only generate list of state emails so if current user was not a state, their email wouldn't get lumped into the list and be sent an email specifically for the states. The check was to make sure if somehow the user passed in was a CMS user it would skip this step.

haworku (approved):

PR #1060, opened 2022-07-15
[MR-2228] Add CMS Dashboard 

haworku:
## Summary
Still in draft but ready for review of api/db work. This is the work that is left:
- [x] wrap in feature flag
- [x] add Cypress tests - added a basic test for one submission. We would need a new approach to dealing with submissions in Cypress via api interaction or seeding the DB to have a robust way to test different combinations of submissions and how they should effect the dashboard.
- [ ] share more UI code between StateDashboard and CMSDashboard components

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2228

#### Test cases covered

see `CMSDashboard.test.tsx` (web tests)
- displays submissions table
- displays submissions table sorted by updatedAt field
- displays submissions table excluding any in progress draft
-  displays each health plan package status tag as expected
- displays program status tags as expected

see `indexHealthPlanPackages.test.ts` for isCMSUser block (api tests)
- returns an empty list if only draft packages exist
- return a list of submitted packages from multiple states

## QA guidance
Add different types of submission in different statues and check the CMS dashbaord


services/app-web/src/pages/CMSDashboard/CMSDashboard.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+            if (currentSubmissionData instanceof Error) {
+                recordJSException(
+                    `indexHealthPlanPackagesQuery: Error decoding proto. ID: ${sub.id} Error message: ${currentSubmissionData.message}`
+                )
+
+                return null
+            }
+            if (sub.status === 'DRAFT') {
+                // should display draft submissions to a CMS user - this is also filtered out on the api side
```
rswerve, line 96: should not, right?

rswerve:
Code looks good to me.  Happy to step through it when it's deployed.

services/app-api/src/resolvers/indexHealthPlanPackages.ts
```diff
@@ -13,38 +13,63 @@ export function indexHealthPlanPackagesResolver(
...
-        if (!isStateUser(user)) {
-            const errMsg = 'user not authorized to fetch state data'
-            logError('indexHealthPlanPackages', errMsg)
-            setErrorAttributesOnActiveSpan(errMsg, span)
-            throw new ForbiddenError(errMsg)
-        }
 
-        // fetch from the store
-        const results = await store.findAllHealthPlanPackages(user.state_code)
+        if (isStateUser(user)) {
```
macrael, line 19: We should be able to just change what store function we call based on the user type and still share all the cleanup code. The error handling and the package.map() should be the same.

macrael (approved):
Looks great! The db stuff makes good sense.

haworku:
@macrael hello!! Yes, I can't use the feature flag yet because I can't get the unit tests passing with it present. I think this is the first example of feature work with feature flags that are a boolean and unit tests together. So yes its not doing anything yet. I'll move this out of draft and into  final review when I get everything working together.

haworku:
I think this PR got in a weird state and I can't re-deploy. Going to close, rebase main and then reopen.

PR #1069, opened 2022-07-19
MR-860: Use CalendarDate where appropriate

rswerve:
## Summary

We have several dates that are typed to the troublesome javascript `Date`, which introduces hour/minute/second/time zone complexity when what we actually want is the calendar date.  In this PR, I created a new type, called CalendarDate--which looks like YYYY-MM-DD--and applied it to all those values where we don't care about clock time.

One interesting finding: some bugs were baked into our tests.  In resubmitPackageStateEmail.test.ts, check out the last test, "renders overall email as expected."  rateDateEnd is 11-31-2021, but the test was expecting 12-01-2021 in the generated string (and passing previously).  This is now working as intended.

#### Test cases covered

No new tests.  Old tests passing.

## QA guidance

Date behavior--entry, save, retrieve, go back, go forward--should be non-hinky.


services/app-web/src/common-code/dateHelpers/calendarDate.ts
```diff
@@ -1,19 +1,35 @@
...
+        return ''
+    }
+    const parsedCalendarDate = date.split('-')
+    const month = parsedCalendarDate[1]
+    const day = parsedCalendarDate[2]
+    const year = parsedCalendarDate[0]
+    return `${month}/${day}/${year}`
+}
+
+function formatDateTime(date: Date | undefined): string {
```
macrael, line 17: Are we still using this function? We probably don’t want the `tz.(‘UTC’)` for this if so since we should probably just resepct the timezone the Date was set in.

services/app-web/src/common-code/healthPlanFormDataType/UnlockedHealthPlanFormDataType.d.ts
```diff
@@ -66,6 +66,208 @@ type RateCapitationType = 'RATE_CELL' | 'RATE_RANGE'
...
+    | '05'
+    | '06'
+    | '07'
+    | '08'
+    | '09'
+    | '10'
+    | '11'
+    | '12'
+
+type Year =
```
macrael, line 116: Might be worth a comment here about how this is a potentail place for bugs if dates outside of these years get entered.

services/app-web/src/formHelpers/formatters.ts
```diff
@@ -29,24 +30,29 @@ const formatYesNoForProto = (
...
 }
 
-const formatFormDateForDomain = (attribute: string): Date | undefined => {
+const formatFormDateForDomain = (
+    attribute: string
+): CalendarDate | undefined => {
     if (attribute === '') {
         return undefined
     }
-    return dayjs.utc(attribute).toDate()
```
macrael, line 49: Yay. glad this utc hack is disappearing.

macrael (approved):
Looks great. So many test changes!

services/app-api/src/emailer/emails/unlockPackageStateEmail.test.ts
```diff
@@ -86,6 +87,9 @@ test('body content is correct', async () => {
 })
 
 test('renders overall email as expected', async () => {
+    console.log('maz sub: ', sub)
```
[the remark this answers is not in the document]
rswerve: fixed

services/app-web/src/common-code/healthPlanFormDataType/UnlockedHealthPlanFormDataType.d.ts
rswerve: resolved offline,  no change

services/app-web/src/common-code/dateHelpers/calendarDate.ts
rswerve: still using, and fixed

PR #1072, opened 2022-07-20
ADR: Form Changing Over Time

macrael:


docs/ADRs/018-changing-form-over-time.md
```diff
@@ -0,0 +1,91 @@
...
+
+We also discussed what to do when someone edits a submission that has become out of date. 
+
+#### Edit using the old schema
+
+Allow a user to re-submit using the same questions that they answered the first time they submitted the form. This would require us maintaining a version of the form inputs that matches old versions of the form. It would preserve all the questions asked at the original time of submission, even if the edits came years later
+
+#### Edit using the new schema
+
+When someone edits an out of date submission, the submission will be migrated to the new schema. This means that our form entry UI does not need to be aware of past versions, and it means that we expect states to answer new questions if they go back in to edit an old submission, bringing it up to date in the process. This is the route we chose to follow.
```
haworku, line 91: I would like to see the edits section fleshed out more. I think there is some complexity missing.

Anytime we are  editing old data using new a new data schema, we need a way to inform the user what their missing fields are.  We don't have form level validation and detailed error messages build into review and submit yet and we haven't had much of a need to since users up until now have been editing versions of data that match the current schema. 

However, once there is old form data at play, we need a way to force users to update the appropriate fields or display useful errors when they don't both client-side and server-side.

docs/ADRs/018-changing-form-over-time.md
```diff
@@ -0,0 +1,91 @@
...
+
+There are a variety of different changes to the form that will require changing the shape of our data. Two simple ones are adding a new field or removing an old one. More complex ones could be changing the type of a field or changing the values of an enum. We identified two broad strategies for dealing with this, one more invasive than the other. 
+
+First we can simply always add optional fields and only remove optional fields. By that I mean optional at the typescript field level. This forces our code that deals with the form data to always check on these fields and deal appropriately with the case where the data is missing. Since this mostly concerns the single page where we display the package summary, it's mostly a question of leaving question/answers off the page when they don't exist in the data. This works in both directions. Displaying an old submission that doesn't have one of the new questions on it, omit it. Displaying a new submission that doesn't have a question that has been removed, omit it. If a field's type needs to change, that can be accommodated by adding a new field and removing the old one. 
+
```
haworku, line 66: Another drawback here we could muddy the waters for what is the source of truth for form schema required fields. Because we still do have required fields we want to return failures for in certain types of form submits (such as submitting the package overall)

docs/ADRs/018-changing-form-over-time.md
```diff
@@ -0,0 +1,91 @@
...
+The drawback, and ultimately the deal breaker, is that we would then lose some actual answers submitted by states in the past. We decided that it was crucial to keep a complete record of what a state submitted in the past regardless of how the form has changed since. Reviews of these contracts take place on a timeline of years, we don't want to lose the record of what was submitted since that might be referenced over time in the adjudication of the submission. 
+
+
+### Schema Migration
+
+Since we want to keep old data intact and we need to be able to change the schema of the data over time, that means that our application will need to be able to deal with loading different schemas of data if it is viewing a recent or an old version of the form. 
+
+There are a variety of different changes to the form that will require changing the shape of our data. Two simple ones are adding a new field or removing an old one. More complex ones could be changing the type of a field or changing the values of an enum. We identified two broad strategies for dealing with this, one more invasive than the other. 
+
```
haworku, line 64:
```suggestion
### Optional Fields
```

docs/ADRs/018-changing-form-over-time.md
```diff
@@ -0,0 +1,91 @@
...
+First we can simply always add optional fields and only remove optional fields. By that I mean optional at the typescript field level. This forces our code that deals with the form data to always check on these fields and deal appropriately with the case where the data is missing. Since this mostly concerns the single page where we display the package summary, it's mostly a question of leaving question/answers off the page when they don't exist in the data. This works in both directions. Displaying an old submission that doesn't have one of the new questions on it, omit it. Displaying a new submission that doesn't have a question that has been removed, omit it. If a field's type needs to change, that can be accommodated by adding a new field and removing the old one. 
+
+The drawback to this is that our summary page has to be able to display every single question ever asked on the form. As the form changes, our summary page will grow continuously, and in a kind of patchwork where different submissions across time will display different subsets of all the questions. We can mitigate this by building tests around rendering the summary for old submissions. 
+
```
haworku, line 68:
```suggestion
### Version the data
```

docs/ADRs/018-changing-form-over-time.md
```diff
@@ -0,0 +1,91 @@
...
+  optional ContractInfo contract_info = 15;
+  repeated Document documents = 16;
+
+  repeated RateInfo rate_infos = 50;
+}
+```
+
+You can see that each field has a type, and those types can include enums or other proto messages. You will also see that everything is optional, that is Protobuf's solution to evolving schemas, any field you ask for might be missing so you have to prepare for that. You will see that we have adopted a similar scheme for managing our changing data throughout the app. 
+
+## Options Considered
```
haworku, line 42: Where the decision made section? It feels like options considered section has both discussion of pros and cons and options as well as making a specific decision. I'm finding it a bit hard to parse out what the proposed decision for the ADR is.

docs/ADRs/018-changing-form-over-time.md
```diff
@@ -0,0 +1,91 @@
...
+
+The drawback to this strategy is that it introduces a lot of overhead to form changes. The more the form changes, the more versions we have to track, bloating our data types and the code that has to deal with them. 
+
+The nice thing about these two strategies is that they are complimentary. We have decided to start by just adding optional fields to the existing FormData type. Our summary page will be able to display old versions of the data simply by omitting questions for any optional fields not present. If at some point the accumulation of old questions on that page become too onerous to maintain, or if we need to drastically change the shape of the form data, we can version the data then, making a clean brake and simplifying things again.
```
haworku, line 73:
```suggestion
The nice thing about these two strategies is that they are complimentary. We have decided to start by just adding optional fields to the existing FormData type. Our summary page will be able to display old versions of the data simply by omitting questions for any optional fields not present. If at some point the accumulation of old questions on that page become too onerous to maintain, or if we need to drastically change the shape of the form data, we can version the data then, making a clean break.
```

docs/ADRs/018-changing-form-over-time.md
```diff
@@ -0,0 +1,91 @@
...
+
+### On Updating Old Data
+
+We also discussed what to do when someone edits a submission that has become out of date. 
+
+#### Edit using the old schema
+
+Allow a user to re-submit using the same questions that they answered the first time they submitted the form. This would require us maintaining a version of the form inputs that matches old versions of the form. It would preserve all the questions asked at the original time of submission, even if the edits came years later
+
+#### Edit using the new schema
```
haworku, line 89:
```suggestion
#### Edit using the new schema and adding form level validations to review and submit page
```

I think this is what this is suggesting?

docs/ADRs/018-changing-form-over-time.md
```diff
@@ -0,0 +1,91 @@
...
+
+Finally we come to the details of how we want to display different versions of this data. There were design questions and technical questions we considered. 
+
+In line with our desire to maintain the complete record of past submissions, it makes sense for us to attempt to display the summary of an old submission in the same state as it was originally submitted. Rather than displaying questions with blank answers, we want to just omit questions and answers all together when a submission is missing a field. This makes an old submission missing a new question appear the same way it did when originally submitted, rather than appearing like it's missing something. 
+
+### On Updating Old Data
+
+We also discussed what to do when someone edits a submission that has become out of date. 
+
+#### Edit using the old schema
```
haworku, line 85:
```suggestion
#### Edit using the old schema, requires versioned form UI and versioned schemas 
```

I think this is what this is saying?

haworku (approved):
This looks good to me the main change I would like to see is pulling out the proposed solution more clearly. Also maybe clarifying if there are options that are mutually exclusive (we have a pick one) or ones that could be used in parallel with eachother.

JasonLin0991 (approved):
Looks great!

docs/ADRs/018-changing-form-over-time.md
MegMurphy: @haworku @macrael  To add my 2 cents on this comment and the resulting thread: I am not sure I agree with the supposition that we will always need to “force” users to update the newly added fields and will need to do an overall form-level validations, in the even that they do not find themselves on an unlocked impacted form page that would trigger page-level validations.

In some cases, we may populate data into newly added fields via a data migration. 

In many other cases, the newly field may be fine being null for previously made submissions, and may only need to be filled it in the event that the submission is unlocked for other reasons. 

If the submission has already been made and is pending and it IS important that CMS go back and capture missing data for a new field from a state, a reasonable approach would be for CMS to unlock the submission and request the state explicitly complete new field. (Yes, technically a state could resubmit without making the desired change, based on how they actually navigated thru that resubmit, but they could technically do that for any edit request; we don't have a way we are necessarily validating that a state ever makes the change that was requested of them before they resubmit). 

While do have existing designs for form-level validations, this is rather low down in the backlog while we prioritize building functionality to alleviate communications and milestone tracking pain in the more immediate future. I'd suggest a reasonable approach for now is to proceed with current approach and evaluate over the course of our expanding pilot if the frequency of incomplete submissions bubbles up as a priority pain point for CMS that necessitates further exploration.

macrael: This might be a good topic for our next DTBM

MegMurphy: No, that was not what I meant. To be clear, if they touch the contract details page I would still expect page-level validations to trigger on that page. 

I thought what Hana was saying was if the state did NOT touch contract details at all, and opened up form directly from the supporting documents page and went to resubmit, they could manage to resubmit without touching the covid question at all. That that seems fine for now, and not a huge issue to be solved.

macrael: I see, so you are ok with us enforcing the page level validations but so long as they don’t visit that page we could still submit without error. 

If that’s the way we go this, it does complicate the story technically a little bit. Right now, the server validates packages when we call the submit or re-submit API to ensure that we are putting complete packages into the db. doing this feature would require relaxing that for re-submissions, at least. Something worth talking about.

PR #1073, opened 2022-07-20
[MR-2228] Add CMS Dashboard

haworku:
## Summary
Add CMS Dashboard behind `cms-dashboard` flag. 

Note: Tests are intentionally skipped, we will need to implement an approach to testing with LD mocks to get these passing consistently regardless of environment.  This will be a separate effort. 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2228

#### Test cases covered

see `CMSDashboard.test.tsx` (web tests)
- displays submissions table
- displays submissions table sorted by updatedAt field
- displays submissions table excluding any in progress draft
-  displays each health plan package status tag as expected
- displays program status tags as expected

see `indexHealthPlanPackages.test.ts` for isCMSUser block (api tests)
- returns an empty list if only draft packages exist
- return a list of submitted packages from multiple states

## QA guidance
Add different types of submission in different statues and check the CMS dashbaord


services/app-api/src/resolvers/indexHealthPlanPackages.ts
```diff
@@ -1,50 +1,62 @@
...
 import { logError, logSuccess } from '../logger'
-import { isStoreError, Store } from '../postgres'
+import { isStoreError, Store, StoreError } from '../postgres'
 import {
     setErrorAttributesOnActiveSpan,
     setResolverDetailsOnActiveSpan,
     setSuccessAttributesOnActiveSpan,
 } from './attributeHelper'
 
+const validateAndReturnHealthPlanPackages = (
```
macrael, line 13: Nice way to do this. I was imagining leaving this below and doing some kind of 
```
let results = undefined
if (isStateUser) {
  results = findAllByState()
} else if (isCMSUser) {
  results = findAll()
}

… do stuff with results
```
but your version doesn’t have to deal with the undefined `results` type which is nice.

macrael (approved):
Looks good

PR #1079, opened 2022-07-22
Change deploy.yml GH permissions for dependabot

mojotalantikite:
## Summary

We'd like to have dependabot PRs run through CI without needing to be manually triggered by an engineer. This gives PRs opened by dependabot write access via the `GITHUB_TOKEN` so that our review apps can be created.


macrael (approved):
lg

PR #1081, opened 2022-07-23
Run app-web serverless serially

mojotalantikite:
## Summary

Maz had been seeing some CI issues in `app-web` where the serverless deploy would hit an OOM. We fixed it in his branch, but it looks like it got reverted with #1075 . This adds it back in order to avoid those occasional GHA OOMs in `app-web`.


macrael (approved):
I guess correctness is more important than speed :p

macrael:
What was the failure that this caused?

mojotalantikite:
@macrael it was starting to cause OOM errors in `app-web` CI builds, so this ends up saving us from that

PR #1082, opened 2022-07-25
[MR-2419] States can upload XLSM files

JasonLin0991:
## Summary
[MR-2419](https://qmacbis.atlassian.net/browse/MR-2419)

- Add all MS macro enabled excel files to accepted files.
- Logging for when a user tries to upload an invalid type.


services/app-web/src/components/FileUpload/FileProcessor/FileProcessor.tsx
```diff
@@ -117,11 +124,12 @@ export const FileProcessor = ({
     const isLoading = status === 'PENDING'
     const isScanning = status === 'SCANNING'
     const isComplete = status === 'UPLOAD_COMPLETE'
+    const fileType = name.slice(name.lastIndexOf('.'))
```
macrael, line 127: Since we’re messing with what is bascially a path, I’d use `path.extname` here

macrael (approved):
looks good to me, definitely take a look at the path module, those functions are good to use anytime they are called for, especially since they will do the right thing on window/mac/linux where things can be different

PR #1095, opened 2022-08-01
[MR-2462] Add Tealium

haworku:
## Summary
Add Tealium to the application as per CMS [page level tracking and Tealium installation instructions](https://confluence.cms.gov/display/BLSTANALYT/Page+Tracking+Template) 

Next steps after PR merges: Check with CMS that implements Tealium verifies our install is working. 

Other useful links:
- [Tealium SPA documentation](https://docs.tealium.com/platforms/javascript/single-page-applications/)
- [Tealium Track Views documentation](https://docs.tealium.com/platforms/javascript/track/)

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2462
https://jira.cms.gov/browse/BLSTANALYT-4173


services/app-web/src/types/index.d.ts
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export {}
+
+declare global {
```
haworku, line 4: Need to extend the window type to be able to interact with utag.

haworku:
FYI I am working on a commit to remove tealium profile out from env secrets. It is not needed there.

services/app-web/src/hooks/useScript.ts
```diff
@@ -1,22 +1,48 @@
...
+
+// Add script pointing to third party URL to bottom of page, with handling for a boolean flag on or off
+const useScript = ({
+    url,
+    id,
+    featureFlag,
+}: {
+    url: string
+    id: string
+    featureFlag: boolean
```
rswerve, line 33: Is this for general purpose scripts?  Feature flags won't always be booleans.

services/app-web/src/hooks/useTealium.ts
```diff
@@ -0,0 +1,104 @@
...
+        // Load utag.sync.js - add to head element - SYNC load from src
+        const initializeTagManagerSnippet = createScript({
+            src: `https://tags.tiqcdn.com/utag/cmsgov/${tealiumProfile}/${tealiumEnv}/utag.sync.js`,
+            id: 'tealium-load-tags-sync',
+        })
+        document.head.appendChild(initializeTagManagerSnippet)
+
+        // Load utag.js - Add to body element- ASYNC load inline script
+        const inlineScript =
+            document.createTextNode(`(function (t, e, a, l, i, u, m) {
```
rswerve, line 47: wut

rswerve (approved):

services/app-web/src/hooks/useTealium.ts
haworku: lol I know. If you look at both the docs from Tealium and the CMS page level tracking instructions - they both say load the script this way (as an inline anonymouse script) so that's why I did it.  The `t,e,a,l,i,u,m` stuff is from the CMS setup specifically.

services/app-web/src/hooks/useScript.ts
haworku: This is meant to be more a true false show the script or not. I will name the variable that `showFlag` instead of `featureFlag` may be clearer.  You would have to do the logic around determining the boolean in the place the hook is used e.g. if the feature flag was something that is not already true false you calculate `featureflag === THE_VALUE_THAT_MEANS_THE_FLAG` and pass that in to this parameter in hook to get a boolean value

PR #1096, opened 2022-08-01
[MR-2237] Rate name shows the programs the rate cert applies to

JasonLin0991:
## Summary
[MR-2237](https://qmacbis.atlassian.net/browse/MR-2237)

- Rate name generation using `generateRateName` can now handle when a submission has `rateProgramIDs`.
- Generating rate names has been moved from the emailer to the resolvers `unlockHealthPlanPackage` and `submitHealthPlanPackage` if submission type is `CONTRACT_AND_RATES` and passed to the emailer.
- If there is no `rateProgramIDs` on a `CONTRACT_AND_RATE` submission it will use the package `programIDs` to generate a name. This is exactly how the rate name format is generated before rate certification programs feature. 

#### Test cases covered
- `unlockHealthPlanPackage.test.ts` and `submitHealthPlanPackage.test.ts`
  - Add test for rate name
  - Add tests `'generates rate name by package programs when rate programs are not specified'`
- Update email tests for change in `generateRateName`
- Update `createAndUpdateTestHealthPlanPackage` helper function to include `rateProgramIDs`
- `healthPlanFormData.test.ts`
   - Update generateRateName tests

## QA guidance

- Please enable `rate-certification-programs` feature flag in LaunchDarkly before testing this. Although this isn't hidden behind the feature flag, selecting programs that apply to a rate certification is.

Please disable `rate-certification-programs` feature flag when done testing.


services/app-api/src/emailer/emailer.ts
```diff
@@ -46,31 +46,37 @@ type Emailer = {
     sendCMSNewPackage: (
         submission: LockedHealthPlanFormDataType,
         submissionName: string,
```
haworku, line 48: Do you think it would work to just stop passing `submissionName` and `rateName` both to each emailer function and instead just pass the `programs`. One parameter instead of 2? I can't tell why we should keep passing in strings instead of just generating the strings within the emailer functions closer to where they are used.

services/app-api/src/resolvers/submitHealthPlanPackage.ts
```diff
@@ -258,8 +259,14 @@ export function submitHealthPlanPackageResolver(
         }
 
         // Send emails!
+        const isContractAndRate =
+            lockedFormData.submissionType === 'CONTRACT_AND_RATES'
+
         const name = packageName(lockedFormData, programs)
         const status = packageStatus(updatedPackage)
+        const rateName = isContractAndRate
```
haworku, line 267: This kind feels like something that should go in the emailer functions not up in the resolver

services/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.ts
```diff
@@ -163,10 +162,14 @@ function programNames(
...
     submission: HealthPlanFormDataType,
-    statePrograms: ProgramArgType[]
+    statePrograms: ProgramArgType[],
+    programIDs?: string[]
 ): string {
     const padNumber = submission.stateNumber.toString().padStart(4, '0')
-    const pNames = programNames(statePrograms, submission.programIDs)
+    const pNames =
+        programIDs && programIDs.length > 0
+            ? programNames(statePrograms, programIDs)
```
haworku, line 171: Can you add a comment in here about why the ternary is needed? I think we could easily forget about this context and look at this function later confused how `programIds` could not exist

services/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.ts
```diff
@@ -180,17 +183,22 @@ function packageName(
...
         rateDateEnd,
         rateDateStart,
+        rateProgramIDs,
     } = rateData
-    let rateName = `${submissionName}-RATE`
+    let rateName = `${packageName(
+        rateData,
+        statePrograms,
+        rateProgramIDs
+    )}-RATE`
```
haworku, line 201: I like this change- its more self contained and more trustworthy to unit test rather than passing in the already formatted `submissionName` as an opaque string

haworku (approved):

haworku:
🐛   I think I found a bug with the emails - getting unknown rate name for the email associated with [this submission](https://d2981i51x3udyf.cloudfront.net/submissions/32bde198-e2f0-42f5-94ca-7f747b6782e0).  The submission looks fine in the review and submit page. Maybe we don't have enough unit tests around the rate names part of the emails? 
Screen Shot 2022-08-02 at 5 23 24 PM

services/app-api/src/emailer/emailer.ts
JasonLin0991: Yea totally doable. I think the reason why I didn't go this way is it seems like repeat code in each emailer function, like how it used to with generating rate name in the emailer functions. Also now that I changed `store.findPrograms()` to find combined `programIDs` and `rateProgramIDs` we should be able to pass that to the emailers. This was the reason for the bug you mentioned below.

haworku: @macrael Can you weigh in here? I know you had thoughts about what the parameters and bounds of the emailer service should be

JasonLin0991:
> 🐛 I think I found a bug with the emails - getting unknown rate name for the email associated with [this submission](https://d2981i51x3udyf.cloudfront.net/submissions/32bde198-e2f0-42f5-94ca-7f747b6782e0). The submission looks fine in the review and submit page. Maybe we don't have enough unit tests around the rate names part of the emails?

`programs`, resulted from `store.findPrograms()`, is used to find program names for selected package programs and rate programs. So the bug occurs because we only had package programs to find program names from. I fixed this by 

- Combining both `programIDs` and `rateProgramIDs`(removing duplicates) from the package.
- Then find `store.findPrograms()` using the combined programIDs.
- Resulting list of programs to find program names in `packageName` and `generateRateName`

services/app-api/src/resolvers/unlockHealthPlanPackage.ts
```diff
@@ -129,16 +130,30 @@ export function unlockHealthPlanPackageResolver(
             throw new Error(errMessage)
         }
 
-        const programs = store.findPrograms(draft.stateCode, draft.programIDs)
-        if (!programs || programs.length !== draft.programIDs.length) {
-            const errMessage = `Can't find programs ${draft.programIDs} from state ${draft.stateCode}, ${draft.id}`
+        const isContractAndRate = draft.submissionType === 'CONTRACT_AND_RATES'
+
+        const combinedProgramIDs = [...draft.programIDs]
+        if (isContractAndRate && draft.rateProgramIDs) {
```
haworku, line 136: Could we just adjust our store function tho return ALL possible programs for the state if no program ids are passed in?  I think that might cut down on complexity here in the resolver, we wouldn't need to generate that specific program ids list here and in each resolver that needs it. 

Instead,  we could just do that filtering and error handling logic in one place- in  the  shared`generateRateName` utility. Just pass in all available programs for the state each time.

haworku (approved):
Emails look good

services/app-api/src/resolvers/unlockHealthPlanPackage.ts
JasonLin0991: We are also checking to make sure the length of the `programs` is the same as the `combinedPrograms`. I believe this was done to make sure we could find all the program names and throw and error if not.

```
    if (!programs || programs.length !== combinedProgramIDs.length) {
        const errMessage = `Can't find programs ${combinedProgramIDs} from state ${draft.stateCode}, ${draft.id}`
        logError('unlockHealthPlanPackage', errMessage)
        throw new Error(errMessage)
    }
```

JasonLin0991: I think we could make a return all programs, but it would require more refactoring than expected. 

We would still need to check to make sure if programs from the `store` and programs from the package matched in length, otherwise we would get `Unknown Program` when generating names. 

We are also throwing this error before updating the db, moving this would still update the db, but not send emails on errors.

```
        if (!programs || programs.length !== combinedProgramIDs.length) {
            const errMessage = `Can't find programs ${combinedProgramIDs} from state ${draft.stateCode}, ${draft.id}`
            logError('unlockHealthPlanPackage', errMessage)
            throw new Error(errMessage)
        }
```

```
if (!programs || programs.length !== combinedProgramIDs.length) {
```

It doesn't seem explicit that this checks to make sure we have matching programs, but `findPrograms()` filters out state programs from our `statePrograms.json` that are not included in programs from the submission. If the length does not match, this means a program there was an issue finding a specific program from either the submission or `statePrograms.json` which would cause errors generating names.

services/app-api/src/emailer/emailer.ts
macrael: I noticed a few things here

1. rateName is being added as a new param to all the email functions, even though some of those functions wrap up the submissionName inside of an options dictionary and some don’t. Ideally we’d be more consistent. Some of the functions take the whole submission and some take an object like “UpdatedEmailData” with just the info we need for this particular email. Until we sort all of that out though, let’s add new options the way the function currently accepts them. 
2. Thinking a little ahead, we are soon going to have packages that have multiple rates associated with them. A single “rateName” will not cut it. I don’t know if an array of rate names will work, but my hunch is we’ll want to be displaying info about each rate individually and just indexing into an array of names feels a bit haphazard. 
3. Setting the rateName to undefined for ContractOnly submissions definitely looks a bit off to me, I don’t love it when a funciton can be called with arguments that don’t actually make sense. Like if you passed a rate name in even though the submission is ContractOnly, what would happen?

This seems like it might be worth a quick discussion to talk about the pro/cons of how we’re passing info into the emailer.

JasonLin0991: > Thinking a little ahead, we are soon going to have packages that have multiple rates associated with them A single “rateName” will not cut it.

This is a very good point I overlooked.

> Setting the rateName to undefined for ContractOnly submissions definitely looks a bit off to me, I don’t love it when a funciton can be called with arguments that don’t actually make sense.

Also a good point

> stop passing submissionName and rateName both to each emailer function and instead just pass the programs.

@haworku suggestion seems like the better way to go? We can generate names as needed in the emailers, especially when we're dealing with multiple rates per package.

> This seems like it might be worth a quick discussion to talk about the pro/cons of how we’re passing info into the emailer.

This would be good.

JasonLin0991:
@macrael @haworku I moved package name and rate name generation to the emailer functions and removed `UpdatedEmailData` for `UpdateInfoType` from domain-models. Renaming this data to `updateInfo` consistently from resolvers to emailer functions should make it less confusing now. 

A couple functions I wanted to get some eyes on are `findAllPackageProgramIds` in `emailer/templateHelpers.ts` and `findProgramsHelper` in `testHelpers/emailHelpers`.  

- `findAllPackageProgramIds` was made because this code is used in many places from resolvers, emailer functions and tests. The only thing that irks me is where the function lives, there is probably a better place for this. 

- `findProgramsHelper` was to mimic `store.findPrograms` but without using `statePrograms.json` and pulling statePrograms from our `mockMNState()` and is only used in tests. I made this because I didn't want to adjust our `findPrograms` function in `store` just for testing, but this also feels messy.

services/app-api/src/emailer/templateHelpers.ts
```diff
@@ -130,6 +126,21 @@ const generateStateReceiverEmails = (
     return pruneDuplicateEmails(stateReceiverEmails)
 }
 
+//Finds all programs ids in a package and combines them into one array removing duplicates.
+const findAllPackageProgramIds = (
+    pkg: UnlockedHealthPlanFormDataType | LockedHealthPlanFormDataType
+): string[] => {
+    const programs = [...pkg.programIDs]
```
macrael, line 133: This could be a good place to use a `Set` instead of an array. then you dont have to do any of the “if .includes()” checks yourself.

services/app-api/src/resolvers/createHealthPlanPackage.ts
```diff
@@ -41,10 +41,7 @@ export function createHealthPlanPackageResolver(
             input.programIDs
         )
 
-        if (
-            programs === undefined ||
-            programs.length !== input.programIDs.length
-        ) {
+        if (programs instanceof Error) {
```
macrael, line 44: I thought the plan was just to pass _all_ of the programs for a given state into the emailer and have it be responsible for the parsing. then we skip dealing with this error all together.

services/app-api/src/emailer/emails/resubmitPackageStateEmail.ts
```diff
@@ -16,20 +16,24 @@ import type { EmailData, EmailConfiguration } from '../'
...
-    resubmittedData: UpdatedEmailData,
-    config: EmailConfiguration
+    updateInfo: UpdateInfoType,
+    config: EmailConfiguration,
+    programs: ProgramType[]
 ): Promise<EmailData | Error> => {
     const isUnitTest = config.baseUrl === 'http://localhost'
     const isTestEnvironment = config.stage !== 'prod'
     const receiverEmails = generateStateReceiverEmails(pkg, user)
+    const packageName = generatePackageName(pkg, programs)
```
macrael, line 26: I think we should move the potential error of not having a program that matches one of the programs we need into the emailer. Right now this call is safe because we checked that all the right programs exist in the resolver. I think it would be better to make that check closer to where we actually care about the programs existing, which is here in the emailer. You could do the same check here, returning an error if we are missing a program, and then handling that error in the resover the same way we handle any errors that the emailer could encounter.

It’s definitely a little six of one, half dozen the other, since we’re making the check either way, but it just trips me up a little that this generatePackageName call could fail to find the program it needs silently if we forget to do the check at the resolver level. It’s another case where I think we have sligtly less code as is, but at the cost of moving the error check outside the layer that actually cares about it.

macrael (approved):
This is a great cleanup. I like it when we can go ahead and make a small refactor in service of an existing PR.

services/app-api/src/testHelpers/emailerHelpers.ts
```diff
@@ -439,10 +489,27 @@ const mockContractAmendmentFormData = (
...
+        rateProgramIDs: ['3fd36500-bf2c-47bc-80e8-e7aa417184c5'],
         ...submissionPartial,
     }
 }
 
+const findProgramsHelper = (
+    stateCode: string,
+    programIDs: string[]
+): ProgramType[] | Error => {
+    const programs = mockMNState().programs.filter((program) =>
```
haworku, line 501: docs: please add a comment about why we put state specific logic in this test helper

services/app-api/src/emailer/templateHelpers.ts
JasonLin0991: Done! 

I thought `Set()` was added in es6 and not available in es5 for some reason. In our `tsconfig.json` in `app-api` the `complierOptions` has target set to es5.

services/app-api/src/emailer/emails/resubmitPackageStateEmail.ts
JasonLin0991: Done! I just moved `findPrograms` and `findAllPackageProgramIds` to the emailer functions. Returning error if `findProgams` encounters one.

services/app-api/src/testHelpers/emailerHelpers.ts
JasonLin0991: This function was removed entirely since we moved `findPrograms` to emailer functions, it was easier to use actual programs for each state. For instance testing `CHIP` submission, I edited the submission to be `MS` state and programs for `MS`. We could create a more robust mock of a state, but I feel this PR has gotten pretty long.

services/app-api/src/emailer/templateHelpers.ts
macrael: looks like you backed out the Set changes, I thought it did work?

I was imagining it more like this:
```
  const programs = new Set(pkg.programIDs)
  if (pkg.submissionType === 'CONTRACT_AND_RATES' && pkg.rateProgramIDs) {
     pkg.rateProgramIDs.forEach( id => programs.add(id))
  }
  return programs
```

I don't think there's anything wrong with returning a set instead of an array, all the iterators will still work.

services/app-api/src/emailer/emails/newPackageCMSEmail.ts
```diff
@@ -25,12 +30,24 @@ export const newPackageCMSEmail = async (
         pkg,
         stateAnalystsEmails
     )
+    const combinedProgramIDs = findAllPackageProgramIds(pkg)
+    //Get program data from combined program ids
+    const programs = findPrograms(pkg.stateCode, combinedProgramIDs)
```
macrael, line 35: Sorry for all the churn, but I actually think that our call to findPrograms specifically should still be in the resolver. I was thinking that we would just grab a list of all the programs for the given state and pass that into the emailer. Oner day that lookup will likely require a database connection and I don't want to have to pass that into the emailer.

macrael (commented):
Let's chat about this, I think I wasn't totally clear in my previous review.

macrael (approved):
This looks great. I like the overall simplification in the submit resolver.

PR #1099, opened 2022-08-02
[MR-2485] Program selection combobox not focused when error clicked

JasonLin0991:
## Summary
[MR-2485](https://qmacbis.atlassian.net/browse/MR-2485)

Seems like the issue was the `name` prop for `Select` component was not applying or incorrectly applying to the `input` element used in the component. Then focusing using the `name` attribute did not work. 

The fix was to pass `inputId` to `Select` component from `react-select` and then focus using `id` instead of `name` attribute. Then refactor forming error object passed into `ErrorSummary` to select `rateProgramIDs/programIDs` by adding `#` in front of each key. As described in the comments in `ErrorSummary` adding an `#` will switch from focusing using `name` to `id` attribute.

## QA guidance

1. Please make sure `rate-certificiation-programs` feature flag is enabled in LaunchDarkly.
2. On the Submission type or Rate Details page, leave the program selection box empty and click continue. 
3. Click on the link the error summary should now scroll to the program select box and highlight it.
4. Please make sure to disable feature flag when finished.

### DIFF
--- a/services/app-web/src/components/ProgramSelect/ProgramSelect.tsx
+++ b/services/app-web/src/components/ProgramSelect/ProgramSelect.tsx
@@ -52,9 +52,8 @@ export const ProgramSelect = ({
             })}
             className={styles.multiSelect}
             classNamePrefix="program-select"
-            id="programSelect"
+            id={`${name}-programSelect`}
             name={name}
-            aria-label="programs (required)"
             options={programOptions}
             isMulti
             ariaLiveMessages={{
--- a/services/app-web/src/components/ProgramSelect/ProgramSelect.tsx
+++ b/services/app-web/src/components/ProgramSelect/ProgramSelect.tsx
@@ -4,6 +4,7 @@ import Select, { AriaOnFocus, Props } from 'react-select'
 import { Program } from '../../gen/gqlClient'
 
 export type ProgramSelectPropType = {
+    name: string
     statePrograms: Program[]
     programIDs: string[]
 }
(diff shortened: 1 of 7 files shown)


services/app-web/src/components/ProgramSelect/ProgramSelect.tsx
```diff
@@ -50,8 +52,8 @@ export const ProgramSelect = ({
             })}
             className={styles.multiSelect}
             classNamePrefix="program-select"
-            id="programIDs"
-            name="programIDs"
+            id="programSelect"
```
haworku, line 55: Wouldn't this be a bit problematic when we have multiple program selectboxes on the same page (such as in the multi-rate world which is coming soon). We would have multiple elements with the same html id on the page.  I am now thinking we should never be hard-coding this id at all fully - maybe make its value something like `${name}-select`  where name is always assigned to something specific like "contractPrograms" or "rate1programs"

JasonLin0991: Doh, good catch.

haworku:
Screen Shot 2022-08-02 at 4 46 20 PM

nit: It looks like the error summary order has fallen out of sync from the way the form is displayed. I believe we intended this order to follow the order of the fields listed on the form. This order is controlled in the RateDetailsSchema file - order matters in that file. Could you please move rate program ids up the other fields since its at the top of the list. 

Also documents should be the very first (documents is added to the list in `RateDetails.tsx` in the `<Error Summary />` error prop.

haworku (approved):

JasonLin0991:
> 

This was a bug on my part. The way I was adding new key value pairs to the error object was adding them to the bottom rather than the top.

PR #1102, opened 2022-08-03
Fix vulnerability in uploads service

mojotalantikite:
## Summary

This pulls in the changes from the quickstart repository to mitigate against potential injection attacks in the `uploads` service.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2499

### DIFF
--- a/services/uploads/src/antivirus.js
+++ b/services/uploads/src/antivirus.js
@@ -45,7 +45,7 @@ function downloadFileFromS3(s3ObjectKey, s3ObjectBucket) {
     }
 
     const tmpFileName = `${crypto.randomUUID()}.tmp`;
-    let localPath = `${constants.TMP_DOWNLOAD_PATH}${tmpFileName}`;
+    let localPath = path.join(constants.TMP_DOWNLOAD_PATH, tmpFileName);
     let writeStream = fs.createWriteStream(localPath);
 
     utils.generateSystemMessage(
--- a/services/uploads/src/antivirus.js
+++ b/services/uploads/src/antivirus.js
@@ -5,6 +5,7 @@
 const AWS = require('aws-sdk');
 const path = require('path');
 const fs = require('fs');
+const crypto = require('crypto');
 const clamav = require('./clamav');
 const s3 = new AWS.S3();
 const utils = require('./utils');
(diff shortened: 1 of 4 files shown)


services/uploads/src/antivirus.js
```diff
@@ -17,6 +18,11 @@ const constants = require('./constants');
  * @return {int} Length of S3 object in bytes.
  */
 async function sizeOf(key, bucket) {
+    console.log('key: ' + key);
+    console.log('bucket: ' + bucket);
+    var sts = new AWS.STS();
```
macrael, line 23: are either of these two varaibles used?

services/uploads/src/antivirus.js
```diff
@@ -33,25 +39,18 @@ async function isS3FileTooBig(s3ObjectKey, s3ObjectBucket) {
...
-    const downloadDir = `/tmp/download`;
-    if (!fs.existsSync(downloadDir)) {
-        fs.mkdirSync(downloadDir);
+    if (!fs.existsSync(constants.TMP_DOWNLOAD_PATH)) {
+        fs.mkdirSync(constants.TMP_DOWNLOAD_PATH);
     }
 
-    let localPath = pathFromObjectKey(s3ObjectKey);
+    const tmpFileName = `${crypto.randomUUID()}.tmp`;
+    let localPath = `${constants.TMP_DOWNLOAD_PATH}${tmpFileName}`;
```
macrael, line 53: ugh, i should have reviewed this in their PR. I wish this was actually shared code!! We should be using path.join here.

macrael (approved):
Looks good. Let’s just keep the code like it is upstream if either of my comments conflict with that.

services/uploads/src/antivirus.js
mojotalantikite: They don't seem to be -- I just copied the files over, but yeah I think we can delete these.

services/uploads/src/antivirus.js
mojotalantikite: yup good catch, I'll fix it

JasonLin0991:
@mojotalantikite I believe you're cypress test is failing because the `rate-certification-programs` was turned on. You should be able to get this passing if we toggle it off.

haworku:
@mojotalantikite sorry that my bad I forgot to toggled off the flag while I was testing like @JasonLin0991 said. also fyi for the cypress github action I'm bumping that on the cypress dependabot branch now after running into an issue

PR #1104, opened 2022-08-05
Print storybook endpoint

JasonLin0991:
## Summary
Print put story book endpoint in `Finish Prep`


macrael (approved):
then we have it

haworku (approved):
thanks for pulling out a standlone PR for this too

PR #1119, opened 2022-08-12
MR-929: Combine code coverage reports into one

rswerve:
## Summary

We run test suites for app-api, app-web, and cypress.  Our coverage for each one has been in its own reports, which doesn't capture overlap or different sections of a file being covered by different test suites.

This ticket was to combine all those coverage reports into one.

Currently, the combined report can be accessed in two ways: locally, by running `yarn run test:coverage` and in CI, in the "combined-test-coverage" artifact.

Code will not be instrumented and coverage numbers will not be generated in val or prod.

#### Screenshots

image

#### Test cases covered

No new tests.  A couple of tests remain skipped or commented out.

## QA guidance

What do we think of the current exclusions?  Some of the low coverage auth modules look possibly skippable.


services/app-web/src/pages/StateSubmission/Documents/Documents.test.tsx
```diff
@@ -944,45 +944,6 @@ describe('Documents', () => {
     })
 
     describe('Document categories checkbox', () => {
-        it('not present on contract only submission, categories default to contract-supporting', async () => {
```
haworku, line 947: Did you mean to comment this out? Seems like a valid test

package.json
```diff
@@ -18,16 +18,84 @@
...
+    "nyc": {
+        "report-dir": "coverage-cypress",
+        "exclude": [
+            "/*.stories.tsx",
+            "**/src/s3",
+            "**/src/api",
+            "**/src/testHelpers",
+            "**/src/localAuth",
+            "**/src/gen",
+            "**/src/components/Header/StateIcon/StateIcon.tsx",
```
haworku, line 52: I think we could add `/constants` folder, `TraceContext.tsx`, `useTealium.ts` for app-web. And I think we could exclude `awsParameterStore.ts` for app-api. 

I see what you are saying about the low level auth functions maybe being an option to exclude, though at the same time it feels like we could be testing some of those using mocks so I'm of two minds about. We could at least be testing our error handling. I'm thinking lets leave it for now

.github/workflows/deploy.yml
```diff
@@ -354,3 +366,76 @@ jobs:
...
+          mkdir -p cypress_temp
+          mkdir -p .nyc_output
+          echo "listing current folder"
+          ls
+          echo "listing a cypress folder"
+          ls ./cypress-json-coverage-1/
+          cp ./cypress-json-coverage-1/coverage-final.json ./cypress_temp/from-cypress-1.json
+          cp ./cypress-json-coverage-2/coverage-final.json ./cypress_temp/from-cypress-2.json
+          cp ./cypress-json-coverage-3/coverage-final.json ./cypress_temp/from-cypress-3.json
+          cp ./cypress-json-coverage-4/coverage-final.json ./cypress_temp/from-cypress-4.json
```
haworku, line 424: This a pretty long script. Can we quickly document whats is happening somewhere - maybe make a new file `docs/testCoverage` with a couple sentences about this approach.

haworku (commented):
I'm not seeing the combined report yet running locally, it seems like my tests are just timing out `yarn run test:coverage`. Were you having issues like that at all, for example with app-api tests not completing or cypress tests not exiting?

haworku (approved):
I do see the lcov report in the CI artifact so ✅

services/app-web/src/pages/StateSubmission/Documents/Documents.test.tsx
rswerve: it's back

package.json
rswerve: added

.github/workflows/deploy.yml
rswerve: added some echo lines to make it clear what's happening.  it's really just a bunch of copying, and a couple of merges...

PR #1121, opened 2022-08-15
MR-2460: Unit testing with Launch Darkly flags

JasonLin0991:
## Summary
[MR-2460](https://qmacbis.atlassian.net/browse/MR-2460)

Added a new jestHelper function `ldUseClientSpy`  that can be used to spy on `useLDClient` and return a flag value we set when calling `ldUseClientSpy`. This is a quick implementation that fixes our issue with skipped and commented out tests due to feature flags.

- The `ldUseClientSpy` will also account for more than one instance of `useLDClient`. 
   - For instance if `SESSION_EXPIRING_MODAL` and `RATE_CERT_PROGRAMS` flags are both in a test, but we only configured `ldUseClientSpy({rate-certification-programs: true})` for `RATE_CERT_PROGRAMS` flag. Then `ldUseClientSpy` will use the `defaultValue` for `SESSION_EXPIRING_MODAL` that was passed into `useLDClient` from the component. 
- You can configure multiple flags in `ldUseClientSpy`. 
- The parameter `featureFlag` for `ldUseClientSpy` is typed based on the featureFlags in `flags.ts` it will update the type whenever you add or remove a featureFlag

**NOTE**: Wherever we use `ldUseClientSpy` in a test, we should also `jest.clearAllMocks()`. I have added `afterEach(() => jest.clearAllMocks())` to test files that use `ldUseClientSpy`.


rswerve:
I love the simplicity of this.  A few notes:
1) ldClient.variation doesn't require a default value, so the helper should at least handle that case.  (This could be complicated, given all the possible types that the default can have.  Maybe just throw an error....)
2) It would be nice to use the same keys that we use when we use ldClient itself, so something like `ldUseClientSpy({featureFlags.RATE_CERT_PROGRAMS, true})` instead of `ldUseClientSpy({ 'rate-certification-programs': true })`.
3) we should document how to use this, especially the bit about clearing mocks, in some appropriate doc.  Maybe the app-web README?

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.test.tsx
```diff
@@ -197,7 +199,7 @@ describe('RateDetails', () => {
         })
     })
 
-    it('progressively disclose new rate form fields as expected', async () => {
+    it('progressively disclose new rate form fields as expected u237fiwb18dgbjdfwd', async () => {
```
rswerve, line 202: the password to your crypto vault?

macrael (approved):
I like this, especially the way that we fall back on the in-code default for any unspecified flags. 

Right now, if a test does not call ldUseClientSpy does that mean that the full LD machinery will be activated? I think we want to avoid our tests calling out to LD servers as much as possible. Could we move the call to ldUseClientSpy inside renderWithProviders? Turning flags into configuration for that method would fit with the rest of the dependencies we are configuring there.

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.test.tsx
JasonLin0991: Whoops, this is just a random string.

JasonLin0991:
> 

Correct me if I'm wrong, but currently our `renderWithProviders` does not have a Launch Darkly Provider, so I think in our unit tests shouldn't be connecting to LD servers, instead it should be returning the defaultValue passed into `useLDClient`. 
cc: @rswerve @haworku

As for moving `ldUseClientSpy` inside the `renderWithProviders`, we can and originally it was in it, essentially I was only `jest.spyOn` the `useLDClient` when `renderWithProviders` was passed configuration for flags. I moved it out because my concern was how to handle clearing mocks after each test. If it was included inside `renderWithProviders` the user may never know that they needed to clear mocks after tests when passing a flag configuration. With this function we could add comments around the function about clearing mocks.

The best solution for this is to somehow make a provider for LD in `renderWithProviders` that can handle all this without having to spy on methods. I suspect making a provider would take much longer than a 2 point story and we needed this to get past the hurdle of failing tests due to feature flags. We could split that work out to a separate story and throw it in the tech backlog though. What do you all think?

JasonLin0991:
> I love the simplicity of this. A few notes:
> 
> 1. ldClient.variation doesn't require a default value, so the helper should at least handle that case.  (This could be complicated, given all the possible types that the default can have.  Maybe just throw an error....)
> 2. It would be nice to use the same keys that we use when we use ldClient itself, so something like `ldUseClientSpy({featureFlags.RATE_CERT_PROGRAMS, true})` instead of `ldUseClientSpy({ 'rate-certification-programs': true })`.
> 3. we should document how to use this, especially the bit about clearing mocks, in some appropriate doc.  Maybe the app-web README?

1. Done, throwing an error in the test, we should always supply a default value when using `useLDClient`?
2. This one I could use some help on if we wanted this. Right now I'm restricting the `key` to the actual string value thats passed into `useLDClient` in order to match the value. Maybe we can pair on this.
3. For now I placed a comment above the function, I think eventually we will want to make a LDProvider for `renderWithProviders` and not have to use `jest.spyOn`.

JasonLin0991:
> LGTM. I would love to see us merge this work as is for incremental improvement. I think this addresses the immediate pain with LD and mocking for tests. We can always continue to develop and grow this approach and we have more feature flag interactions in the application.

I agree. I'll merge this in for now. 

> It would be nice to use the same keys that we use when we use ldClient itself, so something like ldUseClientSpy({featureFlags.RATE_CERT_PROGRAMS, true}) instead of ldUseClientSpy({ 'rate-certification-programs': true }).

@rswerve We can still pair on this if you wanted and create a separate PR for it!

PR #1126, opened 2022-08-16
Reinstate CalendarDate and add migration system

macrael:
## Summary

This is a big one, but half of it isn’t mine. 

OK, hand off:

* this is working except for running the migration in CI
* See the Readme in app-api/protoMigrations for how testing/buildng/adding new migrations goes
* I've tried a couple things to get the migrations running in CI but I think @mojotalantikite will need to take it from there. We need to be able to run the code in migrate_protos.ts, either from the command line or directly from code, and that will need to have the migrations that are in the heathPlanFormDataMigrations directory available in the lambda to be imported as well. 

#### Related issues
https://github.com/CMSgov/managed-care-review/pull/1069
https://github.com/CMSgov/managed-care-review/pull/1075


haworku:
🌮  thanks! 

When this PR gets fleshed out more please put some thought also to the QA plan. I would love to have a concrete steps defined for how another team member can verify this works both in code review but also after the merge/deploy to prod. It might be we need screenshots of some current submissions in prod to validate this worked.

services/app-api/protoMigrations/migrate_protos.ts
```diff
@@ -0,0 +1,247 @@
...
+                await prismaClient.protoMigrationsTable.findMany()
+            const migrations = ranMigrationsTable.map((m) => m.migrationName)
+
+            return migrations
+        },
+
+        async runMigrations(migrations: MigrationType[]) {
+            const revs = await prismaClient.healthPlanRevisionTable.findMany()
+
+            console.log(revs)
```
rswerve, line 54: I assume these logs will be removed...

rswerve:
@macrael Looks like I remember 🙂.  I am seeing a type error on line 435 of ApolloHelpers that's keeping the app from building.  Looks like a draft vs submission issue.

rswerve (approved):
Looks great.  🤞

mojotalantikite:
I'm closing this one since we decided not to use this CalendarDate implementation. The migrator does fix the issue though, as confirmed in testing from #1233.

PR #1127, opened 2022-08-16
MR-2539: Move getting some emails to email parameter store

JasonLin0991:
## Summary
[MR-2539](https://qmacbis.atlassian.net/browse/MR-2539)

This will change `cmsReviewSharedEmails` and `ratesReviewSharedEmails` that was set using ENV variables to be pulled directly from parameter store. This is exactly how we are pulling state analyst emails. 

Also removed `SES_REVIEW_TEAM_EMAIL_ADDRESSES` and `SES_RATES_EMAIL_ADDRESSES` from `app-api/serverless.yml` and `.envrc`

### DIFF
--- a/services/app-api/serverless.yml
+++ b/services/app-api/serverless.yml
@@ -128,7 +128,6 @@ provider:
     SES_REVIEW_HELP_EMAIL_ADDRESS: ${ssm:/configuration/email/reviewHelpAddress}
     SES_RATE_HELP_EMAIL_ADDRESS: ${ssm:/configuration/email/rateHelpAddress}
     SES_DEV_TEAM_HELP_EMAIL_ADDRESS: ${ssm:/configuration/email/devTeamHelpAddress}
-    SES_RATES_EMAIL_ADDRESSES: ${ssm:/configuration/email/ratesAddresses}
     APPLICATION_ENDPOINT: ${self:custom.applicationEndpoint, 'http://localhost:3000'}
     AWS_LAMBDA_EXEC_WRAPPER: /opt/otel-handler
     OPENTELEMETRY_COLLECTOR_CONFIG_FILE: /var/task/collector.yml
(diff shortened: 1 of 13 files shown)


services/app-api/serverless.yml
```diff
@@ -58,9 +58,6 @@ custom:
       '/configuration/default/vpc/subnets/private/c/id': 'offline'
       '/configuration/iam/full_permissions_boundary_policy': 'arn:aws:iam::local:policy/local/developer-boundary-policy'
       '/configuration/email/sourceAddress': ${env:SES_SOURCE_EMAIL_ADDRESS, 'offline'}
-      '/configuration/email/reviewTeamAddresses': ${env:SES_REVIEW_TEAM_EMAIL_ADDRESSES, 'offline'}
-      '/configuration/email/ratesAddresses': ${env:SES_RATES_EMAIL_ADDRESSES, 'offline'}
-      '/configuration/email/reviewHelpAddress': ${env:SES_REVIEW_HELP_EMAIL_ADDRESS, 'offline'}
       '/configuration/email/rateHelpAddress': ${env:SES_RATE_HELP_EMAIL_ADDRESS, 'offline'}
```
mojotalantikite, line 61: Ok, so locally these are being set now in the `getRatesReviewSharedEmailsLocal` func, right? I like that.

JasonLin0991: That is correct.

services/app-api/src/parameterStore/emailParameterStore/ratesReviewSharedEmails/getRatesReviewSharedEmails.ts
```diff
@@ -0,0 +1,23 @@
...
+    } else {
+        //Split string into array using ',' separator and trim each array item.
+        return ratesAddresses.split(',').map((email) => email.trim())
+    }
+}
+
+export const getRatesReviewSharedEmailsLocal = async (): Promise<
+    string[] | Error
+> => [
+    `Rate Submission Reviewer 1" <rate.reviewer.1@example.com>`,
```
mojotalantikite, line 20: Same question here, do we need a leading "?

mojotalantikite (approved):
LGTM, thanks for getting this together so quickly!

haworku (approved):
Looks good but make sure we test emails on the review app before merging with a list of values in parameter store dev env.

JasonLin0991:
> Looks good but make sure we test emails on the review app before merging with a list of values in parameter store dev env.

I tested out the review app, looks like its working now. I had to update the parameter store in dev env, to a list of emails, in order test the changes in `serverless.yml`. I will wait to update VAL until this is merged in thouhg.

PR #1134, opened 2022-08-18
MR-2532: State user cannot submit unlocked contract only package with rate data

JasonLin0991:
## Summary
[MR-2532](https://qmacbis.atlassian.net/browse/MR-2532)

In `submitHealthPlanPackage.ts` resolver, before submission, a check is done for `submissionType === CONTRACT_ONLY` and checks package for any valid rate related data using `hasAnyValidRateData` function. If both are true then `removeRatesData` function is ran to remove rate related data, including documents, before submission.

The `removeRateRelatedDocuments` function will remove any documents with `documentCategories` that are type `RATE` or `RATE_RELATED` only from the **package**. If the document `documentCategories` contains `CONTRACT` or `CONTRACT_ONLY` in addition to `RATE` or `RATE_RELATED`, then it will remove `RATE` and `RATE_RELATED` from `documentCategories` and keep `CONTRACT` or `CONTRACT_ONLY`. This ensures rate and rate supporting documents only will be removed from the package and supporting documents that are both contract and rate related will be converted to contract related only.

Removing rate related data is done on submission to preserve rate related data and documents incase the user needs to revert the submission type back to `CONTRACT_AND_RATES` before submission/resubmission. 

#### Test cases covered

**Added**
- `submitHealthPlanPackage.test.ts`
   - `'removes any rate data from CONTRACT_ONLY submissionType and submits successfully'`
   - `'does not remove any rate data from CONTRACT_AND_RATES submissionType and submits successfully'`

- `healthPlanFormData.test.ts`
   - `'Submit CONTRACT_ONLY with valid rate data: $testDescription'`

**Removed**
- `submitHealthPlanPackage.test.ts`
   - `'returns an error if there are invalid rate details fields for submission type'`
      - This was testing for valid rate data for `CONTRACT_ONLY` submission types, which is outdated, since rate data will be removed from `CONTRACT_ONLY` submission.

<strike>**Cypress**
Enabled code for rate program selection in `fillOutNewRateCertification` and `fillOutAmendmentToPriorRateCertification` commands in `stateSubmissionFormCommands.ts`. Now that `rate-program-certification` flag is enabled for all environments.</strike>

## QA guidance

- Please check this with a combination of documents and supporting documents.

### DIFF
--- a/services/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.test.ts
+++ b/services/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.test.ts
@@ -547,6 +547,7 @@ describe('submission type assertions', () => {
     const contractOnlyWithValidRateData: {
         submission: UnlockedHealthPlanFormDataType
         testDescription: string
+        expectedResult: Partial<UnlockedHealthPlanFormDataType>
     }[] = [
         {
             submission: {
@@ -558,6 +559,14 @@ describe('submission type assertions', () => {
                         documentCategories: [
                             'CONTRACT_RELATED' as const,
                             'RATES_RELATED' as const,
+                        ],
+                    },
+                    {
+                        name: 'contract_supporting_that_applies_to_a_rate_also_2.pdf',
+                        s3URL: 'fakeS3URL',
+                        documentCategories: [
+                            'RATES_RELATED' as const,
+                            'CONTRACT_RELATED' as const,
                         ],
                     },
                     {
@@ -585,6 +594,29 @@ describe('submission type assertions', () => {
                 ],
             },
             testDescription: 'With all valid rate data ',
+            expectedResult: {
+                rateType: undefined,
+                rateDateCertified: undefined,
+                rateDateStart: undefined,
+                rateDateEnd: undefined,
+                rateCapitationType: undefined,
+                rateAmendmentInfo: undefined,
+                rateProgramIDs: [],
+                actuaryContacts: [],
+                documents: [
+                    {
+                        name: 'contract_supporting_that_applies_to_a_rate_also.pdf',
+                        s3URL: 'fakeS3URL',
+                        documentCategories: ['CONTRACT_RELATED' as const],
+                    },
+                    {
+                        name: 'contract_supporting_that_applies_to_a_rate_also_2.pdf',
+                        s3URL: 'fakeS3URL',
+                        documentCategories: ['CONTRACT_RELATED' as const],
+                    },
+                ],
+                rateDocuments: [],
+            },
         },
         {
             submission: {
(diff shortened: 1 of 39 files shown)


services/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.ts
```diff
@@ -228,16 +251,65 @@ const generateRateName = (
     return rateName
 }
 
+const removeRateRelatedDocuments = (
+    documents: SubmissionDocument[]
+): SubmissionDocument[] => {
+    const noRateDocs: SubmissionDocument[] = []
+    documents.forEach((document) => {
+        const ratesRelatedDocIndex =
+            document.documentCategories.indexOf('RATES_RELATED') === -1
```
rswerve, line 260: Is converting to undefined serving any purpose here?  I think you could just get the index and check below if it's -1.

JasonLin0991: Good catch! It was until, I found a bug with the code. Now it's unnecessary left over code.

JasonLin0991:
@rswerve @haworku Now that I think of it, maybe I should disable rate program cypress commands, in `fillOutNewRateCertification` and `fillOutAmendmentToPriorRateCertification`, again and re-enable after this ticket [MR-2525](https://qmacbis.atlassian.net/browse/MR-2525)

services/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.ts
```diff
@@ -228,16 +251,60 @@ const generateRateName = (
     return rateName
 }
 
+const removeRateRelatedDocuments = (
```
haworku, line 254: I'm seeing some business logic in here - please add a unit test for this function.

services/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.test.ts
```diff
@@ -541,4 +543,166 @@ describe('submission type assertions', () => {
...
+                rateAmendmentInfo: undefined,
+                rateProgramIDs: [],
+                actuaryContacts: [],
+                documents: [],
+                rateDocuments: [],
+            },
+        },
+    ]
+    test.each(contractOnlyWithValidRateData)(
+        'Submit CONTRACT_ONLY with valid rate data: $testDescription',
```
haworku, line 701: Can we also test that CONTRACT_AND_RATES packages don't lose any data when removeRatesData is called?

JasonLin0991: I added `'does not remove any rate data from CONTRACT_AND_RATES submissionType and submits successfully'` in `submitHealthPlanPackage.test.ts`

haworku (approved):

PR #1148, opened 2022-08-25
MR-2540: Finish cleaning up email parameter store work

JasonLin0991:
## Summary
[MR-2540](https://qmacbis.atlassian.net/browse/MR-2540)

- Finish retrieving all emails directly from parameter store rather than weaving through serverless. 
- Updated `README` in `app-api` with documentation on parameter store, how we are using it and how to add new parameter stores.
- Refactor `generateCMSReviewerEmails` in `app-api/src/emailer/templateHelpers.ts` and add more comments. 
   - Hopefully this will clarify how we are generating the reviewer emails for each submission type.
   - Added handling for submission types other than `CONTRACT_ONLY` and `CONTRACT_AND_RATES` in preparation of more submission types. For now it will return an error `generateCMSReviewerEmails does not currently support submission type: ${pkg.submissionType}.`. This should hopefully force us to be more specific on generating reviewer emails for each new submission type.
   - Updated parameter store emails display name to better reflect actual email addresses.
      - For example `Reviewer One" <mc-review-qa+reviewer1@truss.works>` is now `"MCOGDMCOActions" <mc-review-qa+MCOGDMCOActions@truss.works>`.
      - This should help us better identify we have the emails correctly mapped to the [Figma](https://www.figma.com/file/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?node-id=105%3A7393) in testing.
      - Currently this change is only in `Dev` environment, I will change `Impl` if this is approved.

**Notes**
- Pulling emails when sending emails not in configuration.
   - @haworku Brought up a very good point with pulling emails from parameter stores directly at dependency configuration in `apollo_gql.ts`. Which was that changing the value of the email parameter stores, other than state analysts, will not update until dependency configuration is done again. 
   - This would be an issue of we had to change the email in parameter store while a user in pilot was currently working in the app, their submission will not have the updated emails.
   - This was also the case when we stored the emails in github secrets, so not something new from our change to parameter store.
   -  At first I agreed that we should treat these email stores like the state analyst email stores, which is being pulled at the resolvers, but after taking a look the refactor work needed I thought otherwise. 
   - It would be a pretty large refactor (emailer functions, configuration, resolvers and all tests) for a temporary solution because we are moving to storing emails in our database. I think it would be better to refactor this when we switch to pulling emails from the database.
   - @haworku @rswerve @mojotalantikite Let me know what you think about moving this refactor to when we move to storing emails to the database.

#### Test cases covered

- `emailParameterStore.test.ts`
   - Every parameter store function has these 3 tests
      - `'returns [type of emails] email as [string or array of strings]'`
      - `'returns error when fetching store value fails'`
      - `'returns error when Type of parameter store value is incompatible'`

- `templateHelpers.test.ts`
   - `'Generate CMS Reviewer email: $testDescription'`
      - `'Contract only submission'`
      - `'Contract and rates submission'`
      - `'Submission with CHIP program specified for contract certification'`
      - `'Submission with CHIP program specified for rate certification'`
      - `'Puerto Rico submission'`
      - `'Error result.'`


mojotalantikite:
Cool, LGTM, thanks for doing this! 

In terms of this happening at dependency configuration time in `apollo_gql.ts`, while that's true, because this is a lambda environment that dependency configuration happens often. Possibly even on every request, due to how low our traffic is. You can't say for sure that it'll happen on every request, because we don't have control of how lambdas are invoked and scheduled (AWS handles that), but those values are being refreshed all the time.

So that's all to say I do agree we should eventually refactor this to lookup the email addresses right when we do the sending, but it's probably not a big deal to wait on that and land this the way that it is. The only thing I'll say is we should keep our eye on application latency after this lands, because we'll be making requests to param store on possibly every lambda invocation (even those unrelated to emailing) to set these values.

mojotalantikite (approved):

haworku:
Thanks @mojotalantikite that is clarifying.

services/app-api/README.md
```diff
@@ -39,12 +39,45 @@ Sets up authentication based on environment variables. Options are local or Amaz
...
+## Adding new a new email parameter stores
+- This requires you to have AWS Console access, which is managed via Active Directory and Cloudtamer.
+- After accessing AWS Console navigate to `System Manager > Parameter Store`.
+- Locate the `Create parameter` button and press to begin adding a new store.
+- Create a new parameter store **Name** following existing patterns for store names.
+    - For configuration emails `/configuration/email/[name]`
+    - For new state analyst emails `/configuration/[state abbreviation]/stateanalysts/email`
+- Choose a `Type` for the store, currently all email parameter stores are either `String` or `StringList`
+- Input values for the parameter store.
+    - **It's highly recommended to input values in `Managed Care Dev` and `Managed Care Impl` that imitate the `Type` of the parameter store. e.g. parameter store `Type` of `StringList` should have the value `string1@example.com,string1@example.com,string1@example.com`**
```
haworku, line 76: Thanks for adding docs on this part as well

services/app-api/src/emailer/templateHelpers.ts
```diff
@@ -78,36 +78,63 @@ const generateCMSReviewerEmails = (
...
-        pkg.submissionType === 'CONTRACT_AND_RATES' &&
-        pkg.stateCode !== 'PR' &&
-        !includesChipPrograms(programIDs)
-    ) {
-        return pruneDuplicateEmails(contractAndRateReviewerEmails)
-        //
-    } else if (includesChipPrograms(programIDs)) {
-        return pruneDuplicateEmails(chipReviewerEmails)
-    }
+    // dmcpEmail is added in both CONTRACT_ONLY and CONTRACT_AND_RATES, in case it wasn't included in cmsReviewSharedEmails.
```
haworku, line 101: I appreciate the named variables, it helps a lot. I would also like to see a set of table tests  for `templateHelpers.ts` specifically the `generateCMSReviewerEmails` function. I know we should be also testing this through the templates ...  but whenever I see a utility function manipulating an array in several ways I see an opportunity to also quickly unit test and hit each logic branch (e.g. submission type with chip programs, submission type without chip, etc).

haworku (approved):
I don't see anything blocking. Appreciate this work!

services/app-api/src/emailer/templateHelpers.ts
JasonLin0991: Added tests for `generateCMSReviewerEmails` in a new `templateHelpers.test.ts`. Tests include...
- `'Contract only submission'`
- `'Contract and rates submission'`
- `'Submission with CHIP program specified for contract certification'`
- `'Submission with CHIP program specified for rate certification'`
- `'Puerto Rico submission'`
- `'Error result.'`

PR #1170, opened 2022-09-01
MR-2320: Enable CSV export

rswerve:
## Summary

Product would like a way to gain a window into the database in production, so this ticket creates an endpoint, accessible only to CMS users, which returns a CSV file of all the revisions in the prod db.

#### Test cases covered

Some simple tests around the Reports component.

## QA guidance

I don't think we have a CMS user for preview deployments, correct?

@mojotalantikite Access to the Reports page is restricted by our routing, but I assume that anyone can hit the reports endpoint and get a response back.  Is that correct, and can we do anything about it?


mojotalantikite:
> @mojotalantikite Access to the Reports page is restricted by our routing, but I assume that anyone can hit the reports endpoint and get a response back. Is that correct, and can we do anything about it?

So yeah, that's correct, as of now anyone can hit that endpoint and get the data. We can fix that by adding an authorizer to the handler config, which will attach it to the API gateway for us. I'll leave a comment at the spot in this PR that you could do that at right now.

rswerve:
> > @mojotalantikite Access to the Reports page is restricted by our routing, but I assume that anyone can hit the reports endpoint and get a response back. Is that correct, and can we do anything about it?
> 
> So yeah, that's correct, as of now anyone can hit that endpoint and get the data. We can fix that by adding an authorizer to the handler config, which will attach it to the API gateway for us. I'll leave a comment at the spot in this PR that you could do that at right now.

Thank you.  That will limit access to logged in users, and there's no way at this level to discriminate between CMS and state users, right?

mojotalantikite (approved):

mojotalantikite:
> > > @mojotalantikite Access to the Reports page is restricted by our routing, but I assume that anyone can hit the reports endpoint and get a response back. Is that correct, and can we do anything about it?
> > 
> > 
> > So yeah, that's correct, as of now anyone can hit that endpoint and get the data. We can fix that by adding an authorizer to the handler config, which will attach it to the API gateway for us. I'll leave a comment at the spot in this PR that you could do that at right now.
> 
> Thank you. That will limit access to logged in users, and there's no way at this level to discriminate between CMS and state users, right?

At this point no, there's no way at the API gateway level to distinguish between the two. If we wanted to do that, we'd have to make a call to Cognito to list the users and then filter to see what the user's role is. There are a couple functions in the domain-model called `isStateUser` and `isCMSUser` that we use in other places to verify what type of user we're dealing with. I'd guess we'd have to do something similar in this handler.

services/app-api/src/handlers/reports.ts
```diff
@@ -0,0 +1,116 @@
...
+            transforms.flatten({
+                objects: true,
+                arrays: false,
+                separator: ',',
+            }),
+        ],
+    })
+    const csv = await parser.parse(bucket)
+
+    return {
```
haworku, line 105: suggestion: consider returning error code for the errors or capture the errors and console.log them in the handler (those logs go to cloudwatch). 

I remember doing that for the email submit lambda because otherwise I was noticing the fail reason was difficult to find if you just throw. I could be off though ... maybe you had no issue with this.

haworku:
> I don't think we have a CMS user for preview deployments, correct?
fy8 we do have a CMS user for review app deployments its Review App CMS User in 1Password

services/app-api/src/handlers/reports.ts
haworku: oops just ran through this on deploy app and getting uncaught error when clicking link. can't see specific message but its in my dev tools console

rswerve: I don't have the review app cms user in my password vault, so I'll take a look when we can get that sorted out...

services/app-api/serverless.yml
```diff
@@ -144,6 +144,15 @@ functions:
...
 
+  reports:
+    handler: src/handlers/reports.main
+    events:
+      - http:
+          path: reports
+          method: get
+          cors: true
+          authorizer: aws_iam
+
```
mojotalantikite, line 155: Ah, you're gonna want to attach the Prisma layer for querying the DB -- also the OTEL layer for us to get telemetry data:
```
layers:
      - !Ref PrismaClientEngineLambdaLayer
      - arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-nodejs-amd64-ver-1-2-0:1
```

Check out line 234 in the graphql handler for an example.

rswerve:
Closing in favor of https://github.com/CMSgov/managed-care-review/pull/1183

PR #1175, opened 2022-09-02
Add protobuf data migrator to CI

mojotalantikite:
## Summary

This refactors the protocol buffers migration tool and adds it to run in CI via our existing migrations lambda handler. 

Some major changes to be aware of:
-  I've moved all the migrator code to the `app-proto` service. During code generation I copy generated code into the necessary places.
-  I've pulled out the functions from the proto migrator CLI into a lib that we can import and call from other code. I call it directly from the lambda handler, rather than building it as a CLI and shell-ing out to it as it was originally written.
- I've added the data migration files to the migration lambda layer. Since we're actually using Prisma directly now from the migration handler, I had to also add the Prisma query engine to the migration layer.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2553
https://qmacbis.atlassian.net/browse/MR-2571


haworku (approved):
Looks good. Reminder, we spoke IRL about
- adding a step to take the db snapshot before starting migration
- adding additional comments to app-proto README about whats going on in the proto scripts and when these are called 


Something else I noticed- there is a README already in `app-proto/protoMigrations` can we just check that text is still correct and link from main proto README? I think those docs could get lost.

rswerve (approved):

PR #1183, opened 2022-09-08
MR-2320: Enable CSV Export

rswerve:
## Summary

Product would like a way to gain a window into the database in production, so this ticket creates an endpoint, accessible only to CMS users, which returns a CSV file of all the revisions in the prod db.

#### Test cases covered

Some simple tests around the Reports component.  

## QA guidance

Should be able to log into the review app as a cms user (now in 1password! 🙂 ), go to /reports and download a csv.


haworku (approved):
LGTM. Will leave a comment on slack about another field we may want to include but let's see what product says.

PR #1186, opened 2022-09-09
MR-2525: Implement launch darkly in cypress testing

JasonLin0991:
## Summary
**[MR-2525](https://qmacbis.atlassian.net/browse/MR-2525)**

In draft mode, testing implantation out of local deployment. Would like some feed back before starting on docs.
- Refactor `featureFlag` object in `app-web/src/common-code/featureFlags/flags.ts` and update everywhere that uses it.
   - Added default value to the object, this way it's centrally located. This also allows us generate default values for cypress tests.
- New LD commands added in `tests/cypress/support/launchDarklyCommands.ts`
   - **_`interceptFeatureFlags`_**: This command allows you to intercept LD calls and set flag values. It also generates a `featureFlagStore.json` file in `tests/cypress/fixtures/stores` with default feature flag values along with any specific values passed in. 
   - **_`featureFlagStore.json`_**: This file is used to store our flag values so later in the test we can refer back to it. This is currently used in `fillOutNewRateCertification` and `fillOutAmendmentToPriorRateCertification` commands to toggle code specifically for feature flags on.
   - **_`stubFeatureFlags`_**: This command intercepts all of LD api calls and returns our own response. This will remove the need for waiting on actual API calls. This command is being used on all specs in `beforeEach` to intercept these calls on all tests and it will setup default feature flag values in `featureFlagStore.json`.
   - **_`getFeatureFlagStore`_**: This command reads the `featureFlagStore.json` and returns an object of the feature flag values. You can pass in an array of specific feature flags to return those values instead of the entire set of flags. It's important this command is only called after `stubFeatureFlags` or `interceptFeatureFlags` is called or risk not finding the `featureFlagStore.json` file. 
   - All these commands are Typed using the `featureFlag` object and types in `app-web/src/common-code/featureFlags/flags.ts`. Like the unit tests this will restrict flags to existing ones and help with autocompletion.
- Launch Darkly implementation into Cypress documentation
   - Located in `docs/technical-design/launchDarklyTestingApproach.md`.
   - Updated `017-feature-flag-lifecycles`.
   - Added a section for Unit Tests. Not completed, but placed here so we do not forget and I have created a [ticket MR-2592](https://qmacbis.atlassian.net/browse/MR-2592) in the backlog tech maintenance to complete it.

#### Test cases covered

Cypress
- `rateDetails.spec.ts`
   - `'can get fill out rate details with rate-certification-program feature flag on'`


haworku (approved):
This looks good. What I like about this is when we remove flags because they have been fully turned on its just a matter of removing that flag (which will trigger the type warnings for where its used), removing the conditionals, and leaving the code inside (since that is now the accepted behavior for the application overall).

Once we do a round of feature flag deprecation for rate-certification-programs I think this would be excellent work to present to frontend end guild or elsewhere.

tests/cypress/fixtures/stores/featureFlagStore.json
```diff
@@ -0,0 +1 @@
+{"enable-health-endpoint":true,"session-expiring-modal":true,"session-expiration-minutes":30,"modal-countdown-duration":2,"app-api-graphql-errors":false,"rate-certification-programs":true}
```
haworku, line 1: Reminder to remove this fully from the PR.

tests/cypress/support/launchDarklyCommands.ts
```diff
@@ -0,0 +1,76 @@
...
+            })
+    ).as("LDClientStream");
+
+    // Intercept feature flag calls and generates default feature flag values in store.
+    cy.interceptFeatureFlags()
+})
+
+//Command to get feature flag values from the featureFlagStore.json file.
+Cypress.Commands.add('getFeatureFlagStore', (featureFlags?: FeatureFlagTypes[]) => {
+    cy.readFile('tests/cypress/fixtures/stores/featureFlagStore.json').then((store: Record<FeatureFlagTypes, FlagValueTypes>): Partial<Record<FeatureFlagTypes, FlagValueTypes>> => {
```
haworku, line 66: Suggestion: add a catch and a clear console.error with what went wrong and how to address so its easier to debug what happened if that file doesn't get written. as you point out in your notes - there is an dependency between the functions

haworku (approved):

tests/cypress/support/launchDarklyCommands.ts
JasonLin0991: Looks like `catch` is not valid in cypress commands, [docs](https://docs.cypress.io/guides/core-concepts/conditional-testing.html#Error-Recovery). But the command does return an informative error in cypress, partly why I left out the error handling.

haworku: ah got it! Thank you.

JasonLin0991:
@haworku @rswerve @mojotalantikite I just added LD Cypress implementation to our ADR let me know what you all think.

docs/architectural-decision-records/017-feature-flag-lifecycles.md
```diff
@@ -22,14 +22,17 @@ A feature typically can be added behind a flag if:
 
 Flags are first added to Launch Darkly via their UI. As an authenticated Launch Darkly user, navigate to our feature flags and click “create flag”. Choose a human readable flag name and a ‘-’ delimited flag key. The flag key is what we use inside our code. Make sure to mark the flag as client side if it is meant to be user visible (rather than API only). If the flag is permanent, mark this as well.
 
-We keep a list of flags in a constants file in `common-code/flags` and all new flags must be added to this file. For example, if you added a flag key of ‘new-testing-flag’, you’d add it to the file as follows:
+We keep a list of flags in a constants file in `app-web/src/common-code/featureFlags` and all new flags must be added to this file. For example, if you added a flag key of ‘new-testing-flag’ with a default value, you’d add it to the file as follows:
```
haworku, line 25: 👋 Hey! i don't think this is well suited in this  ADR. An architectural decision record is supposed to be a straightforward comparison of different options.  [More about how Truss uses an ADR.](https://github.com/trussworks/Engineering-Playbook/blob/f0cdfd5e91e6ef70fd6a311ff882ad7281a1aeba/docs/documentation/adr.md)

The only time an ADR is updated (usually) is if we make a different decision.

These docs you are adding are implementation details for testing with LD. This goes in the technical design docs section. That section is for more general docs about how we are going to implement something with a less rigid format.  You could either add to the existing doc on testing approach or if this feels odd there --- feel free to start a launch darkly technical design doc. 

🗒️  I will also take an action item to **add the template we use for ADRs to the repository.**   ADR 0018 was a bit odd because it didn't follow the template but you will see other ADRs follow the template properly and we should just make this explicit so its clearer in the future

PR #1189, opened 2022-09-09
Add site maintenance banner and feature flag

haworku:
## Summary
This came up during incident response.  We want to use LD to easily toggle on a site maintenance flag when we have an issue to communicate to users (in this case the API was down for an unexpected reason due to AWS outage).

#### Related issues
n/a 

#### Test cases covered
Added tests for flag on and off state.

### DIFF
--- a/services/app-web/src/common-code/featureFlags/flags.ts
+++ b/services/app-web/src/common-code/featureFlags/flags.ts
@@ -6,6 +6,10 @@ export const featureFlags = {
     /*
      Toggles the /health api endpoint
     */
+    SITE_MAINTENANCE_BANNER: 'site-maintenance-banner',
+    /*
+     Toggles the /health api endpoint
+    */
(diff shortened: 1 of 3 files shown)


JasonLin0991 (approved):
Looks good. I have some suggestions we can talk about at RSG or something, but for now I think we need get this in?

Suggestions

- Not relying on manual feature flag toggling to show the banner. Maybe we do some checks in app-api and/or app-web and display the banner when something doesn't pass.
- Should we disable certain UI when banner is showing? Like `Sign In` button.

haworku:
@JasonLin0991 Feel free to add to RSP agenda. Good topic, good questions. 

For clarity - This is really only a  fallback for  cases where we don't have error handling built out yet, as happened today. Agree we should have errors already built out but its good to have an option for cases that aren't handled yet.

PR #1191, opened 2022-09-12
Sign in error alert - when user logged out due to inactivity 

haworku:
## Summary
User should know when they are logged out of IDM due to inactivity 
#### Related issues
https://qmacbis.atlassian.net/browse/MR-53

#### Screenshots

Screen Shot 2022-09-12 at 3 05 39 PM


services/app-web/src/hooks/useTealium.ts
```diff
@@ -29,8 +29,8 @@ const useTealium = (): {
     // Add Tealium setup
     // this effect should only fire on initial app load
     useEffect(() => {
-        // Do not add tealium for local dev
-        if (process.env.REACT_APP_STAGE_NAME === 'local') {
+        // Do not add tealium for local dev or review apps
+        if (process.env.REACT_APP_AUTH_MODE !== 'IDM') {
```
haworku, line 33: The tealium changes in this PR are due to realizing that on review apps (which have a random domain) Tealium tags will not load properly because it does not recognize the URL. This causes unnecessary API calls and React updates because we keep trying to find tealium and fail.

services/app-web/src/gqlHelpers/apolloErrors.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+const isLikelyUserAuthError = (
+    error: ApolloError,
+    isAuthenticated: boolean
+) => {
+    const { networkError } = error
+    if (networkError?.name === 'ServerError') {
+        const serverError = networkError as ServerError
+        if (serverError.statusCode === 403 && isAuthenticated) {
```
rswerve, line 55: could just return the conditional here

services/app-web/src/pages/Errors/ErrorBoundaryRoot.tsx
```diff
@@ -3,8 +3,12 @@ import { recordJSException, recordSpan } from '../../otelHelpers/tracingHelper'
 import { useEffect } from 'react'
 import { EmptyHeader } from '../../components/Header/EmptyHeader'
 import styles from '../App/AppBody.module.scss'
-// It is possible to have different error boundary fallback components at different levels of the hierarchy.
-// This is the root error boundary. If it is visible, this means that the React frontend has crashed entirely and the entire tree is gone.
+/* The error boundary is a fallback UI to catch errors
+    See React docs: https://reactjs.org/docs/error-boundaries.html
+    It is possible to have different error boundary fallback components at different levels of the hierarchy.
+    This component is the root error boundary. I visible, this means that the React frontend has crashed fully and the entire tree is gone.
```
rswerve, line 9:
```suggestion
    This component is the root error boundary. If visible, this means that the React frontend has crashed fully and the entire tree is gone. 
```

rswerve (approved):
this worked as expected for me locally

PR #1207, opened 2022-09-21
MR-2613: endpoint docs

rswerve:
## Summary

Added a doc to walk someone through creating endpoints, either in GraphQL or AWS.

## QA guidance

Does it make sense?  Too much or not enough info?  Anything I missed or got wrong?

### DIFF
--- a/docs/technical-design/creatingEndpoints.md
+++ b/docs/technical-design/creatingEndpoints.md
@@ -38,6 +38,40 @@ query dataExport {
 
 We're telling GraphQL that we want a query called dataExport, which returns a field called "name".
 
+#### The GraphQL schema
+
+_**We'll also add our query and its return type to `/app-graphql/src/schema-graphql`**_
+
+First, we'll add the type
+
+```graphql
+type DataExport {
+    name: String!
+}
+```
+
+And then, on the `Query` type, we'll add our new method
+
+```graphql
+dataExport: DataExport
+```
+
+#### Generating our types
+
+Once we've told GraphQL what query we want and described that data's return type, we can generate type information that can be shared between the web and api layers. We have a generated file at `/app-web/src/gen/gqlClient.tsx`, which contains graphQL types that we can use in the front end.
+
+_**On the command line, navigate to `/services/app-graphql`.**_  
+_**Run `yarn gqlgen`**_
+
+The generated file isn't checked into source control, but if you look inside, you'll now find a new type that you can use on the front-end if necessary.
+
+```typescript
+export type DataExport = {
+    __typename?: 'DataExport'
+    name: Scalars['String']
+}
+```
+
 #### The domain model
 
 We use Typescript in our app, so all the data flowing from GraphQL through the api layer to the front end must also have a type. We keep the types that are shared between the api layer and the front end in our domain models.
@@ -149,14 +183,11 @@ And also add it to the Resolvers object. We've made a query, so this line will b
 dataExport: dataExportResolver(store),
 ```
 
-### Generating code
+### Generating prisma code
 
-Once we have our backend code in place, we need to run a couple of commands to generate glue types and methods.
+Once we have our backend code in place, if we've modified the prisma schema (we did not, in this example), we need to run another generate command.
 
-_**On the command line, navigate to `/services/app-graphql`.**_  
-_**Run `yarn gqlgen`**_
-
-_**Again on the command line, navigate to `/services/app-api`**_  
+_**On the command line, navigate to `/services/app-api`**_  
 _**Run `yarn prisma generate`**_
 
 ### Getting data on the front end
--- a/docs/technical-design/creatingEndpoints.md
+++ b/docs/technical-design/creatingEndpoints.md
@@ -4,7 +4,7 @@ There are two ways that we create endpoints in our app. One uses GraphQL, the ot
 
 **When we want to fetch or update information relevant to the app itself--think of submissions, or user info--we'll typically use GraphQL.**
 
-**When we want to communicate with a service like New Relic, download a file, or, for example, to set up a health-check endpoint, we'll use a handler.** 
+**When we want to communicate with a service like New Relic, download a file, or, for example, to set up a health-check endpoint, we'll use a handler.**
 
 The relative merits of each approach should become clear below.
 
@@ -202,9 +202,9 @@ We also provide a `path`. This handler will run when someone navigates to the si
 
 The `authorizer` line tells AWS that this method can only be invoked by a user who is logged in and has permissions set in AWS IAM. This will typically flow from someone being a state or CMS user, with permission to interact with the main site.
 
-`layers` tie this handler to services that it needs. In this case, it needs database access, so there's a Prisma layer, and it's also being tied to `otel`, our open telemetry service, for observability.
+`layers` tie this handler to services that it needs. A lambda layer is basically a collection of code that stays in AWS and doesn't need to be uploaded or updated with each deployment. It's an efficient way to make dependencies available without having to upload and build them each time. In the case of our example, we need database access, so we include our Prisma layer, and we're also tying in to `otel`, our open telemetry service, for observability.
 
-In order to communicate with those layers, we have to specify which `vpc`, or virtual private cloud, this handler will run within.
+In order to communicate with those layers, we have to specify which `vpc`, or virtual private cloud, this handler will run within. AWS services are basically islands, and vpcs are a way to enable secure communication between some of the islands under our control. By saying that our handler belongs to a particular vpc, we are telling AWS which services it can communicate with.
 
 API Gateway handlers have a hard timeout of 30 seconds, which we cannot modify or exceed. If your function takes longer than that to run, you'll need to put some of the functionality in non-ApiGateway Lambdas and consume the output in your endpoint/handler.
 
--- a/docs/technical-design/endpoints.md
+++ b/docs/technical-design/endpoints.md
@@ -1,6 +1,6 @@
 # Endpoints
 
-There are two ways that we create endpoints in our app. One uses GraphQL, the other creates handlers for AWS Lambda.
+There are two ways that we create endpoints in our app. One uses GraphQL, the other creates AWS Lambda handlers.
 
 When we want to fetch or update information relevant to the app itself--think of submissions, or user info--we'll typically use GraphQL.
 
@@ -38,9 +38,9 @@ We're telling GraphQL that we want a query called dataExport, which returns a fi
 
 #### The domain model
 
-We use Typescript in our app, so all the data flowing from GraphQL through the api layer, to the front end, must also have a type. We keep the types that our shared between the api layer and the front end in our domain models.
+We use Typescript in our app, so all the data flowing from GraphQL through the api layer to the front end must also have a type. We keep the types that are shared between the api layer and the front end in our domain models.
 
-_**Create a new domain model in `/app-api/src/domain-models/DataExport.ts`**_ Let's call it DataExport.ts and put the following type in it.
+_**Create a new domain model in `/app-api/src/domain-models/DataExport.ts`**_ Put the following type in it.
 
 ```typescript
 export type DataExportType = {
@@ -77,6 +77,8 @@ export async function dataExport(
 
 Recall the `schema.prisma` file. We're saying, in the `State` model (`client.state`, in this snippet), get the first record. From that result, give us the `name` field, and we'll assign it to a key of `name` in the object that we return. (Note that our object key can be whatever we want (as long as it matches the types we just made). It doesn't have to match the property name from the result, but in many cases that will make the most sense.)
 
+Also note that here we need the DataExportType that we just created in the domain model.
+
 #### Adding our method to the store
 
 Once we've defined this Prisma method, we can add it to our store, which will in turn make it available to our resolvers.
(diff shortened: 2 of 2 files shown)


docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
...
+        subnetIds: ${self:custom.privateSubnets}
+```
+
+The code that runs when this handler is called is in `/app-api/src/handlers/reports.ts`.
+
+The handler points to `/app-api/src/handlers/reports.main`. We're telling the handler where to look for the file that contains the code that will run in the lambda (`src/handlers`) and also the name of the function (`main`). In `reports.ts` the exported handler is called `main`. The location and exported method of `reports.ts` have to match what's specified on the `handler:` line in our YAML file.
+
+We also provide a `path`. This handler will run when someone navigates to the site's base name, slash reports. For prod, it would be `https://mc-review.onemac.cms.gov/reports`.
+
+The `authorizer` line tells AWS that this method can only be invoked by a user who is logged in and has permissions set in AWS IAM. This will typically flow from someone being a state or CMS user, with permission to interact with the main site.
```
JasonLin0991, line 197: How or where do we specify which type of user has access to this endpoint? In the reports endpoint I believe it is restricted to CMS users?

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
...
+
+`layers` tie this handler to services that it needs. In this case, it needs database access, so there's a Prisma layer, and it's also being tied to `otel`, our open telemetry service, for observability.
+
+In order to communicate with those layers, we have to specify which `vpc`, or virtual private cloud, this handler will run within.
```
JasonLin0991, line 201: `layers` and `vpc` are still a bit mysterious to me. Is there anyway to flesh this out more? For instance, the need for a `prisma` layer to access the service is straightforward, but where would I go in order to find the specifics for the layers `PrismaClientEngineLambdaLayer` and `arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-nodejs-amd64-ver-1-2-0:1` for `otel`. Same question for `vpc`.

JasonLin0991 (changes requested):
Just a couple of comments to prevent me from pinging @mojotalantikite in the future when I need to add an endpoint.

Otherwise everything else seems to be good

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
+# Endpoints
```
haworku, line 1:
```suggestion
# How to create a new endpoint
```
Can you also change the title of this file to be a bit clearer its a how to not a comprehensive list of available endpoints.

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
...
+
+You can think of this as the mirror-image of the GraphQL query, but for the application code.
+
+### Retrieving from the database
+
+#### The Prisma query
+
+Now let's get the data that we want.
+
+_**Create a new file: `/app-api/src/postgres/dataExport.ts`**_. We use Prisma to query the database.
```
haworku, line 59:
```suggestion
_**Create a new file: `/app-api/src/postgres/dataExport.ts`**_. We use [Prisma to query the database](https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries).
```

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
...
+        name: result?.name,
+    }
+}
+```
+
+Recall the `schema.prisma` file. We're saying, in the `State` model (`client.state`, in this snippet), get the first record. From that result, give us the `name` field, and we'll assign it to a key of `name` in the object that we return. (Note that our object key can be whatever we want (as long as it matches the types we just made). It doesn't have to match the property name from the result, but in many cases that will make the most sense.)
+
+Also note that here we need the DataExportType that we just created in the domain model.
+
+#### Adding our method to the store
```
haworku, line 82:
```suggestion
#### Adding our method to the Store

We handle the database actions through an exported "Store", rather than calling the methods directly.
```
Do you think this kind of change could be helpful? Not sure if it adds clarity or not.

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
...
+
+```typescript
+dataExport: () => dataExport(client),
+```
+
+When we access the store in our app, the dataExport method we defined above will be available to us.
+
+### Creating the resolver
+
+We use Apollo Client to interface with our data source and to provide information from the store to the front-end; the methods to do that are called "resolvers".
```
haworku, line 109:
```suggestion
We use Apollo Client to interface with our data source and to provide information from the store to the front-end; the methods to do that are called ["resolvers"](https://www.apollographql.com/docs/apollo-server/data/resolvers/).  
```

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+### Generating code
+
+Once we have our backend code in place, we need to run a couple of commands to generate glue types and methods.
+
+_**On the command line, navigate to `/services/app-graphql`.**_  
+_**Run `yarn gqlgen`**_
+
+_**Again on the command line, navigate to `/services/app-api`**_
```
haworku, line 155: 🤔 I *think* running `yarn prisma generate` should just happen first?  In the example you give, I don't think there is a db change, so you could prisma generate right away.  I mention this because once a developer runs the command they get the benefit of those new types while building the api part, which could be helpful to them.

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
...
+
+_**On the command line, navigate to `/services/app-graphql`.**_  
+_**Run `yarn gqlgen`**_
+
+_**Again on the command line, navigate to `/services/app-api`**_  
+_**Run `yarn prisma generate`**_
+
+### Getting data on the front end
+
+One result of the `generate` commands is to create a hook that we can use on the front end to call our new resolver. The name of the hook will be of the form _useWhatYourResolverIsCalledQuery_ (it will end with "Mutation" if that's what you've created). You can pull it in and get data.
```
haworku, line 160:
```suggestion
One result of the `yarn gqlgen` command is a typed [`useQuery` React hook](https://www.apollographql.com/docs/react/data/queries/) that we can use on the front end to call our new resolver. The name of the hook will be of the form _useWhatYourResolverIsCalledQuery_ (it will end with "Mutation" if that's what you've created). You can pull it in and get data.
```

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
...
+
+```typescript
+import { useDataExportQuery } from '../../gen/gqlClient'
+const { data, loading, error } = useDataExportQuery()
+```
+
+## Lambda
```
haworku, line 167:
```suggestion
## Lambda

We use AWS Lambda to build RESTful endpoints in our application. When we create a lambda handler, serverless will add a new Lambda function for us in AWS. If we configure it correctly, that function will have access to things like our postgres database, which itself is hosted on AWS.
```

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Serverless configuration
+
+When we create a handler, we're telling serverless to create a new Lambda function for us in AWS. If we configure it correctly, that function will have access to things like our postgres database, which itself is hosted on AWS. Here's one existing handler configuration, found in `/services/app-api/serverless.yml`.
```
haworku, line 171:
```suggestion
Here is an example of serverless configuration for a Lambda that handles an api request, found in `/services/app-api/serverless.yml`.
```

Deleted sections were just moved up. Was trying to match what you did in the GraphQL section - you had an introduction paragraph first.

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
...
+
+We also provide a `path`. This handler will run when someone navigates to the site's base name, slash reports. For prod, it would be `https://mc-review.onemac.cms.gov/reports`.
+
+The `authorizer` line tells AWS that this method can only be invoked by a user who is logged in and has permissions set in AWS IAM. This will typically flow from someone being a state or CMS user, with permission to interact with the main site.
+
+`layers` tie this handler to services that it needs. In this case, it needs database access, so there's a Prisma layer, and it's also being tied to `otel`, our open telemetry service, for observability.
+
+In order to communicate with those layers, we have to specify which `vpc`, or virtual private cloud, this handler will run within.
+
+API Gateway handlers have a hard timeout of 30 seconds, which we cannot modify or exceed. If your function takes longer than that to run, you'll need to put some of the functionality in non-ApiGateway Lambdas and consume the output in your endpoint/handler.
```
haworku, line 203: Great little detail. Kind of thing its good to know before you start 😆

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
+# Endpoints
+
+There are two ways that we create endpoints in our app. One uses GraphQL, the other creates AWS Lambda handlers.
+
+When we want to fetch or update information relevant to the app itself--think of submissions, or user info--we'll typically use GraphQL.
```
haworku, line 5:
```suggestion
**When we want to fetch or update information relevant to the app itself--think of submissions, or user info--we'll typically use GraphQL.**
```

docs/technical-design/endpoints.md
```diff
@@ -0,0 +1,213 @@
...
+
+When we want to communicate with a service like New Relic, download a file, or, for example, to set up a health-check endpoint, we'll use a handler. The relative merits of each approach should become clear below.
```
haworku, line 7:
```suggestion
**When we want to communicate with a service like New Relic, download a file, or, for example, to set up a health-check endpoint, we'll use a handler.** 

The relative merits of each approach should become clear below.
```

docs/technical-design/endpoints.md
haworku: Bolded these lines for emphasis --  I feel like in a long doc it can be easy to miss the key points. I feel like these two sentences are really the key point if you come here wondering "what should I do" lol.

haworku (approved):
This is great, I don't think its too much info, it felt like the step by step guide. All of my comments are suggestions except please change the file name and file title for clarity.

Now I am thinking what are ways we can also add docs to better annotate and explain serverless config in general because those files are quite dense.

docs/technical-design/endpoints.md
rswerve: These lines contain the sum total of my knowledge of layers and vpcs 😆  @mojotalantikite is there a quick summary you can think to include here?

docs/technical-design/endpoints.md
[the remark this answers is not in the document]
rswerve: This is another question I think @mojotalantikite can answer.

docs/technical-design/endpoints.md
rswerve: @haworku when you say "happen first"...where do you think the command should be run in the process described in the doc?

haworku: I believe it is the first step to the graphql docs. 

If you make a change the prisma schema at all as part of your development of the api you would need to run `yarn prisma generate` to get the new client and types. If you don't make schema changes, you can skip the script all together. The dev tool is already running that command to run the api.  That's my understanding at least. @mojotalantikite please correct me if I am mistaken.

docs/technical-design/endpoints.md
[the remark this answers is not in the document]
mojotalantikite: so we actually can't restrict by the user type directly in the API Gateway right now with the AWS IAM authorizer type we're using. If we moved to using an authorizer that we had more control over we could do that, but in it's current implementation it's just going to let anyone through that has signed in. 

If we want to limit someone by user type at this point, we'd need to do it in the handler itself. Roughly, we'd get the user from Cognito, check it's type, and then allow/deny access based on it being a state or CMS user. Maybe we should write some middleware that does that?

docs/technical-design/endpoints.md
mojotalantikite: Sure! So VPCs we don't have access to control at CMS, so we have one in each environment that was provisioned for us by them. The identifiers we need are stored in Parameter Store, and we use those to configure the VPC that some of these lambdas need access to. The gist of it is: if a handler needs to talk to the DB, it needs to have the VPC config.

That's a good point about the Prisma client engine layer, we don't really have docs about that. It's a layer that is built from the `app-api/scripts/prepare-prisma-layer.sh` script, which is needed to communicate to the DB.

As for the OTEL layer, that is one that is made by [AWS and we just import.](https://aws-otel.github.io/docs/getting-started/lambda/lambda-js) We can write some docs about it -- there's actually an updated layer that we might want to schedule an upgrade to our OTEL tooling for.

docs/technical-design/endpoints.md
mojotalantikite: There's actually a couple things that need to happen. `yarn prisma generate` will generate the schemas, but it won't make a new SQL command that alters the DB. 

If you need to add anything to the schema and alter the DB, you can do that and work with it locally using [`prisma db push`](https://www.prisma.io/docs/concepts/components/prisma-migrate/db-push). Once you know you have the change you want with the schema, you can do [`prisma migrate dev --name name_of_db_change`](https://www.prisma.io/docs/concepts/components/prisma-migrate/get-started). That will generate the SQL code that runs in the DB to change the tables properly for your new schema.

JasonLin0991 (approved):

docs/technical-design/creatingEndpoints.md
```diff
@@ -149,14 +183,11 @@ And also add it to the Resolvers object. We've made a query, so this line will b
...
 ```
 
-### Generating code
+### Generating prisma code
 
-Once we have our backend code in place, we need to run a couple of commands to generate glue types and methods.
-
-_**On the command line, navigate to `/services/app-graphql`.**_  
-_**Run `yarn gqlgen`**_
+Once we have our backend code in place, if we've modified the prisma schema (we did not, in this example), we need to run another generate command.
```
haworku, line 188: Yep, nice!

PR #1225, opened 2022-09-28
MR-2231: State can add an additional rate cert

JasonLin0991:
## Summary
[MR-2231](https://qmacbis.atlassian.net/browse/MR-2231)

Allow state users to add additional rate certifications.

#### Test cases covered
- Tests for this new feature will be added in another small feature.

## QA guidance
**The new additional rate feature is hidden behind the `multi-rate-submissions` feature flag.**

There have been changes in how the API checks rate data. Please make sure to check
- Changing submission to `contract only` and submitting removes rate data.
   - You can check this by changing it back to contract rates after submission, there should not be any rate data prefilled in the rate details page.
- There has been a change in the domain model since we added `rateInfos`. Since both `makeDocumentKeyLookupList.test.ts` and `makeDocumentDateLookupTable.test.ts` both used  protobuf encoded string, the tests failed because the encoded string did not reflect the changes. I refactored both theses tests to use encoded ( `toProtobuffer()` ) mock revision form data. This way it would reflect changes to both protobuf and domain. 
   - **Please test to make sure documents still show up correctly on the review and summary pages with single rate certifications**
- CSV data dumps will also need QA


services/app-web/src/common-code/healthPlanFormDataType/UnlockedHealthPlanFormDataType.d.ts
```diff
@@ -66,6 +66,17 @@ type RateCapitationType = 'RATE_CELL' | 'RATE_RANGE'
 
 type ManagedCareEntity = 'MCO' | 'PIHP' | 'PAHP' | 'PCCM'
 
+type RateInfoType = {
+    rateType?: RateType
```
haworku, line 70: 🔍 I would like to take another look at types here. A couple questions

- I notice that for both unlocked and locked we are adding additional types but still keeping the old rate related types as well. That's double the types. Why aren't we removing the types we won't use anymore? 
-  Is there a reason to give `RateInfoType` such a loose type with less optional fields? Stronger typing could better reflects its state in locked submission (with required fields where we expect them to be present). The rateInfos used in unlockedSubmissions seem to be to be `Partial<RateInfoType>`.

services/app-web/src/common-code/proto/healthPlanFormDataProto/toDomain.ts
```diff
@@ -320,6 +356,7 @@ const toDomain = (
 
     // Protos support multiple rate infos for now, but we only support one in our domain models
     // so if there are multiple we'll drop the extras.
+    // This should be removed once all multi-rate UI features have been implemented. We should then only use rate data from rateInfos list of rate data.
```
haworku, line 359: When would it be removed? I'm confused why not now. Or which feature are you thinking this should be shipped with?

JasonLin0991: I was thinking in [MR-2527](https://qmacbis.atlassian.net/browse/MR-2527). But I can do that now and leave UI work for MR-2527. I think this will mean it's blocked by this PR.

services/app-web/src/common-code/healthPlanFormDataType/UnlockedHealthPlanFormDataType.d.ts
JasonLin0991: > I notice that for both unlocked and locked we are adding additional types (rateInfos) but still keeping the old rate related types as well. That's double the types. Why aren't we removing the types we won't use anymore? I was assuming we would remove the old fields since we will be relying on the array going forward. Let me know what you are thinking.

The review and summary pages still use the old fields. Switching them over is a part of another task [MR-2527](https://qmacbis.atlassian.net/browse/MR-2527). But I think I could get the pages to use the first rate from `rateInfos` for now and leave the UI work for MR-2527.

> Is there a reason to give RateInfoType such a loose type with less optional fields? Stronger typing could better reflects its state in locked submission (with required fields where we expect them to be present). The rateInfos used in unlockedSubmissions seem to be to be Partial<RateInfoType>.

That's a good point, I'll make those changes. I'm wondering if that would cause issues with save as draft, but I'll test it out.

JasonLin0991: @haworku There were more Backend work than I initially thought. So there will be more changes coming.

haworku: Sounds good no worries. I know this PR was still in draft and not yet in review, thought I would take a look. Let me know if there's anything you want to pair on as well. I think we underestimated this ticket, this is a significant chunk of work you took on, thanks for getting things in order!

services/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.ts
```diff
@@ -54,25 +54,37 @@ const hasValidContract = (sub: LockedHealthPlanFormDataType): boolean =>
...
+        sub.rateInfos.length === 0
+            ? false
+            : sub.rateInfos.every((rateInfo) => {
+                  const validBaseRate =
+                      rateInfo.rateType !== undefined &&
+                      rateInfo.rateDateCertified !== undefined &&
+                      rateInfo.rateDateStart !== undefined &&
+                      rateInfo.rateDateEnd !== undefined &&
+                      (rateInfo.rateProgramIDs !== undefined ||
+                          rateInfo.rateProgramIDs !== [])
```
rswerve, line 67: ! == [] will always be true.  I think you want a length check here.

JasonLin0991: You are 100% correct. If we do change this to check for length though, existing submissions in prod will no longer be able to submit without adding rate programs. State users would not know that the submission failed because of missing rate program sunless they navigate back to `Rate Details` page. 

We could fix with multiple ways like our migrations tool, front end logic or backend logic, but we should separate this into another ticket, [MR-2680](https://qmacbis.atlassian.net/browse/MR-2680),  and combine the fix with addressing existing submissions

services/app-web/src/components/FileUpload/FileUpload.tsx
```diff
@@ -94,8 +98,17 @@ export const FileUpload = ({
     const inputRequired = inputProps['aria-required'] || inputProps.required
     // update fileItems in parent
     React.useEffect(() => {
-        onFileItemsUpdate({ fileItems })
-    }, [fileItems, onFileItemsUpdate])
+        if (JSON.stringify(fileItems) !== JSON.stringify(previousFileItems)) {
+            onFileItemsUpdate({ fileItems })
+        }
+    }, [fileItems, previousFileItems, onFileItemsUpdate])
```
JasonLin0991, line 104: @haworku Here is a change in FIleUpload that will need some extra scrutiny. Because `onFileItemsUpdate` is actually setting the state in `RateDetials` component, it was causing an infinite re-render when there was multiple rate certifications. Checking the `stringify` data of `fileItems` and `previousItems` prevented this use effect from calling `onFileItemsUpdate` unless there was an actual change in files or file data.

haworku (approved):
Things look good to me. I did play with things a bit in a review app, I ran the contract to contract and rate pattern and also looked at supporting documents. FIleUpload change looks fine to me. 

 It also sounds like the types cleanup  will be addressed in another story, that was my only lingering concern.

PR #1229, opened 2022-09-29
MR-2464: Fill in rate program names from programIDs

rswerve:
## Summary

We now allow users to assign programs to rates (rateProgramIDs), as we've always done for submissions as a whole (with programIDs).  But that means older submissions don't have user-selected program names for their rates.  This PR falls back to program IDs to fill in program names for rates from older submissions.

#### Test cases covered

Added a test to check that the fallback works.  The real test will be pulling up older submissions in a deployed environment and seeing program names in the rate detail section.


haworku (approved):
Code change looks good. I checked the existing `generateRateName` function too just to make sure things are in sync...  looks like it already handles this (fallback to programIds if rate programids don't exist)

PR #1230, opened 2022-09-29
Improve page view logging in Tealium 

haworku:
## Summary
The page_view events were not being properly associated in google analytics. There was some bad or just confusing data (things like `undefined: Home`, `Home: Home`, `Dashboard:Dashboard`  being logged).  I added a function to better determine the name of pages, particularly with a shared route (such as ROOT `/`)

I've also added new types `RouteTWithUnknown` to reflect a type that is already used in the app currently in a few places. 

### Related issues
This is a small part of MR-2572 I can complete now without waiting for response to Blast ticket. 
https://qmacbis.atlassian.net/browse/MR-2572 
 
### Screenshots
Screen Shot 2022-09-29 at 5 34 48 PM

## QA guidance
If you navigate through the app locally you can see the mock Tealium event console log showing what page name would be logged. 

**Note to reviewers 😄 :** 
Even though the page names are now more correct you will see they are happening twice on certain pages. This is _not_ unexpected and could mess up analytics but I wasn't able to address this quick re-render bug easily. I tried a couple things to cleanup the effect but couldn't pinpoint what was causing the issue (what is causing AppBody to unmount). I think it is one of the contexts (pagecontext or authcontext) but also couldn't work around that either - here's an [article](https://javascript.plainenglish.io/react-context-why-am-i-getting-unnecessary-re-renders-b61836d224a7) I was referencing. 

I welcome code review on this now if you can see where the bug is (or have suggestions).  However, since this was a bug present with the existing implementation on prod, I don't think this issue should block the incremental improvement.

### DIFF
--- a/services/app-web/src/components/DynamicStepIndicator/DynamicStepIndicator.tsx
+++ b/services/app-web/src/components/DynamicStepIndicator/DynamicStepIndicator.tsx
@@ -1,10 +1,10 @@
 import { StepIndicator, StepIndicatorStep } from '@trussworks/react-uswds'
 
-import { PageTitlesRecord, RouteT } from '../../constants/routes'
+import { PageTitlesRecord, RouteTWithUnknown } from '../../constants/routes'
 
 export type DynamicStepIndicatorProps = {
-    formPages: RouteT[]
-    currentFormPage: RouteT | 'UNKNOWN_ROUTE'
+    formPages: RouteTWithUnknown[]
+    currentFormPage: RouteTWithUnknown
 }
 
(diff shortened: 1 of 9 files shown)


rswerve (approved):
Looks good and walking through I see the appropriate console logs.

FYI, I've used [Why Did You Render](https://www.npmjs.com/package/@welldone-software/why-did-you-render) with some luck in the past.

haworku:
@rswerve thanks! I will def check that out. Def was reaching the limit React performance tools and console log.

PR #1233, opened 2022-10-03
Clean up protobuf data migrate

mojotalantikite:
## Summary

When testing the protobuf data migrator against the CalendarDate PR in #1126 there were some issues with reading and applying the migration file. This change fixes the migrator. 

I was able to confirm in another [branch](mt-migrate-path) that the code from #1126 migrates the dates to avoid the off by one error it introduced.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2463


services/app-api/tsconfig.json
```diff
@@ -1,17 +1,19 @@
...
+        "target": "es2021",
+        "moduleResolution": "node",
+
         "baseUrl": "/",
         "experimentalDecorators": true,
         "emitDecoratorMetadata": true,
         "preserveConstEnums": true,
-        "sourceMap": true,
-        "target": "es5",
+        "sourceMap": false,
```
haworku, line 13: Just looking at the scale of these changes to tsconfig for an unrelated effort is like reading the rationale for https://qmacbis.atlassian.net/browse/MR-2672 in code form.

services/app-api/src/handlers/postgres_migrate.ts
```diff
@@ -101,8 +97,14 @@ export const main: APIGatewayProxyHandler = async () => {
 
     // run the data migration. this will run any data changes to the protobufs stored in postgres
     try {
-        const migrator = newDBMigrator(dbConnectionURL)
-        await migrate(migrator, '/opt/nodejs/')
+        execSync(
```
haworku, line 100: just checking my understanding - we are not able to use MacRae's fav pattern of dependency injection anymore here _because_ it would require that cross service import to get the protos. Is that right? This seems like the area of our code in this fix directly related to https://qmacbis.atlassian.net/browse/MR-2671

services/app-api/src/handlers/postgres_migrate.ts
```diff
@@ -2,10 +2,6 @@ import { APIGatewayProxyHandler } from 'aws-lambda'
 import { RDS } from 'aws-sdk'
 import { execSync } from 'child_process'
 import { getDBClusterID, getPostgresURL } from './configuration'
-import {
-    migrate,
-    newDBMigrator,
-} from '../../../app-proto/protoMigrations/lib/migrator'
 
 export const main: APIGatewayProxyHandler = async () => {
```
haworku, line 6: nit: This is outside the scope of the change but  I'm wondering about AWS lambda practices here. To me this handler could you add a better 200 return info. Right now it's very basic - not that use for logging purpose later.  I feel like it would be helpful if we returned some kind of info around which migration returned successfully.  what do you think @mojotalantikite 
```
return {
        statusCode: 200,
        body: JSON.stringify('successfully migrated'),
        headers: {
            'Access-Control-Allow-Origin': '*',
            'Access-Control-Allow-Credentials': true,
}
```

haworku (approved):

services/app-api/src/handlers/postgres_migrate.ts
mojotalantikite: yeah, I had to move away from the pattern in the handler itself because of issues from the interplay between the lambda environment, the webpack bundle, and the import of compiled scripts from `app-proto`.

services/app-api/src/handlers/postgres_migrate.ts
mojotalantikite: I can definitely switch this up a bit. I'd just have to record which parts of the migrate handler ran (schema vs data) and return that. Shouldn't be that hard, happy to do it here.

mojotalantikite:
I started adding in some of the success returns, but the more I thought about it the more cases started coming up. Pushed that follow on work to this ticket, so we're not blocking the bug fix on it: https://qmacbis.atlassian.net/browse/MR-2692

PR #1245, opened 2022-10-06
First rate-certification in multi-rate submission on Rate Details page should not be removable.

JasonLin0991:
## Summary
[MR-2231](https://qmacbis.atlassian.net/browse/MR-2231)
First rate-certification in multi-rate submission on Rate Details page should not be removable.


haworku (approved):
Hehe. Yup 👍 good catch

PR #1251, opened 2022-10-12
Fix system error when adding additional rate when existing rates have incompletion errors.

JasonLin0991:
## Summary
[MR-2231](https://qmacbis.atlassian.net/browse/MR-2231?focusedCommentId=114581)

The new array is added when `FileUpload` component calls `onFileItemsUpdate` to update `fileItemsMatrix` with the files for the specific rate certification. `getDocumentsError` ran before a new array was added into `fileItemsMatrix` state so using any array method caused crashes. 

The fix was to push in new empty array into `fileItemsMatrix` when clicking on `Add another rate certification`.

### DIFF
--- a/services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
+++ b/services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
@@ -1175,6 +1175,18 @@ export const RateDetails = ({
                                                         const newRate =
                                                             rateInfoFormValues()
                                                         push(newRate)
+                                                        setFileItemsMatrix(
+                                                            (fileMatrix) => {
+                                                                const newMatrix =
+                                                                    [
+                                                                        ...fileMatrix,
+                                                                    ]
+                                                                newMatrix.push(
+                                                                    []
+                                                                )
+                                                                return newMatrix
+                                                            }
+                                                        )
                                                         setFocusNewRate(true)
                                                     }}
                                                     ref={newRateButtonRef}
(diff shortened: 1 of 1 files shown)


services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
```diff
@@ -1175,6 +1175,18 @@ export const RateDetails = ({
                                                         const newRate =
                                                             rateInfoFormValues()
                                                         push(newRate)
+                                                        setFileItemsMatrix(
+                                                            (fileMatrix) => {
+                                                                const newMatrix =
```
rswerve, line 1180: In one place this gets called `newFiles`.  Is it actually different there?
Also, this looks fine; just trying to understand what happens to the empty array that gets pushed.  Does it ever get populated?

JasonLin0991: I think you are referring to `newFiles` in the `onFileItemsUpdate` prop in `FileUpload`? This is actually when uploaded files from `FileUpload` files are inserted into the `fileItemsMatrix` state, but on initial load of the component it will pass in whatever is returned from the `FileUpload`. For new rate certification this would be `undefined` since the other prop `initialItems` in this component uses `fileItemsMatrix`. This is why the `getDocumentsError` causes the bug.

The empty array should not do anything, our checks for valid files handles empty arrays. The empty array will be overwritten when new files are uploaded using the `splice()` method. This is done in the function for `onFileItemsUpdate` prop in `FileUpload`

```
setFileItemsMatrix((fileMatrix) => {
        const newFiles = [...fileMatrix]
        newFiles.splice(index,1, fileItems)
        return newFiles
    }
)
```

rswerve: Gotcha on the empty array.  For the naming, I'm referring to line 577 of this file in this PR.  Since the array gets called `newFiles` there, it makes me think it's an array of files, not an array of arrays.  Is that correct?  If it's still an array of arrays, we should probably call it `newMatrix` there too.

rswerve (approved):

PR #1254, opened 2022-10-13
Add unit and Cypress tests for multi-rate submissions

haworku:
## Summary

- Add unit tests for multi-rate submission to jest and Cypress
- Wrap each rate cert in a `<Fieldset />` for a11y reasons. These are groups of form fields.
- Clear out necessary logging from hooks used in tests.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2678

#### Test cases covered
Unit tests for handles multiple rates
- renders add another rate button, which adds another set of rate certification fields to the form
- accepts documents on second rate
- cannot continue without selecting rate type for a second rate
- cannot continue if no documents are added to the second rate
- progressively disclose new rate form fields on the second rate

Cypress tests
- can unlock and resubmit multi-rate submissions
- can add and remove multiple rate certifications and navigate to and from rate details page

## QA guidance

n/a this is a testing task


services/app-web/src/hooks/useStatePrograms.ts
```diff
@@ -8,7 +8,7 @@ const useStatePrograms = (): Program[] | [] => {
 
     if (loggedInUser && loggedInUser.__typename === 'StateUser') {
         statePrograms = loggedInUser.state.programs
-    } else {
+    } else if (loggedInUser && loggedInUser.__typename === 'CMSUser') {
```
haworku, line 11: I made this more restrictive because otherwise `undefined` loggedInUser from tests were getting caught here and flooding logs with unnecessary print statements

services/app-web/src/hooks/useTealium.ts
```diff
@@ -101,15 +101,13 @@ const useTealium = (): {
     useEffect(() => {
         // Do not add tealium for local dev or review apps
         if (process.env.REACT_APP_AUTH_MODE !== 'IDM') {
-            console.log(`mock tealium page view: ${tealiumPageName}`)
+            // console.log(`mock tealium page view: ${tealiumPageName}`)
```
haworku, line 104: I commented out these since they are very useful when debugging tealium but add a lot of noise otherwise

I removed the progrmming error console.error because this will log in tests that are using useTealium and that seems excessive (but also not really worth mocking).

services/app-web/src/pages/Landing/Landing.test.tsx
```diff
@@ -6,6 +6,8 @@ import {
 import { Landing } from './Landing'
 
 describe('Landing', () => {
+    afterAll( () => jest.clearAllMocks())
```
haworku, line 9: noticed we are using a feature flag mock here but forgot to clear so added this

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.test.tsx
```diff
@@ -454,6 +500,395 @@ describe('RateDetails', () => {
...
+            return screen.getAllByTestId('rate-certification-form')
+        }
+
+        const lastRateCertificationFromList = (screen: Screen) => {
+            return rateCertifications(screen).pop()
+        }
+
+        it('renders add another rate button, which adds another set of rate certification fields to the form', async () => {
+            renderWithProviders(
+                <RateDetails
```
JasonLin0991, line 597: Should we add another test like this, but removing a rate certification?

JasonLin0991 (changes requested):
Looks good, just one question about possibly adding one more test.

JasonLin0991 (approved):
Thanks for the new test!

JasonLin0991:
@haworku Could not figure out the reason for cypress failing tests when trying to add a second rate certification. 

- The tests start failing to `findByText('Submissions')` in the `'can unlock and resubmit'` tests, on cold start. 
- It starts to fail when adding testing code to navigate to the `Review and submit` page after the submission has been unlocked a second time.
- Running cypress locally, connected to review deployment, on cold start does pass, it's only in CD where it fails.

I removed the second test entirely and commented out the code in the first test that adds a second rate certification.

haworku:
Got it. I think we should merge what we have here so some rate tests exist and close the ticket (since we are covering multi rate in jest tests and in the new rate details test). Let's connect beginning of next week about who can take on getting that multi-rate  unlock resubmit test commented back in.

PR #1262, opened 2022-10-17
Automatically Prune Aurora Snapshots

mojotalantikite:
## Summary

We create Aurora snapshots before every data migration run just in case we need a restore point. We also have snapshots created automatically for other actions that take place in AWS. This has led to us hitting resource limits in `dev` before, so we needed a way to clean up those resources periodically.

This adds a lambda function that runs on a cron schedule to scan for snapshots older than 30 days, pruning them when they are found.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2589


mojotalantikite:
One thing to mention is that I ended up using the [v3 JS aws-sdk](https://docs.aws.amazon.com/sdkref/latest/guide/version-support-matrix.html) for this handler. It's been GA for almost 2 years now and thought we might want to start looking at adopting it for new code.

haworku (approved):

haworku:
Looks good, let me know the plan for how we verify this after it merges @mojotalantikite

mojotalantikite:
@haworku thanks, I'll merge now then! I'll take a look at our val/prod environments after the cron run and make sure it's only getting rid of snapshots older than 30 days (I confirmed this in the dev env already).

PR #1274, opened 2022-10-19
MR-2698: Generate rate name when Protobuf is constructed

rswerve:
## Summary

Up until now, we've been generating the rate name on the fly wherever it's been needed (eg. in review and submit, or when constructing emails).  But we decided that it made more sense to add it as a property of rateInfos, which we would set once, and then reference wherever we need it.  This should be simpler and less error prone, as we now call `generateRateName` only once in the app.

#### Screenshots

Everything should work exactly as before

#### Test cases covered

Now we call `generateRateName` in our tests, and make sure it matches what's already in the form data.

#### QA Guidance

There are still several places where we reference `rateInfos[0]`.  Those will all be converted to work with an index in MR-2527, where we will show multiple rate certs on review and submit and submission summary pages.


services/app-web/src/common-code/proto/healthPlanFormDataProto/toProtoBuffer.ts
```diff
@@ -195,6 +227,11 @@ const toProtoBuffer = (
                                   doc.documentCategories
                               ),
                           })),
+                          rateProgramName: generateRateName(
+                              domainData,
+                              domainData.rateInfos[index],
+                              findRatePrograms(domainData, index)
+                          ),
```
JasonLin0991, line 234: Could we pass the `rateInfo` from the map function instead of `domainData.rateInfos[index]` since it would be the same data.

I think we can also just pass all the programs for a given state in the third parameter in `generateRateName()`. `generateRateName()` calls `packageName()` which will use package programs if rate programs do not exists and then `programName()` will find the name of the program from the id.

services/app-api/src/postgres/insertHealthPlanPackage.ts
```diff
@@ -81,7 +81,7 @@ export async function insertHealthPlanPackage(
         managedCareEntities: [],
         federalAuthorities: [],
     }
-
+    console.log('draft in insert: ', draft)
     const protobuf = toProtoBuffer(draft)
```
JasonLin0991, line 85: console.log for testing?

JasonLin0991 (changes requested):
Just some questions in `toProtoBuffer`. I really like that we no longer need to call the `generateRateName()` function all the time in the emailer functions.

services/app-api/src/postgres/insertHealthPlanPackage.ts
rswerve: thanks.  removed

services/app-web/src/common-code/proto/healthPlanFormDataProto/toProtoBuffer.ts
rswerve: great suggestions; much cleaner now

services/app-api/src/resolvers/unlockHealthPlanPackage.test.ts
```diff
@@ -575,13 +576,21 @@ describe('unlockHealthPlanPackage', () => {
...
+        expect(mockEmailer.sendEmail).toHaveBeenNthCalledWith(
+            4,
             expect.objectContaining({
                 subject: expect.stringContaining(`${name} was unlocked`),
                 //Rate name should have defaulted back to using package programs to generate name
                 bodyHTML: expect.stringContaining(rateName),
             })
         )
+        // expect(mockEmailer.sendEmail).toHaveBeenCalledWith(
+        //     expect.objectContaining({
```
haworku, line 588: nit: i'm guessing this meant to be commented in or else removed

services/app-proto/src/health_plan_form_data.proto
```diff
@@ -178,6 +178,7 @@ message RateInfo {
   repeated Document rate_documents = 8;
   optional RateCapitationType rate_capitation_type = 9;
   repeated string rate_program_ids = 10;
+  optional string rate_program_name = 11;
```
JasonLin0991, line 182: `rate_program_name` could be confused for a single program name or a list of program names. Something like `rate_certification_name` or just `rate_name` might describe what this string is. Sorry for not catching this on the previous review.

JasonLin0991 (approved):
Thanks for the changes. One more suggestion, but not blocking.

services/app-api/src/resolvers/unlockHealthPlanPackage.test.ts
rswerve: thx

services/app-proto/src/health_plan_form_data.proto
rswerve: makes sense.  changed to rate_certification_name

PR #1276, opened 2022-10-20
MR-2238: State can only add additional actuary on contact page

JasonLin0991:
## Summary
[MR-2238](https://qmacbis.atlassian.net/browse/MR-2238)

- Add new fields `addtlActuaryContacts` and `addtlActuaryCommunicationPreference`  to protoBuffer schema 
- UI on `Contacts` page and `ContactsSummarySection.tsx` for multi-rate-submissions flag on
- Toggling between UI
- Fix some errors in passing tests
- Added unselected `Actuary Communication Preferences` error to error summary..

Notes
- Additional actuaries are not directly related to any specific rate certification, instead they are actuaries that apply to the entire package as a whole. Previously, when dealing with single rate submissions, we pulled out actuaries from the first rate certification in `rateInfos`. Now additional actuaries will have its own field, `addtlActuaryContacts` in the `proto` schema and `domain` model at the top level of the schema and model.

- In addition `Actuary Communication Preferences` will also have its field, `addtlActuaryCommunicationPreference` at the top level. Instead of being pulled from any specific rate certification.

- Most of the logic in `Contact.tests.tsx` has not changed because our previous domain model was using a top level field `actuaryContacts` to pull data into and save into the first rate certification in `rateInfos` of the proto data. This happens in `toProtoBuffer` and `toDomain`. This field did not exist in the protoBuffer schema at the top level.

#### Test cases covered
Unit Tests:
- Contacts.test.tsx
   - `'text renders correctly when multi-rate-submissions flag is turned on'`
- ContactsSummarySection.test.tsx
   - `'renders correct text for actuary contact fields when multi-rate-submissions flag is on'`
- UploadedDocumentsTable.test.tsx
   - Fixed passing tests that logged errors
- SupportingDocumentsSummarySection.test.tsx
   - Fixed passing tests that logged errors

Cypress Tests:
- contacts.spec.ts
   - `'can navigate to and from contacts page with contract and rates submission with multi-rate-submissions flag on'`


services/app-web/src/common-code/proto/healthPlanFormDataProto/toDomain.ts
```diff
@@ -441,12 +443,12 @@ const toDomain = (
...
         rateProgramIDs: rateInfo?.rateProgramIds ?? [],
-        actuaryCommunicationPreference: enumToDomain(
+        addtlActuaryCommunicationPreference: enumToDomain(
             mcreviewproto.ActuaryCommunicationType,
-            rateInfo?.actuaryCommunicationPreference
+            addtlActuaryCommunicationPreference
         ),
         stateContacts: cleanedStateContacts,
-        actuaryContacts: parseActuaryContacts(rateInfo),
+        addtlActuaryContacts: parseActuaryContacts(addtlActuaryContacts),
```
haworku, line 451: 🔍 This changes here looks risky to me. If we were to merge without the other related tickets I think we would cause a regression, because we discard the `rateInfo.[ACTUARY_FIELD]`  from the domain model data all together. 

However, those same fields are where old submissions are storing actuary contacts.

To make this change more able to merge by independently (without also requiring https://qmacbis.atlassian.net/browse/MR-2233) I believe we should still be reading from the rateInfos as a fallback. 

Something like 
        addtlActuaryContacts: parseActuaryContacts(addtlActuaryContacts) ?? parseActuaryContacts(rateInfo)

haworku: resolved with [9bdeae2](https://github.com/CMSgov/managed-care-review/pull/1276/commits/9bdeae2339ccbad98f2a0c16d117a2274cb7c753)

haworku (approved):
Things look good.

haworku:
I think we can close this in favor of #1292 but will let you do the honors @JasonLin0991

PR #1280, opened 2022-10-21
MR-2698: Fix for old data

rswerve:
Meghan says that the name isn't showing up for old data.  I added a fallback to generating the name when it's not already on the submission.

### DIFF
--- a/.github/actions/setup_env/action.yml
+++ b/.github/actions/setup_env/action.yml
@@ -9,7 +9,7 @@ runs:
       id: yarn-cache-dir-path
       run: echo "::set-output name=dir::$(yarn cache dir)"
 
-    - uses: actions/setup-node@v3
+    - uses: actions/setup-node@v2.5.0
       with:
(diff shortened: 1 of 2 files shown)


haworku (approved):

haworku:
Thank you for getting a PR up! Going to merge this to address prod bug. 

I want to pair to make a follow on PR  though about a couple things
-  I believe this could print `undefined` to the user which seems like something we should avoid
- Would like to try doing this in `toDomain` rather than in React components (there might be other places we need the rate name down the road, we would have to remember  use this function in each place)
- Can we get a failing unit test for the prod bug case? Not sure but that would be ideal to prevent this from happening again. I think this is the second time we have had to do a quick fix for rate names.

rswerve:
> 

Oh, for sure.  This was literally the quickest possible fix for a prod issue.  I'm already working on unit tests for old protos, and yeah, this logic will move with the rest of the logic we just moved 🙂 into the rate name generator.

PR #1283, opened 2022-10-24
MR-2742: Test old protos on Submission Summary page

rswerve:
## Summary

We need some kind of testing around loading old protos without breaking the UI.  Currently, a user can go directly to a submission (which contains a proto), bypassing any data handling that we do in toProtoBuffer that's been added since the proto was created. 
This PR takes old protos that we save to the file system whenever our data model changes, and loads them into the UI and does a snapshot test.

**NOTE**: When you run these locally, you'll need to set the jest timezone to UTC.  For example, I run tests locally by navigating to `services/app-web` and running `yarn test`.  I would need to run `TZ=UTC yarn test` instead.


services/app-web/src/pages/SubmissionSummary/SubmissionSummary.test.tsx
```diff
@@ -333,6 +335,55 @@ describe('SubmissionSummary', () => {
             })
         })
 
+        it('loads old protos correctly', async () => {
```
haworku, line 338:
```suggestion
        it('loads outdated health plan packages with old protos as expected', async () => {
```

haworku:
Please change our `test` script to adjust for timezone since it looks like this setting is required for snapshots to reliably pass. This is in the app-web `package.json`, can be changed to  `"test": "TZ=UTC react-scripts test"`

rswerve:
> 

Good suggestions; thanks.  Pushed em up and will merge when it passes...

PR #1288, opened 2022-10-25
MR-2233: State can add certifying actuary on rate details page

JasonLin0991:
## Summary
[MR-2233](https://qmacbis.atlassian.net/browse/MR-2233)

- Add certifying actuary contact form for each rate certification section on the Rate Details page.
- Add `ActuaryContactFields` component and refactor `Contacts.tsx` to use his component.
- Update `rateInfos` domain model to include `actuaryContacts` and `actuaryCommunicationPreference` fields.
- Update tests for domain model changes.
- Update `hasValidRates` to check for valid actuary contacts data.
- Update `updateToVersion3()` to convert the second actuary contact and beyond to `addtlActuaryContacts`.

#### Test cases covered

Unit tests
- `RateDetails.test.tsx`
   - `handles multiple rates`: Update tests with assertions for certifying actuary contact fields

- `healthPlanFormData.ts`:
   - `'hasValidRates evaluates as expected'`: Update tests to cover missing actuary contact fields

- `toLatestProtoVersion.test.ts`
   - `'should convert actuary contacts after the first to addtlActuaryContacts and communication preference as expected'`
   - `'should return empty array for addtlActuaryContacts when only a single actuary contact exists'`

- Cypress Tests
   - `contacts.spec.ts`: Update to fill out rate details before filling out contacts.
   - `fillOutNewRateCertification` and `fillOutAmendmentToPriorRateCertification`: Code to fill out certifying actuary contact fields

## QA guidance

**Because we are running `toLatestVersion()` based on proto version. These tests can only be done on older submission data with a previous proto version. I have created two submission on VAL (FL) that can be used to test these two scenarios in preparation for PO/Design review.**
[Single actuary contact](https://mc-review-val.onemac.cms.gov/submissions/16deca1b-0c97-4d39-9bd6-d766a1177169)
[Multiple actuary contacts](https://mc-review-val.onemac.cms.gov/submissions/ed68a272-47d5-4b28-bfa7-b76130e69b93)

- Old actuary data can be displayed correctly with feature flag on.
1. <strike>Turn `multi-rate-submissions` flag **off** and fill out a contract and rates submission with more then 1 actuary contact.</strike>
2. <strike>Submit, then unlock.</strike>
2. Turn feature flag on
3. Navigate to Rate details page and verify certifying actuary is correct for the first ( and only rate).
4. Click continue and verify additional actuary information is correct.
  
- Submissions with old actuary data can still be resubmitted when feature flag is off.
1. Turn `multi-rate-submissions` flag **off** and fill out a contract and rates submission with more then 1 actuary contact.
2. Submit, then unlock.
3. Edit actuary contacts on Contacts page and continue to submit.


services/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.ts
```diff
@@ -64,7 +64,18 @@ const hasValidRates = (sub: LockedHealthPlanFormDataType): boolean => {
...
+                          rateInfo.rateProgramIDs !== []) &&
+                      rateInfo.actuaryContacts.every(
+                          (actuaryContact) =>
+                              actuaryContact.name !== undefined &&
+                              actuaryContact.titleRole !== undefined &&
+                              actuaryContact.email !== undefined &&
+                              ((actuaryContact.actuarialFirm !== undefined &&
+                                  actuaryContact.actuarialFirm !== 'OTHER') ||
+                                  (actuaryContact.actuarialFirm === 'OTHER' &&
+                                      actuaryContact.actuarialFirmOther !==
```
JasonLin0991, line 76: This is kind of confusing, let me know if anyone has a suggestion. Every contact should have every field, if `actuarialFirm` is `OTHER` then `actuarialFirmOther` should not be undefined.

services/app-api/src/resolvers/submitHealthPlanPackage.ts
```diff
@@ -231,7 +231,7 @@ export function submitHealthPlanPackageResolver(
         const submissionResult = submit(draftResult)
 
         if (isSubmissionError(submissionResult)) {
-            const errMessage = 'Incomplete package cannot be submitted'
+            const errMessage = submissionResult.message
             logError('submitHealthPlanPackage', errMessage)
             setErrorAttributesOnActiveSpan(errMessage, span)
```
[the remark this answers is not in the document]
haworku: What do the possible error messages look like and where are they used? Does the user ever see it?

I like the more specific error message for logging for sure. 

But I _believe_ we are  displaying this thrown err message  to the user too. Peeking in the submit modal we have code like this.

>  if (result instanceof Error) {
>             setModalAlert({
>                 heading: modalValues.errorHeading,
>                 message: result.message,
>        })

Want to make sure that any api messaging we directly surface is to users is human readable (and run by design). We should be confident know most the possible errors that api message could be before we surface to users. For that reason, I would keep the more detailed logging you have added but still pass the vague error message text we originally had to the thrown UserInputError. Its right below this code block you highlighted.

 Let's talk about field specific error messaging with design, its a good topic for frontend design meeting. Y

services/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.ts
haworku: Outside the scope of this PR, we need to rework this function.

l am very wary of expanding the way we are using this hasValidateRates function to include checking nested fields. We aren't unit testing these functions either and these functions block/allow submission.

The chance for bugs increases as we load logic in here. And it might not be needed - we already have multiple validation schemas in the app. Kind of feels like we should be reading from one of those schema to check  rates - not hard coding the validation data in yet another place.  

Since this is new functionality (we never had isValidContacts) and never checked actuary contacts at this stage before, I suggest we do something very simple for now and revisit with a follow on PR. Just spot check that the actuary contacts have email fields and an actuary firm exists, that's it. 

 @rswerve do you have thoughts here? 

@rswerve

haworku (approved):

PR #1294, opened 2022-10-27
MR-2702: RateInfo on new submissions should have saved id

JasonLin0991:
## Summary
[MR-2702](https://qmacbis.atlassian.net/browse/MR-2702)

We already had RateInfo `id` field in the protobuf, this is to now generate a UUID for this field in `toProtoBuffer` and add the field to the domain model.

#### Test cases covered

- `healthPlanFormDataEncoding.test.ts`
   - `'encodes to protobuf and generates rate id'`
   - `'encodes to protobuf and back to domain model without corrupting existing rate info id'`


rswerve (approved):
nice tests

PR #1298, opened 2022-10-31
Remove excessive console logs

haworku:
## Summary
Comment out the console log here. I do think it would be nice to have OTEL know when an old proto is opened (if we had to debug unexpected behavior with toLatestVersion) but not sure right now where that should connect into the active spans. 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2761


mojotalantikite (approved):

haworku:
Done: https://qmacbis.atlassian.net/browse/MR-2764

PR #1299, opened 2022-10-31
MR-2753:  Move generation of rate certificate name to toDomain

rswerve:
## Summary

This is a clean up ticket, to move the generation of a missing property out of react components and into toDomain, so that we can do it once and consume the correct information throughout the front end.

We had hoped to do this in our toLatestVersion protobuf handling, but rate name generation relies on too much information only available after the proto has been converted to the domain model for it to be feasible to run at the proto level.

As it is, I find this ad hoc and messy.  I'm leaning toward preferring a data migration over this code fix.  

Opening this as a draft so folks can see what it is, and we can discuss in DTBM tomorow.

#### Test cases covered

Existing tests should catch any problem here, but I also added a test for when we unwrap an old proto, to make sure the rateCertificationName is being generated.


haworku (approved):
LGTM 🌮 

As per your comment @rswerve , if you want to do this in a data migration, I would support that too. We have a data migration we have to do any way to also add rate ids to old submissions. We could do both together even if that's of interest and folks want to pair (if I'm understanding right, they are addressing similar issues, adding missing rate data to single rate submissions that were added to the db before the multi-rate epic). https://qmacbis.atlassian.net/browse/MR-2714 is @JasonLin0991 ticket

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
```diff
@@ -154,15 +138,15 @@ export const RateDetailsSummarySection = ({
                 {submission.rateInfos.map((rateInfo) => {
                     return (
                         <React.Fragment
-                            key={`${rateName(rateInfo)}${JSON.stringify(
-                                rateInfo.rateDocuments
-                            )}`}
+                            key={`${
+                                rateInfo.rateCertificationName
+                            }${JSON.stringify(rateInfo.rateDocuments)}`}
                         >
```
JasonLin0991, line 144: This maybe an issue, but using these two pieces of data may not guarantee a unique key? I might opted here to add an `index` to the mix and as a part of my ticket [MR-2714](https://qmacbis.atlassian.net/browse/MR-2714) I could switch this to use the `id` in each rate info.

rswerve: ID would be great; are you saying you could roll that into your PR?  And while I don't think this guarantees a unique key, I think it's also very unlikely that we'd get a dupe on the same page.

rswerve:
Rather than handle this in code, we're going to do a migration on the few db entries that are missing the rate certificate name.

rswerve:
Re-opening, because this fixes the problem and nothing we do currently will likely outlast more universal changes to how we use protobufs.

JasonLin0991 (approved):
Looks good.

PR #1317, opened 2022-11-09
MR-2714: rate id migration

JasonLin0991:
## Summary
[MR-2714](https://qmacbis.atlassian.net/browse/MR-2714)

Migration to generate a UUID for field `id` in `rateInfo` on  contract and rates packages that contain a rate certification.

I tested this migration in this [Draft PR](https://github.com/CMSgov/managed-care-review/pull/1312). In the PR description I have posted the package data before and after migration.

#### Test cases covered

`unlockedWithALittleBitOfEverything.test.ts`
- `'0001_rate_id_migration'`
   - `'correctly generates rate ids for old contract and rates submissions'`
   - `'does not override existing rate ids'`
- `'0000_initial_migration'`
   - `'version 2022-08-19 matches the expected values'`
      - Refactored to match pattern in `'0001_rate_id_migration'`


haworku (approved):
LGTM. I see tests checking old protos, we did manual testing on review app, also see diff for the snapshot test is just for the new proto.

services/app-web/src/common-code/proto/healthPlanFormDataProto/unlockedWithALittleBitOfEverything.test.ts
```diff
@@ -122,11 +122,6 @@ describe('0001_rate_id_migration', () => {
             throw migratedFormData
         }
 
-        //Rate certification name generation is a part of a different migration, so to test our rate id migration we will
-        // manually generate a rate id.
-        migratedFormData.rateInfos[0].rateCertificationName =
-            'MCR-MN-0005-SNBC-RATE-20220621-20221021-AMENDMENT-20210523'
```
haworku: Oh we were able to remove this once you merged main? Can you share more about what happened here because I was going to write a ticket about this pattern today. Is it not needed then to do these kind of workarounds in proto tests generally?

JasonLin0991: Yes! It was actually @rswerve rate certification name versioning work that made old data symmetrical with new domain model. Prior, old data did not have rate certification and once you encode, decode and convert to domain model a certification name was generated and now is asymmetrical.

PR #1318, opened 2022-11-10
Add how to change HealthPlanFormData doc

haworku:
## Summary
This is a follow on from the versioning versus migrations technical design doc.

### DIFF
--- a/docs/technical-design/howto-health-plan-form-data-changes.md
+++ b/docs/technical-design/howto-health-plan-form-data-changes.md
@@ -1,0 +1,30 @@
+# How to handle changes in the Health Plan Form Data protocol buffer
+
+## Summary
+
+| Data Layer     | Adding a field | Removing a field | Changing the type of a field |
+| ----------- | ----------- | ----------- | ----------- |
+| Protocol Buffers    |  Go for it!   | Go for it! However, deprecate, don't destroy | Go for it! Test for data loss.
+| Domain Models   | Required field? If so, handle fallbacks for old data and test for data loss      | Go for it! | Handle fallbacks for old data and test for data loss
+
+## Further discussion
+
+### Protobufs
+
+Relevant filepath: health_plan_form_data.proto
+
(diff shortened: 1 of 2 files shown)


mojotalantikite (approved):

JasonLin0991 (approved):

docs/technical-design/howto-health-plan-form-data-changes.md
```diff
@@ -0,0 +1,34 @@
...
+| Domain Models   | Use caution for required fields, you will need to handle fallbacks for old data. Straightforward for optional fields.     | Deletion is okay but deprecate if product wants to display this data to users on old submissions. | Use caution here, you will need to handle fallbacks.
+
+## Further discussion
+
+### Protobufs
+
+Relevant filepath: 
+- [health_plan_form_data.proto](/services/app-proto/src/health_plan_form_data.proto)
+
+Many changes to the protobuf schema itself are straightforward because protos are flexible, have unique field numbers, and use optional field. This means that adding a field or deprecating a field (`[deprecated = true]`) does not effect the ability of the proto decoder. If a newer schema is used to read data written with an older schema, the proto decoder drops fields that are no longer in the schema. It also will not error for fields that are present in new schema but missing from old data (they ar just assigned to undefined or empty list).
```
rswerve, line 17:
```suggestion
Many changes to the protobuf schema itself are straightforward because protos are flexible, have unique field numbers, and use optional fields. This means that adding a field or deprecating a field (`[deprecated = true]`) does not affect the ability of the proto decoder. If a newer schema is used to read data written with an older schema, the proto decoder drops fields that are no longer in the schema. It also will not error for fields that are present in the new schema but missing from old data (they are just assigned to undefined or empty list).
```

docs/technical-design/howto-health-plan-form-data-changes.md
```diff
@@ -0,0 +1,34 @@
...
+
+Changing the data *type* of a field in the schema is also possible, but there is a risk that new values lose precision or is truncated. See discussion of [updating a message type](https://developers.google.com/protocol-buffers/docs/proto3#updating). For example, removing a `repeated` marker from a field will preserve only the X item in the list.
```
rswerve, line 19:
```suggestion
Changing the data *type* of a field in the schema is also possible, but there is a risk that new values lose precision or are truncated. See discussion of [updating a message type](https://developers.google.com/protocol-buffers/docs/proto3#updating). For example, removing a `repeated` marker from a field will preserve only the X item in the list.
```

PR #1319, opened 2022-11-10
MR-2680: API does not correctly check for rate programs when submitting contract and rates package.

JasonLin0991:
## Summary
[MR-2680](https://qmacbis.atlassian.net/browse/MR-2680)

- Fixed `isValidRates()` in `/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.ts` to correctly check `rateProgramIDs` for empty array.
- Updated tests for `isValidRates()`
- Migration for old `submitted` `contract_and_rates` packages without `rateProgramIDs`
- Fixed `0001_initial_migration` to not override `protoVersions` > 1 with `2`, especially in review deploy and unit testing. On Val and Prod, there is a record of the migrations ran, so it should not get ran again.

**NOTE**: There is no good way to test out older data in review app `db` against the migrations. Going back to a version of the app where rate programs were not implemented would be challenging because of how much the app had changed. Hopefully the unit tests in `unlockedWithALittleBitOfEverything.test.ts` will be sufficient.

#### Test cases covered

`unlockedWithALittleBitOfEverything.test.ts`
- `'0002_rate_programs_migration'`
   - `'correctly copies package programs to rate programs for old contract and rates submissions'`
   - `'does not override existing rate programs'`

`submitHealthPlanPackage.test.ts`
- Removed `generates rate name by package programs when rate programs are not specified` - You can no longer submit contract and rate packages without rate programs

`unlockHealthPlanPackage.test.ts`
- Removed `generates rate name by package programs when rate programs are not specified` - You can no longer submit contract and rate packages without rate programs

## QA guidance

**NOTE**: There will not be a perfect way to test this on VAL. Previous work had the review pages generate rate name using the package programs if rate programs do not exist so we can't use that to check which older submission data is the correct old data. 

**But checking when rate program was implemented we can narrow down which submissions on Val could be old data. [MR-2086](https://github.com/CMSgov/managed-care-review/pull/1087), was the PR for rate programs and was merged in on `July 29, 2022`. So any submissions on val where it was last updated before then should be old data. **

There will be two tests for these submissions, although one is less reliable than the other.

1. Unlocking one of these submissions and going to the `Rate Details` page, should then have the `Rate Program` drop down populated with whatever the package program was.**
2. Unlocking one of these submissions and resubmit, without navigating to the `Rate Details` page, should allow you to resubmit. If the migration was not successful, then you should not be able to resubmit.


haworku (approved):
Excuse the delay, this looks good. I downloaded the CSV off prod to be able to check this, similarly we can keep an eye on the June submissions (which are the earliest in the app) as well

PR #1321, opened 2022-11-14
Save Users from Cognito into Aurora

mojotalantikite:
## Summary

Currently we rely on Cognito to lookup users in our graphql lambda handler, which we then pass down into our Apollo server. In order to start doing more with users besides auth, we need to start storing data about those users to Postgres. This PR begins to add User data to Postgres Aurora.

In this PR we're mainly concerned with getting the data out of Cognito and into a Users table in Postgres. When a request comes into our API we check to see if we've stored any data on that user into Postgres before. If not, we look up the user in Cognito and then store that metadata into the Users table in Postgres. If the user does exist in Postgres, we turn it into the same type that the rest of the application expects and return it.

Some things to keep in mind when reviewing this: 
* We use the same user ID that Cognito uses as the user ID in the Postgres User's table. We can use the same ID to lookup in either Cognito or in Postgres.
* There is an existing `UserType` that we use in our application, but Prisma generates a `User` type off of the model, so we have to do some conversions. Eventually it'd be nice to just rely on the Prisma generated type.
* ~~The PR branch deployments doesn't get the `identities` key from Cognito, as that comes from IDP. This is the field that includes the user's EUA ID, which is something we want to store. This only exists in higher dev/val/prod environments, so I added a field to PR branches to mimic this (otherwise we don't have an EUA ID in PR, which we need for testing). I don't love this and am open to suggestions here.~~ We decided to drop storing the eua ID, as we don't immediately need it and it's complicating those code paths.
* We need to be able to get the `Store` in the cognito auth funcs, so there's a bit of code duplication to instantiate the store in the middleware. Also open to suggestions here.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2493


haworku:
@macrael Could you be one of the reviewers here since some of this is code you are familiar with? Was thinking maybe you would have opinions on the local dev stuff and `Store` handling as well.

I will also review!

services/app-api/src/authn/cognitoAuthn.ts
```diff
@@ -126,45 +144,153 @@ export function userTypeFromAttributes(attributes: {
...
+    }
+
+    if (user.role === 'STATE_USER') {
+        return ok({
+            role: 'STATE_USER',
+            email: user.email,
+            givenName: user.givenName,
+            familyName: user.familyName,
+            name: user.givenName + ' ' + user.familyName,
+            euaID: user.euaID,
```
haworku, line 187: Are we certain state users have an eua? I thought EUA was only something CMS users would have since they use EUA sso login when they access IDM. State users just go directly through IDM. 

Also, if its true that only CMS users have EUA, I start to wonder do we really even need to track EUA in our app ? What is the value of that?

services/app-api/src/testHelpers/storeHelpers.ts
```diff
@@ -69,6 +69,12 @@ function mockStoreThatErrors(): Store {
         getAllRevisions: async () => {
             return genericStoreError
         },
+        getUser: async (_ID) => {
```
haworku, line 72: `findUser` a per [design-patterns.md](https://github.com/CMSgov/managed-care-review/blob/main/docs/technical-design/design-patterns.md)

services/app-api/src/authn/cognitoAuthn.ts
```diff
@@ -126,45 +144,153 @@ export function userTypeFromAttributes(attributes: {
             role: 'CMS_USER',
             email: attributes.email,
             name: fullName,
+            givenName: attributes.given_name,
+            familyName: attributes.family_name,
+            euaID: euaID,
```
haworku, line 149: I think these two functions are handling the admin user differently. We have code on the prisma side that handles the admin but it looks like the cognito side doesn't have this and admin would fall in the error "unsupported user role" block.

haworku:
Question - @mojo Should I make a ticket around the user type issue? What would the description of that ticket be, is our goal to get ride of the app defined user type all together?

services/app-api/src/domain-models/UserType.d.ts
```diff
@@ -1,19 +1,34 @@
...
+type UserType = StateUserType | CMSUserType | AdminUserType
 
 type StateUserType = {
     role: 'STATE_USER'
     email: string
     name: string
     state_code: string
+    givenName: string
+    familyName: string
+    euaID: string
```
macrael, line 13: State Users definitely dont’ have an EUA ID. I think if we don’t have anything we’re planning to use the EUA id for rn and it’s complicating this work then it makes sense to me to drop it. I’m trying to think of why we would use someone’s EUA down the road, maybe we’d need it to communicate who did some action to a third party system?

haworku: Iagree, I think we can drop EUA. We could also use email address to communicate who did an action

mojotalantikite: euaID has been dropped from everywhere

mojotalantikite:
> 



> Question - @mojo Should I make a ticket around the user type issue? What would the description of that ticket be, is our goal to get ride of the app defined user type all together?

Good call, I've dropped a ticket into the epic for this -- thanks!: https://qmacbis.atlassian.net/browse/MR-2808

services/app-api/src/authn/cognitoAuthn.ts
haworku: fyi I made a ticket where this could be handled (adding full admin user workflow) https://github.com/CMSgov/managed-care-review/pull/1329

PR #1325, opened 2022-11-15
MR-2683: State indicates if rate cert is shared

JasonLin0991:
## Summary
[MR-2683](https://qmacbis.atlassian.net/browse/MR-2683)

- Proto/Domain Work 
   - Update `health_plan_form_data.proto` with new `rateInfo` field `packages_with_shared_rate_certs`.
   - Saved previous version of the proto in `app-proto/archive`
   - Add `packagesWithSharedRateCerts` field in `toProtoBuffer.ts`
   - Add `packagesWithSharedRateCerts` field `RateInfoType` in `UnlockedHealthPlanFormDataType.d.ts`
   - Add `packagesWithSharedRateCerts` in `unlockedHealthPlanFormDataSchema.ts`
   - Add `packagesWithSharedRateCerts` in `toDomain.ts`
   - Bump `CURRENT_PROTO_VERSION` to `4` in `toLatestVersion.ts`
   - Update `RateDetails.tsx` with new `packagesWithSharedRateCerts` field in `rateInfo`

- Feature flag work
   - Add new `Rates across submissions` feature flag in LD
   - Add new `rates-across-submissions` flag into `flags.ts`
   - Add feature flag to `RateDetails.tsx`

- Shared check box on Rate Details page
   - Add a checkbox, under document upload form, to identify file as shared
      - `sharedRateCert` field in `RateDetailsSchema.ts`, used to specify if rate certification is shared. This field is not saved to the health plan form data, instead it's an optional field used to toggle on work in [MR-2685](https://qmacbis.atlassian.net/browse/MR-2685)
   - Checkbox hidden behind `rates-across-submissions` flag.

#### Test cases covered

- Unit tests
   - `RateDetails.test.tsx`
      - `'correctly checks shared rate certification checkboxes on each rate certification'`
   - Update tests and mock data with `packagesWithSharedRateCerts` field

- Cypress
   - Add `cy.stubFeatureFlags()` to  `generateReport.spec.ts` before each test.


services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
```diff
@@ -648,6 +660,31 @@ export const RateDetails = ({
...
+                                                                    checked={
+                                                                        rateInfo.sharedRateCert
+                                                                    }
+                                                                    onChange={(
+                                                                        e
+                                                                    ) =>
+                                                                        setFieldValue(
+                                                                            e
+                                                                                .target
+                                                                                .name,
```
JasonLin0991, line 681: This checkbox is only used for toggling the UI work in [MR-2685](https://qmacbis.atlassian.net/browse/MR-2685), linking to other submissions. 

The thought here was to have a value in `formik` to track the checkbox on each rate cert and toggle on without modifying the new `packagesWithSharedRateCerts` field. Thew new `packagesWithSharedRateCerts` will be storing the linked submission ID's and saved to the health plan form data so I was trying to avoid using this field as a toggle, although it is still used to set the initial value of `sharedRateCert`.

I believe this means that [MR-2685](https://qmacbis.atlassian.net/browse/MR-2685) will be blocking [MR-2684](https://qmacbis.atlassian.net/browse/MR-2684).  I could saved `sharedRateCert` field into the health plan form data, but I thought it was an unnecessary and redundant field once MR-2685 was finished. Then we could use `packagesWithSharedRateCerts` to display the `Shared` tag on the summary and review/submit pages

Thoughts? If this is good, then I can update the sprint by switching out MR-268 with MR-2685?

rswerve: I think the checkbox makes sense, and it's fine to wait on 2684 until 2685 is finished.  Just to check my understanding:
The state of the checkbox determines whether we should see a "please select the shared submissions" input (2685), and once there are shared submissions, because they're available throughout the app with the form data, we can use the shared submissions array to determine whether we show a "shared" tag (2684).

JasonLin0991: 💯!

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
```diff
@@ -648,6 +660,31 @@ export const RateDetails = ({
                                                                 }
                                                             />
                                                         </FormGroup>
+                                                        {ratesAcrossSubmissions && (
+                                                            <FormGroup>
+                                                                <Checkbox
+                                                                    id={`sharedRateCheckBox-${index}`}
```
rswerve, line 666: Didn't we recently add an `id` to rateInfo?  We could use that here.

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
```diff
@@ -88,6 +89,8 @@ export interface RateInfoFormType {
     rateDocuments: SubmissionDocument[]
     actuaryContacts: ActuaryContact[]
     actuaryCommunicationPreference?: ActuaryCommunicationType
+    packagesWithSharedRateCerts: string[]
+    sharedRateCert: boolean
```
rswerve, line 93: I find it's helpful to name variables that hold booleans to something that starts with "is" or "has," so we immediately  know we're dealing with a boolean.  `hasSharedRateCert`?

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
JasonLin0991: We do! But not when adding a new rate certification since `id` is generate in `toProtoBuffer.ts`. I'm already generating a `uuid` `key` in rateInfo, so I'll use that!

rswerve (approved):
looks great!

PR #1329, opened 2022-11-17
Add retroactive ADR on developer testing accounts for state user

haworku:
## Summary

This is an already drafted ADR I found in GDrive that I think we meant to add to the repository. I've added the ADR and implemented the change it suggests (in terms of excluding American Samoa from requests). I used the date from the last edited time on the ADR.

Also cleaned up the naming of a couple of store methods based on the [design-patterns.md](https://github.com/CMSgov/managed-care-review/blob/main/docs/technical-design/design-patterns.md) suggestions.


JasonLin0991 (commented):
We removed `AS` from `findAllRevisions`. For acceptance testing, should we find another way to dump that csv data?

haworku:
@JasonLin0991 I figured if we needed to get AS data down the road,  we could always change things but by default our api requests should all be excluding that data. I didn't see a use case for AS data?  Are you saying you want AS in there for doing verification of something related to AS on prod, can def add it back to that request.

rswerve (approved):

PR #1333, opened 2022-11-18
Add error handling if LD provider fails

haworku:
## Summary
This is coming out of today's pairing session around Cypress. We should be logging any error that would crash the app even if its unlikely to happen 🌟

### DIFF
--- a/services/app-web/src/index.tsx
+++ b/services/app-web/src/index.tsx
@@ -134,7 +134,7 @@ if (ldClientId === undefined) {
     )
 })().catch((e) => {
     recordJSException(
-        'LDProvider failed, could not initialized application' + e
+        'LDProvider failed, could not initialize the application' + e
     )
     throw new Error('Could not initialize the application: ' + e)
 })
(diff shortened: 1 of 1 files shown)


services/app-web/src/index.tsx
```diff
@@ -132,6 +133,9 @@ if (ldClientId === undefined) {
         </React.StrictMode>
     )
 })().catch((e) => {
+    recordJSException(
+        'LDProvider failed, could not initialize the application' + e
+    )
     throw new Error('Could not initialize the application: ' + e)
```
JasonLin0991, line 139: Do we know if this only catches the LDProvider errors? If so this is good, otherwise a generic error like the thrown one maybe better.

services/app-web/src/index.tsx
```diff
@@ -132,6 +133,9 @@ if (ldClientId === undefined) {
         </React.StrictMode>
     )
 })().catch((e) => {
+    recordJSException(
```
JasonLin0991, line 136: Do have have access to `otel` at this point in the app?

haworku: nope we don't i was thinking that when I was coming back from PT today. We can't do this lol!

PR #1332, opened 2022-11-18
MR-2685: State links other submissions to a shared rate cert

JasonLin0991:
## Summary
[MR-2685](https://qmacbis.atlassian.net/browse/MR-2685)

States can link a rate certification to other  `CONTRACT_AND_RATES` submissions.

- Add `PackageSelect` component, tests, and story-book.
   - `PackageSelect` is similar to `ProgramSelect` with the addition of text for loading and error states of queries that feed it data.
- Refactor `ProgramSelect.tsx` component
   - Moved component to the `Select` directory where `PackageSelect` component lives.
   - Refactored style sheets to be common between select components.
   - Minor refactor to `ProgramSelect.test.tsx` and fixed not failing errors during testing.
- Add `PackageSelect` component to `RateDetails.tsx` for linking submissions
   - using `useIndexHealthPlanPackagesQuery` to fetch submission data for `PackageSelect`.
      - `skip` to toggle query on feature flag state.
      - using `onCompleted` option to parse data once query is finished.
      - logging error with `onError` option.
   - Update `Yup` schema for `packagesWithSharedRateCerts` errors.

#### Test cases covered

Unit tests
- `RateDetails.test.tsx`
   - Refactored where existing tests live in the `describe ` blocks
   - `'correctly checks shared rate certification checkboxes and selects shared packages'`
   - `'cannot continue when shared rate checkbox is checked and no package is selected'` 
- `ProgramSelect.test.tsx`
   - `'displays program options'`
   - `'can select and return programs in an array'`
   - `'can remove all selected programs'`

PR #1336, opened 2022-11-21
[MD-2685] Add label and link to related submissions PackageSelect on Rate Details page 

haworku:
## Summary
 - Add label and  "View all submissions" link to the PackageSelect'
 - Wrap the related fields in same Form Group - so error styles show for the entire block
- Export types for select options

#### Related issues
Add on to #1332 
https://qmacbis.atlassian.net/browse/MR-2685

### Screenshots
Here's how error styling will look now that both related fields are in same parent element. 
Screen Shot 2022-11-21 at 10 55 24 AM


rswerve:
Looks good.  I see the error message, and it clears when I select a submission.

rswerve (approved):

PR #1338, opened 2022-11-21
[MR-2587] Only return 5 packages in dropdown list. Also add re-useable utilities.

haworku:
#### Related issues
https://qmacbis.atlassian.net/browse/MR-2687

#### Test cases covered

I did not add tests for this specifically and relied on existing test.

 This could be a Cypress assertion (check in E2E tests when many submissions are present that only 5 are displayed in dropdown) but considering what is also going on cypress decided to let things be. 

## QA guidance
- Check that we only display contract and rate submission and of these 5 appear in list

### DIFF
--- a/services/app-graphql/src/queries/indexHealthPlanPackages.graphql
+++ b/services/app-graphql/src/queries/indexHealthPlanPackages.graphql
@@ -22,9 +22,13 @@ query indexHealthPlanPackages {
                         createdAt
                         unlockInfo {
                             updatedAt
+                            updatedBy
+                            updatedReason
                         }
                         submitInfo {
                             updatedAt
+                            updatedBy
+                            updatedReason
                         }
                         formDataProto
                     }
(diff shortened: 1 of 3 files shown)


services/app-graphql/src/queries/indexHealthPlanPackages.graphql
```diff
@@ -22,9 +22,13 @@ query indexHealthPlanPackages {
                         createdAt
                         unlockInfo {
                             updatedAt
+                            updatedBy
```
haworku, line 25: I made changes here to  be able to use the shared frontend utility `getCurrentRevisionFromHealthPlanPackage`. Want to avoid domain to base 64  calls in React code, we should be trying to isolate that error handling to one place.

haworku:
@rswerve This is a rates across subs PR that could merge tomorrow with feature flag I think. If you check things out and approve feel free to hit the merge button too while I am out.

haworku:
Closing in favor of  #1349

PR #1342, opened 2022-11-22
Remove Cypress checks for dev/val on promote to prod workflow

haworku:
## Summary
This came out of today's engineering sync - I took the AI to further simplifying our promote workflow and chance of flakes by removing Cypress from steps. 

Welcome reviews or adjustments to the PR, not sure how I feel about skipping val when I see it in code. Maybe we just skip dev. 

#### Related issues
Follow on to #1337 and #1341 after more discussion

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -142,49 +142,8 @@ jobs:
       react_app_auth_mode: IDM
       nr_license_key: ${{ secrets.NR_LICENSE_KEY }}
 
-<<<<<<< Updated upstream
-  dev-cypress:
-    name: dev - cypress
-    needs: [promote-app-dev]
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


rswerve:
> ## Summary
> This came out of today's engineering sync - I took the AI to further simplifying our promote workflow and chance of flakes by removing Cypress from steps.
> 
> Welcome reviews or adjustments to the PR, not sure how I feel about skipping val when I see it in code. Maybe we just skip dev.
> 
> #### Related issues
> Follow on to #1337 and #1341 after more discussion

I don't think we should think of this as a permanent fix, and I also think we have to skip both if we're really going to test the hypothesis that using the same spec file more than once is the issue.  Approving.

rswerve (approved):

haworku:
Ah right. Thank you @rswerve for helping me keep track of the thread ⭐

PR #1344, opened 2022-11-22
Sync statePrograms copies

rswerve:
## Summary

Just noticed that we made some updates to the statePrograms file in app-api.  But we also have a copy of that file in app-web 🤦 .  This just brings app-web up to date.

### DIFF
--- a/services/app-api/src/data/statePrograms.json
+++ b/services/app-api/src/data/statePrograms.json
@@ -1,1184 +1,1184 @@
 {
-  "states": [
-    {
-      "name": "Alabama",
-      "programs": [
+    "states": [
         {
-          "id": "276702fe-f835-46cd-9c97-e52940447f22",
-          "fullName": "Alabama Coordinated Health Network",
-          "name": "ACHN"
+            "name": "Alabama",
+            "programs": [
+                {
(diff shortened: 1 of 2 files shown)


haworku:
😵‍💫 I forgot about that. cc: @macrael 

It feels... very likely this could happen again. We should adjust the import-programs script to change the files in both places.

haworku (approved):
Hard to tell whats going on with the diff - but spot checked a few.

macrael:
Ohh! it’s indenting by 4 spaces instead of 2

macrael:
@rswerve can you re-save both files as 2-spaces indented for now and then we can merge this to make sure we don’t break anything. We can do a cleanup afterwards

rswerve:
> @rswerve can you re-save both files as 2-spaces indented for now and then we can merge this to make sure we don’t break anything? We can do a cleanup afterwards

This is being formatted by Prettier.  Do we want different indenting for json?  https://github.com/CMSgov/managed-care-review/blob/main/.prettierrc#L3

macrael:
No you are probably right, we can switch over to 4 spaces. Should be a fix we can make to the generator program to make it output 4 spaces as well and get on the program.

PR #1349, opened 2022-11-23
Rates across submissions (issue numbers in description)

rswerve:
MR-2587 replacement for #1338 
MR-2684
MR-2686

## Summary

Finishes up several of the rates across submissions tickets.  Incorporates @haworku's code for 2587.

We should now see a "SHARED" tag on documents that the user has told us are shared with other submissions.
We should see a "Linked submissions" column when such a document exists.
That column should contain the name of the other submission, and a link to it.

We should see the most recent 5 submissions in the dropdown in RateDetails where we can select the linked submission.

Includes a refactor of the `packagesWithSharedRateCerts` type to include both the name and the id of a package.  Since this is behind a feature flag and there are no prod submissions using the old type, this change doesn't require any migration work.

#### Screenshots

image

#### Test cases covered

New unit tests for when we show the shared rate info, and also for when we don't (when it's not the latest submission, and when there are no shared rates).

### DIFF
--- a/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
+++ b/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
@@ -14,12 +14,20 @@ import { usePreviousSubmission } from '../../../hooks/usePreviousSubmission'
 import styles from '../SubmissionSummarySection.module.scss'
 import {
     HealthPlanFormDataType,
+    packageName,
     RateInfoType,
 } from '../../../common-code/healthPlanFormDataType'
 import { Program } from '../../../gen/gqlClient'
 import { useLDClient } from 'launchdarkly-react-client-sdk'
+import { useIndexHealthPlanPackagesQuery } from '../../../gen/gqlClient'
+import { recordJSException } from '../../../otelHelpers'
 import { featureFlags } from '../../../common-code/featureFlags'
 import { Link } from '@trussworks/react-uswds'
+import { getCurrentRevisionFromHealthPlanPackage } from '../../../gqlHelpers'
+import { SharedRateCertDisplay } from '../../../common-code/healthPlanFormDataType/UnlockedHealthPlanFormDataType'
+
+type PackageName = string
+type RatePackageNamesLookup = { [id: string]: PackageName } // packages names keyed by their package id
 
 export type RateDetailsSummarySectionProps = {
     submission: HealthPlanFormDataType
@@ -38,6 +46,9 @@ export const RateDetailsSummarySection = ({
     submissionName,
     statePrograms,
 }: RateDetailsSummarySectionProps): React.ReactElement => {
+    const [ratePackageNamesLookup, setRatePackageNamesLookup] =
+        React.useState<RatePackageNamesLookup | null>(null)
+
     // Launch Darkly
     const ldClient = useLDClient()
     const showMultiRates = ldClient?.variation(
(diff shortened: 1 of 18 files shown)


haworku (approved):
Thanks for getting this all together and testing the "when not to show case". LGTM

services/app-web/src/components/SubmissionSummarySection/UploadedDocumentsTable/UploadedDocumentsTable.tsx
```diff
@@ -166,6 +183,18 @@ export const UploadedDocumentsTable = ({
...
                             <td>{documentCategory}</td>
+                            {showSharedInfo
+                                ? packagesWithSharedRateCerts &&
+                                  packagesWithSharedRateCerts.map((item) => (
+                                      <td>
+                                          <NavLink
+                                              to={`/submissions/${item.packageId}`}
+                                          >
+                                              {item.packageName}
+                                          </NavLink>
```
JasonLin0991, line 194: Users can link to rate certifications in draft submissions where the package name could change and submitted packages could be unlocked and changes could also affect the package the name.

The upside is not having to make the query for the linked submissions to get the name. The downside is users may get confused if the linked package names don't match what is on the dashboard and when visiting the submission. I think in this case it might be good to just do the query and get that data to get the most up-to-date package name for the linked submission. @haworku @macrael any comments on this? Maybe I'm overthinking it? 

Also if the user had visited the dashboard page. I think the queried data would be cached anyways and not make another api call for the data. That would cut down on the actual api calls.

JasonLin0991 (commented):
Just a quick question about storing the package name at the time of linking submissions.

services/app-web/src/components/SubmissionSummarySection/UploadedDocumentsTable/UploadedDocumentsTable.tsx
haworku: Hey, @JasonLin0991 I think you are talking about the ratename - this is about the package name (looks something like MCR-MN-PMAP). 

@rswerve and I looked at this together last week at this together last week and thought there were few options 1. adjust what we store in the proto 2. make another request on the review-submit page to the index api -  display this link and handle error states if things go wrong 3.  adjust the initial  fetch draft submission graphql query to also return all related packages. 

We chose #1 because this was intended to be a quick feature, we know these types of associations between postgres and proto are already a hack, and we didn't see any major drawbacks.

haworku: Ah, I see what you are saying, thanks for clarifying.

haworku: Addressed in [6376b81](https://github.com/CMSgov/managed-care-review/pull/1349/commits/6376b81379132a04b0f8a41b76e7971cf3fc4549)

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
```diff
@@ -46,8 +64,86 @@ export const RateDetailsSummarySection = ({
...
+            }
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            const [currentRevision, currentSubmissionData] =
+                currentRevisionPackageOrError
+
+            // Exclude active submission package and contract_only from list.
+            if (
+                currentSubmissionData.submissionType === 'CONTRACT_AND_RATES' &&
+                currentRevision.submitInfo?.updatedAt &&
+                pkg.id
```
JasonLin0991, line 119: Currently linked draft submissions will not update the package name when the linked draft submission changes its programs It will update the package name once the linked draft submission is submitted. Is this intended?

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
```diff
@@ -197,7 +199,8 @@ export const RateDetails = ({
                         })
                     }
                 })
-            setPackageOptions(packages)
+            const abbreviatedPackagesList = packages.slice(0, 5)
```
JasonLin0991, line 202: I'm seeing the earliest 5 submissions instead of the most recent 5. I wonder if we need to reverse then slice or should we do a sort at an earlier part of this and be confident we have the correct order?

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
```diff
@@ -163,26 +164,27 @@ export const RateDetails = ({
             data?.indexHealthPlanPackages.edges
                 .map((edge) => edge.node)
                 .forEach((sub) => {
-                    const currentRevision = sub.revisions[0]
```
JasonLin0991, line 166: We could sort by date of submitted and draft submissions after `.map((edge) => edge.node)`.

JasonLin0991 (changes requested):
A couple of things I found.

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
haworku: This is a great catch! Thank you - I cleaned this up by just making that lookup a lot simpler. [f2eda5b](https://github.com/CMSgov/managed-care-review/pull/1349/commits/f2eda5bb9c1a76a35d7fc6ffc044c07eb206e596)

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
haworku: Did a few things to address this
- Added the sorting by `updatedAt` quick fix [cf17705](https://github.com/CMSgov/managed-care-review/pull/1349/commits/cf1770534403f207c93ce65bbea6c6b06b008952). 
- Looked more into react-select when I realized the only way to limit the visible options _but still show all the options in the typeahead_ is not by limiting the options list like I was doing but adding more dynamic handling. As a result looked into AsyncSelect with defaultOptions https://react-select.com/async#defaultoptions but decided it was a lot of code (we end up having to re-implement the search logic rather than using built in behavior) for what is basically a styling motivated change.
-  Decided to go with shrinking the size of the dropdown instead and also ran that by design [37b1e50](https://github.com/CMSgov/managed-care-review/pull/1349/commits/37b1e5006ddba194ef0af4487ca8ad3dde1b3298) 
- I made sure that how we sort and pick the current revision is always the same [e0e422f](https://github.com/CMSgov/managed-care-review/pull/1349/commits/e0e422fedb4a57973ce112283d4307fd57225f00) Part of what made debugging hard is we have a few ways we were doing this.

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
JasonLin0991: Looks good! This does bring something up, **that we could address at a later time and not blocking for this PR**, linked submissions that changed from `CONTRACT_AND_RATES` to `CONTRACT_ONLY`. Would it even be a common occurrence? How or should we handle that?

JasonLin0991 (approved):
✅

PR #1350, opened 2022-11-29
Cleanup stateprograms.json usage

macrael:
## Summary

In our last update to the statePrograms.json I failed to update both instances of the file in our repo. This PR is to try and consolidate that down to a single file so that we don’t need to keep the two files synced anymore.


haworku (approved):
LGTM

do we have docs on how to generate stateprograms at all to update? maybe we should have a readme in  our scripts folder or a doc specifically about CLI scripts. Also looks like theres a line about state programs in the submisssion-diagram.md

PR #1351, opened 2022-11-29
Turn on Aurora before Cypress run

mojotalantikite:
## Summary

Jason had noticed that Cypress was encountering issues from time to time due to Aurora taking a while to start up. This adds a run of our migrator lambda in the `Finishing Prep`, which will be a no-op, but due to connecting to Aurora will cause it to turn back on if it is scaled down to off.

This only happens in PR branches.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2743


mojotalantikite:
I didn't notice the exact error that was encountered in the ticket that made us want to do this work, but now I'm wondering if it happens to be the same one we continue to see where the application can't be loaded at all (which I suspect is from geo-location restrictions and have an open ticket with GitHub about). 

@JasonLin0991 do you happen to remember what that error you were seeing was?

JasonLin0991:
> I didn't notice the exact error that was encountered in the ticket that made us want to do this work, but now I'm wondering if it happens to be the same one we continue to see where the application can't be loaded at all (which I suspect is from geo-location restrictions and have an open ticket with GitHub about).
> 
> @JasonLin0991 do you happen to remember what that error you were seeing was?

Yes! If you look in `/tests/cypress/integration/cmsWorkflow/unlockResubmit.spec.ts` there is a whole section of commented out code that tests multi-rate submissions, enabling those tests should start causing flaky tests. Here is the [PR](https://github.com/CMSgov/managed-care-review/pull/1254#issuecomment-1285626027) where I mentioned the issue.

JasonLin0991:
@mojotalantikite The test still fail with the same errors posted in this [PR Comment](https://github.com/CMSgov/managed-care-review/pull/1254#issuecomment-1285626027). So possibly not a cold start issue. I think it's still good to turn on Aurora before the cypress run though.

haworku (approved):
This looks good to me. 

@JasonLin0991 in the future I would have been v comfortable with you just approving this PR when you reviewed it. I think the idea came from you initially, was agreed upon by all of eng, so don't feel like there is someone else who needs to 👍  things.

PR #1361, opened 2022-12-05
Fix user return in Aurora code path

mojotalantikite:
## Summary

When testing out the newly merged Users table work in #1321 , the shared CMS dev user seems to be having issues with inserting into Aurora. This fix should get that user back up and running while we investigate.


macrael (approved):
I don’t understand why there are so many commits here for a hot fix but I think it’s alright

mojotalantikite:
Closed for #1362 , the branch on this one wasn't caught up to `main` I guess.

PR #1362, opened 2022-12-05
Fix cognitoAuth with Aurora inserts

mojotalantikite:
## Summary

When testing out the newly merged Users table work in #1321 , the shared CMS dev user seems to be having issues with inserting into Aurora. This is due to a difference in attributes for a state vs CMS user. 

This adds a default value to the insert for a CMS user (state is empty), and will return the user from cognito if there is a failure in this code path (so as to not bring down the application fully).

### DIFF
--- a/services/app-api/src/postgres/insertUser.ts
+++ b/services/app-api/src/postgres/insertUser.ts
@@ -23,7 +23,7 @@ export async function insertUser(
                 familyName: user.familyName,
                 email: user.email,
                 role: user.role,
-                stateCode: user.stateCode,
+                stateCode: user.stateCode ?? '',
             },
         })
         console.log('insert user return: ' + val)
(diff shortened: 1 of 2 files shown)


services/app-api/src/postgres/insertUser.ts
```diff
@@ -23,7 +23,7 @@ export async function insertUser(
                 familyName: user.familyName,
                 email: user.email,
                 role: user.role,
-                stateCode: user.stateCode,
+                stateCode: user.stateCode ?? '',
```
macrael, line 26: And when we pull a CMS user out of the table, we ignore the stateCode field? Is this field required at the Prisma layer?

macrael (approved):

services/app-api/src/postgres/insertUser.ts
mojotalantikite: stateCode is an optional in the Prisma schema

mojotalantikite: actually it looks like Prisma uses `null` when it has an optional, fixing this

mojotalantikite:
Closing for #1363 -- we can't use 'cognito' as it is a reserved string in AWS.

PR #1367, opened 2022-12-06
Add graphql resolvers/mutations to User

mojotalantikite:
## Summary

This adds the update and delete mutations to assign/delete a `State` from a `User`. Since we don't have an admin panel built out these have no front-end components at this time (e.g. to assign a `State` to a `User`).

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2494


services/app-api/src/postgres/postgresStore.ts
```diff
@@ -67,6 +69,16 @@ type Store = {
     findUser: (id: string) => Promise<User | StoreError>
 
     insertUser: (user: InsertUserArgsType) => Promise<User | StoreError>
+
+    updateUserAssignedState: (
+        userID: string,
+        state: State
```
macrael, line 75: I’m hesitant for us to have the Prisma types in the type signatures here. The idea behind our doman models is that they are the types used for calls between modules, which lets us separate our implementation at the prisma layer from everywhere else we use them. When you pick this up can we pair a little and work out how we want to handle this? I appreciate that ts is duck typed so for the easy case where the types line up we don’t have to manually convert even if we do have to duplicate the types.

services/app-api/src/authn/cognitoAuthn.ts
```diff
@@ -122,62 +119,24 @@ export function userTypeFromAttributes(attributes: {
...
 
     if (roles.includes(CMS_ROLE_ATTRIBUTE)) {
         return ok({
             role: 'CMS_USER',
             email: attributes.email,
-            name: fullName,
             givenName: attributes.given_name,
             familyName: attributes.family_name,
-        })
+        } as CMSUserType)
```
macrael, line 134: i’m mildly surprised this is neccecary since having the role specified should force it to be the right type

services/app-api/src/postgres/findUser.ts
```diff
@@ -1,17 +1,33 @@
...
         const findResult = await client.user.findUnique({
             where: {
                 id: id,
             },
         })
-        return findResult ?? ({} as User)
+
+        switch (findResult?.role) {
+            case 'ADMIN_USER':
+                return findResult as AdminUserType
```
macrael, line 23: it would be nice if we didn’t have to use as here, and since this is repeated code in each of the handlers maybe we can have a more strict converter.

macrael (approved):
looks great, funny to see my early, out of js prac under bar variable names. I dunno if there’s a slicker way to convert between types so I dont’ think that needs to hold us up but I noticed it.

services/app-api/src/domain-models/UserType.d.ts
```diff
@@ -1,29 +1,26 @@
...
-    state_code: string
+    stateCode: string
     givenName: string
     familyName: string
 }
 
 type CMSUserType = {
+    id?: string
     role: 'CMS_USER'
     email: string
```
haworku, line 15: 🤔 @macrael  Should adding `states` CMSUserType be an add on ticket after this work?  I get that it would require changing a bunch of stuff in tests so seems out of scope here. Does seem like kind of blocker to us actually using the states that are going to be surfaced on the new prisma user type  in the future though...

macrael: I ended up doing this. I went ahead and added resovlers to call these apis and that required adding stateAssignments so I could actually get that data out of the API

services/app-api/src/authn/cognitoAuthn.ts
```diff
@@ -227,26 +187,19 @@ export async function userFromCognitoAuthProvider(
...
-            if (isStoreError(result)) {
-                console.error(
-                    `Could not insert user: ${JSON.stringify(result)}`
-                )
-                return cognitoUserResult
-            }
-            return userTypeFromUser(result)
-        } catch (e) {
-            console.error(`Could not insert user: ${JSON.stringify(e)}`)
+        const result = await store.insertUser(userToInsert)
```
haworku, line 193: 🔴 I think we need the `try...catch` still here right-- and then inside the catch use `convertPrismaErrorToStoreError`.  Also we should log to OTEL.

I don't think isStoreError covers things that seems like something different peeking at the code

services/app-api/src/domain-models/StateCodeType.ts
```diff
@@ -0,0 +1,64 @@
...
+    'UT',
+    'VA',
+    'VT',
+    'WA',
+    'WI',
+    'WV',
+    'WY',
+] as const
+
+type StateCodeType = typeof StateCodes[number] // iterable union type
```
haworku, line 56: Nice. Thee frontend could use this too. There is a StateCodeType there as well  used for displaying the icon

services/app-api/src/testHelpers/gqlAssertions.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    if (res.errors.length > 1) {
+        console.error('Got Multiple Errors: ', res.errors)
+        throw new Error('response returned multiple errors')
+    }
+
+    return res.errors[0]
+}
+
+// assertAnErrorCode returns the error code from the only error's extensions
```
haworku, line 18:
```suggestion
// assertAnErrorExtensions returns the error code from the only error's extensions
```

haworku (approved):
Code looks good. Was interesting to see how the users work looks with the third user and also with more error handling learned a lot reading this. Let me know if there's anything else to check in terms of "verifying" things.

PR #1370, opened 2022-12-08
Fix SES email sending

mojotalantikite:
## Summary

This fixes a bug where sending email notifications through SES was throwing an error. This likely cropped up due to moving from the AWS v2 SDK to the newer v3 SDK and a change on how lambdas are invoked.

Instead of invoking a lambda handler to do the SES sending, this just uses SES on it's own.

### Related Issues
https://qmacbis.atlassian.net/browse/MR-2867

### DIFF
--- a/services/app-api/serverless.yml
+++ b/services/app-api/serverless.yml
@@ -145,6 +145,8 @@ layers:
 functions:
   email_submit:
     handler: src/handlers/email_submit.main
+    layers:
+      - arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-nodejs-amd64-ver-1-2-0:1
 
   health:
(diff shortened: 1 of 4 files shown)


macrael (approved):
This seems like a good fix. I’d be curious what actually changed about lambda invoking then

mojotalantikite:
> I’d be curious what actually changed about lambda invoking then

It seemed like it was around the [invocation params](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-lambda/interfaces/invocationrequest.html), since the lambda handler wasn't even showing that it was being run at all (I wasn't seeing any runs of the email lambda in the aws console). As I was starting to fix it I started questioning why were even doing that in the first place.

PR #1374, opened 2022-12-09
CMS Dashboard filter Epic

JasonLin0991:
## Summary
[Filter CMS Dashboard](https://qmacbis.atlassian.net/browse/MR-2468)

This PR covers the entire `Filter CMS Dashboard`. Most of the work was in [MR-2618](https://qmacbis.atlassian.net/browse/MR-2618), so the rest of the tickets in that epic were fairly easy to implement.

[MR-2618](https://qmacbis.atlassian.net/browse/MR-2618) CMS filters dashboard by state
- When I am logged in as a CMS user then I see a filter for “State/Territory” with a dropdown that contains all states, DC, and Puerto Rico. 
- When I expand drop down then I can select multiple states/territories from dropdown
and the list of submissions dynamically updates to show submissions made for states that I selected 
- When I make a selection then the state filters I have selected appear as chips in the state filter field 

[MR-2619](https://qmacbis.atlassian.net/browse/MR-2619) CMS filters dashboard by submission type
- When I am logged in as a CMS user then I see a filter for “submission type” with a dropdown that contains the following options: 
   1. Contract Only
   2. Contract and Rates 
- When I expand drop down then I can select one or more submission types in the list 
- When I make a selection then the submission type filters I have selected appear as chips in the submission type filter field

[MR-2620](https://qmacbis.atlassian.net/browse/MR-2620) CMS can clear all filters
- If I am on the CMS dashboard and I select “clear all” then any selected filters are cleared and all submissions appear

[MR-2710](https://qmacbis.atlassian.net/browse/MR-2710) CMS sees no results message when no submissions match filter selection
- When a user selects a set of filters for which there are no results  then they see a message that indicates there are no results for the selected filters and a prompt to clear their filters, per the design 

[MR-2711](https://qmacbis.atlassian.net/browse/MR-2711) CMS can see a count of how many filters have been applied
- when I am viewing the submission dashboard  I want to quickly see at a glance how many filters have been selected so that I know I am looking at a sub-set of state submissions.

[MR-2712](https://qmacbis.atlassian.net/browse/MR-2712) CMS can expand/collapse filter accordion 
- When I navigate to the dashboard then all the filters are collapsed by default within an accordion 
- When I expand out the accordion then I can see all the filter options

#### Screenshots

**NOTE:** There is a design change for the `Clear filters` button. Design agreed to move the button to inside the accordion instead of in the header. This is because the `Accordion` component from React USWDS is a button, so it would cause the accordion to close when clicking the `Clear filters` button and cause accessibility issues.

#### Test cases covered

- ➕ `FIlterAccordion.test.tsx`
   - 'renders filter accordion and expected filters'
   - 'clears all filters when Clear filters button is clicked'
- ➕ `FilterSelect.test.tsx`
   - 'displays filter options'
   - 'displays selected options'
- ➕ `HealthPlanPackageTable`
   - 'renders table with expected number of submissions'
   - 'displays no submission text when no submitted packages exist'
   - 'displays submissions table with expected headers for cms users'
   - 'does not display State and Submission type columns for state users'
   - 'displays submissions table sorted by that revisions last updated column'
   - 'has correct submission links for cms users'
   - 'has correct submission links for state users'
   - 'displays the expected program tags for current revision that is submitted/resubmitted'
   - 'should display filters on dashboard page when showFilters is true'
   - 'should not display filters on dashboard page when showFilters is false'
   - 'can filter table by submission state'
   - 'can filter by state and submission type'
   - 'should clear all filters when clear filter button is clicked'
   - 'displays no results found when filters return no results'

- `CMSDashboard.test.tsx`
   - ➖  'displays no submission text when no submitted packages exist'
   - ➖  'displays submissions table when submitted packages exist'
   - ➖  'displays submissions table with expected headers'
   - ➖  'displays submissions table sorted by that revisions last updated column'
   - ➖  'displays the expected program tags for current revision that is submitted/resubmitted'
   - ➕  'should display filters on cms dashboard page when feature flag is on'

- `StateDashboard.test.tsx`
   - ➕ 'should not display filters on state dashboard page'


services/app-web/src/components/FilterAccordion/FilterAccordion.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+        expect(
+            screen.getByRole('button', { name: 'Clear filters' })
+        ).toBeInTheDocument()
+    })
+    it('clears all filters when Clear filters button is clicked', async () => {
+        renderWithProviders(
+            <FilterAccordion
+                filterTitle={'Testing filter accordion'}
+                children={filters}
+            />,
```
JasonLin0991, line 96: Here I'm passing an array of filter components in the `children` prop, instead of nesting inside the component. The `renderWithProviders` does not like it when I nest the `FilterSelect` inside the `FilterAccordion`. It has to do with the way I'm using `React.cloneElement` in `FilterAccordion` to pass in `toggleClearFilter` prop to its children `FilterSelect`.

services/app-web/src/components/FilterAccordion/FilterAccordion.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    const handleClearFilters = () => {
+        setToggleClearFilter(!toggleClearFilter)
+        if (onClearFilters) onClearFilters()
+    }
+
+    const childrenWithToggleProps = React.Children.map(children, (child) => {
+        return React.cloneElement(child as React.ReactElement, {
+            toggleClearFilter,
+        })
```
JasonLin0991, line 26: I am controlling the `Clear filters` button in this component instead of the parent component. `toggleClearFilter` state is what updates the children component by passing it in using `React.cloneElement`. This way the entire `FilterAccordion` and `FilterSelect` components are not reliant on the parent and reusable outside of the `HealthPlanPackageTable` component.

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
```diff
@@ -0,0 +1,331 @@
...
+    })
+
+    const stateColumn = reactTable.getColumn('stateName')
+    const submissionTypeColumn = reactTable.getColumn('submissionType')
+
+    // Filter options based on table data instead of static list of options.
+    // const stateFilterOptions: FilterOptionType[] = Array.from(stateColumn.getFacetedUniqueValues().keys()).map(state => ({
+    //     value: state,
+    //     label: state
+    // }))
```
JasonLin0991, line 218: I still need to touch base with design about how we want to display the options. The commented out code here will only display unique options contained in the tableData instead of a static list.

JasonLin0991:
I'm not sure if we needed cypress tests for this one. Considering it's only front end logic and most of the functionality is tested in unit tests.

services/app-web/src/components/FilterAccordion/FilterAccordion.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState } from 'react'
+import styles from './FilterAccordion.module.scss'
+import { Accordion, Button } from '@trussworks/react-uswds'
+
+export type FilterAccordionPropType = {
+    onClearFilters?: () => void
+    filterTitle: string
+    children: React.ReactElement[]
```
haworku, line 8: 🔍 suggestion: related to ensuring types are descriptive to intended use case for this component
- `onClearFilters` should be required or the button display should be conditional based on presence of this prop. This change would avoid unexpected UX on click.
- `children` seem to be `FilterSelect[]`. The component assumes children are "FilterSelect" elements that would accept  a `toggleClearFilter` prop, not a generic react element of any type.

services/app-web/src/components/FilterAccordion/FilterSelect/FilterSelect.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+        )
+
+        //With dropdown closed expect only selected options to be on the page
+        await waitFor(() => {
+            expect(screen.queryByText('Apple')).toBeInTheDocument()
+            expect(screen.queryByText('Grape')).toBeInTheDocument()
+            expect(screen.queryByText('Orange')).not.toBeInTheDocument()
+            expect(screen.queryByText('Banana')).not.toBeInTheDocument()
+        })
+    })
```
haworku, line 102: todo: Since this is intended to be reuseable component I would like to see unit tests for other props doing work here
- `toggleClearFilter`
- `label`

services/app-web/src/components/FilterAccordion/FilterSelect/FilterSelect.tsx
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }, [toggleClearFilter, prevToggleClearFilter, selectInputRef])
+
+    return (
+        <div className={styles.filter} data-testid={`${name}-filter`}>
+            {label && <label htmlFor={`${name}-filterSelect`}>{label}</label>}
+            <Select
+                ref={selectInputRef}
+                options={filterOptions}
+                loadingMessage={() => 'Loading submissions...'}
```
haworku, line 75: todo: `loadingMessage` should be surfaced as a prop or removed from the component since its more for an async load anyway.

services/app-web/src/components/FilterAccordion/FilterSelect/FilterSelect.tsx
```diff
@@ -0,0 +1,95 @@
...
+        focused,
+        isDisabled,
+    }): string => {
+        return `You are currently focused on option ${focused.label}${
+            isDisabled ? ', disabled' : ''
+        }`
+    }
+
+    //This component just wraps the menu list inside a div, so we can set a data-testid and run jest tests.
+    const MenuList = (props: MenuListProps<FilterOptionType>) => {
```
haworku, line 50: nitpick: This could be declared outside the component to avoid unnecessary re-rendering. I only know this cuz we ran into some issues with cypress that were caused by sub components declared inside a component being re-rendered unnecessarily so now I make an effort to declare outside.

haworku (commented):
Here's an initial review from reading code FilterAccordion and FilterSelect. Will continue with rest of PR and also double checking things in the deployed app on Monday. So far 👍

services/app-web/src/components/FilterAccordion/FilterAccordion.tsx
JasonLin0991: The intent of `onClearFilters` was surface a function that would get triggered when the `Clear filters` button is clicked. For instance if `HealthPlanPackageTable` had a function that is called when filters are cleared, you can pass that function into `FilterAccordion` to be called when the `Clear filter` button is called. The idea was to allow `FilterAccordion` component to clear its filters without having to write logic in its parent component. As for displaying the button, we can put another prop to have it show or hide.

haworku: @JasonLin0991 Gotcha, I think I understand the intent. My comment is trying to address that display of the associated button seems dependent on that prop. We don't want unexpected UX where a button is visible but nothing happens when you click it (which could happen right now).

 To address that I would either make `onClearFilters` required  or make the button UI dependant on that prop *`onFlearFilters && <Button /> `  (probably the second one, for most flexibility). I don't think any additional props are needed...

services/app-web/src/pages/CMSDashboard/CMSDashboard.tsx
```diff
@@ -1,75 +1,40 @@
...
 import styles from '../StateDashboard/StateDashboard.module.scss'
 import { recordJSException } from '../../otelHelpers/tracingHelper'
 import {
     handleApolloError,
     isLikelyUserAuthError,
 } from '../../gqlHelpers/apolloErrors'
-import { ErrorAlertFailedRequest, ErrorAlertSignIn } from '../../components'
-
-// We only pull a subset of data out of the submission and revisions for display in Dashboard
-// Depending on submission status, CMS users look at data from current or previous revision
```
haworku, line 30: todo: Please preserve this comment ^ somewhere. 

How `submittedAt` and `updatedAt` are calculated is different between cms and state user (since cms can't see unlocked submissions being edited or draft submissions). Want to make sure this is noted.

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
```diff
@@ -0,0 +1,333 @@
...
+        [styles.submittedTag]: isSubmitted(status),
+        [styles.draftTag]: status === 'DRAFT',
+        [styles.unlockedTag]: status === 'UNLOCKED',
+    })
+
+    const statusText = isSubmitted(status)
+        ? SubmissionStatusRecord.SUBMITTED
+        : SubmissionStatusRecord[status]
+
+    return <Tag className={tagStyles}>{statusText}</Tag>
```
haworku, line 82: todo: Use the new shared `<InfoTag />` component. This will also change the color slightly of Unlocked which is intended, the color it is right now is off from what is in Figma. The program tag and the unlocked tag are supposed to be same color

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.test.tsx
```diff
@@ -0,0 +1,647 @@
...
+
+        await waitFor(() => {
+            expect(
+                screen.getByText(/You have no submissions yet/)
+            ).toBeInTheDocument()
+            expect(screen.queryByRole('table')).toBeNull()
+        })
+    })
+
+    it('displays submissions table with expected headers for cms users', async () => {
```
haworku, line 130: suggestion: Still separate out the cms and state user specific behavior is `describe` blocks. Right now things are intermingled its a bit hard to code and review as well as read/review tests and get a mental model of the expected behavior.

haworku (commented):
Take another look at [MR-2711](https://qmacbis.atlassian.net/browse/MR-2711), otherwise things are looking pretty good.

services/app-web/src/components/FilterAccordion/FilterAccordion.tsx
haworku: We paired on this in person, ✅

services/app-web/src/pages/CMSDashboard/CMSDashboard.tsx
```diff
@@ -97,10 +62,16 @@ export const CMSDashboard = (): React.ReactElement => {
         }
     }
 
-    if (loginStatus === 'LOADING' || !loggedInUser || loading || !data) {
+    if (
+        loginStatus === 'LOADING' ||
+        !loggedInUser ||
+        loading ||
+        !data ||
+        !loggedInUser.__typename
```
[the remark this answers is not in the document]
JasonLin0991: Type error for passing user `__typename` into `HealthPlanPackageTable` because `__typename` could could be `undefined`. Now that we are passing the entire user in, this isn't needed anymore.

services/app-web/src/components/FilterAccordion/FilterSelect/FilterSelect.test.tsx
JasonLin0991: Added some test for `label`. As for the `toggleClearFilter` it's being tested in `FilterAccordion.test.tsx` and `HealthPlanPackage.test.tsx`, but I was having issues figuring out how to test this toggle in the stand alone component. Particularly how to update the prop without rendering and clearing out the selection, as currently re-renders would clear the selection out. But this component is mean to be only used with `FilterAccordion`.

services/app-web/src/components/FilterAccordion/FilterAccordion.tsx
JasonLin0991: Recap our zoom convo here. The `Clear filters` button is functional independently from the `onClearFilters` prop. The `onClearFilters` prop is  surfaced to the parent incase there needs to be a side effect triggered when clicking on the `Clear filters` button. 

If we wanted we could hide the button when there are no children in the component. Although the `FilterAccordion` should not be an empty accordion.

haworku (approved):

PR #1375, opened 2022-12-09
Add Cypress monitor hack for Chrome in Linux

mojotalantikite:
## Summary
 Some of our Cypress runs were throwing an error like this:
```
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
```

After some searching I [found a fix](https://repo1.dso.mil/platform-one/big-bang/apps/library-charts/gluon/-/issues/7) that creates a virtual monitor so that Cypress will run there, as there is an issue with Chrome on Linux that this works around.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -343,6 +343,13 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


macrael (approved):
good hack

PR #1382, opened 2022-12-13
Fix documents not showing up on review and summary pages when running app locally.

JasonLin0991:
## Summary
Documents were not showing up on the `review and submit` and `summary` pages when running the app locally. This was because `newLocalS3Client()` in `s3Local.ts` needed to set a `region` in the `S3Client`


rswerve (approved):

haworku (approved):

services/app-web/src/s3/s3Local.ts
```diff
@@ -22,6 +22,7 @@ export function newLocalS3Client(
             secretAccessKey: 'S3RVER', // pragma: allowlist secret; pre-set by serverless-s3-offline
         },
         endpoint: endpoint,
+        region: 'us-east',
```
rswerve, line 25: I approved this, but as I'm debugging the snapshot issue, I see that we actually use an environment variable for this value.  `process.env.REACT_APP_S3_REGION`.  We should switch to that here, too, I think.

rswerve: In fact, now that I look some more 🙂, I think there may be some other issue going on here, because we shouldn't need (I think) a region locally.  See the comment here: https://github.com/CMSgov/managed-care-review/blob/main/services/app-web/src/index.tsx#L68

rswerve (changes requested):

services/app-web/src/s3/s3Local.ts
rswerve: Let's chat about this with @haworku in the morning.  As you discovered, the client requires a region be set (but you'll find it works locally no matter what you set it to, as long as it's not undefined) but it seems strange to be setting a bogus region for a local client.

haworku: If this still isn't resolved can we chat about it in informal eng sync? I think other eng too would be good to have in on the discussion since its also conceptually about how we handle local dev versus deployed env with our AWS clients in the code.

macrael (approved):

services/app-web/src/s3/s3Local.ts
JasonLin0991: @maz Could you can give this a quick approval, thanks!

rswerve (approved):

PR #1385, opened 2022-12-14
Fix OTEL endpoint requests from React

mojotalantikite:
## Summary

While working on the Cypress issues I noticed that requests from React to the OTEL endpoint were getting 403s. I'm not sure when this started, but switching to use the application endpoint with the authorizer instead of the api gateway endpoint clears those up it seems. This was working fine in the past, so I'm not sure what changed with the API gateway to make this start failing all of a sudden.

### DIFF
--- a/services/app-api/serverless.yml
+++ b/services/app-api/serverless.yml
@@ -189,6 +189,7 @@ functions:
               - x-amzn-trace-id
               - b3
               - traceparent
+          authorizer: aws_iam
 
   graphql:
     handler: src/handlers/apollo_gql.graphqlHandler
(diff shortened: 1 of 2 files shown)


haworku (approved):
This might be impossible  cuz its an issues with OTEL but is there a way to add alerting when the otel request is failing heheh. Or like heath check our collector urls? Or do something in AWS to get notified about 403s?

mojotalantikite:
> This might be impossible cuz its an issues with OTEL but is there a way to add alerting when the otel request is failing heheh. Or like heath check our collector urls? Or do something in AWS to get notified about 403s?

Oh yeah, we should set something up to alert if this is happening. Since this is a separate endpoint which is analytics only, we could probably just alert on any 403 to it, as it's not an endpoint that would have legit 403s from users trying to use the app. I made a ticket for it: https://qmacbis.atlassian.net/browse/MR-2927

PR #1386, opened 2022-12-15
Update Cypress CI components

mojotalantikite:
## Summary

Cypress was having issues with GPU in Linux on our containers, and this PR does a few things:
* Upgrades the Cypress container for a newer Chrome
* Pins the Linux version we run in
* Is explicit about `testing-library` versions
* Matches the chromedriver between local and CI
* Sets an autogenerated build ID for Cypress runs, which is recommended by Cypress for matrix runs in CI. This should fix the issues we've been seeing in `promote` where the Cypress run has already been marked as complete.
 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2897


macrael (approved):
whack each mole

haworku (approved):
Thx for the PR description, that helped.

PR #1388, opened 2022-12-16
Add ability to control IP allowlist on WAF

mojotalantikite:
## Summary

Since we were seeing issues with IP addresses being geo-ip'd inconsistently, and therefore traffic blocked by the CMS WAF geo-restriction rules, this add the ability to add IPs to an allowlist rule on the API gateway. This will likely be used mainly for allowing IP blocks from GitHub Actions access to `dev`. In order to accomplish this, we have to drop the WAF plugin and use the rules directly in our CloudFormation configs.

Open question: do we want this to merge into the higher environments, or do we want this to be restricted only to the dev environment (do we see the need to add IPs to an allowlist in higher environments)?

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2874

### DIFF
--- a/services/app-api/serverless.yml
+++ b/services/app-api/serverless.yml
@@ -281,6 +281,18 @@ resources:
       Properties:
         ResourceArn: arn:aws:apigateway:${self:provider.region}::/restapis/${self:custom.appApiGatewayId}/stages/${sls:stage}
         WebACLArn: ${self:custom.appApiWafAclArn}
+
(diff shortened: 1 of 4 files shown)


mojotalantikite:
after talking to @macrael I think we're gonna just let this sit and see if the geoip stuff crops back up (haven't seen it in a while tbh) now that we have a bunch of other Cypress fixes landed. We can reopen and merge if we see it again.

PR #1391, opened 2022-12-19
[MR-2931] Fix email unlocked submission link

macrael:
## Summary

Simple rendering bug, we were omitting an “/edit” in the link in our email to states after unlock. I changed all the urls in emails to be generated based on our routes in the FE so if we change those in the future the emails will correctly update. 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2931

#### Test cases covered

I added a test case for this one editable link we share out, the snapshot tests look correct for the other urls which are all to the submission summary page instead of review and submit.

## QA guidance

Unlock a sumbission and make sure the email sent to to the state has a working link in it.


macrael:
This appears to have broken a bunch of FE tests, i think b/c somehow I’ve imported a different version of react-router that’s being used there and making errors? I’m trying to figure it out.

macrael:
ok i think I got it working. Some weirdness in cross-version compat.

rswerve (approved):
Logged in, submitted, got the email, it linked correctly back to the submission.

PR #1394, opened 2022-12-21
MR-308 Move contract type buttons to submission type page

rswerve:
## Summary

To support future branching workflows, we need users to select a contract type earlier in the process.  This PR moves them to the first page of user responses.

In addition, there was a big refactor of the proto snapshot tests, which simply don't work properly when they include async calls.  I rewrote them as regular assertions (which also didn't work in a loop, so there's a lot of duplication now).  Even these will probably need to be refactored, when we fully understand what's going on with our saved protos, which brings us to...

...the saved protos are odd, because some of them were no different than their predecessor (I've deleted those), which shouldn't really happen.

#### Screenshots

image

#### Test cases covered

Wrote new tests for the submission type page, and refactored cypress shared commands to handle the new flow.


JasonLin0991 (changes requested):
`Contract action type` should be move to the `SubmissionTypeSummary` section. 


Just a reminder, on the summary pages, the `Submitted` details is on its own row which is intended. Incase you saw it and wondered if this was intentional.

services/app-web/package.json
```diff
@@ -13,7 +13,7 @@
         "clean": "rm -rf node_modules && yarn cache clean",
         "format": "prettier --write src/**/*.ts{,x}",
         "lint": "tsc --noEmit && eslint src/ --max-warnings=0",
-        "test:once": "react-scripts test --watchAll=false --runInBand --ci",
+        "test:once": "TZ=UTC react-scripts test --watchAll=false --runInBand --ci",
```
macrael, line 16: oof that must be a fun one

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetailsSchema.ts
```diff
@@ -1,82 +1,83 @@
...
+                return Boolean(value && value.length > 0)
+            }
+        ),
+        federalAuthorities: Yup.array().test(
+            'authoritySelection',
+            'You must select at least one authority',
+            (value) => {
+                return Boolean(value && value.length > 0)
+            }
+        ),
```
macrael, line 65: There might be a way to put all of these in a conditional inside this yup object instead of making the yesNoError take a parameter

macrael (approved):
nice work

rswerve:
> `Contract action type` should be move to the `SubmissionTypeSummary` section. 
> 
> Just a reminder, on the summary pages, the `Submitted` details is on its own row which is intended. Incase you saw it and wondered if this was intentional. 

Good call.  Moved it over 🙂

rswerve:
> `Contract action type` should be move to the `SubmissionTypeSummary` section. 

oops, forgot to close the comment in the other reply

PR #1396, opened 2022-12-22
Initial documentation of graphql.schema

macrael:
## Summary
This PR includes doc strings in graphql.schema for all of our GraphQL operations. 

I’ve included a list of all errors that a user can trigger, but did not list the known unexpected server errors. We need a way to view this documentation outside of the schema, next. 

Documentation is included as string literals inside the schema file, which is the standardized way of documenting GraphQL.

#### Related issues

https://qmacbis.atlassian.net/browse/MR-2849


services/app-graphql/src/schema.graphql
```diff
@@ -4,36 +4,139 @@ schema {
...
+    CMSUsers can find all the packages that do not have the DRAFT status
+    """
     indexHealthPlanPackages: IndexHealthPlanPackagesPayload!
 }
 
 type Mutation {
+    """
+    createHealthPlanPackage creates a new HealthPlanPackage.
+
+    The new HealthPlanPackage is created with a single revision and with
```
rswerve, line 42: I think someone new to the project would wonder what a "revision" is in this context.

services/app-graphql/src/schema.graphql
```diff
@@ -4,36 +4,139 @@ schema {
...
+    - id
+    - stateCode
+    - stateNumber
+    - createdAt
+    - updatedAt
+
+    Errors:
+    - ForbiddenError:
+        - A non-state user called this
+        - A state user from a different state called this.
```
rswerve, line 74: What can a CMS user do, or not?

rswerve: You mention it in a previous paragraph; just wondering if we want to be comprehensive in the Errors subsection.

services/app-graphql/src/schema.graphql
```diff
@@ -73,71 +178,156 @@ type UnlockHealthPlanPackagePayload {
...
+    "A nickname for the program"
     name: String!
+    "The full name for the program"
     fullName: String!
 }
 
+
+"""
+HealthPlanPackageStatus tracks the editing vs. submitted status of the package.
+It is not intended to track the overal status of the package through the review process
```
rswerve, line 241:
```suggestion
It is not intended to track the overall status of the package through the review process
```

services/app-graphql/src/schema.graphql
```diff
@@ -73,71 +178,156 @@ type UnlockHealthPlanPackagePayload {
...
+as that is fleshed out.
+
+State Machine:
+```
+DRAFT -> SUBMITTED -> UNLOCKED -> RESUBMITTED
+                         ^               |
+                         |---------------|
+```
+
+This status is used to determine wether the package is currently being edited by a state user
```
rswerve, line 251:
```suggestion
This status is used to determine whether the package is currently being edited by a state user
```

rswerve (approved):
I completely love this.  There are some repeating data paths that feel like they could be captured in a chart or state machine, but I think getting the info down is the key step.

🎉

services/app-graphql/src/schema.graphql
macrael: I’ll change non-state user to “CMSUser” since that’s the only other type of user we have right now and is more clear, does that cover it?

haworku (approved):

PR #1402, opened 2022-12-22
Add GitHub OIDC

ben-harvey:
## Summary

This change updates the `deploy` and `destroy` workflows, and the reusable workflows they call, to use short-term credentials through GitHub's OIDC provider rather than long-term credentials through a service user.  The advantages of this approach are: 
- OIDC roles can be managed directly by the development team, unlike service users that are managed via support tickets.
- OIDC-based credentials are short-lived and don't need to be stored as GitHub Secrets and rotated periodically as static credentials do.

This change adds a new serverless service called `github-oidc`, which creates the AWS resources that enable GitHub OIDC.  This service is meant to be bootstrapped manually once for each deployment environment.  After this, further changes to the OIDC service role (e.g. to add new permissions for an AWS API) are made through the standard automated deployment process, which creates an ephemeral OIDC service role for each feature branch that contains changes to this service. The ephemeral role can be used to test the updated permissions. For more details and examples, please see the `README` for the service.

The `github-oidc` service is adapted from the [QuickStart template](https://github.com/CMSgov/macpro-quickstart-serverless/pull/542) to fit MC-Review's use case.  I tried to make it easy for the team to follow the principle of least privilege by configuring two parameters:
- `gitHubActionsAllowedAwsActions`: a list of AWS actions that workflows are allowed to perform.  I derived the list of allowed AWS APIs by first using Access Advisor to see recently used AWS actions for the current IAM service user, then adding a few APIs to the list until the `deploy` workflow succeeded.  If you try to access any other AWS services in a workflow using the OIDC-based credentials, the workflows will fail due to permission errors until you add the corresponding actions to this list. 
- `managedPolicyArns`: if the above becomes a hassle, the team can add the AWS-managed `AdministratorAccess` policy to this list, which will allow all AWS actions in workflows (this is the current setup with the IAM service user).

### Rollout
The intent is to roll this change out to the `dev` environment only, then enable it for `val` and `prod` in a separate change once we are confident that things are working as expected.  To achieve this, I: 
- bootstrapped the `dev` environment as documented in the `README`
- pinned the `uses` reference for the reusable workflows in the `promote.yml` workflow to the commit hash of HEAD  directly before this change.  This is necessary since I've changed the inputs for the reusable workflows called by `promote`, and a reference to `main` would break the `promote` workflow once this change is merged to `main`.  My plan is to revert these references back to `main` once we are ready to roll out this change to `val` and `prod`. 

Note that we'll also need to remove the pins to this branch for the reusable workflows called by `deploy`.
#### Related issues

https://qmacbis.atlassian.net/browse/MR-2860

#### Screenshots

N/A

#### Test cases covered

N/A

## QA guidance
Manual testing notes:

  - I ran multiple successful deploys using OIDC.  I made sure these included deploys with and without changes to the `github-oidc` service, to test the conditional job-running logic in `deploy-infra-to-env`
  - I deployed a stage of the service that added new permission for an AWS API, then successfully called that API via the updated OIDC role later in the `deploy` workflow (with a pause to let the role settle).  I also confirmed that attempting to call that API using an OIDC role without the added permission failed as expected.
  - I bootstrapped the `dev` environment following the README steps and confirmed that all infra and app services deploy successfully using the `dev` OIDC role
  - I temporarily updated the triggers for the `destroy` workflow to test it out


.github/workflows/deploy-app-to-env.yml
```diff
@@ -35,16 +36,19 @@ jobs:
       - name: Setup env
         uses: ./.github/actions/setup_env
 
+      - name: Configure AWS credentials
```
macrael, line 39: yay for standard GH actions for this stuff, looks great.

macrael (approved):
Ben this looks great. I assume you applied this manually to dev so that this could run correctly in the review app? 

Let’s chat about the manual aspect of this, if we could have it be self-updating then that yaml might better belong in one of our services where a serverless deploy would be responsible for updating it in the future

.github/workflows/promote.yml
```diff
@@ -118,7 +118,8 @@ jobs:
 
   promote-infra-dev:
     needs: [build-prisma-client-lambda-layer, unit-tests]
-    uses: CMSgov/managed-care-review/.github/workflows/deploy-infra-to-env.yml@main
+    # TODO update inputs/secrets and pin back to main after verifying OIDC in dev
+    uses: CMSgov/managed-care-review/.github/workflows/deploy-infra-to-env.yml@10f745b1ba147dbef9b4161bb41d9c4aa1b4f2b9
```
macrael, line 122: Do these need to have the environment: tags set to work?

macrael (approved):
Looks good to me, do we need to be passing environment: in the promote script now?

.github/workflows/promote.yml
ben-harvey: Sort of, @macrael. Once the pin is moved back to `main`, then environment is a required input for `deploy-infra-to-env`.  The way you do environments and environment secrets with reusable workflows is weird. The [tldr; in this article](https://colinsalmcorner.com/consuming-environment-secrets-in-reusable-workflows/#tldr) is good at clarifying this.

PR #1403, opened 2022-12-22
Console info

rswerve:
This was already merged but somehow there's a stray console.log, that I just removed.


macrael (approved):

rswerve:
I think this needs a new branch name.  I'll just roll it into a subsequent pr if the linter hasn't told someone else to do it before then.

PR #1413, opened 2022-12-29
Use lerna commands in dev tool

mojotalantikite:
## Summary

Now that we've upgraded to Lerna 6, we can start using it's [new features](https://blog.nrwl.io/lerna-reborn-whats-new-in-v6-10aec6e9091c) (caching, build order dependencies, etc).

This moves the `dev` tool to call the lerna commands directly, rather than using yarn commands from a `cwd`. It also adds a basic caching config.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2498
https://qmacbis.atlassian.net/browse/MR-1181


mojotalantikite:
As I was doing this I noticed a few things:
* We use `yarn install` in a lot of places, with a `cwd` to the service. Since we're using yarn workspaces for a monorepo, a single `yarn install` is all we need, it's going to do it for everything no matter what. Maybe we find a place to just do this once?
* We have a lot of yarn scripts entries that probably could be deleted. Can we clean these out?
* It'd be nice to standardize around a naming scheme in our yarn scripts. Some are of the form `task:subtask`, some use `_`, etc.

I think we should lean more on using these monorepo tools so we can start getting the benefits of caching and standardization.

haworku:
@mojotalantikite It looks like you are starting to do also complete this ticket - https://qmacbis.atlassian.net/browse/MR-1181

Can we just wrap that work fully in this PR? 

That would mean adding documentation to this PR. We want to describe the lerna commands a bit more, the intended way you install and remove package with lerna, etc. Those docs could also easily cover some of what you are raising in https://github.com/CMSgov/managed-care-review/pull/1413#issuecomment-1367483259 as well - re: how to use yarn scripts, standardizing the naming schema (🙏 yes please lets do this now), etc etc

mojotalantikite:
@haworku oh awesome, thanks for pointing out that ticket! I'm down, I'll pull it into the sprint. I'm also curious about what that linked yarn 2 upgrade ticket would take.

nx.json
```diff
@@ -0,0 +1,34 @@
+{
+    "tasksRunnerOptions": {
```
macrael, line 2: was ist das

macrael (approved):
Looks good. We can actually probably drop all the “yarn install” commands. That’s run in `dev` itself before we even get to the `dev.ts` script and since we have lerna there’s no need to re-do it in each service anymore.

macrael (approved):
so nice to standardize across our services

PR #1417, opened 2022-12-30
Fix submission link flake

JasonLin0991:
## Summary

This PR is an attempt to fix the flaky `unlockResubmit.spect.ts` test. The flake revolves around clicking on a submission link.
My suspicion was the link itself may not have the `href` url at the time of the click. This could be because the link is generated from a function and cypress clicks on the link so quickly that it doesn't give the app enough time to generate the link.

Screen Shot 2022-12-30 at 1 53 15 PM

The code above is what generates the link. Could be `react-table` rendering the table, then having to do some work to generate the table rows and links that causes our bug.

I'm checking for a `href` attribute to exist, not empty, and contain the common text `submission` that every submission link has.

I ran cypress in CI several times and seems to not have the flake appear again.


macrael (approved):
good test for this hypothesis

PR #1419, opened 2023-01-03
Create wrapper for fetchHealthPlanPackage and implement on the submission summary page

macrael:
## Summary

Trying to clean up our pages that are doing a bunch of post fetch logic unpacking our protos

I started with SubmissionSummary and will implement this wrapper on the other pages that call this function for this PR. If we like the approach we can refactor other GQL calls. 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2952

#### Test cases covered
I wrote some new tests that I believe cover the logic that I ripped out of the useEffect, please make sure they make sense. 

## QA guidance

Making sure that packages in different submission statuses render for the right people is key. A draft should be invisible to a CMS user, an unlocked package should redirect a state user to the edit flow, etc.


JasonLin0991 (approved):
Looks good!

services/app-web/src/gqlHelpers/fetchHealthPlanPackageWrapper.ts
```diff
@@ -0,0 +1,191 @@
...
+        error: new Error('UNEXPECTED APOLLO BEHAVIOR, NO DATA'),
+    }
+}
+
+// This is ugly, but incredibly useful.
+// We want to hide our protobuf parsing from our pages. The GQL
+// query returns, several layers down, HealthPlanRevisions which have
+// a formDataProto property on it holding the encoded proto string.
+// We want our wrapper to return the identical query to graphql, but
+// with all the HealthPlanRevisions having an additional `formData`
```
rswerve, line 70: Putting this here for posterity, since we already chatted about this in Zoom: this, and the type magic it requires, made me go "hmmm," and one reason is that it might make what we return less flexible in the future.

services/app-web/src/gqlHelpers/fetchHealthPlanPackageWrapper.ts
```diff
@@ -0,0 +1,191 @@
...
+    data: FetchHealthPlanPackageQuery | undefined
+}): FetchHealthPlanPackageResult {
+    if (loading) {
+        return {
+            status: 'LOADING',
+        }
+    }
+
+    if (error) {
+        return {
```
haworku, line 46: 🔴 issue: We are missing out on uniform OTEL logging here. There are several types of errors that can happen that can log for before returning things to the component. This is what the  new`handleApolloError` function is intended for.

services/app-web/src/gqlHelpers/fetchHealthPlanPackageWrapper.ts
```diff
@@ -0,0 +1,191 @@
...
+            variables: {
+                input: {
+                    pkgID: id,
+                },
+            },
+        })
+    )
+
+    if (result.status === 'SUCCESS') {
+        const pkg = result.data.fetchHealthPlanPackage.pkg
```
haworku, line 122: 💡 No action needed but a thought - This `parsedPackage` code seems like the most extendable part of the wrapper. It seems relevant for any query... now or in future...  that would return a health plan package with a submission and revisions. At quick glance though one issue is we have no real type for ParsedPackageType.  interesting problem to poke at in the future

haworku (approved):
I see some improvements to how we handle invalid URLs, tests look good to me as well.

services/app-web/src/gqlHelpers/fetchHealthPlanPackageWrapper.ts
haworku: I also feel uncomfortable with type gymnastics we do but I feel  this isn't too much of a transgression (side eye `toDomain`). I wonder if some kind type magic will be required for anything intended to be re useable  because we have no authoritative type for the true package (submission + revisions) with that data scattered between postgres and the proto (e.g. the graphql types for metadata + the domain model types for the form data) it becomes difficult for frontend to get any type without either mashing things together or inferring types. 

I'd be curious what you mean about less flexible in the future? I didn't see that  yet want to know what you are seeing.

rswerve: The main thing I was looking at was the recursive type replacement below, which I think is caused by putting the unwrapped data back on the package itself.  What we discussed yesterday was returning the data here as its own freestanding type.

services/app-web/src/gqlHelpers/apolloQueryWrapper.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+type WrappedApolloResultType<
+    ResultType extends WrappableApolloResultsType,
+    AdditionalVars = Record<string, unknown>
+> = Omit<ResultType, 'data' | 'loading' | 'error'> & {
+    result: ApolloResultType<NonNullable<ResultType['data']>, AdditionalVars>
+}
+
+// Apollo's useQuery returns three independent variables to track loading, error, and data. In our usage
```
rswerve, line 35: This is such weird behavior from Apollo.  Love this wrapper.

rswerve: I guess it's not _weird_ exactly--other libraries do it, too, but I prefer the wrapped method.

services/app-web/src/gqlHelpers/fetchHealthPlanPackageWrapper.ts
```diff
@@ -0,0 +1,131 @@
...
+        )
+        recordJSException(err)
+        console.error(err)
+        return {
+            status: 'ERROR',
+            error: err,
+        }
+    }
+
+    const formDatas: { [revisionID: string]: HealthPlanFormDataType } = {}
```
rswerve, line 68: `formDataRevisions`?

services/app-web/src/pages/StateSubmission/StateSubmissionForm.tsx
```diff
@@ -207,114 +175,49 @@ export const StateSubmissionForm = (): React.ReactElement => {
...
 
-        return specificContent ?? <GenericErrorPage />
+    if (fetchResult.status === 'ERROR') {
+        recordJSException(fetchResult.error)
+        console.error('Error from API fetch', fetchResult.error)
+        return <GenericErrorPage /> // api failure or protobuf decode failure
     }
 
-    if (!fetchLoading && !submissionAndRevisions) {
+    const { data, formDatas, documentDates, documentLists } = fetchResult
```
rswerve, line 192: I'm just agreeing with myself here, but I think this kind of thing shows the power of separating the bit of information we want from the query--the types are much less work when we add it on the back end, and the parsing is easier on the front end.

rswerve (approved):

services/app-web/src/gqlHelpers/fetchHealthPlanPackageWrapper.ts
macrael: Great callout, I actually ended pulling all the otel error logging out of the wrapper and instead am explicitly handling the errors from inside the pages themselves. I think that trying to do as much of that error-logging as possible from those pages directly instead of inside library code is going to make thing easiest. In the page itself is where we have all the context neccecary to log errors correctly, an easy example of that is that we can pass isAuthenticated to handleApolloError with confidence. I’d love to chat about this at our next eng meeting. 

Another way this makes sense to me is that the wrapper returns the error, and our pages that call this have to respond to that error, I think it makes the most sense to assume as a page that you need to fully handle logging the error rather than assuming that that has been done.

services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummary.tsx
```diff
@@ -40,102 +32,61 @@ export const SubmissionRevisionSummary = (): React.ReactElement => {
...
                 <Loading />
             </GridContainer>
         )
     }
 
-    if (data && !submissionAndRevisions) return <Error404 /> // api request resolves but are no revisions likely because invalid submission is queried. This should be "Not Found"
-    if (error || !packageData || !submissionAndRevisions)
+    if (fetchResult.status === 'ERROR') {
+        // Log something about the error?
+        recordJSException(fetchResult.error)
```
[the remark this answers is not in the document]
macrael: See below, I actually did the opposite, but we could add a new helper function to handle an ApolloError | Error more easly.

services/app-web/src/gqlHelpers/fetchHealthPlanPackageWrapper.ts
haworku: Sure! I'm fine with that boundary, just having it consistent and using the apollo error helper so we get robust errors is my interest.

PR #1420, opened 2023-01-03
fix: MR-2933 Formatting in state submission confirmation email. Add formatEmailAddress

haworku:
## Summary
Addresses issues where emails display to users in a strange format.

This [bug]( https://qmacbis.atlassian.net/browse/MR-2933 )  was likely introduced when we moved toward formatting emails in parameter store with an alias. 

Other changes:
- pulled out state assistance template block into its own partial
- added `/formatters` folder 
- added new formatter functions  for `isEmailAddress` and `includesEmailAddress` to check for email regex. 
- Adjust `pruneDuplicateEmails` to partially account for matching emails with different aliases

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2933

#### Test cases covered
Add `test.each` matrix tests for each of the utility functions.  Run through different types of strings (includes email alias, doesn't include alias, is not an email, etc

## QA guidance
- Make a new submission with duplicate emails for the state contact. Also, add your own email as a state contact (to ensure you can see new state email). 
-  The email content should show all emails properly formatted in the text content. Also, there should be no duplicate emails in the receivers list.


JasonLin0991 (approved):
Looks good.

There seems to be another bug `generateStateRecieverEmails`. I posted about it in this [slack thread](https://cmsgov.slack.com/archives/C014CRU197U/p1672951724057769). Not sure if you wanted to tackle it in this PR or make a new ticket.

haworku:
@JasonLin0991 thank you!

PR #1424, opened 2023-01-04
Memoize column data in reactTable to avoid re-mounting table cells

macrael:
## Summary

This should fix our flakey Cypress click in the unlockResubmit test.

Without memozing the column data structure, it’s re-creating it on every render, and I _think_ since that includes all the cell() functions that is causing react to re-mount all our cells on every render. By memoizing, react won’t lose track of the elements

### DIFF
--- a/services/app-web/src/pages/StateDashboard/StateDashboard.tsx
+++ b/services/app-web/src/pages/StateDashboard/StateDashboard.tsx
@@ -66,7 +66,7 @@ export const StateDashboard = (): React.ReactElement => {
                 getCurrentRevisionFromHealthPlanPackage(sub)
 
             if (currentRevisionDataOrError instanceof Error) {
-                return null
+                return
             }
             const [_, currentSubmissionData] = currentRevisionDataOrError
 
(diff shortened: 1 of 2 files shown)


services/app-web/src/pages/StateDashboard/StateDashboard.tsx
```diff
@@ -66,7 +66,7 @@ export const StateDashboard = (): React.ReactElement => {
                 getCurrentRevisionFromHealthPlanPackage(sub)
 
             if (currentRevisionDataOrError instanceof Error) {
-                return null
```
haworku, line 69: This should return something. How about the `<ErrorAlertFailedRequest />` UI used in CMS dashboard. Want to avoid returning null from a page level component in general, we should be using fall back error pages.

macrael: Yeah, this is why I try and avoid using forEach() because I always get confused about where the return will go. This is fine for now but hopefully can be better handeled by another wrapper.

PR #1425, opened 2023-01-04
MR-2881: Add count text for filters

rswerve:
## Summary

We want to display a bit of text when users are using filters, so that they either see a total count of submissions, when no filters are applied, or a x of y count, when some are.  There should be no message when filters are applied but don't return anything.

#### Screenshots

image
image
image

#### Test cases covered

Added lines to existing filter tests to also check the text when we filter.

### DIFF
--- a/services/app-web/src/common-code/formatters/pluralizer.ts
+++ b/services/app-web/src/common-code/formatters/pluralizer.ts
@@ -1,11 +1,11 @@
 export const pluralize = (word: string, count: number): string => {
-    const isPlural = count !== 1;
+    const isPlural = count !== 1
     switch (word) {
-        case "do":
-            return isPlural ? "do" : "does";
-        case "does":
-            return isPlural ? "do" : "does";
+        case 'do':
+            return isPlural ? 'do' : 'does'
+        case 'does':
+            return isPlural ? 'do' : 'does'
         default:
-            return isPlural ? word + "s" : word;
+            return isPlural ? word + 's' : word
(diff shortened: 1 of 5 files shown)


JasonLin0991 (approved):
Looks great!

JasonLin0991:
@rswerve I take it all back, it's not looking good yet. When no submission are found it should display `Displaying 0 of 5 submissions`

Figma
Screenshot 2023-01-04 at 6 13 46 PM

JasonLin0991 (changes requested):
@rswerve I take it all back, it's not looking good yet. When no submission are found it should display `Displaying 0 of 5 submissions`

Figma
Screenshot 2023-01-04 at 6 13 46 PM

rswerve:
> @rswerve I take it all back, it's not looking good yet. When no submission are found it should display `Displaying 0 of 5 submissions`
> 
> 

Thanks, I'll fix that.

rswerve:
> @rswerve I take it all back, it's not looking good yet. When no submission are found it should display `Displaying 0 of 5 submissions`
> 
> Figma 

Actually, we need to check with design.  This is in the AC for the ticket: `If submission list has been filtered to 0, then do not display this line and display the blank state design. `

rswerve:
Got clarification from design and pushed up new changes @JasonLin0991

JasonLin0991 (approved):
💯. Thanks for double checking with design!

PR #1428, opened 2023-01-06
fix: MR-2542 Ensure that trailing spaces in email address inputs do not trigger errors

haworku:
## Summary
Looking for feedback on the approach. I tried addressing the bug in two ways
- Using `.trim` in yup validations. This ensures validation error is simply not raised. See 482e326ea69c9bc0f161dbfe9ab047410aca1fe5 
- Direct remove leading and trailing whitespace from the input value itself `onBlur` to prevent this class of issue from arising, at least where our app-wide components are used . See 6ccda52dc6f5825503d1c6d447da43b887eaad4d 

Other change in this PR is slimming contacts and actuary contacts JSX by using `<FieldTextInput />` to remove extraneous markdown. 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2542

## QA guidance
- Besides testing bugfix to email address inputs, worth running through contacts and actuary contacts UI, make sure I didn't cause a regression here. Seems like we are testing with unit tests and Cypress pretty thoroughly here which is good to see.

### DIFF
--- a/services/app-web/src/components/Form/FieldTextInput/FieldTextInput.tsx
+++ b/services/app-web/src/components/Form/FieldTextInput/FieldTextInput.tsx
@@ -3,6 +3,8 @@ import { useField } from 'formik'
 import { Label, TextInput, FormGroup } from '@trussworks/react-uswds'
 import { PoliteErrorMessage } from '../../'
 
+import styles from './FieldTextInput.module.scss'
+import classNames from 'classnames'
 /**
  * This component renders a ReactUSWDS TextInput component inside of a FormGroup,
  * with a Label and ErrorMessage.
@@ -16,16 +18,19 @@ import { PoliteErrorMessage } from '../../'
 
 export type TextInputProps = {
     label: string
+    variant?: 'TOPLEVEL' | 'SUBHEAD' // subhead used for forms where fields could be nested under a larger heading or label
     id: string
     hint?: React.ReactNode
     error?: string
     showError: boolean
     name: string
     type: 'text' | 'email' | 'number' | 'password' | 'search' | 'tel' | 'url'
-} & JSX.IntrinsicElements['input']
+} & React.ComponentProps<typeof TextInput>
 
 export const FieldTextInput = ({
+    className,
     label,
+    variant = 'TOPLEVEL',
     id,
     hint,
     error,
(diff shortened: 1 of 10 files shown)


services/app-web/src/components/Form/FieldYesNo/FieldYesNo.tsx
```diff
@@ -23,7 +23,7 @@ export type FieldYesNoProps = {
     hint?: React.ReactNode
     showError?: boolean
     id: string
-    variant?: 'PRIMARY' | 'SECONDARY' // secondary variant used for nested fields yes/no fields under an overarching heading
+    variant?: 'TOPLEVEL' | 'SUBHEAD' // subhead variant used for nested fields yes/no fields under an overarching heading
```
haworku, line 26: @macrael FYI saw your comment after I merged but... your wish is granted 🪄 https://github.com/CMSgov/managed-care-review/pull/1390#discussion_r1053541129

rswerve (approved):
I like the onBlur functionality.  Will keep the spaces from causing any problems downstream.

I entered a bunch of contacts with leading and trailing spaces.  No problem on submit.

services/app-web/src/components/Form/FieldTextInput/FieldTextInput.tsx
```diff
@@ -16,35 +18,59 @@ import { PoliteErrorMessage } from '../../'
...
+            [styles.nestedField]: variant === 'SUBHEAD',
+        },
+        className
+    )
+
+    // Latch into onBlur to do any input cleaning
+    // Initial use case is to trim away trailing whitespace in email addresses
+    const customOnBlur = (e: React.FocusEvent<HTMLInputElement, Element>) => {
+        if (!e) return
+        helpers.setValue(field.value.trim())
```
JasonLin0991, line 56: This `customOnBlur` does not seem to be working for email input on the contacts page. When not focused on the email input, it still retains leading and trailing spaces. 

You can continue to supporting documents with leading trailing spaces, but **it does trim**, even when `trim()` is removed from the the Yup validation in `Contacts.tsx`. I think is a graphical issue only.

JasonLin0991 (changes requested):
Looks good. just one issue with the email input.

JasonLin0991 (approved):
Awesome! Thanks for those changes.

PR #1435, opened 2023-01-10
MR-2962 Add missing fields handling across review and submit

haworku:
## Summary
Add missing fields UI across the submission summary page.
1. State users should see red warning text
2. CMS users should not see missing fields at all, they can are entirely from summary

New small components:
- DataDetailMissingField
- DataDetailCheckboxList
- DataDetailDataRange
- DataDetailContactField

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2962

This extends  behavior developed for #1390 across the app.

#### Test cases covered
Jest
- All submission summary section have tests to ensure missing fields data does not render for submitted submissions

Cypress 
- in dashboard > submission summary test added one line assertion to double check no missing field UI pops up.

## QA guidance
- Best way to test is to create a new submission and then immediately jump to review-and-submit page and check that the missing field label populates all over


services/app-web/src/components/DataDetail/DataDetail.stories.tsx
```diff
@@ -9,41 +12,64 @@ export default {
...
+
+export const WithCheckboxList = Template.bind({})
+WithCheckboxList.decorators = [(Story) => ProvidersDecorator(Story, {})]
+WithCheckboxList.args = {
+    id: 'crystal-ball',
+    label: 'The secret to the future',
+    data: ['this', 'that', 'the other thing'],
+}
+
+// Having trouble getting these stories to render without crashing story
```
haworku, line 46: I can't get these stories rendering properly. I am thinking it doesn't like the use of subcomponents as props. I have been referencing [storybook docs ](https://storybook.js.org/docs/react/writing-stories/stories-for-multiple-components/)but not getting anywhere. Leaving it here in case anyone sees what I'm missing, otherwise once approve I will make a commit removing these stories all together.

services/app-web/src/components/DataDetail/DataDetailContactField/DataDetailContactField.tsx
```diff
@@ -0,0 +1,47 @@
...
+import {
+    ActuaryContact,
+    StateContact,
+} from '../../../common-code/healthPlanFormDataType'
+import { getActuaryFirm } from '../../SubmissionSummarySection'
+import { DataDetailMissingField } from '../DataDetailMissingField'
+
+type Contact = StateContact | ActuaryContact
+
+function isCertainActuaryContact(contact: Contact): boolean {
```
macrael, line 11: You could make this function a type-check with the return type `: contact is ActuaryContact` and I think you could lose the `as`’s in this file. https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
```diff
@@ -151,66 +126,84 @@ export const ContractDetailsSummarySection = ({
...
+                                    dict={ModifiedProvisionsRecord}
+                                />
+                            }
+                        />
+
+                        <DataDetail
+                            id="unmodifiedProvisions"
+                            label="This contract action does NOT include new or modified provisions related to the following"
+                            explainMissingData={!isSubmitted}
+                            data={
```
macrael, line 198: I wonder if using `children` rather than a `data` prop would make all this cleaner. Might make storybook happy if your theory is right

macrael (approved):
looks good, I like adding some structure to our dd's

services/app-web/src/components/DataDetail/DataDetail.stories.tsx
JasonLin0991: I poked around `DataDetailContactField.tsx` and looks like its actually `getActuaryFirm()` causing the Storybook issue.

```
  {isCertainActuaryContact(contact) && (
      <>
          <br />{' '}
          <span>{getActuaryFirm(contact as ActuaryContact)}</span>
      </>
  )}
```

If you move that function into `DataDetailContactField.tsx` everything works fine. I'm not sure exactly why. Possibly not being able to call out that function from storybook?

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
haworku: Yeah I was thinking about that. Thought it was a lot of code change to do for the goal of this PR but looking at things again its worth exploring at least if that would fix storybook. Would be good to know

services/app-web/src/components/DataDetail/DataDetail.stories.tsx
haworku: I really think something is broked with our storybook. Subcomponents are not working as documented. I think its something odd going on maybe with imports in our monorepo structure.

PR #1440, opened 2023-01-12
Add caption prop, scope tag to HealthPlanPackageTable

haworku:
## Summary
Add a11y handling tweaks to table
- optional caption prop 
- `scope` attribute to specify that th header cell is a header for a column
- Use `showFilters` prop to determine if filtering language used in aria-live so its not possible to get into state where filters are off but text announces as if filters in place. 

See https://github.com/trussworks/accessibility/wiki/Tables

#### Related issues
related to #1436 

#### Test cases covered
- Added new test for caption. Had to edit several tests to pass filters prop for when we are testing filters

## QA guidance
Running through table in screenreader.

### DIFF
--- a/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.stories.tsx
+++ b/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.stories.tsx
@@ -111,8 +111,8 @@ WithCaption.args = {
     caption: 'Table Data',
 }
 
+// CMS User experience
 export const WithFilters = Template.bind({})
-
 WithFilters.decorators = [(Story) => ProvidersDecorator(Story, {})]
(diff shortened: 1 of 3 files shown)


haworku:
@JasonLin0991 feel free to merge to your branch or else i'll merge to main when your work goes through. i felt like this went beyond the "small change" I promised a bit so figured i would open the PR in case

JasonLin0991 (approved):
Looks good. 

I made similar changes with showing filter count which is what the conflict is. I used the user role, where as you used `showFilters` props. I think you had the better idea of using `showFilters`. When you resolve the conflicts we can use `showFilters` instead. Thanks!

haworku:
@JasonLin0991 fine to merge this? Can you give me another 👍 i ended up having to push new commits cuz rebasing was annoying

PR #1441, opened 2023-01-12
Upgrade react-uswds to 3.x

haworku:
## Summary
Bump react-uswds a major version to 3.x. Also includes react-uswds Node 18 support in `engines` field.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2522

## Tests
- The long hold up on this PR was trying to **get tests passing for things related to rate amendments**. Specifically the date range pickers on those pages were not longer being properly edited with our existing commands, both in jest and in cypress. had to write selectors that were more specific (since date pickers render two inputs, an internal and an external input, make sure external input is being changed) and also clarify which input should be edited when there were multiple Start date and end date inputs both rate dates and rate effective dates. 
- Interestingly, the selectors that worked in jest weren't working in cypress, so had to re- implement there another way.

## QA guidance
- Tests pass
- visual spot checks of changed components


services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.test.tsx
```diff
@@ -853,198 +852,6 @@ describe('RateDetails', () => {
                 expect(continueButton).toHaveAttribute('aria-disabled', 'true')
             })
         })
-
-        it('progressively disclose new rate form fields on the second rate', async () => {
```
haworku, line 857: I deleted some unnecessary and long running tests. For example this logic is handled with the single rate test, there isn't a need to do the same test for multiple rates since its all shared logic and components at a high level.

tests/cypress/support/stateSubmissionFormCommands.ts
```diff
@@ -274,19 +289,28 @@ Cypress.Commands.add('fillOutNewRateCertification', () => {
...
-    cy.findAllByLabelText('Start date').eq(1).type('03/01/2024')
-    cy.findAllByLabelText('End date').eq(1).type('03/01/2025')
+
+/* 
+    There are currently multiple date range pickers on the page with the same label names (start date, end date) and different headings
+    Preferred approach would be targeting by via findBy* or a custom data-cyid attribute
+    However, surfacing custom attributes on the nested inputs in third party component DateRangePicker not possible in current react-uswds version 
+    For now using targeting by html id (anti-pattern)
+*/
+    cy.get(`[id="rateInfos.${id}.rateDateStart"]`).type('02/01/2023')
```
macrael, line 309: is there a parent div or something we could use to use .within here to differentriate between? Or is that only in testing library itself? No worries if not but that’s my only thought.

macrael (approved):
yay looks great i like the date helper

tests/cypress/support/stateSubmissionFormCommands.ts
haworku: @macrael Yeah, good point. Looking at this code again, maybe what I should do is put the `data-cyid` attribute I was struggling to get on the date inputs on the overarching form group div instead. Then use `within`.  Without a code change there isn't anything unique about the parent since its just a generic form group. 

I'm thinking to merge this as is but will keep looking at things since I'm going to do more uswds upgrade stuff next

PR #1447, opened 2023-01-18
MR-2623: Store filter state in URL

rswerve:
This moves filter state into the URL, and makes the filter multiselects controlled components, driven by the URL.  It also ties the table filters to the URL state.

One test deleted because the functionality was tested elsewhere.

Considered adding tests for the URL itself, but, analogizing to useState, that felt like testing implementation details.  If the filters work, the URL is working.


services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.test.tsx
```diff
@@ -93,6 +93,9 @@ const submissions: PackageInDashboardType[] = [
 ]
 
 describe('HealthPlanPackageTable for CMS User (with filters)', () => {
+    beforeEach(() => {
+        window.location.assign('#')
```
macrael, line 97: is this to clear any state leftover between tests?

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
```diff
@@ -249,11 +310,14 @@ export const HealthPlanPackageTable = ({
                     {showFilters && (
                         <FilterAccordion
                             onClearFilters={() => {
+                                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                                //@ts-ignore
```
macrael, line 314: what does typescript complain about here?

macrael (approved):
I think it’s usually worth putting some tests in when you have a serializer and deserializer to make sure that they do what we expect. Agreed that with this implementation testing the filters themselves will confirm that the whole system is working, but specifically it might be good to have some tests that the URLs we generate are what we expect. 

Testing on the review app I got `https://d25wmdnp5vk5z8.cloudfront.net/#filters=stateName%3DMinnesota%26submissionType%3DContract%2520action%2520and%2520rate%2520certification`

is it possible for us to not urlencode the `=` and some of the other non-space stuff that is OK in urls?

I think it’s totally fine to merge and demo this as is and pretty up the URLs afterwards

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.test.tsx
rswerve: yup

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
rswerve: `type "never" has no call signature`  jotai types the set function as a `never`  I'm not sure why.

rswerve:
> is it possible for us to not urlencode the = and some of the other non-space stuff that is OK in urls?

Messed with this a fair bit, actually, and jotai is encoding everything and doesn't surface an option to not encode.  So I think the answer is maybe, if we use a different library.

rswerve:
> it might be good to have some tests that the URLs we generate are what we expect

Sure.  I'll roll that into the ticket to fix the focus issues.

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
rswerve: To be clear, there are a bunch of issues/discussions in jotai about this, and I've tried the various proposed solutions; no luck so far, but I think it's fixable with the correct incantation.

PR #1448, opened 2023-01-18
MR-2970: All submitters should receive email on submission, unlock and resubmission

JasonLin0991:
## Summary
[MR-2970](https://qmacbis.atlassian.net/browse/MR-2970)

Submitters were not getting emails when packages were unlocked because the function `generateStateReceiverEmails` was using `user` context to get the submitter email. The bug occurs when unlocking a submission done by a CMS user, the user here would never be the submitter which would only be a state user.

The fix was to get the submitter email from the revision data and pass it to the emailer. This is done using a new function `packageSubmitters`, which will return an array of all the submitter emails in the package.

Now that we no longer use `generateStateReceiverEmails` to grab the current user, the function can be removed and getting state contacts could be done with a one liner in each state emailer function.

#### Test cases covered

- `healthPlanPackage.test.ts`
   - `packageSubmitters` - 'returns expected package submitters'

- `submitHealthPlanPackage.test.td`
   - `send state email to state contacts and current user on valid resubmission` --> `'send state email to state contacts and all submitters on valid resubmission'`

- `unlockHealthPlanPackage.test.ts`
   - `send state email to state contacts and all submitters when unlocking submission succeeds`

## QA guidance

**NOTE**: Note sure how we can test packages with multiple submitters since we don't seem to have multiple accounts for the same state in the review deploy. There are unit tests that do test for multiple submitters on a submission.


services/app-api/src/resolvers/healthPlanPackage/unlockHealthPlanPackage.test.ts
```diff
@@ -468,6 +468,91 @@ describe('unlockHealthPlanPackage', () => {
...
+        const stateReceiverEmails = [
+            'james@example.com',
+            'notspiderman@example.com',
+            ...sub.stateContacts.map((contact) => contact.email),
+        ]
+
+        // email subject line is correct for CMS email.
+        // Mock emailer is called 4 times, 2 for the first unlock, 2 for the second unlock.
+        // From the pair of emails, the second one is the state email.
+        expect(mockEmailer.sendEmail).toHaveBeenNthCalledWith(
```
macrael, line 543: I like that you are checking sendEmail and able to confirm the toAddresses directly, rather than just intercepting the call to sendUnlockPackage… 

I’m always a bit hestiant to use “toHaveBeenCalledWith” but I think this makes a lot of sense. I wonder if we could bulid out a test emailer that you could pull the raw email texts out of and assert agasint those instead.

macrael (approved):
This looks great. I’m glad for the pkg-> formData rename

services/app-api/src/resolvers/healthPlanPackage/unlockHealthPlanPackage.test.ts
JasonLin0991: > I wonder if we could bulid out a test emailer that you could pull the raw email texts out of and assert agasint those instead.

That would be an interesting idea.

> I’m always a bit hestiant to use “toHaveBeenCalledWith” ...

Could you elaborate more on this?

I'll merge this now so we can get the fix in, but please feel free to respond.

PR #1451, opened 2023-01-19
fix: MR-2962 MR-314 for product acceptance

haworku:
## Summary
Small bugfix from product acceptance. See conversation on each ticket and commit history for context.

#### Related issues

- https://qmacbis.atlassian.net/browse/MR-314
- https://qmacbis.atlassian.net/browse/MR-2962

#### Test cases covered

- shows missing field error on amended provisions when expected
- does now show missing field error on amended provisions when fields are valid

## QA guidance
Manual QA for MR-314 steps
- Log in as a state user and view the review/submit page on an unlocked submission where this field was not filled out on initial submission (you can get here by turning flag off, adding sub, unlocking, then turning flag on).
-  As expected, there is an error message shown on the review and submit page indicating this field is required. 
- Click edit on the submission type section and navigate to submission type page. 
- Leave the risk-based question as blank and select “save as draft”. 
- You are navigated back to the dashboard. From dashboard, click back into the submission you just edited.
- You should notice that the risk-based question is still empty and be able to fill out

### DIFF
--- a/services/app-web/src/components/Form/FieldYesNo/FieldYesNo.tsx
+++ b/services/app-web/src/components/Form/FieldYesNo/FieldYesNo.tsx
@@ -26,23 +26,6 @@ export type FieldYesNoProps = {
     variant?: 'TOPLEVEL' | 'SUBHEAD' // subhead variant used for nested fields yes/no fields under an overarching heading
 } & JSX.IntrinsicElements['input']
 
-export type FieldYesNoFormValue = 'YES' | 'NO' | undefined // Use for formik string value
-export type FieldYesNoUserValue = 'Yes' | 'No' | undefined // Use for user facing display
-
-// Helpers used for reading and writing from db boolean types to formik string type of YES and NO
-export const booleanAsYesNoFormValue = (bool?: boolean): FieldYesNoFormValue =>
-    bool ? 'YES' : bool === false ? 'NO' : undefined
-
-export const booleanAsYesNoUserValue = (bool?: boolean): FieldYesNoUserValue =>
-    bool ? 'Yes' : bool === false ? 'No' : undefined
-
-// Use for db stored value
-export const yesNoFormValueAsBoolean = (
-    maybeString: FieldYesNoFormValue | string
-) => {
-    return maybeString === 'YES' ? true : false
-}
-
 export const FieldYesNo = ({
     name,
     label,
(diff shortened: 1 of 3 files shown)


services/app-web/src/components/Form/FieldYesNo/FieldYesNo.tsx
```diff
@@ -105,3 +88,24 @@ export const FieldYesNo = ({
...
+export type FieldYesNoUserValue = 'Yes' | 'No' | undefined // Use for user facing display
+export type FieldYesNoBoolean = true | false | undefined // Use for sending to backend
+
+export const booleanAsYesNoFormValue = (bool?: boolean): FieldYesNoFormValue =>
+    bool ? 'YES' : bool === false ? 'NO' : undefined
+
+export const booleanAsYesNoUserValue = (bool?: boolean): FieldYesNoUserValue =>
+    bool ? 'Yes' : bool === false ? 'No' : undefined
+
+export const yesNoFormValueAsBoolean = (
```
haworku, line 103: The only real diff to code is in this function block

services/app-web/src/components/Form/FieldYesNo/FieldYesNo.tsx
```diff
@@ -105,3 +88,24 @@ export const FieldYesNo = ({
         </Fieldset>
     )
 }
+
+// HELPERS
+export type FieldYesNoFormValue = 'YES' | 'NO' | undefined // Use for formik string value
```
macrael, line 93: I feel like I’ve tried and failed before but it would be nice if we could figure out how to make these fields store true/false instead of YES NO in formik just to cut down on the different representations here. 

Would it make sense to use a map here to make it really explicit the relationship between these? Or would that just make this more complicated?

`const fieldValueMapping: { true: { userValue: ‘Yes’, formikValue: ‘YES’}, false: … undefined: …}`

macrael (approved):
looks good, I think it makes sense that there are 3 different ways to represent this data

services/app-web/src/components/Form/FieldYesNo/FieldYesNo.tsx
haworku: I like the idea of making things even more explicit.

The issues with the dictionary I see are that booleans as keys won't work and then we still have to handle fallbacks. The functions currently all have a fallback or default value to use if nothing matches. I don't think you could use the dictionary directly for those reasons _(not sure if that's what you were intending though, did I understand correctly?)_. 

Another random idea: a single exported function that takes any valid values for this field something like `getYesNoValue(value, type) => value `. 

Value is `FieldYesNoUserValue | FieldYesFormValue |  FieldYesNoBooleanValue`
Type is some new enum we make to represent each possible type.

Does that seem better?

PR #1452, opened 2023-01-19
Add screen reader announcement when entering table.

JasonLin0991:
## Summary
[MR-2872](https://qmacbis.atlassian.net/browse/MR-2872)

Missed this AC from the previous PR for this ticket.

### DIFF
--- a/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.module.scss
+++ b/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.module.scss
@@ -51,3 +51,15 @@
     color: $theme-color-base-ink;
     padding-top: units(3);
 }
+
+.visuallyHidden {
+    position: absolute !important;
+    width: 1px !important;
+    height: 1px !important;
+    padding: 0 !important;
+    margin: -1px !important;
+    overflow: hidden !important;
+    clip: rect(0,0,0,0) !important;
+    white-space: nowrap !important;
+    border: 0 !important;
+}
--- a/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.module.scss
+++ b/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.module.scss
@@ -51,15 +51,3 @@
     color: $theme-color-base-ink;
     padding-top: units(3);
 }
-
-.visuallyHidden {
-    position: absolute !important;
-    width: 1px !important;
-    height: 1px !important;
-    padding: 0 !important;
-    margin: -1px !important;
-    overflow: hidden !important;
-    clip: rect(0,0,0,0) !important;
-    white-space: nowrap !important;
-    border: 0 !important;
-}
--- a/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
+++ b/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
@@ -332,11 +332,11 @@ export const HealthPlanPackageTable = ({
                         </tbody>
                         <caption className={caption ?? styles.srOnly}>
                             {caption || 'Table'}
-                            {showFilters && (
-                                <span className={styles.srOnly}>
-                                    {`, ${filtersApplied}, ${submissionCount}.`}
-                                </span>
-                            )}
+                            <span className={styles.srOnly}>
+                                {`${
+                                    showFilters && `, ${filtersApplied}`
+                                }, ${submissionCount}.`}
+                            </span>
                         </caption>
                     </Table>
                     {!hasFilteredRows && (
--- a/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
+++ b/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
@@ -290,7 +290,6 @@ export const HealthPlanPackageTable = ({
                         </div>
                     </div>
                     <Table fullWidth>
-                        {caption && <caption>{caption}</caption>}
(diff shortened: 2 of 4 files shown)


services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
```diff
@@ -331,6 +330,12 @@ export const HealthPlanPackageTable = ({
                                 </tr>
                             ))}
                         </tbody>
+                        <caption className={caption ?? styles.visuallyHidden}>
+                            {caption || 'Table'}
+                            <span
```
JasonLin0991, line 335: This was moved from above the `<thead>` to the bottom of the `<Table>`. This was done because when visually hiding the `<caption>` there is a thick border below the column headers. Moving `<caption>` to the bottom removed the thick border and when not visually hiding, the caption still displays above the column headers.

JasonLin0991: Also when tabbing into the table, in VO, it will announce the caption twice. Not sure if this was standard, because I would not find anything about this. Navigating into this table using `control + option + directional arrows` does not announce the caption twice.

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.module.scss
```diff
@@ -51,3 +51,15 @@
     color: $theme-color-base-ink;
     padding-top: units(3);
 }
+
+.visuallyHidden {
```
haworku, line 55: There is a shared `.srOnly` class that is global. I think this is what we need? I didn't notice the borders sorry about that

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
```diff
@@ -331,6 +330,12 @@ export const HealthPlanPackageTable = ({
                                 </tr>
                             ))}
                         </tbody>
+                        <caption className={caption ?? styles.srOnly}>
+                            {caption || 'Table'}
+                            <span
+                                className={styles.srOnly}
+                            >{`, ${filtersApplied}, ${submissionCount}.`}</span>
```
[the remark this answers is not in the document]
JasonLin0991: Yes! good catch!

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
```diff
@@ -331,6 +330,12 @@ export const HealthPlanPackageTable = ({
                                 </tr>
                             ))}
                         </tbody>
+                        <caption className={caption ?? styles.srOnly}>
+                            {caption || 'Table'}
```
haworku, line 334: 🔴  Captions have intended usage - they should be [specific to the table content](https://www.w3.org/WAI/tutorials/tables/caption-summary/). E.g. in this case "submissions" would be a good fallback caption for this component.

Also it seems like with your change we are going to always include a caption. So then does that man `caption` prop has changed its meaning? Now instead of showing or caption or not its becoming more `showCsutomCaption` and is used for optionally displaying a custom caption to the user. Is that the intent ?

macrael: We’re in the roadmap meeting @haworku :p

haworku (approved):

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
JasonLin0991: This is a great point out! 

`"Captions should be specific to the table content"`, so maybe we should update it to fallback to `Submission table`. When using VO, without the `table` part it will announce `Submission, 0 filters applied, displaying 2 of 2 submissions`, which sounded confusing and did describe the context of what the submission was in. Hence the `Table` as a default, but zooming out from just the table, the component itself is scoped to a table of submissions so `Submission table` would make more sense.

> Now instead of showing or caption or not its becoming more `showCsutomCaption` and is used for optionally displaying a custom caption to the user. Is that the intent ?

If I'm interpreting your question correctly, yes. I was trying to preserve the use of caption for sighted users too incase we did want to have a custom caption that should be displayed for sighted users. The fallback is mainly used for VO users, since we do not have a caption for the table by default it has to announce something before the filter/submission count otherwise it would sound confusing. Without a `caption` passed, the fallback will still be hidden for sighted users and when one is passed it will be shown. The filter/submission count will always be hidden. As for the prop name `caption` --> `showCustomCaption` I think that it still showing a caption or not for sighted users, but for VO users it will be `showCustomCaption`.

haworku: Ah, okay so the reason "Submissions" may be enough and we don't need to append "table" is because the screenreader announces the caption and then the word "table" already. This is how captions and things like aria-labels work - they provide an [accessible name](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#naming_effectively) which is announced, and then then the semantic html name is announced.
> Do NOT include a WAI-ARIA role name in the accessible name. For example, do not include the word "button" in the name of a button, the word "image" in the name of an image, or the word "navigation" in the name of a navigation region. Doing so would create duplicate screen reader output since screen readers convey the role of an element in addition to its name. 

That guidance applies here. `table` is already the role implicitly. We can also do whatever for now and check in our user test about this. 

I'm fine with following whatever you decide about prop naming but that makes sense. Thanks for clarifying the intent just wanted to be sure I wasn't assuming things.

PR #1455, opened 2023-01-20
Fix CI skipping script when sha can't be found

mojotalantikite:
## Summary

Occasionally the CI skipping script will throw an error that the git sha that it is comparing against can't be found. This adds a check for that and will run CI deploys for all services if it encounters that error.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-100


macrael (approved):
terriffic

PR #1460, opened 2023-01-23
fix: MR-2962 handle amended provisions empty state

haworku:
## Summary
Another pass to fix MR-2962. Follow on from #1451. This time added special prop to checkbox list to allow empty list behavior when two checkbox lists are related like amended provisions

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2962

#### Test cases covered

Adjusted the test cases to get properly failing tests for the bug with checkbox list
- renders missing field error when list is empty by default
- renders null when `displayEmptyList` and list is empty
- 
## QA guidance
- Check review and submit pages for different configurations of amended provisions checkbox list (contract details > amendment needs to be selected). Only if both lists are empty should we show empty state error message.

### DIFF
--- a/services/app-web/src/components/DataDetail/DataDetail.tsx
+++ b/services/app-web/src/components/DataDetail/DataDetail.tsx
@@ -5,7 +5,7 @@ import { DataDetailMissingField } from './DataDetailMissingField'
 export type DataDetailProps = {
     id: string
     label: string
-    children?: JSX.Element | string | string[]
+    children?: React.ReactNode | string[]
     explainMissingData?: boolean // Display fallback text when data is undefined or null. Should be true on review-and-submit
 }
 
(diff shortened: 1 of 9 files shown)


macrael:
@haworku I think that logic on the review and submit page probably is showing "no" for unset just because that page was written assuming that the form was complete by the time you got to it. If you wanted to change that to explicitly check for false vs. undefined that would make sense, and still show incomplete if any are undefined?

haworku:
@macrael Oh, gotcha. Seems like we should unify then. See [eaaf0e1](https://github.com/CMSgov/managed-care-review/pull/1460/commits/eaaf0e1adba27e75994c9fbf1b5a61e417061bcc)

Just to be explicit, this would change UX submission summary as well for CMS users,  not just for review and submit for state users since its a shared component. For submission summary with this change, we drop missing fields from the page (following the pattern elsewhere).

macrael (approved):
I confirmed that this fixes the bug where if you leave one side out entirely the other showed that something was missing. 

I’m not sure what we would do if we added a new checkbox to the modifiedProvisions list though. In that case we would want to show an error because a specific value was missing from the list, not just that the list was empty as the current DataDetail component is set up to alert on. 

I suspect that the fix for that will probably best come with some of our reconciliation of the various schemas we have for the formData. I would imagine our one “complete form data” parser to return a list of keys of things that are missing that we could pass into the display components. 

All of which is to say lets merge this and keep in mind this case when thinking about how to have better validators in the future.

PR #1462, opened 2023-01-23
Upgrade Aurora Serverless v1 to Postgres 11

mojotalantikite:
## Summary

Postgres 10 is EOL in AWS Aurora, so we need to run the manual in-place upgrade to Postgres 11 on dev/val/prod. This change sets the pg version to 11.13, which is the version that AWS does an in place upgrade to, so that all PR branches get the same version as is deployed in our higher environments. 

Once the dev/val/prod DBs have been upgraded in the AWS console we can merge this. This needs to land before January 31.

We have a separate ticket to do the more involved upgrade to Aurora Serverless v2 with a newer Postgres release.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2577

### DIFF
--- a/services/postgres/serverless.yml
+++ b/services/postgres/serverless.yml
@@ -116,6 +116,7 @@ resources:
       Properties:
         Engine: aurora-postgresql
         EngineMode: serverless
+        EngineVersion: '11.18'
         DatabaseName: '${self:custom.databaseName}'
         MasterUsername: !Sub '{{resolve:secretsmanager:${PostgresSecret}::username}}'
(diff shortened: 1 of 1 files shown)


mojotalantikite:
I've tested this a few times: the in-place upgrade takes about 15 minutes to complete, during which time the DB is unavailable and the application will be down. Merging in this version pin change after the upgrade is basically a no-op for the upgraded DB, but all new deploys in PR branches will deploy a new DB with the version matching this in-place upgrade.

macrael (approved):

PR #1466, opened 2023-01-25
Fix: Screen reader not reading dynamic table caption text.

JasonLin0991:


haworku (approved):
lgtm.

 fyi @rswerve i know you had code maybe close to clear filters button

JasonLin0991:
> lgtm.
> 
> fyi @rswerve i know you had code maybe close to clear filters button

Not much changed with the work @rswerve did. Pulling out the code for updating/clearing filters into functions, the only additional side effect was to set `tableCaption` state to null in these functions to unmount and then remount the caption in `useEffect`.

I'll merge now, but @rswerve also ping me on slack if I broke anything.

PR #1467, opened 2023-01-26
update maintenance banner

macrael:
## Summary

I made a new launch darkly flag to replace the current LD flag that has variations instead of just being a bool. Setup that flag to control the new scheduled maintenance alert vs. the unschedule maintenance alert. 

Moved that alert up from the landing page to AppBody. 
* This means that it shows no matter where someone is in the app
* I haven’t investigated if that would mean there would be data loss if someone left a tab open in our app when maintenance started, I think that risk is worth investigating but not before our first maintenance tomorrow that we have notified people about. 
* It means that whatever page someone is on disappears, replaced by the banner. This too is something we can revisit after tomorrow but would require a lot more changes (each page would need to display the banner on its own) which I’d rather not do before tomorrow.

Also pass a bool down to Header to remove the login link whenever we are under maintenance. 

#### Test cases covered
@JasonLin0991 I had to call the useClientSpy with `{}` even when I didn’t want to override the default value, let’s talk about that I think it would be best if you didn’t have to swizzle the method if you just want the defaults. 

I moved the tests up to AppBody as well

## QA guidance

You should be able to modify this variable in Dev and see it reflected in the review app.

### DIFF
--- a/services/app-web/src/pages/App/AppBody.test.tsx
+++ b/services/app-web/src/pages/App/AppBody.test.tsx
@@ -1,6 +1,7 @@
 import { screen, waitFor } from '@testing-library/react'
 
 import {
+    ldUseClientSpy,
     renderWithProviders,
     userClickSignIn,
 } from '../../testHelpers/jestHelpers'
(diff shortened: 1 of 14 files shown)


haworku (approved):
Approving! Thank you 👏 

A couple things I would like to see addressed though 
- Current UI looks quite odd to me. There could be ways to address even without a global component
- fix ` ErrorAlertFailedRequest` so its using same content. Quick glance at storybook looks like that alert still uses old lang pattern for mailto

haworku:
> Or have a separate flag that let's us turn the API off.
thats a nice follow on - I think it should be a separate flag. It would be good to have that toggle for other reasons too.

services/app-web/src/pages/App/AppBody.test.tsx
```diff
@@ -72,6 +73,46 @@ describe('App Body and routes', () => {
...
+        expect(
+            await screen.findByText(
+                /MC-Review is currently unavailable due to technical issues/
+            )
+        ).toBeInTheDocument()
+    })
+
+    it('does not display maintenance banner when flag is off', async () => {
+        ldUseClientSpy({})
+        renderWithProviders(<AppBody authMode={'AWS_COGNITO'} />, {
```
JasonLin0991, line 98: > I had to call the useClientSpy with {} even when I didn’t want to override the default value, let’s talk about that I think it would be best if you didn’t have to swizzle the method if you just want the defaults.

I think the `jest.resetAllMocks()` at the top level `describe` does not apply to each test in the `describe` block for 'Environment specific banner'

```
describe('App Body and routes', () => {
    afterEach(() => {
        jest.resetAllMocks()
    })
```

Adding a `jest.resetAllMocks()` in the `beforeEach()` should clear out the `useClientSpy` mock that was called in the previous test.

```
    describe('Environment specific banner', () => {
        const OLD_ENV = process.env
        beforeEach(() => {
            jest.resetModules() // Most important - clears the cache
            process.env = { ...OLD_ENV } // Make a copy
        })
```

JasonLin0991: @macrael I ran the tests without calling `useClientSpy({})` and it still passed. Maybe we can pair and you could show me what is going on.

services/app-web/src/pages/App/AppBody.tsx
```diff
@@ -39,10 +73,16 @@ export function AppBody({
...
+                authMode={authMode}
+                setAlert={setGlobalAlert}
+                disableLogin={siteUnderMaintenance}
+            />
             <main id="main-content" className={styles.mainContent} role="main">
                 {globalAlert && globalAlert}
-                {loginStatus === 'LOADING' ? (
+                {maintenanceBanner ? (
+                    maintenanceBanner
+                ) : loginStatus === 'LOADING' ? (
```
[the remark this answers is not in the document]
macrael: That’s a way better idea, I just didn’t think of it. Lemme see if I can swing that real quick

services/app-web/src/pages/App/AppBody.test.tsx
macrael: Let’s pair on this, what I’m seeing is that if you don’t call ldUseClientSpy anywhere in your test, all of our flag calls just return undefined because the launch darkly client isn’t setup.

PR #1469, opened 2023-01-26
Upgrade to node v18.13.0 and update uploads service

macrael:
## Summary

Upgrade to the highest node version supported by AWS

This PR has grown beyond just node, it also includes porting the uploads service to typescript, and now adding an audit service. 

#### Related issues

https://qmacbis.atlassian.net/browse/MR-2278


macrael:
Also includes upgrading serverless and serverless-offline

services/uploads/src/updateAVDefinitions.test.ts
```diff
@@ -0,0 +1,42 @@
+import { NewClamAV } from "./clamAV"
+import { NewTestS3UploadsClient } from "./s3"
+import { updateAVDefinitions } from "./updateAVDefinitions"
+import path from 'path'
+
+describe('updateAVDefinitions', () => {
```
mojotalantikite, line 6: 🎉 🎉 🎉

services/uploads/src/clamAV.ts
```diff
@@ -0,0 +1,212 @@
...
+        scanLocalFile: (path: string) => scanLocalFile(fullConfig, path),
+        fetchAVDefinitionsWithFreshclam: (workdir: string) => fetchAVDefinitionsWithFreshclam(fullConfig, workdir),
+    }
+}
+
+/**
+ * Uploads the AV definitions to the S3 bucket.
+ */
+async function uploadAVDefinitions(config: ClamAVConfig, s3Client: S3UploadsClient, workdir: string) {
+    // delete all the definitions currently in the bucket.
```
mojotalantikite, line 51: I'd be interested for us to time this whole process of running `freshclam` to get definitions, clearing out the bucket, and then copying the new ones to the s3 bucket again. We could probably save some time by just calling the [s3 sync command](https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html) with the CLI, which is sort of like `rsync` in that we'll only copy the diff. Not sure we'd want to re-implement that with the JS SDK though, probably just use the tool directly.

Maybe we add some timing logs in this PR, ship it, and then decide if it makes sense to do some follow on work.

services/app-api/package.json
```diff
@@ -56,6 +56,7 @@
         "path-to-regexp": "^6.2.1",
         "protobufjs": "^6.11.2",
         "request": "^2.88.2",
+        "serverless": "^3.26.0",
```
mojotalantikite, line 59: Probably don't need this in deps if we have it in devDeps?

services/uploads/serverless.yml
```diff
@@ -86,34 +88,48 @@ layers:
...
   avScan:
-    handler: src/antivirus.lambdaHandleEvent
+    handler: src/avScan.avScanLambda
     name: ${self:service}-${sls:stage}-avScan
     timeout: 300 # 300 seconds = 5 minutes. Average scan is 25 seconds.
     memorySize: 3008
     layers:
       - !Ref ClamDefsLambdaLayer
-      - arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-nodejs-amd64-ver-1-0-1:1
+      - arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-nodejs-amd64-ver-1-8-0:2
```
mojotalantikite, line 97: I'm guessing this is working fine since we're not actually instrumenting anything with OTEL in this service? This is the layer upgrade I want to schedule so that we can bump all our OTEL libs.

services/uploads/src/auditUploads.ts
```diff
@@ -0,0 +1,114 @@
+import { Context, S3Event } from 'aws-lambda';
+import { rm, stat, readdir } from 'fs/promises'
+import path from 'path'
+
+import { NewS3UploadsClient, S3UploadsClient } from './s3'
+
+import { NewClamAV, ClamAV } from './clamAV'
+
+export async function auditUploadsLambda(event: S3Event, _context: Context) {
```
mojotalantikite, line 9: Thanks for including this! It was the first thing I thought of when I read your pings this morning, and you already did it ❤️ .

Do we want to set this on a cron schedule for the lambda? There's an example of that serverless config in the `cleanup` lambda in `app-api`. Couldn't hurt to have a semi-regular full bucket av scan happening.

mojotalantikite (approved):
Oh man, so excited the AV scanning stuff made it over to TypeScript! ❤️ for doing it.

services/uploads/src/clamAV.ts
macrael: It’s only 3 or 4 files so I think it’s pretty quick. That could be sweet though, we’re syncing in both directions, uploading the new ones and downloading them for clamscan

services/uploads/src/auditUploads.ts
macrael: that’s a good idea. Need to figure out how for it to report errors to us then

services/uploads/src/updateAVDefinitions.test.ts
macrael: For some reason it was having its connection refused to the local s3 instance we have running, gotta chase that down, but there’s no reason we can’t test all this code locally.

services/uploads/serverless.yml
macrael: I’ll back this down, I tried an upgrade when I didn’t know what the problems were yet

PR #1470, opened 2023-01-26
MR-2994 Fix focus on rerender

rswerve:
## Summary

There were a few outstanding issues from putting the filter state into the URL.

1) Focus was lost on the clicked element on rerender.  This fixes that.
2) A Typescript warning was being silenced. Using react-table's provided types and letting Jotai infer from those fixed that issue.
3) The URL is still URI encoded and hard to read.  This was partially addressed.  We can turn off encoding _within_ the values, but not between the values.  For example, what was 
 ```Contract%2520action%2520and%2520rate%2520certification``` 
is now  
```Contract+action+and+rate+certification```

But we still have substrings like  ```stateName%3DVirginia%252CMinnesota```

and that's because Jotai [runs `toString` on the URLParams](https://github.com/jotai-labs/jotai-location/blob/adabca165c3486cade23efa238f1f27d16255ff1/src/atomWithHash.ts#L91).  This is actually a reasonable thing for them to do.

#### Test cases covered

Added some tests to make sure the URL was what we expect, and one test to check focus on the "clear filters" button.  I could **not** get focus to work properly for react-select inside jest.  I gave up after the better part of a day.

### DIFF
--- a/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
+++ b/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useRef } from 'react'
 import {
     createColumnHelper,
     flexRender,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx [hunk 1]
```diff
@@ -148,8 +142,38 @@ export const HealthPlanPackageTable = ({
...
-    const [columnFilters, setColumnFilters] =
-        useAtom<ColumnFilter[]>(columnHash)
+    const lastClickedElement = useRef<string | null>(null)
+    const [columnFilters, setColumnFilters] = useAtom(columnHash)
+
+    /* we store the last clicked element in a ref so that when the url is updated and the page rerenders
+        we can focus that element.  this useEffect (with no dependency array) will run once on each render. 
+        Note that the React-y way to do this is to use forwardRef, but the clearFilters button is deeply nested 
+        and we'd wind up passing down the ref through several layers to achieve what we can do here in a few lines 
+        with DOM methods */
```
macrael, line 152: I apprecaite this comment! Good to be so explicit when doing something unexpected.

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
```diff
@@ -312,18 +336,20 @@ export const HealthPlanPackageTable = ({
...
         column.setFilterValue(
             selectedOptions.map((selection) => selection.value)
         )
     }
 
     const clearFilters = () => {
+        lastClickedElement.current = 'clearFiltersButton'
         setTableCaption(null)
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        //@ts-ignore
```
macrael, line 326: also yay

macrael (approved):
fab

PR #1480, opened 2023-02-01
MR-2885: CMS and state see sidebar navigation with empty q&a table

JasonLin0991:


tests/cypress/integration/stateWorkflow/questionsAnswers/questionAnswers.spec.ts
```diff
@@ -0,0 +1,146 @@
+describe('Q&A', () => {
+    beforeEach(() => {
+        cy.stubFeatureFlags()
+    })
+
+    it('cannot navigate to and from Q&A page with cms-questions flag off', () => {
+        cy.interceptFeatureFlags({
```
haworku, line 7: (issue: blocking question): It seems like there's a very small part of this test's assertions that we are actually using. Can we find a solution where we don't add a test that makes a whole submission to achieve this? I want to avoid adding new standalone tests that are for something small like this because of cypress flakes and runtime.
```
 //Try to visit Q&A page
    cy.visit(`/summary/${submissionId}/q&a`)

     // Look for 404 text on page
      cy.findByRole('heading', { name: '404 / Page not found', level: 1}).should('exist')
```
I suggest attaching those assertions to an existing test for now since we don't have stubbed request yet. For example can we add this to the bottom of  `can navigate to and from Q&A page` and switch the flag right before? Let me know if you think this is workable. I would to more precise and focused with our cypress stuff since we have had ...issues... in the past.

services/app-web/src/pages/SubmissionSideNav/SubmissionSideNav.tsx
```diff
@@ -0,0 +1,181 @@
...
+                            >
+                                <svg
+                                    className="usa-icon"
+                                    aria-hidden="true"
+                                    focusable="false"
+                                    role="img"
+                                >
+                                    <use xlinkHref={`${sprite}#arrow_back`} />
+                                </svg>
+                                {loggedInUser?.__typename === 'StateUser' ? (
```
haworku, line 146: 🎨 style nit: The blue on these links look wrong now. Maybe be nested scss issue. Let me know if you want to look together 
Screenshot 2023-02-03 at 4 28 53 PM

haworku (approved):
Thanks for this great setup! I didn't realize this PR was out of draft, just caught it now. Please take a look at the issues I flagged as blocking

services/app-web/src/pages/App/AppRoutes.tsx
```diff
@@ -67,10 +72,26 @@ const StateUserRoutes = ({
...
+                            }
+                            element={<QuestionsAnswers />}
+                        />
+                    )}
+                    <Route
+                        path={RoutesRecord.SUBMISSIONS_SUMMARY}
+                        element={<SubmissionSummary />}
+                    />
+                </Route>
+                {showQuestionsAnswers && (
```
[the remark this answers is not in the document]
JasonLin0991: We wouldn't be able to wrap `SUBMISSIONS_UPLOAD_QUESTION` and `SUBMISSIONS_QUESTIONS_AND_ANSWERS` together because `SUBMISSIONS_QUESTIONS_AND_ANSWERS` is a nested route inside the `SubmissionSideNav` layout route.

tests/cypress/integration/stateWorkflow/questionsAnswers/questionAnswers.spec.ts
JasonLin0991: Good suggestion!

services/app-web/src/pages/SubmissionSideNav/SubmissionSideNav.tsx
JasonLin0991: Nice catch. I check val to make sure it was just my branch, but looks like the link color is also incorrect there too.

Screenshot 2023-02-06 at 11 24 07 AM

JasonLin0991: Looks like the fix was to remove `variant="unstyled"` attribute from the `Link` component.

services/app-web/src/pages/App/AppRoutes.tsx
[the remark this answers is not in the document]
haworku: Oh yes! Didn't see that.

haworku (approved):

services/app-web/src/pages/SubmissionSideNav/SubmissionSideNav.tsx
haworku: Yeah these discrepancies with styling is something I was noticing too with the USWDS work. Plus also feels like the unstyled variant shouldnt override custom css on a component passed in by user 🤔 that might be something worth changing in react-uswds.

Thanks for making the change!

PR #1494, opened 2023-02-07
Switch to Launch Darkly Federal

mojotalantikite:
## Summary

This moves us to using the Launch Darkly Federal instance. That installation is behind CMS SSO at https://mo-idp.cms.gov .

All the feature flags have been copied over from our current install to the fed environment.

I've added new keys to parameter store to store our API keys in dev, val, and prod, which is read into the `LD_SDK_KEY` and `REACT_APP_LD_CLIENT_ID` env variables. All devs will need to update their local SDK keys (in .envrc.local) to use the new keys, which can be found in the settings of the Launch Darkly web app.

### Testing

It'd be great if a few devs could bang on this in both local and dev/PR.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2626
https://qmacbis.atlassian.net/browse/MR-2627
https://qmacbis.atlassian.net/browse/MR-2628


mojotalantikite:
A couple other things that made it into this PR:
- OTEL containers were erroring locally as I was working on this, and I needed to fix up the config to clear out some noise
-  The cors configs on recent versions of serverless were updated to include headers they previously weren't including, so the config can be less verbose

I can pull both of those out into other PRs if people prefer, they were just left in my branch from my hacking.

README.md
```diff
@@ -309,6 +309,12 @@ We currently are using Open Telemetry for distributed tracing, with our OTEL exp
 
 You should now be at our New Relic dashboard where all our OTEL metrics are being displayed.
 
+## Launch Darkly
+
+We currently use the CMS Federal (.us) install of Launch Darkly to manage our feature flags. This can be accessed through the [CMS SSO portal](https://mo-idp.cms.gov) if you have the appropriate EUA job code.
```
macrael, line 314: Is there a real url we can put here that redirects through the mo-idp url? Or do you have to go to the portal to get in (I still haven’t gotten access yet)

macrael (approved):
Looks good, no need to split out this work into different PRs, I’m always a fan of cleaning things up as you go, I’d only do multiple PRs if it got really big

README.md
mojotalantikite: This is what they gave to us, but it looks like if I hit app.launchdarkly.us and give it our `@teamtrussworks` email address it'll redirect for us. I'll update these docs.

haworku (approved):

PR #1495, opened 2023-02-10
Productionize Audit Bucket Function

macrael:
## Summary
Still a WIP but I’m making the audit bucket function work right. 

* Will split up scanning files into their own lambdas so it all happens in parallel and doesn’t hit the 15 minute lambda time limit. 
* Will fix any mis-tagged files to be tagged correctly. 
* Has local tests against the local s3 instance uploads is already running

The file structure is still very much a work in progress, it will be cleaned up once the funciton itself is working. 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3028


dev_tool/src/test/uploads.ts
```diff
@@ -0,0 +1,27 @@
+import { spawn } from 'child_process'
+import { requireBinary } from '../deps.js'
+
+function checkUploadsDeps() {
+    const helpString =
+        'ClamAV must be installed and in $PATH in order to test uploads. You can install with `brew install clamav`'
```
mojotalantikite, line 6: Should we add `clamav` into the list of packages to install with `brew` under the `Local Tooling` section of our Readme?

services/uploads/src/deps/clamAV/clamAV.test.ts
```diff
@@ -0,0 +1,80 @@
+import path from 'path'
+import { NewTestS3UploadsClient } from '../s3'
+import { NewClamAV, parseInfectedFiles } from './clamAV'
+import { rm, mkdtemp } from 'fs/promises'
+
+describe('clamAV', () => {
+    it('parses clamscan output correctly', () => {
+        const tests = [
+            {
+                output: `/Users/macrae/ccode/mc-review/managed-care-review/services/uploads/src/clamAV/testData/badList.csv: Win.Test.EICAR_HDB-1 FOUND
```
mojotalantikite, line 10: I know these paths don't matter for this parsing test, but do we want to make the path string a little more generic than your local checkout? Just thinking of a junior dev 5 years from now wondering what this path is and who `macrae` is 😂

mojotalantikite (approved):
Looks great, thanks for getting all of this taken care of! 

Once I got the jest cache cleared out local tests worked fine.

dev_tool/src/test/uploads.ts
macrael: done

PR #1496, opened 2023-02-13
MR-2990: Add index and create questions GQL operations

JasonLin0991:
## Summary
[MR-2990](https://qmacbis.atlassian.net/browse/MR-2990)

- Add `createQuestion` and `indexQuestions` GQL operations
- Users added into postgres db when logging in on `local` env.
   - This is needed for testing `createQuestions` resolver for relating questions to users in the database
- **Apollo Sandbox** connection documentation
- Commented out parts of the graphql schema that are not used yet.
- Update diagram of database in technical design docs.
   - [Link](https://github.com/CMSgov/managed-care-review/blob/jl_mr-2990_add_index_and_create_questions_gql_operations/docs/technical-design/database-diagram.md) to the updated diagram since it's hard to see it from the `Files changed` tab.

#### Test cases covered

- `createQuestion.test.ts`
   - `'returns question data after creation'`
   - `'allows question creation on UNLOCKED and RESUBMITTED package'`
   - `'returns an error package status is DRAFT'`
   - `'returns an error if a state user attempts to create a question for a package'`
   - `'returns error on invalid package id'`
- `indexQuestions.test.ts`
   - `'returns question data after creation'`
   - `'allows question creation on UNLOCKED and RESUBMITTED package'`
   - `'returns an error package status is DRAFT'`
   - `'returns an error if a state user attempts to create a question for a package'`
   - `'returns error on invalid package id'`


services/app-api/src/postgres/questionAnswers/insertQuestion.ts
```diff
@@ -0,0 +1,62 @@
...
+import { convertPrismaErrorToStoreError, StoreError } from '../storeError'
+import { Question, CreateQuestionInput } from '../../domain-models'
+import { v4 as uuidv4 } from 'uuid'
+
+export async function insertQuestion(
+    client: PrismaClient,
+    questionInput: CreateQuestionInput,
+    user: CMSUserType
+): Promise<Question | StoreError> {
+    const questionData = {
```
macrael, line 12: Seems fine to box this up into its own variable, but not required. Was this from an earlier attempt at just passing the object into create()?

services/app-api/src/postgres/questionAnswers/insertQuestion.ts
```diff
@@ -0,0 +1,62 @@
...
+                    include: {
+                        stateAssignments: true,
+                    },
+                },
+            },
+        })
+
+        const createdQuestion: Question = {
+            ...result,
+            addedBy: result.addedBy as CMSUserType,
```
macrael, line 55: Maybe we could just drop the `include: { addedBy: …` from above and just return the `user` that the function was called with? No need to query for that since we already have it.

services/app-api/src/resolvers/questionAnswers/createQuestion.test.ts
```diff
@@ -0,0 +1,303 @@
...
+                    ],
+                },
+            },
+        })
+
+        // Expect no errors when creating question for resubmitted package
+        expect(createdQuestionOnResubmit.errors).toBeUndefined()
+
+        // Expect two questions from indexQuestions
+        const indexQuestionsResult = await cmsServer.executeOperation({
```
macrael, line 128: Sweet, way to check the work with both APIs together.

services/app-api/src/resolvers/questionAnswers/createQuestion.test.ts
```diff
@@ -0,0 +1,303 @@
...
+                '5c10fe9f-bec9-416f-a20c-718b152ad633',
+                '037af66b-81eb-4472-8b80-01edf17d12d9',
+            ],
+            riskBasedContract: false,
+            submissionType: 'CONTRACT_ONLY',
+            submissionDescription: 'A real submission',
+            contractType: 'BASE',
+        }
+
+        const draftPkg = await stateServer.executeOperation({
```
macrael, line 203: We should have a helper test function for this, `createTestHealthPlanPackage`, any reason not to use that instead of manually executing the CREATE_HEALTH_PLAN_PACKAGE request?

macrael: In general, I’ve mostly found that I only use the .executeOperation fucntion on the resolver i’m actively testing and try and use test helper functions for everything else.

services/app-api/src/resolvers/questionAnswers/createQuestion.test.ts
```diff
@@ -0,0 +1,303 @@
...
+            ],
+        }
+
+        const createdQuestion = await stateServer.executeOperation({
+            query: CREATE_QUESTION,
+            variables: { input },
+        })
+
+        expect(createdQuestion.errors).toBeDefined()
+        expect(
```
macrael, line 257: This is where the error helper functions in `gqlAssertions.ts` could be helpful

services/app-api/src/resolvers/questionAnswers/createQuestion.test.ts
```diff
@@ -0,0 +1,303 @@
...
+            createdQuestion.errors && createdQuestion.errors[0].message
+        ).toBe('user not authorized to create a question')
+    })
+    it('returns error on invalid package id', async () => {
```
macrael, line 261: Since you added the NOT_FOUND extension, let’s have the test that checks that comes through

services/app-api/src/resolvers/questionAnswers/indexQuestions..ts
```diff
@@ -0,0 +1,71 @@
+import { isStateUser, IndexQuestionsPayload } from '../../domain-models'
```
macrael, line 1: This filename has an extra `.` in it

services/app-api/src/resolvers/questionAnswers/indexQuestions.test.ts
```diff
@@ -0,0 +1,209 @@
...
+        const server = await constructTestPostgresServer()
+        const cmsServer = await constructTestPostgresServer({
+            context: {
+                user: testUserCMS,
+            },
+        })
+
+        const submittedPkg = await createAndSubmitTestHealthPlanPackage(server)
+
+        await cmsServer.executeOperation({
```
macrael, line 28: You could create a helper function for creating questions like we did for setting up the packages. For test helper functions it’s OK if they throw an error if something goes wrong since that will end the test ASAP

services/app-api/src/resolvers/questionAnswers/indexQuestions.test.ts
```diff
@@ -0,0 +1,209 @@
...
+        })
+
+        expect(result.errors).toBeDefined()
+        if (result.errors === undefined) {
+            throw new Error('annoying jest typing behavior')
+        }
+        expect(result.errors).toHaveLength(1)
+        const resultErr = result.errors[0]
+
+        expect(resultErr?.message).toBe(
```
macrael, line 179: Check that the code == ‘FORBIDDEN’ here, too

services/app-api/src/testHelpers/gqlHelpers.ts
```diff
@@ -83,6 +84,9 @@ const constructTestPostgresServer = async (opts?: {
 
     const prismaClient = await sharedTestPrismaClient()
     const postgresStore = opts?.store || NewPostgresStore(prismaClient)
+
+    await insertUserToLocalAurora(postgresStore, context.user)
```
macrael, line 88: what happens if the user already exists? will this silently fail? We should check if there is an error an throw it if it’s not because the user already exists.

macrael (approved):
This is stellar work. The only thing that might be missing is an update to the mappers in our codegen.yml file. See how HealthPlanPackage is defined there? Try and do the same for Question and see if anything breaks. It obviously works just fine for now, but we might want  that mapping setup in the future. That mapping is required if we ever want to implement a QuestionResolver on its own like we do for those other Types. 

Ok, actually, after typing that out, I think you can leave it out until we have a need for that, but I want to make sure you see that spot as a potential place for changes in the future since it’s kindof invisible.

services/app-api/src/resolvers/questionAnswers/createQuestion.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        if (isStoreError(result)) {
+            const errMessage = `Issue finding a package of type ${result.code}. Message: ${result.message}`
+            logError('createQuestion', errMessage)
+            setErrorAttributesOnActiveSpan(errMessage, span)
+            throw new Error(errMessage)
+        }
+
+        if (result === undefined) {
+            const errMessage = `Issue finding a package with id ${input.pkgID}. Message: Result was undefined`
```
macrael, line 36: Quibble: I think this message (and the same one for indexQ) could be more specific to what happened, which is that they tried to ask a question on a package that does not exist. Saying “Result was undefined” is an implementation detail to how we defined the postgres function and not really helpful to someone making the query, I think.

services/app-api/src/testHelpers/gqlHelpers.ts
mojotalantikite: it looks like he does `lookupUserAurora()` inside `insertUserToLocalAurora()`, but if we want to do it the way that we're currently doing it we'd want to do the checking before calling the insert func. 

It's a good question though, do we prefer doing the insert or should we call [`upsert`](https://www.prisma.io/docs/concepts/components/prisma-client/crud#update-or-create-records), which will just update the entry if it exists and don't do the check at all?

services/app-api/src/postgres/questionAnswers/insertQuestion.ts
JasonLin0991: I don't remember exactly what I was trying to do. I think it may have been left over from a previous version where `insertQuestion` calls another function to do the actual store and that was the object being passed in. Rolled all of that into `insertQuestion`, so that variable is not needed. Although I will keep the question mapping as a variable instead of putting it in the query.

services/app-api/src/testHelpers/gqlHelpers.ts
JasonLin0991: > what happens if the user already exists? will this silently fail? We should check if there is an error and throw it if it’s not because the user already exists.

It would silently fail to store the user and return the `context.user`. The function is used in local deployment, mimicking what we do in other environments. 

Your question brings up something I overlooked: if `lookupUserAurora` or storing the user results in an error, it would also cause question tests to fail. The larger issue, I think, is that failing to store a user in any environment will also cause `createQuestions` to fail because it will not be able to relate a question to a user in the database.

@macrael @mojotalantikite @rswerve @haworku Maybe we should discuss this in an informal meeting?

JasonLin0991: > It's a good question though, do we prefer doing the insert or should we call [upsert](https://www.prisma.io/docs/concepts/components/prisma-client/crud#update-or-create-records), which will just update the entry if it exists and don't do the check at all?

Also a good topic to hit on in the informal meeting.

macrael: Let’s bring this to that meeting, but it doesn’t need to block merging this PR since the current system works. Also it’s totally fine for that insert to fail the tests, anything breaking during a test should fail the test.

PR #1497, opened 2023-02-14
MR-2798: Tweaks to our CSV report

rswerve:
## Summary

As detailed in the ticket, the team wanted some tweaks to the CSV that we're using as a basis for reports about what's in the system.  As discussed with Meghan in Slack, there are a few departures from the ticket:
1. We don't need another column for the submission date.  That data is already in the report.
2. We decided NOT to include a column for UNLOCKED and RESUBMITTED status.  Meghan can get that info from existing columns.
3. We are excluding DRAFT status, _except_ when it's an unlocked submission.


services/app-api/src/handlers/reports.ts
```diff
@@ -91,10 +109,26 @@ export const main: APIGatewayProxyHandler = async () => {
             console.error('Error decoding revision')
             throw new Error(`Error generating reports array`)
         } else {
-            bucket.push(revision)
+            // add the package name to the revision
```
macrael, line 112: This all looks right. 

I haven’t looked through this handler before, how does this get invoked? I don’t see anything limiting this to ADMIN users which I thought was a requirement for this.

macrael (approved):

services/app-api/src/handlers/reports.ts
rswerve: It's restricted to CMS users at the route level.

macrael: Can you show me? I see us restricting to logged in users in serverless.yml but I don’t see any restriction based on user role

rswerve: Added a bit to restrict the endpoint itself to cms and admin users 🙂

macrael: Nice, yay!

PR #1501, opened 2023-02-14
Fix up yarn workspaces

mojotalantikite:
## Summary

Looks like the OIDC service got it's own package lockfile which throws off the monorepo. This cleans that up to let `yarn` take ownership of the service.


macrael (approved):
I guess nothing new ended up in our yarn.lock after this

PR #1507, opened 2023-02-16
Set our reviewapp deployments to be transient

macrael:
## Summary

Something changed in GH and our “deployment” lines in PRs have stopped having links in them that work pointing to the review app. 

All of our deployments are set to the environment: reviewapps, and it looks like now when a new review app deploys, it sets all previous deployments for “reviewapps” to be inactive. 

First try is to set each deployment to be “transient” which prevents that auto-inactive problem. But then we might want to delete them in destroy or something. We’ll see.

Also, setting environment: dev as part of our OIDC work seems to [unavoidably](https://github.com/actions/runner/issues/2120) create a deployment each time, which is why there are 3 extra “deployment” notifications per deployment in the PR history.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -76,12 +76,13 @@ jobs:
         run: |
           echo "${{steps.changed-services-action.outputs.changed-services}}"
 
-      - uses: chrnorm/deployment-action@releases/v2
+      - uses: chrnorm/deployment-action@v2
         name: Create GitHub deployment
         id: ghdeployment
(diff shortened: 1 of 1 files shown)


haworku:
I wonder if the work from #1512 will impact this or resolve it? I don't know enough about whats going on here just noticed a possible related PR open

macrael:
@haworku that’s going to address a different issue, I’m hoping this PR just makes our deployment links clickable again

rswerve (approved):

PR #1511, opened 2023-02-17
Pin postgres to 11.16 

mojotalantikite:
## Summary

We were contacted by CMS support to tell us that we had some DBs deployed with the old, unsupported Postgres 10.x line in AWS. For whatever reason AWS will still deploy the unsupported version of 10.x Postgres that's been EOL in AWS when making new DBs.

All the DBs in dev/val/prod were updated to 11.x already, but new PR review environments are getting the 10.x line. This pins the version to 11.16.

### DIFF
--- a/services/postgres/serverless.yml
+++ b/services/postgres/serverless.yml
@@ -116,6 +116,7 @@ resources:
       Properties:
         Engine: aurora-postgresql
         EngineMode: serverless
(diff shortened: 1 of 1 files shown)


haworku (approved):

mojotalantikite:
With #1545 merged this is no longer a need, we're running aurora v2 now.

PR #1518, opened 2023-02-23
Upgrade OTEL layer and libs to latest (1.9.1)

mojotalantikite:
## Summary

This upgrades us to using the latest [AWS Distro for OTEL layer](https://aws-otel.github.io/docs/getting-started/lambda/lambda-js). It also includes all the lib and config upgrades necessary for `app-web` and `app-api`.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3050


services/app-web/serverless.yml
```diff
@@ -74,7 +74,7 @@ custom:
         export REACT_APP_S3_DOCUMENTS_BUCKET=${self:custom.s3_documents_bucket_name}
         export REACT_APP_S3_QA_BUCKET=${self:custom.s3_qa_bucket_name}
         export REACT_APP_STAGE_NAME=${sls:stage}
-        export REACT_APP_OTEL_COLLECTOR_URL=${self:custom.application_endpoint_url}/otel
+        export REACT_APP_OTEL_COLLECTOR_URL=${self:custom.api_url}/otel
```
haworku, line 77: This change confuses me. Can you explain this a bit like aren't these totally different urls

haworku (approved):
You mentioned this upgrade improved the logs would like to know more about that sometime or whats the improvement we see

macrael (approved):

services/app-web/serverless.yml
mojotalantikite: Ah yeah, good catch. You're right, they're totally different in review apps, but in dev/val/prod they're the same. We weren't getting frontend logs in review apps, but I needed to turn them on for testing this. For example, in review apps it was evaluating to the cloudfront distro with the `application_endpoint_url` var, but in dev/val/prod it evaluates to the set cloudfront domain (e.g `mc-review-val.onemac.cms.gov`) which is the same as the `api_url`. 

I think it's worth keeping it on for times people want to reference frontend logs in review apps in New Relic.

Also of note, the real place we have cared about `REACT_APP_OTEL_COLLECTOR_URL` is in `app-api` where we use it to set the value to the collector of the lambda layer.

PR #1522, opened 2023-02-28
MR-2888: CMS and state can view CMS questions

JasonLin0991:
## Summary
[MR-2888](https://qmacbis.atlassian.net/browse/MR-2888)

AC:
- A list of all questions and responses is displayed on the Q&A page.
- if a document is uploaded by you, say “you”. if a document is uploaded by CMS say “CMS”
   - When the [graphql-explorer embed work](https://qmacbis.atlassian.net/browse/MR-2959) is merged in, you should be able to test this out locally using `createQueston` mutation and modifying the headers in explorer by giving the `cognito` header user details. Specifically, the user id needs to be different. There is a unit test for this so  I don't think it's necessary to test it.
- Table is ordered by latest (by createdAt) on top
- `"No grouping of questions is done, just a complete list with all the relevant data"`
   - This PR does group question and their responses on a table in order to add the `Upload response` button

Notes:
- Querying questions are no longer being done in `findHealthPlanPackage` store. Instead, it is done in `healthPlanPackageResolver`.
- Refactored getting document URL into a hook, so it can be reused on the QA page.
- Take note in `UploadQuestion.tsx` and `UploadResponse.tsx`, when a question or response is successfully created we are `cache.evict()` the health plan data in order to force a refetch once navigated back to the QA page.
- I started renaming all `QuestionAnwers` files and functions to `QuestionResponse`.

#### Test cases covered

Cypress Test:
- `questionResponse.spec.ts`
   - `'cms-questions flag toggles navigation to and from Q&A page'` replaced with `'can add and display cms questions'`

Unit Tests:
- `indexQuestion.test.ts`
   - `'returns package with question'` replaced with `'returns package with questions and responses'` 

- `SubmissionSideNav.test.tsx`
   - `'shows a generic error page when user is undefined'`

- `QuestionResponse.test.tsx`
   - `'renders expected questions correctly'`
   - `'renders division sections correctly with no questions'`

- `QATable.test.tsx`
   - `'renders question correctly as a CMS user that created the question'`
   - `'renders question correctly as a CMS user that did not create the question'`
   - `'renders question correctly as a state user'`
   - `'renders multiple documents and links correctly'`


JasonLin0991:
This PR is waiting on [MR-2887](https://github.com/CMSgov/managed-care-review/pull/1524). Just realized one of the AC is to display responses which is MR-2887.

services/app-api/src/resolvers/healthPlanPackage/healthPlanPackageResolver.ts
```diff
@@ -2,40 +2,63 @@ import { packageStatus, packageSubmittedAt } from '../../domain-models'
...
-    },
+            if (state === undefined) {
+                throw new Error('State not found in database: ' + packageState)
+            }
+            return state
+        },
+        questions: async (parent, args, context) => {
+            const { span } = context
+            const pkgID = parent.id
+            const result = await store.findAllQuestionsByHealthPlanPackage(
```
haworku, line 51: if we are always including questions on the health plan package response (which is my understand of the intent behind this code change)  I don't think we have to do this is in the separate call to store in the resolver. 

We can just prisma `include` the questions field on the db level query.  In other words, I would handle this directly in the initial prisma queries - things like findHealthPlanPackage or the findAll functions. That way the error handling also is handled there in an existing place.

JasonLin0991: > if we are always including questions on the health plan package response
This db query for questions only happens if the graphql query includes questions in the query. So queries like `indexHealthPlanPackages` and `fetchHealthPlanPackage` which doesn't won't call that function in the `healthPlanPackage` resolver.

The cool thing is if you wanted questions in those queries you just need to include the question field in those queries without having to go into those resolvers and manually query the db.

JasonLin0991: > We can just prisma include the questions field on the db level query.

We could do this! I actually went down this road until @macrael showed me the light of adding the question field to the `healthPlanPackageResolver`. Also adding this to `fetchHealthPlanPackage` would also grab questions when on the submission form.

haworku: ohhh I see thats nice. Gotcha! I understand now the benefits. i like that composability element and didn't realize it wouldn't be called every time. sweet

macrael: @JasonLin0991 I’d love for you to demo this in our next eng hangout meeting, it’s neat!

services/app-api/src/resolvers/questionResponse/index.ts
```diff
@@ -0,0 +1,2 @@
+export { indexQuestionsResolver } from './/indexQuestions.'
```
macrael, line 1: extra slashes here

services/app-graphql/src/schema.graphql
```diff
@@ -26,6 +26,21 @@ type Query {
...
+
+    If a package with the given ID cannot be found, this query returns undefined
+    CMS users cannot fetch a DRAFT HealthPlanPackage
+
+    Errors:
+    - ForbiddenError:
+    - A state user from a different state called this.
+    - A CMSUser attempted to fetch a DRAFT HealthPlanPackage
+    """
+    fetchHealthPlanPackageWithQuestions(
```
macrael, line 40: This should not be neccessary. we should still have a single fetchHealthPlanPackage Query that can be called either requesting questions or not.

services/app-web/src/hooks/useDocument.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useS3 } from '../contexts/S3Context'
+import { BucketShortName } from '../s3/s3Amplify'
+import { SubmissionDocument } from '../common-code/healthPlanFormDataType'
+
+type UnionDocumentType = Document | SubmissionDocument
+type ParsedDocumentWithLinkType =
+    | ({ url: string | null } & Document)
+    | SubmissionDocument
+
+const useDocument = () => {
```
macrael, line 11: is this stuff merged in Hana’s earlier PR or is it new with this PR? You might want to do a merge from main

services/app-web/src/pages/SubmissionSideNav/SubmissionSideNav.tsx
```diff
@@ -54,8 +81,80 @@ export const SubmissionSideNav = () => {
...
+                    questions.OACTQuestions.edges
+                ),
+            },
+        }
+
+        return divisionQuestions
+    }
+
+    const { result: fetchResult } =
+        useFetchHealthPlanPackageWithQuestionsWrapper(id, async (data) => {
```
macrael, line 134: I’m surprised you’re using the callback here instead of parsing questions out of the result data below like we get the package results etc.

macrael (approved):
I think we don’t need that fetchHealthPlanPacakgeWithQuestions Query added to our API, we should be able to call the fetchHPP Query using our separate query file. 

This looks great though, I think my priority is trying to get this in for demo tomorrow, so feel free to merge as is and we can work out the better way to handle the query file stuff after demo.

services/app-web/src/hooks/useDocument.tsx
JasonLin0991: This is new from this PR.

PR #1523, opened 2023-02-28
MR-2950: Embed graphql explorer into app to verify api only work.

JasonLin0991:
## Summary
[MR-2950](https://qmacbis.atlassian.net/browse/MR-2959)

### DIFF
--- a/services/app-web/package.json
+++ b/services/app-web/package.json
@@ -74,6 +74,7 @@
     },
     "dependencies": {
         "@apollo/client": "^3.4.15",
+        "@apollo/explorer": "^3.0.0",
         "@cypress/code-coverage": "^3.10.0",
         "@cypress/instrument-cra": "^1.4.0",
         "@opentelemetry/api": "^1.4.0",
@@ -110,6 +111,7 @@
         "sass": "^1.49.11",
         "typescript": "^4.4.4",
         "url-parse": "^1.5.2",
+        "use-deep-compare-effect": "^1.8.1",
         "uswds": "^2.13.1",
         "uuid": "^9.0.0",
         "web-vitals": "^3.0.1",
--- a/services/app-web/src/constants/routes.ts
+++ b/services/app-web/src/constants/routes.ts
@@ -6,6 +6,7 @@ const ROUTES = [
     'ROOT',
     'AUTH',
     'DASHBOARD',
+    'GRAPHQL_EXPLORER',
     'HELP',
     'REPORTS',
     'SUBMISSIONS',
@@ -33,6 +34,7 @@ const RoutesRecord: Record<RouteT, string> = {
     ROOT: '/',
     AUTH: '/auth',
     DASHBOARD: '/dashboard',
+    GRAPHQL_EXPLORER: '/graphql-explorer',
     HELP: '/help',
     REPORTS: '/reports',
     SUBMISSIONS: '/submissions',
@@ -88,6 +90,7 @@ const PageHeadingsRecord: Partial<Record<RouteTWithUnknown, string>> = {
 const PageTitlesRecord: Record<RouteT | 'UNKNOWN_ROUTE', string> = {
     ROOT: 'Home',
     AUTH: 'Login',
+    GRAPHQL_EXPLORER: 'GraphQL explorer',
     HELP: 'Help',
     REPORTS: 'Reports',
     DASHBOARD: 'Dashboard',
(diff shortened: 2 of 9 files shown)


services/app-web/src/pages/GraphQLExplorer/GraphQLExplorer.tsx
```diff
@@ -0,0 +1,144 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+`
+
+export const GraphQLExplorer = () => {
+    const endpointUrl = 'http://localhost:3030/local/graphql'
```
haworku, line 113: is this supposed to be hard coded to local? I was understanding that we were going to embed into dev/val so we could test dev/val api

services/app-web/src/constants/routes.ts
```diff
@@ -33,6 +34,7 @@ const RoutesRecord: Record<RouteT, string> = {
     ROOT: '/',
     AUTH: '/auth',
     DASHBOARD: '/dashboard',
+    GRAPHQL_EXPLORER: '/graphql-explorer',
```
macrael, line 37: Let’s prefix this with `/dev/` in case we think of more things we want to build out for developers in the future

macrael (approved):
I’m excited to try this on dev. I’m also curious about Hana’s question about setting the endpoint.

services/app-web/src/pages/GraphQLExplorer/GraphQLExplorer.tsx
```diff
@@ -0,0 +1,144 @@
+import { ApolloExplorer } from '@apollo/explorer/react'
+import styles from './GraphQLExplorer.module.scss'
+import { useQuery, gql } from '@apollo/client'
+import { Loading } from '../../components'
+import { useAuth } from '../../contexts/AuthContext'
+import { GenericErrorPage } from '../Errors/GenericErrorPage'
+
+const INTROSPECTION_QUERY = gql`
```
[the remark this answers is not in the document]
JasonLin0991: I could not figure out how to get our graphql schema as a string to pass into the `schema` prop in the `ApolloExplorer` component. I saw it could also accept an `IntrospectionQuery` type and did some googling to find out it was a query to get the entire schema. 

For query string itself is from the Apollo Explorer tool when using it from the browser. Since it was able to figure out our schema by just connecting to it, I figured it did an introspection query. Looked at the network tab and copied the introspection query it sent.

I could use some ideas on where to put this thing.

services/app-web/src/pages/GraphQLExplorer/GraphQLExplorer.tsx
JasonLin0991: Good catch, forgot about this one.

services/app-web/src/pages/GraphQLExplorer/GraphQLExplorer.tsx
```diff
@@ -0,0 +1,144 @@
...
+    return (
+        <div className={styles.background}>
+            <ApolloExplorer
+                className={styles.explorer}
+                endpointUrl={endpointUrl}
+                schema={data}
+                initialState={{
+                    headers: {
+                        'cognito-authentication-provider':
+                            JSON.stringify(loggedInUser),
```
JasonLin0991, line 134: @mojotalantikite This sets up authorization for making the query and works on `local` and `dev`. My question is, is this going to work on `val`?

macrael: I hope this isn’t doing anything in the review apps, this should only work in local dev. But if this PR is working in the review app, I kind of expect it to work in dev/val/prod b/c it means that its sending the actual auth cookie with the request.

JasonLin0991: `I hope this isn’t doing anything in the review apps`

I thought this was supposed to work in all environments except `prod`? That way acceptance testing could use this.

macrael: I'm not able to make a request work in the review app, first hurdle is that CORS is blocking that 'cognito-authentication-provider' header, so I think you should only add that in local dev.

JasonLin0991:
Moving this back to draft. Not working in the review apps.

JasonLin0991:
Works now!

services/app-web/package.json
```diff
@@ -109,6 +110,7 @@
         "sass": "^1.49.11",
         "typescript": "^4.4.4",
         "url-parse": "^1.5.2",
+        "use-deep-compare-effect": "^1.8.1",
```
rswerve, line 113: Is this being used somewhere?

JasonLin0991: Yes! `@apollo/explorer` requires `use-deep-compare-effect` to also be installed.

rswerve:
Able to query in the review app.

image

🎉

PR #1525, opened 2023-03-01
Make updatedAt tags used for testing uploads service be valid

macrael:
## Summary

I added updatedAt tags for the av definition files to make the test run smoothly. Those tags were invalid S3 tags (they had parens in them) so this fixes that and has a little test to make sure. 

#### Test cases covered

Tested the tag value. I should have been doing ISO string anyway. 

## QA guidance

Do you see errors in the avUploadDefinitions service


mojotalantikite (approved):

haworku:
What's going on with this PR? Is this branch still open?

PR #1527, opened 2023-03-03
(WIP) Instrument Uploads

mojotalantikite:
## Summary
This adds instrumentation to the uploads service and is primarily concerned with av scanning.

#### Related issues

https://qmacbis.atlassian.net/browse/MR-2791


haworku (approved):
Things looks good I did poke at the app and zip uploads a bit too. 

Just to capture something from our IRL convo - this PR is not just OTEL improvements and monitoring anymore, its also bringing back a refactor to uploads service that was reverted in #1526 . We think we are ready for this, we have addressed the issue with Cypress that was seen that was blocking work.

macrael:
I’m not sure we’ve resolved that issue? The underlying issue is that this PR caused our average file scan time to jump from 30 seconds to 2 minutes. I would really like to try and fix that problem before we merge this work in, that’s a pretty big jump from a UX perspective not just a problem becaue it hurt our tests.

haworku:
@macrael Ohhhh okay. That wasn't clear to me from the revert; I understood it was an Cypress issue we were mainly trying to resolve to bring this code back. Sounds you and @mojotalantikite need to negotiate then what to do here then, I will bow out I don't have the context.

mojotalantikite:
Yeah, sorry for not being clear about that, I'm actively trying to figure out what's going on with the slowdown in the scanning. Definitely don't want to merge this if it makes everything slow! I'll let everyone know when that is done, but if it seems like it's going to take too long I'll make sure we just land this with the revert from `main`.

mojotalantikite:
Ok, so it looks like our main problem was that the `updateAvDefinitions` lambda wasn't working properly, which was throwing off the scan. I fixed that up and scans are back around 20-30 seconds. Still not awesome, but where they were before at the very least.

PR #1528, opened 2023-03-06
MR-2955: Display CMS users in a table

rswerve:
## Summary

Added a read-only table of CMS users at the route `/settings, visible only to admins.  Eventually, we'll want this table to be the editable source of truth for information about our users, including their email addresses and division assignments.

Also added Iroh, our first local admin user.

@mojotalantikite Do we want a uniqueness constraint on emails in the User table?  In looking at the table after our tests run, I see lots of users with the same email.

#### Screenshots

image

#### Test cases covered

Added a test for the table and contents at `/settings`, as well as some to cover the new admin user.

## QA guidance

If you log in as Iroh, locally and navigate to `/settings`, you should see a table of CMS users.


yarn.lock
```diff
@@ -30250,6 +30250,11 @@ typescript@4.8.4:
   resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.9.4.tgz#a2a3d2756c079abda241d75f149df9d561091e78"
   integrity sha512-Uz+dTXYzxXXbsFpM86Wh3dKCxrQqUcVMxwU54orwlJjOpO3ao8L7j5lH+dWfTwgCwIuM9GQ2kvVotzYJMXTBZg==
 
+typescript@^4.9.5:
+  version "4.9.5"
+  resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.9.5.tgz#095979f9bcc0d09da324d58d03ce8f8374cbe65a"
+  integrity sha512-1FXk9E2Hm+QzZQ7z+McJiHL4NW1F2EzMu9Nq9i3zAaGqibafqYwCVU6WyWAuyQRRzOlxou8xZSyXLEN8oKj24g==
```
JasonLin0991, line 30256: Curious about this change in the yarn.lock, whats going on here? I always thought that changes in this lock file coincided with changes in the package.json.

services/app-web/src/localAuth/LocalLogin.tsx
```diff
@@ -46,12 +47,20 @@ const localUsers: LocalUserType[] = [
         role: 'CMS_USER',
         stateAssignments: [],
     },
+    {
+        id: 'user4',
+        email: 'iroh@example.com',
+        givenName: 'Iroh',
+        familyName: 'Coldstart',
```
JasonLin0991, line 54: Do we have a login on the review app for this user?

macrael: I think that will require an addition to add_cypress_test_users.ts

rswerve: Not as far as I know.  @mojotalantikite @macrael how do we add new review app users?

JasonLin0991 (approved):
Looks good, but trying to view it in the review app. Not sure if we have a user there or on val that is admin.

services/app-web/src/localAuth/LocalLogin.tsx
rswerve: done

services/app-api/src/resolvers/healthPlanPackage/indexHealthPlanPackages.ts
```diff
@@ -52,7 +53,7 @@ export function indexHealthPlanPackagesResolver(
                 user.stateCode
             )
             return validateAndReturnHealthPlanPackages(results, span)
-        } else if (isCMSUser(user)) {
+        } else if (isCMSUser(user) || isAdminUser(user)) {
             const results = await store.findAllHealthPlanPackagesBySubmittedAt()
```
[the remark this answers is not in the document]
JasonLin0991: @rswerve Just a heads up (In case you haven't run into this), after allowing Admin users to fetch a single health plan the summary page works. Going to an unlocked submission leads to a 404 page. I encountered this on a local deployment.

macrael (approved):
looks great

PR #1531, opened 2023-03-07
OIDC for higher environments

ben-harvey:
## Summary

This change enables GitHub OIDC for `val` and `prod` by:
- removing the pin so that `promote` now uses the updated versions of the `deploy-*-to-env` reusable workflows, and updating the inputs and secrets for those workflows
- adding some logic to choose the correct bootstrapped OIDC stage name based on the environment when the `github-oidc` service is deployed.  This is necessary because the `dev` environment uses the stage name `main` (for `val` and `prod` the stage name and environment name match)

Note:  I manually bootstrapped the `val` and `prod` OIDC roles following the instructions in the README 

## QA guidance

This is a tricky one to test manually. Most of our QA comes from the fact that OIDC is working smoothly in `dev`.
 I did test the changes to `deploy-infra-to-env` locally to ensure that the script for setting the OIDC stage name based on the environment worked correctly.

### DIFF
--- a/.github/workflows/deploy-infra-to-env.yml
+++ b/.github/workflows/deploy-infra-to-env.yml
@@ -46,12 +46,12 @@ jobs:
           # if the environment isn't one of [dev, val, prod], fail the job
           run: |
             echo "stage-name=$(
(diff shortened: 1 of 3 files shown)


mojotalantikite (approved):
Cool, looking good! Glad that you were able to get val and prod bootstrapped, still not sure what was going on when I attempted it.

PR #1540, opened 2023-03-10
Sync SecHub findings with Jira issues

ben-harvey:
## Summary

This change adds a workflow that runs on a cron and syncs Security Hub findings from the prod environment with Jira issues. See the inputs to the syncing action for details on the Jira configuration.  For more on the behavior of the syncing action, see https://github.com/Enterprise-CMCS/security-hub-visibility

There's also a small change to the composite action for getting AWS creds so that it can be called by workflows that don't pass a `changed-services` input.

### Authentication
- AWS
  - uses OIDC to get prod credentials.  To support this, `securityhub` is added to the list of allowed AWS actions for the OIDC role
- Jira
   - uses credentials for a Jira service user stored as Actions secrets to perform Jira actions

## QA guidance
Manual testing: using static credentials (`PROD_AWS_SECRET_ACCESS_KEY`, etc), I ran the workflow manually and confirmed that issues are created and added to the correct Jira project, epic, and sprint.  I manually closed an issue without resolving the underlying finding and saw that the issue was reopened by automation. 

Targeting a MAC-FC test environment (because I didn't want to mess with prod findings), I manually resolved a finding and saw that the corresponding issue was automatically closed.

### DIFF
--- a/.github/workflows/sechub-jira-sync.yml
+++ b/.github/workflows/sechub-jira-sync.yml
@@ -1,0 +1,54 @@
+name: Sync Security Hub findings and Jira issues
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '25 1 * * *' # daily at 6:25 am EST
+
+permissions:
+  id-token: write
+
+jobs:
+  sync:
+    name: Run sync
+    environment: prod
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v3
+
+      - name: Get AWS credentials
+        uses: ./.github/actions/get_aws_credentials
+        with:
+          region: ${{ secrets.AWS_DEFAULT_REGION }}
+          account-id: ${{ secrets.PROD_AWS_ACCOUNT_ID }}
+          stage-name: prod
+
+      - name: Sync Security Hub and Jira
+        uses: Enterprise-CMCS/security-hub-visibility@v1.0.1
+        with:
+          jira-token: ${{ secrets.JIRA_TOKEN }}
+          jira-username: ${{ secrets.JIRA_USERNAME }}
+          jira-host: qmacbis.atlassian.net
+          jira-project-key: MR # add issues to the 'MC-Review' project
+          jira-epic-key: MR-3092 # add issues to the '[Tech Maintenance] Security Hub' epic
+          jira-custom-fields: '{ "customfield_10006": 925 }' # add issues to the 'Tech Maintenance' sprint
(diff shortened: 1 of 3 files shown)


.github/workflows/sechub-jira-sync.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Sync Security Hub and Jira
+        uses: Enterprise-CMCS/security-hub-visibility@v1.0.1
+        with:
+          jira-token: ${{ secrets.JIRA_TOKEN }}
+          jira-username: ${{ secrets.JIRA_USERNAME }}
+          jira-host: qmacbis.atlassian.net
+          jira-project-key: MR # add issues to the 'MC-Review' project
+          jira-epic-key: MR-3092 # add issues to the '[Tech Maintenance] Security Hub' epic
+          jira-custom-fields: '{ "customfield_10006": 925 }' # add issues to the 'Tech Maintenance' sprint
```
macrael, line 35: lol @ JIRA, super accessible API there.

.github/workflows/sechub-jira-sync.yml
```diff
@@ -0,0 +1,52 @@
...
+          jira-ignore-statuses: Done
```
macrael, line 36: what does this config var do?

macrael (approved):
yay

.github/workflows/sechub-jira-sync.yml
ben-harvey: @macrael It tells the tool what Jira statuses to ignore when fetching all Security Hub issues from Jira. `Done` is the default status to ignore. Here's what it would look like with and without setting this:

### with ignore `Done`:
- SecHub:  finding A, finding B
- Jira: issue A (`To Do`), issue B (`Done`)

The tool pulls A,B from SecHub and A from Jira.  It creates a new issue for finding B since issue B was ignored on the pull and so finding B doesn't have an issue backing it from the tool's perspective.  This is good since the finding is not resolved yet and we want someone to take action on it.

### without ignore `Done`:
- SecHub:  finding A, finding B
- Jira: issue A (`To Do`), issue B (`Done`)

The tool pulls A,B from SecHub and A,B from Jira.  The tool does nothing since SecHub and Jira are in sync from its perspective.  This is bad since finding B is not yet resolved but is marked as `Done`, so nobody takes action to resolve it.

Perhaps an even better behavior would be to have the tool notice the `Done` status in the second case and set it back to `To Do` if the underlying finding wasn't resolved.

PR #1541, opened 2023-03-11
MR-2906 MR-2901 Add breadcrumbs to upload pages and success banners

haworku:
## Summary
1. Display breadcrumbs  with links back to submissions root and dashboard. Move `UploadResponse` and `UploadQuestions` into the submissions Outlet to enable referencing package data on these pages. 
2.  Add success banners on submit that display on main Q&A page.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2901
https://qmacbis.atlassian.net/browse/MR-2906
#### Test cases covered
Updated tests to handle submissions context sharing via outlet

## QA guidance

- See ticket AC.

### DIFF
--- a/services/app-web/src/pages/App/AppRoutes.tsx
+++ b/services/app-web/src/pages/App/AppRoutes.tsx
@@ -80,24 +80,24 @@ const StateUserRoutes = ({
                 />
                 <Route element={<SubmissionSideNav />}>
                     {showQuestionResponse && (
-                        <Route
-                            path={
-                                RoutesRecord.SUBMISSIONS_QUESTIONS_AND_ANSWERS
-                            }
-                            element={<QuestionResponse />}
-                        />
+                        <>
+                            <Route
+                                path={
+                                    RoutesRecord.SUBMISSIONS_QUESTIONS_AND_ANSWERS
+                                }
+                                element={<QuestionResponse />}
+                            />
+                            <Route
+                                path={RoutesRecord.SUBMISSIONS_UPLOAD_RESPONSE}
+                                element={<UploadResponse />}
+                            />
+                        </>
                     )}
                     <Route
                         path={RoutesRecord.SUBMISSIONS_SUMMARY}
                         element={<SubmissionSummary />}
                     />
                 </Route>
-                {showQuestionResponse && (
-                    <Route
-                        path={RoutesRecord.SUBMISSIONS_UPLOAD_RESPONSE}
-                        element={<UploadResponse />}
-                    />
-                )}
                 <Route
                     path={RoutesRecord.SUBMISSIONS_REVISION}
                     element={<SubmissionRevisionSummary />}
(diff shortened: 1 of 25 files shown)


services/app-web/src/pages/App/AppRoutes.tsx
```diff
@@ -80,24 +80,24 @@ const StateUserRoutes = ({
...
+                            />
+                            <Route
+                                path={RoutesRecord.SUBMISSIONS_UPLOAD_RESPONSE}
+                                element={<UploadResponse />}
+                            />
+                        </>
                     )}
                     <Route
                         path={RoutesRecord.SUBMISSIONS_SUMMARY}
                         element={<SubmissionSummary />}
```
haworku, line 98: @JasonLin0991 Another route I think could eventually be in "sidebar (aka the locked package submission root) -is the previous revision page. I noticed right now it does its own fetch but we could move to outlet and also have it share the fetch for submission summary which includes revisions. thoughts? I would do that kind of work in a separate add-on PR

JasonLin0991: Sounds good!

services/app-web/src/pages/QuestionResponse/QuestionResponse.tsx
```diff
@@ -52,6 +56,7 @@ export const QuestionResponse = () => {
     return (
         <div className={styles.background}>
             <GridContainer className={styles.container}>
+                <QuestionResponseSubmitBanner submitType={submitType} />
```
JasonLin0991, line 59: We have been using the pattern of not rendering components if it's toggled off. I like in `StateDashboard.tsx`

```
  {justSubmittedSubmissionName && (
      <SubmissionSuccessMessage
          submissionName={justSubmittedSubmissionName}
      />
  )}
```
Any particular reason we are not doing that here? Just curious.

JasonLin0991 (approved):
It looks good, but could you merge in main to get tests passing. I would like to test out the review app before you merge in.

services/app-web/src/pages/QuestionResponse/QuestionResponse.tsx
haworku: I will change my code. I like consistency, thanks for pointing it out. Either works, of course.

In terms of why - my thinking was `submitType` is treated as an enum in the banner component and for that reason its not the same as a generic string like submitted name, which is displayed directly in the banner regardless of value as long as it exists.  I wanted to the banner component to handle the business logic around that enum in same place. I just put the undefined handling next that handling what happens if invalid value like `submitType=foobar`  come in the through URL.

JasonLin0991:
Looks good. The cypress error is because in `questionResponse` we are looking for the wrong header now.

```
    cy.findByRole('heading', {
        name: `Minnesota ${submissionName} Upload questions`,
    }).should('exist')
```

PR #1544, opened 2023-03-13
Add a few console logs to our email sending flow

rswerve:
## Summary

In trying to diagnose whether emails were sent, we ran into some gaps in our logging.  This PR adds a couple of lines, and also adds the team's email address to the recipients, so we can have another signal about whether a send happened.

### DIFF
--- a/services/app-api/src/emailer/awsSES.ts
+++ b/services/app-api/src/emailer/awsSES.ts
@@ -75,8 +75,6 @@ async function sendSESEmail(
         return await ses.send(command)
     } catch (err) {
         console.error(JSON.stringify(err))
-        const { requestId, cfId, extendedRequestId } = err.$$metadata
-        console.info({ requestId, cfId, extendedRequestId })
         return err
     }
 }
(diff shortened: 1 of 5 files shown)


services/app-api/src/emailer/awsSES.ts
```diff
@@ -69,7 +69,9 @@ async function sendSESEmail(
     params: SendEmailRequest
 ): Promise<SendEmailResponse | SESServiceException> {
     try {
+        console.info('SendEmailCommand params: ', params)
         const command = new SendEmailCommand(params)
+        console.info('SendEmailCommand: ', command)
         return await ses.send(command)
     } catch (err) {
         console.error(JSON.stringify(err))
```
macrael, line 77: Can you also undo our destructuring below? Just console.info(err.$$metadata) if that’s still interesting to us. The error destructuring meant that the error that got bubbled up was much less informative than it could be.

macrael (approved):

services/app-api/src/emailer/emails/newPackageStateEmail.ts
```diff
@@ -26,6 +26,7 @@ export const newPackageStateEmail = async (
     const receiverEmails = pruneDuplicateEmails([
         ...stateContactEmails,
         ...submitterEmails,
+        ...config.cmsReviewSharedEmails,
```
haworku, line 29: Please add this line to all state emails (unlock, resubmit state emails as well)

PR #1545, opened 2023-03-13
Upgrade Postgres Aurora to v2

mojotalantikite:
## Summary

This adds the upgraded Postgres Aurora v2 instances to serverless. These upgrades were done by hand (the only way available) and imported into Cloud Formation templates. Merging this moves us to using the newly upgraded postgres systems.

A follow on PR will be made to decommission the old DBs once we're satisfied with this upgrade.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2976


macrael (approved):
What does it mean that you aren’t removing the Aurora v1 resources yet here?

mojotalantikite:
> What does it mean that you aren’t removing the Aurora v1 resources yet here?

I just mean this PR isn't going to delete the old database. I want to make sure promote happens ok and if not that we have a DB to revert to.

PR #1546, opened 2023-03-13
Add initial email settings endpoint and return config

haworku:
## Summary
Details of investigation posted on the bug ticket. 

TL;DR not sure what's going on, add email settings endpoint and more errors to apollo_gql load to try and pinpoint where we are losing these variables.

Other changes
- pulled out shared helper for validating stringlist variables from parameter store and added unit tests
-  mass renaming of an email config variable that was still had an old name and was confusing me. 
- remove rates review parameter store helper function - dead code this variable no longer used 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3115

#### Test cases covered
- new tests for `validateAndReturnValueArray`

## QA guidance
- Emails on test box still work as expected. You only need to trigger one contract and rates new submission to check this.
- Go to graphql explorer and hit the fetchEmailSettings endpoint


macrael (approved):
Right now we’d need a submission to go through to actually see the results of these emails, right? 

How about you go ahead and make a new GQL Query that returns the email config? It could call a method in the emailer package that gets all the email config out and returns it. Then we could test the config without having to trigger a submission. That could then be used by the /settings page to display what’s up

haworku:
@macrael Yes! Sounds good, I was thinking to start on settings page next.

services/app-api/src/parameterStore/emailParameterStore/helpers.ts
```diff
@@ -0,0 +1,27 @@
...
+    const { type, value } = storeResponse
+    if (type !== 'StringList') {
+        const errorMessage = `Parameter store ${name} value of Type ${type} is not supported`
+        return new Error(errorMessage)
+    }
+
+    //Split string into array using ',' separator and trim each array item.
+    const stringArray = value.split(',').map((email) => email.trim())
+
+    if (stringArray.length === 0) {
```
haworku, line 20: Just pulling out that block is new logic I added yesterday. It would error already on apollo_gql where parameter store is called if the parameter is somehow empty. Before we only bailed out if it was empty or wrong type. Eventually we can use this kind of  fucntion for every parameter I just started with divisions.

haworku:
FYI @rswerve I made changes to Settings page. Just throwing email settings table underneath for now. Can implement tabs later I was thinking. Let me know if you have any concerns - I know you are still working on this page as well.

services/app-api/src/resolvers/emails/fetchEmailSettings.ts
```diff
@@ -0,0 +1,32 @@
...
+} from '../attributeHelper'
+
+export function fetchEmailSettingsResolver(
+    emailer: Emailer
+): QueryResolvers['fetchEmailSettings'] {
+    return async (_parent, __, context) => {
+        const { user, span } = context
+        if (!isAdminUser(user)) {
+            const errMessage =
+                'Non-admin user not authorized to fetch a settings'
```
rswerve, line 19:
```suggestion
                'Non-admin user not authorized to fetch settings'
```

rswerve (approved):

PR #1489, opened 2023-02-06
MR-2980 Add new division specific email config

haworku:
## Summary
Product would like us to use real email addresses in the help text variables on VAL since they are displayed in emails that pilot test users onboarding see. The problem is we have intentionally used environment specific email configuration up until today.

**Change in this PR (and in AWS Parameter Store):**
- Assign new config variables to each division email address in AWS
- Use the new variables in business logic (see `generateCMSReviewerEmailers`). 
- Deprecate `ratesAddresses`. We can delete entirely from AWS once this lands in prod.
- Add documentation
- 
#### Related issues
https://qmacbis.atlassian.net/browse/MR-2980

### Configuration Changes
NEW
>#### /configuration/email/dmcp
> #### /configuration/email/oact
> #### /configuration/email/dmco
DEPRECATED
> #### /configuration/email/ratesAddresses 

#### Test cases covered
New tests
- `filterChipAndPRSubmissionReviewers` and `includesChipProgram` have unit tests now
- each type of email has unit tests for `to addresses list includes all division emails for contract and rate package` as well as `to addresses list does not include help addresses`

## QA guidance
- 🔍 Check that email senders are still working as expected. Also, could someone please spot double check the oact, dmcp and dcmo config in dev/val/prod in AWS Parameter Store? A second person going through and making sure things look proper there before I merge would be nice. 
- 📓  Read documentation make sure its accurate
- 🐛 This has to merged through to fix the bug. Only then I can change the help address sender variables in VAL safely.

PR #1556, opened 2023-03-16
Fix pruneDuplicateEmails

haworku:
## Summary

@rswerve putting up a failing test for now. I will try to push a commit to fix but in case I don't get there before meetings start wanted to have visibility.

The cases the function needs to handle:

- prune duplicate emails due to alias 
- prune duplicate emails due to casing
- maintain aliased emails that are unique
- maintain raw emails that are unique even if they have uppercase (this is what was broken I believe - I think #1489 and caused divisions to be removed only in prod)

**Maz (@rswerve ) adds**: You were right.  Uppercasing was breaking pruneDupes....  I'm now comparing everything as lowercase, and returning whichever casing we see first in the array (so I tweaked your test to account for that).  All tests are passing.

### DIFF
--- a/services/app-api/src/emailer/formatters/formatEmailAddresses.test.ts
+++ b/services/app-api/src/emailer/formatters/formatEmailAddresses.test.ts
@@ -1,11 +1,11 @@
 import {
     formatEmailAddresses,
-    isEmailAddress,
+    hasAlias,
     includesEmailAddress,
     pruneDuplicateEmails,
 } from './formatEmailAddresses'
 
-describe('isEmailAddress', () => {
+describe('hasAlias', () => {
     const validEmails = ['foo@bar.com', 'foo_bar@bar.com', 'foo+bar@foo.com']
 
     const notEmails = [
@@ -17,7 +17,7 @@ describe('isEmailAddress', () => {
     test.each(validEmails)(
         'given %s as a valid email, returns true',
         (firstArg) => {
-            const result = isEmailAddress(firstArg)
+            const result = !hasAlias(firstArg)
             expect(result).toBe(true)
         }
     )
@@ -25,7 +25,7 @@ describe('isEmailAddress', () => {
     test.each(notEmails)(
         'given %s which is not an email address, returns false',
         (firstArg) => {
-            const result = isEmailAddress(firstArg)
+            const result = !hasAlias(firstArg)
             expect(result).toBe(false)
         }
     )
@@ -81,7 +81,7 @@ describe('formatEmailAddresses', () => {
 describe('pruneDuplicateEmails', () => {
     const sampleEmailLists = [
         [
-            // multiple aliased email addresses that duplicate an email elsewhere on list are pruned
+            // multiple aliased email addresses: keep the first aliased version of each raw email address
             [
                 '"Foo Bar 1" <foo@bar.com>',
                 'foo@bar.com',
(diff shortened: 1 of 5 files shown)


services/app-api/src/emailer/formatters/formatEmailAddresses.test.ts
```diff
@@ -99,9 +99,15 @@ describe('pruneDuplicateEmails', () => {
             ['"Foo Bar 1" foo@bar.com', '"Foo Bar 2" <foo@bar.com>'],
             ['"Foo Bar 1" foo@bar.com', '"Foo Bar 2" <foo@bar.com>'],
         ],
+        // simple email addresses that are duplicates due to casing are pruned and lowercase preferred
         [
-            ['foo@bar.com', 'foo@nothing.com', 'bar@foo.com', 'bar@foo.com'],
-            ['foo@bar.com', 'foo@nothing.com', 'bar@foo.com'],
+            ['Foo@bar.com', 'Fo.b-r@bar.com', 'fo.b-r@bar.com', 'foo@bar.com'],
+            ['Foo@bar.com', 'Fo.b-r@bar.com'],
```
JasonLin0991, line 105: Could we also add tests with different casing in the email and alias? Just want to make sure we test a range of possibilities.

rswerve: You mean something like this?
```javascript
 ['"foo bar 1" FOO@BAR.COM', '"Foo Bar 2" <foo@bar.com>'],
 ['"foo bar 1" FOO@BAR.COM', '"Foo Bar 2" <foo@bar.com>'],
```

JasonLin0991: Yes. I think you could edit the first set of emails in `sampleEmailLists` to include casing in alias emails.

JasonLin0991: The idea I had in mind was to have a set of emails with all the scenarios so we can rigorously test the function.

services/app-api/src/emailer/formatters/formatEmailAddresses.test.ts
```diff
@@ -91,17 +91,40 @@ describe('pruneDuplicateEmails', () => {
                 '"Bar 1" <bar@foo.com>',
                 '"Bar 2" <bar@foo.com>',
             ],
-            ['foo@bar.com', 'bar@foo.com'],
+            ['"Foo Bar 1" <foo@bar.com>', '"Bar 1" <bar@foo.com>'],
         ],
-        // multiple aliased email addresses that do not duplicate a raw email string elsewhere in list will remain
-        // this is because there is not way to determine which to prefer
+        // if there are multiple aliased emails, keep the first one
```
haworku, line 96: Nice, I like the clarity this little adjustment brings. And that we can test for it

services/app-api/src/emailer/formatters/formatEmailAddresses.ts
```diff
@@ -30,30 +30,38 @@ const formatEmailAddresses = (str: string) => {
...
-            return emails.indexOf(email) === index
-        } else {
-            // return unique email address, prefer lowercased if duplicates exist
-            return emails.indexOf(email.toLowerCase()) === index
+const pruneDuplicateEmails = (emails: string[]): string[] => {
+    const uniqueEmails: { [key: string]: string } = {}
+    emails.forEach((currentEmail: string) => {
+        const rawEmail = formatEmailAddresses(currentEmail).toLowerCase()
+        const emailInObject = uniqueEmails[rawEmail]
+        /* ['Jane Johnson <jane@example.com>', 'jane@example.com', 'Bill Smith <BILL@example.com>']
```
haworku, line 44: Suggestion: also add an example in this comment with a unique raw email so its abundantly clear what the `emailInObject` dictionary looks like for that case.

haworku (commented):
javascript star⭐ I think use of dictionary is nice here for a number of reasons. I can't approve cuz its my branch

services/app-api/src/emailer/formatters/formatEmailAddresses.ts
rswerve: good call.  done.

PR #1565, opened 2023-03-23
MR-3116: Email errors should not block submit, unlock and resubmit to complete. 

JasonLin0991:
## Summary
[MR-3116](https://qmacbis.atlassian.net/browse/MR-3116)

`submitHealthPlanPackage.ts` and `unlockHealthPlanPackage.ts` resolvers now throw `GraphQL` errors when sending emails fails.

On the front-end, users will no longer be shown errors when submitting, unlocking, or resubmitting fails due to a failure in sending emails.

Extras:
- `useCreateHealthPlanPackageMutation` in  `SubmissionType.tsx` is now updating the cache more consistent with the Q&A mutations.
- Adding some hard coded graphQL error codes in `errors.ts`. Follows the same style we used in `flags.ts` so we have a central place where we can store error codes. I'm using this for making the unlock and submit mock mutations more consistent with errors from the server. I think @mojotalantikite was talking about having something like this when working on this [spike ticket](https://qmacbis.atlassian.net/browse/MR-3182).

Notes:
- I did not tackle refactoring a lot of the error handling around `mutiationWrappersForUserFriendlyErrors.ts` and other places in this ticket. I was thinking the [spike ticket](https://qmacbis.atlassian.net/browse/MR-3182) should be done first and tickets to refactor error handling will be sprung from it.

#### Test cases covered

Unit tests
- `unlockSubmitModal.test.tsx`
   - `'redirects if submission succeeds, but failed sending emails'`
   - `'does not display modal alert banner error if unlock succeeds, but fails sending emails'`

### DIFF
--- a/services/app-web/src/components/Modal/UnlockSubmitModal.test.tsx
+++ b/services/app-web/src/components/Modal/UnlockSubmitModal.test.tsx
@@ -92,6 +92,52 @@ describe('UnlockSubmitModal', () => {
                 )
             )
         })
+        it('redirects if submission succeeds, but failed sending emails', async () => {
+            let testLocation: Location
+            const modalRef = createRef<ModalRef>()
+            const handleOpen = () =>
+                modalRef.current?.toggleModal(undefined, true)
+            renderWithProviders(
+                <UnlockSubmitModal
+                    healthPlanPackage={mockCompleteDraft()}
+                    submissionName="Test-Submission"
+                    modalType="SUBMIT"
+                    modalRef={modalRef}
+                    setIsSubmitting={mockSetIsSubmitting}
+                />,
+                {
+                    apolloProvider: {
+                        mocks: [
+                            submitHealthPlanPackageMockError({
+                                id: mockCompleteDraft().id,
+                                errorCode: 'EMAIL_ERROR',
+                            }),
+                        ],
+                    },
+                    routerProvider: {
+                        route: `draftSubmission/${
+                            mockCompleteDraft().id
+                        }/review-and-submit`,
+                    },
+                    location: (location) => (testLocation = location),
+                }
+            )
+
+            await waitFor(() => handleOpen())
+            const dialog = screen.getByRole('dialog')
+            await waitFor(() => expect(dialog).toHaveClass('is-visible'))
+
+            await userEvent.click(screen.getByTestId('submit-modal-submit'))
+
+            await waitFor(() =>
+                expect(testLocation.pathname).toBe(`/dashboard`)
+            )
+            await waitFor(() =>
+                expect(testLocation.search).toBe(
+                    `?justSubmitted=Test-Submission`
+                )
+            )
+        })
         it('displays modal alert banner error if submit api request fails', async () => {
             const modalRef = createRef<ModalRef>()
             const handleOpen = () =>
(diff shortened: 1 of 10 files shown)


services/app-web/src/components/Modal/UnlockSubmitModal.tsx
```diff
@@ -129,7 +134,52 @@ export const UnlockSubmitModal = ({
...
+        if (result instanceof Error && result.cause === 'EMAIL_ERROR') {
+            modalRef.current?.toggleModal(undefined, false)
+            if (modalType !== 'UNLOCK' && submissionName) {
+                navigate(`/dashboard?justSubmitted=${submissionName}`)
+            } else {
+                // Updating the cache here with unlockInfo before manually re-fetching query.
+                // This will prevent the loading animation to happen and have up-to-date unlock banner.
+                // We cannot update cache and re-fetch query inside the mutation because it returns an apollo
+                // error on failing emails.
+                const pkg = healthPlanPackage as HealthPlanPackage
```
macrael, line 147: ahh interesting this is a real difference with our throw an error approach. I hadn’t considered this. I think it’s possible to return errors alongside with values in GraphQL but I don’t know how to do that with Apollo, and it would break the wrapper I wrote anyway if we did.

macrael: This cache code is so verbose every time, feels like something that should be in a package instead.

services/app-web/src/gqlHelpers/mutationWrappersForUserFriendlyErrors.ts
```diff
@@ -36,26 +36,37 @@ const handleApolloErrorsAndAddUserFacingMessages = (
             ? ERROR_MESSAGES.submit_error_generic
             : ERROR_MESSAGES.unlock_error_generic
 
+    const options = {
+        cause: {},
```
macrael, line 40: what is the cause option for an Error? Looks like we are translating the GQL ‘code’ into an Error cause?

services/app-web/src/constants/errors.ts
```diff
@@ -17,8 +18,17 @@ const ERROR_MESSAGES = {
     question_missing_field: 'Your question is missing information.',
     question_error_generic: 'Error attempting to add question.',
     response_error_generic: 'Error attempting to add response.',
+    email_error_generic: 'Error attempting to send email.',
 }
 
 const MAIL_TO_SUPPORT = 'mc-review@cms.hhs.gov'
 
-export { MAIL_TO_SUPPORT, ERROR_MESSAGES }
+const GRAPHQL_ERROR_MESSAGES = {
```
haworku, line 26: 🔍 This mapping of graphql error codes to message is not resource agnostic.

 BAD_USER_INPUT or FORBIDDEN could come down on graphql endpoints that are related to all kinds of things (fetching users, editing users, fetching settings, fetching questions) and the messages wouldn't make sense. I suggest still relying on ERROR_MESSAGES with specifically named keys for error messages for now. Something like mapping generic errors for each graphql code seems like a change that should be part of the spike on unifying error messaging.

macrael (approved):
This is great. Throwing an error with a custom code makes good sense to me. Funny that that complicates the cache issue.

services/app-web/src/components/Modal/UnlockSubmitModal.test.tsx
```diff
@@ -313,6 +359,52 @@ describe('UnlockSubmitModal', () => {
...
+                    modalType="UNLOCK"
+                    healthPlanPackage={mockSubmittedHealthPlanPackage()}
+                />,
+                {
+                    apolloProvider: {
+                        mocks: [
+                            unlockHealthPlanPackageMockError({
+                                id: mockUnlockedHealthPlanPackage().id,
+                                reason: 'Test unlock summary',
+                                errorCode: 'EMAIL_ERROR',
```
haworku, line 378: This is is really great to see. I like how you figured out how to test this 🌮

services/app-web/src/constants/errors.ts
JasonLin0991: That is a good point. `BAD_USER_INPUT` and `FORBIDDEN` should be more generic. 

Currently `GRAPHQL_ERROR_MESSAGES` is only used in unit tests. I guess it should be living where the tests are, but once we figure out the approach we could make it a less test-focused resource.

services/app-web/src/components/Modal/UnlockSubmitModal.tsx
JasonLin0991: > This cache code is so verbose every time, feels like something that should be in a package instead.

Could you explain the package idea more? Like, centralize all the cache code into one place and export as functions?

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
```diff
@@ -87,28 +89,33 @@ export const SubmissionType = ({
...
+                                __typename: 'IndexHealthPlanPackagesPayload',
+                                totalCount:
+                                    indexHealthPlanPackages.totalCount + 1,
+                                edges: [
+                                    {
+                                        __typename: 'HealthPlanPackageEdge',
+                                        node: pkg,
+                                    },
+                                    ...indexHealthPlanPackages.edges,
+                                ],
```
haworku, line 118: 💡  I would like to see a technical design doc about what our cacheing approach is since we are integrating it across the app. I feel like there are decisions that have been made or implemented that are not yet reflected in docs. For example, when are using `readQuery` `writeQuery` and when are we `refetchQueries`? 

 We could link [this](https://www.apollographql.com/docs/react/v2/caching/cache-interaction/#updating-after-a-mutation) in the docs. I think this sprint could be a good time to address while its fresh since it also came up last sprint.

services/app-web/src/gqlHelpers/mutationWrappersForUserFriendlyErrors.ts
JasonLin0991: > Looks like we are translating the GQL ‘code’ into an Error cause?

Yep, I needed a way to know what kind of error it was. Specifically, an `EMAIL_ERROR`, in order to let the user complete the submission/unlock. Using the `cause` option seemed like a good place to put error codes. I guess I could have made it `cause.code`?

@haworku can clarify this, but I think `handleApolloErrorsAndAddUserFacingMessages` is returning `new Error` instead of `new ApolloError` or `new GraphQLError` because the server can return `Error` or `ApolloError` so this is handling it by converting all to `Error`?

There is a better way, but I wanted to leave that work till after the spike in server-side errors.

haworku (approved):

PR #1577, opened 2023-03-27
[MR-2803] Edit user division

macrael:
## Summary

I thought there was a PR for this already but I dont’ see it. 

One non obvious fix I made: I included the id of the user in the updateUserQuery.graphql. The Apollo cache for objects operates on the keys “__typename” and “id” so without returning the id in the query it didn’t know that the object had changed in the indexUsers query. By including the ID the indexUsersQuery automatically updates after the update mutation succeeded. 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2803

#### Test cases covered

I need to add a couple tests I think.


services/app-web/src/pages/Settings/CMSUsersTable/CMSUsersTable.tsx
```diff
@@ -1,122 +1,220 @@
...
+                        cmsUserID: userID,
+                        stateAssignments: [],
+                        divisionAssignment: division,
+                    },
+                },
+            })
+
+            if (res.errors) {
+                console.error('Errored attempting to update user: ', res.errors)
+                handleGQLErrors(res.errors)
```
haworku, line 186: What is the UI when your update fails?

haworku:
I can review this tomorrow am. @JasonLin0991 if you want it earlier for your work feel free to review.

services/app-api/prisma/schema.prisma
```diff
@@ -43,21 +43,22 @@ model State {
...
-  stateCode        String?
-  questions        Question[]
-  responses        QuestionResponse[]
+  id                 String             @id
+  givenName          String
+  familyName         String
+  email              String
+  role               Role
+  divisionAssignment Division?
+  stateAssignments   State[]
```
JasonLin0991, line 52: Would it be useful to have a `updatedAt` field for this `User` table since we can now edit it?

services/app-api/src/authn/cognitoAuthn.ts
```diff
@@ -138,6 +138,7 @@ export function userTypeFromAttributes(
             givenName: attributes.given_name,
             familyName: attributes.family_name,
             stateAssignments: [],
+            divisionAssignment: attributes.division,
         })
```
JasonLin0991, line 142: Will `localAuthn.ts` need similar logic?

JasonLin0991 (approved):
Looks good. Just a few non-blocking questions.

services/app-api/prisma/schema.prisma
```diff
@@ -87,16 +88,22 @@ model QuestionResponseDocument {
...
 model QuestionResponse {
-  id            String             @id
-  createdAt     DateTime           @default(now())
-  updatedAt     DateTime           @updatedAt
-  question      Question           @relation(fields: [questionID], references: [id])
-  addedBy       User               @relation(fields: [addedByUserID], references: [id])
+  id            String                     @id
+  createdAt     DateTime                   @default(now())
+  updatedAt     DateTime                   @updatedAt
+  question      Question                   @relation(fields: [questionID], references: [id])
```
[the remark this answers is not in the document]
macrael: It’s possible that was required if updatedAt was added to the table after the initial creation of the table. It’s not neccecary in the normal course of things

services/app-api/prisma/schema.prisma
macrael: great idea

services/app-api/src/authn/cognitoAuthn.ts
macrael: Doesn’t look like it, we aren’t parsing the user out the same way.

haworku:
Good work here MacRae, thank you!  Good debugging.

> The Apollo cache for objects operates on the keys “__typename” and “id” so without returning the id in the query it didn’t know that the object had changed in the indexUsers query. 
Makes sense. I've added this to the [cacheing documentation ](https://qmacbis.atlassian.net/browse/MR-3187) ticket as a comment. These are the classes of bugs we start to have when we implement caching. We should get ahead of it with patterns.

PR #1578, opened 2023-03-27
sendSESEmail not throwing error for senEmail to catch

JasonLin0991:
…g moved to resolver.

### DIFF
--- a/services/app-api/src/emailer/awsSES.ts
+++ b/services/app-api/src/emailer/awsSES.ts
@@ -74,7 +74,6 @@ async function sendSESEmail(
         console.info('SendEmailCommand: ', command)
         return await ses.send(command)
     } catch (err) {
-        console.error(JSON.stringify(err))
         return err
     }
 }
--- a/services/app-api/src/emailer/emailer.ts
+++ b/services/app-api/src/emailer/emailer.ts
@@ -14,6 +14,7 @@ import {
     UnlockedHealthPlanFormDataType,
 } from '../../../app-web/src/common-code/healthPlanFormDataType'
 import { UpdateInfoType, ProgramType } from '../domain-models'
+import { SESServiceException } from '@aws-sdk/client-ses'
 
 // See more discussion of configuration in docs/Configuration.md
 type EmailConfiguration = {
@@ -100,12 +101,15 @@ function newSESEmailer(config: EmailConfiguration): Emailer {
         sendEmail: async (emailData: EmailData): Promise<void | Error> => {
             const emailRequestParams = getSESEmailParams(emailData)
 
-            try {
-                await sendSESEmail(emailRequestParams)
-                return
-            } catch (err) {
-                return new Error('SES email send failed. ' + err)
+            const emailResult = await sendSESEmail(emailRequestParams)
+
+            if (emailResult instanceof SESServiceException) {
+                return new Error(
+                    'SES email send failed. Error: ' +
+                        JSON.stringify(emailResult)
+                )
             }
+            return
         },
         sendCMSNewPackage: async function (
             formData,


services/app-api/src/emailer/awsSES.ts
```diff
@@ -74,7 +74,6 @@ async function sendSESEmail(
         console.info('SendEmailCommand: ', command)
         return await ses.send(command)
     } catch (err) {
-        console.error(JSON.stringify(err))
         return err
```
haworku, line 77: Wait.. I'm a little confused now. I thought we were moving to throwing errors from here and not continuing to return them?

JasonLin0991: Oops forgot this. Too eager to get this up for PR.

haworku (approved):
Works for me. Are you validating the change the same way as the other work  with the broken proto?

JasonLin0991:
> Works for me. Are you validating the change the same way as the other work with the broken proto?

I'm working on a test in the resolver to verify the change in error handling. I'll ping you when I finish and push it up.

services/app-api/src/testHelpers/emailerHelpers.ts
```diff
@@ -62,7 +64,18 @@ function testEmailer(customConfig?: EmailConfiguration): Emailer {
         config,
         sendEmail: jest.fn(
             async (emailData: EmailData): Promise<void | Error> => {
-                console.info('Email content' + JSON.stringify(emailData))
+                try {
+                    await testSendSESEmail(emailData)
+                } catch (err) {
+                    if (err instanceof SESServiceException) {
+                        return new Error(
+                            'SES email send failed. Error: ' +
```
haworku, line 72:
```suggestion
                            'SES email send failed. Error is from Amazon SES. Error: ' +
```

JasonLin0991:
@macrael From our 1:1, I made some of the changes you suggested. 

- Changed `GraphQLError` code to `INTERNAL_SERVER_ERROR` instead of `EMAIL_ERROR`. 
- I moved `EMAIL_ERROR` to a new field `extensions.argumentName` following the pattern in `createHealthPlanPackage.ts`

I also refactored `submitHealthPlanPackage` and `unlockHealthPlanPacakge` resolvers code to log and throw an error if both State and CMS emails fail. Previously if both emails failed we would only be logging the error from sending CMS emails before throwing an error.

cc: @haworku

macrael:
Looks great, one thing:

> * I moved `EMAIL_ERROR` to a new field `extensions.argumentName` following the pattern in `createHealthPlanPackage.ts`

I think you should use a different extension name. “argumentName” is referencing the argument in the request that is responsible for the 400 error. So if you specify any programIDs in your createHPP mutation that don’t exist on that state, we set `argumentName` to be `“programID”` to indicate that that is where the problem lies. 

In this case, there isn’t an argument in the request that was responsible for the error so we should call it something else.

JasonLin0991:
> In this case, there isn’t an argument in the request that was responsible for the error so we should call it something else.

@macrael I switched it to `cause`.

services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.ts
```diff
@@ -225,21 +230,36 @@ export function submitHealthPlanPackageResolver(
         ) {
             const errMessage = `Attempted to submit an already submitted package.`
             logError('submitHealthPlanPackage', errMessage)
-            throw new UserInputError(errMessage) // TODO: This is should be a custom ApolloError such as INVALID_PACKAGE_STATUS or ACTION_UNAVAILABLE, not user input error since doesn't involve form fields the user controls
+            throw new GraphQLError(errMessage, {
+                extensions: {
+                    code: 'INVALID_PACKAGE_STATUS',
```
haworku, line 235: After our informal eng meeting I think we said this is supposed to be `INTERNAL_SERVER_ERROR` for code as well right?  INVALID_PACKAGE_STATUS is the cause not the code is where we landed. 

And USER_INPUT_ERROR (what was originally here in the diff) is a code but the wrong one for this case.

PR #1594, opened 2023-04-03
MR-2427: Add a User audit table

rswerve:
## Summary

Now that we have a UI to change user settings, we want a database table to track those changes.  This PR includes the updated prisma schema, and a generated migration file.

## QA guidance

What to you think of the type for `priorValue`?  I set it to JSON, because that seemed most flexible, given that the potential changes could be to strings or arrays, but I'm open to suggestions.


haworku:
@rswerve I think we never groomed the original[ ticket](https://qmacbis.atlassian.net/browse/MR-2427)so there's no AC I'm seeing but I believe the intent behind writing it was not only that we have an audit table in place but that we start  updating it everytime the user table changes. 

Are you wrapping the actual logging of changes into this PR as well or that gonna be a separate effort this sprint?

haworku:
Schema looks good to me, personally I like the  `priorValue` flexible value because it seems most future proof. Is there a way to  make sure its not empty would be my only suggestion

services/app-api/prisma/schema.prisma
```diff
@@ -43,18 +43,33 @@ model State {
...
+  id              String      @id
+  createdAt       DateTime    @default(now())
+  updatedAt       DateTime    @default(now()) @updatedAt
+  modifiedUserId  String      @map("userId")
+  user            User        @relation(fields: [modifiedUserId], references: [id])
+  updatedBy       User        @relation("UpdatedBy", fields: [updatedByUserId], references: [id])
+  updatedByUserId String
+  action          AuditAction
+  description     String
+  priorValue      Json?
```
JasonLin0991, line 72: See this [comment](https://github.com/Enterprise-CMCS/managed-care-review/pull/1594#issuecomment-1496522763) about `priorValue`

services/app-api/prisma/schema.prisma
```diff
@@ -43,18 +43,33 @@ model State {
...
+  questions             Question[]
+  responses             QuestionResponse[]
+  audits                UserAudit[]
+  updatedByAuditEntries UserAudit[]        @relation("UpdatedBy")
+}
+
+model UserAudit {
+  id              String      @id
+  createdAt       DateTime    @default(now())
+  updatedAt       DateTime    @default(now()) @updatedAt
```
JasonLin0991, line 65: Do we still need `@default(now())` for `updatedAt`? Since this is a new table, we would not have old records without this field. Or am I recalling how this works completely wrong.

rswerve:
Due to some migration weirdness in CI, closing in favor of [ma_2427_audit_migrate](https://github.com/Enterprise-CMCS/managed-care-review/pull/1603)

PR #1602, opened 2023-04-05
Add more OTEL to s3 error paths

mojotalantikite:
## Summary

We had a user encounter some issues with file uploads today, but there were no errors in the OTEL logs to indicate their particular issue. This adds some error recording to the error paths of the s3 context in hopes of getting more insight from that level of `app-web`.


haworku:
The first stop for logging s3 errors should be the S3 library -  `S3Amplify.ts`. Can you check that file make sure we didn't miss something there obvious? That code is really old maybe those recordExceptions aren't working. But basically we should have already had logging for any s3 caused issues.  All of these other paths that touch S3 are farther down in the React code

haworku (approved):
Otherwise looks good.

PR #1603, opened 2023-04-06
MR-2427: Create user audit table and hook up to updateCMSUser

rswerve:
## Summary

As discussed previously, now that we have UI to change user settings, we want a new table to track those changes.  This PR also wires up the change method, so that the new table is populated when a user setting changes.

#### Screenshots

image

#### Test cases covered

Added a test to update the division assignment twice and check the new and prior values in the user and audit tables.

## QA guidance

Note that we're no longer enforcing the presence of an argument in `UpdateCMSUserInput` in the graphql schema, because now we can pass in either state or division, so this check has moved to the resolver.


services/app-api/src/resolvers/user/updateCMSUser.ts
```diff
@@ -54,34 +62,38 @@ export function updateCMSUserResolver(
...
         }
-
         const result = await store.updateCmsUserProperties(
             cmsUserID,
             stateAssignmentCodes,
-            divisionAssignment
+            currentUser.id,
+            divisionAssignment,
+            'Updated user assignments' // someday might have a note field and make this a param
         )
```
JasonLin0991, line 97: Could you update the way we are throwing errors if `store.updateCmsUserProperties` returns an error? We added a new [error-handling](https://github.com/Enterprise-CMCS/managed-care-review/blob/main/docs/technical-design/error-handling.md) doc in our `technical-desgin` docs.

Take a look at `submitHealthPlanPackage` resolver to see how we handle errors from the DB.

services/app-api/prisma/schema.prisma
```diff
@@ -43,18 +43,33 @@ model State {
...
+  responses             QuestionResponse[]
+  audits                UserAudit[]
+  updatedByAuditEntries UserAudit[]        @relation("UpdatedBy")
+}
+
+model UserAudit {
+  id              String      @id @default(uuid())
+  createdAt       DateTime    @default(now())
+  updatedAt       DateTime    @default(now()) @updatedAt
+  modifiedUserId  String      @map("userId")
```
JasonLin0991, line 66: I made this comment on the past PR and mistakenly strike it out.

> Do we still need `@default(now())` for `updatedAt`? Since this is a new table, we would not have old records without this field. Or am I recalling how this works completely wrong

So this question still stands.

JasonLin0991 (approved):

services/app-api/prisma/schema.prisma
rswerve: Honestly, I'm not sure we even need an updatedAt column, because each row will be written and never changed.  But I didn't want to break any auto-checkers that we introduce in the future--since we discussed always having createdAt and updatedAt in each table--so I left it in.  But I think you're right that we don't need to specify the default.  I'll take it out.

services/app-api/src/resolvers/user/updateCMSUser.ts
rswerve: Done.  Thanks for the pointer to that doc.

services/app-api/src/postgres/user/updateCmsUserProperties.ts
```diff
@@ -46,9 +72,27 @@ export async function updateCmsUserProperties(
                     },
                 },
             }),
+            client.userAudit.create({
```
macrael, line 75: You could probably do this in a single .update() command by using `updatedByAuditEntries: { create: {...}` if you wanted to get rid of the transaction here. It's a 6 of one situation though, I don't see strong reasons to prefer one over the other.

services/app-api/src/resolvers/user/updateCMSUser.test.ts
```diff
@@ -97,6 +98,112 @@ describe('updateCMSUser', () => {
...
+            throw new Error('no data')
+        }
+
+        const secondUpdateToUser = secondUpdateRes.data.updateCMSUser.user
+        expect(secondUpdateToUser.email).toBe(newUser.email)
+        expect(secondUpdateToUser.divisionAssignment).toBe(
+            secondDivisionAssignment
+        )
+
+        const auditLogs = await prismaClient.userAudit.findMany({
```
macrael, line 185: nice, I think this is totally appropriate for a test to check the db directly for something we haven't exposed anywhere.

services/app-api/src/resolvers/user/updateCMSUser.ts
```diff
@@ -39,56 +43,81 @@ export function updateCMSUserResolver(
...
+        }
+
         // validate division assignment and throw an error if invalid
-        if (divisionAssignment !== undefined) {
+        if (divisionAssignment) {
             if (!isValidCmsDivison(divisionAssignment)) {
                 const errMsg = 'Invalid division assignment'
                 logError('updateCmsUser', errMsg)
                 setErrorAttributesOnActiveSpan(errMsg, span)
                 throw new UserInputError(errMsg, {
```
macrael, line 63: I _think_ that b/c you are throwing a UserInputError here, there is no need to specify code: 'BAD_USER_INPUT', unless adding extensions overrides the code to nothing if not set. 

We probably want to have our own errors library that sets all of this with a function at some point.

services/app-api/src/resolvers/user/updateCMSUser.ts
```diff
@@ -39,56 +43,81 @@ export function updateCMSUserResolver(
...
         )
         if (isStoreError(result)) {
             if (result.code === 'INSERT_ERROR') {
                 const errMsg = 'cmsUserID does not exist'
                 logError('updateCmsUser', errMsg)
                 setErrorAttributesOnActiveSpan(errMsg, span)
-                throw new UserInputError(errMsg, { argumentName: 'cmsUserID' })
+                throw new UserInputError(errMsg, {
+                    argumentName: 'cmsUserID',
+                    code: 'BAD_USER_INPUT',
```
macrael, line 118: I think this one needs to be in `extensions: {` still

macrael (approved):

services/app-api/src/postgres/user/updateCmsUserProperties.ts
rswerve: I like the `transaction` as a signal that these _need_ to happen together.

services/app-api/src/resolvers/user/updateCMSUser.ts
rswerve: Looking at the docs, it seems that `extensions: {` is an error _output_ convention, and not used while throwing.  I think  we can add whatever fields we want to the object passed in as the second parameter.

https://www.apollographql.com/docs/apollo-server/v3/data/errors/#throwing-errors

I'll hold off merging in case @JasonLin0991 knows something I don't know, otherwise I'll update the error handling doc as part of this PR, to make this clear.

JasonLin0991: I believe when using error classes from `apollo-server-lambda`, the second parameter is the `extensions` object. These error classes also already set the error codes so in this case `UserInputError` is already set to `BAD_USER_INPUT` and the extra `code:  BAD_USER_INPUT` is not needed.

From `apollo-server-lambda`
```typescript
export class UserInputError extends ApolloError {
  constructor(message: string, extensions?: Record<string, any>) {
    super(message, 'BAD_USER_INPUT', extensions);

    Object.defineProperty(this, 'name', { value: 'UserInputError' });
  }
}
```

services/app-api/src/resolvers/user/updateCMSUser.ts
JasonLin0991: Correct! You don't need to specify the `code` in when using these error classes from `apollo-server-lambda`. What should be changed is the second argument. I believe the entire second argument is the `extensions` object, so in this case, the error object would contain an extension inside the extension. 

```typescript
throw new UserInputError(errMsg, {
   argumentName: 'divisionAssignment',
   argumentValues: divisionAssignment,
   cause: 'INVALID_DIVISION_ASSIGNMENT'
})
```

services/app-api/src/resolvers/user/updateCMSUser.ts
rswerve: Yeah, we agree.  I think this, in the error handling doc, is not right though.  What do you think @JasonLin0991?
```typescript
throw new GraphQLError('Email failed.', {
    extensions: {
        cause: 'DB_ERROR',
    },
})
```

PR #1607, opened 2023-04-10
Validate that migrations are wrapped in transactions

mojotalantikite:
## Summary

We've decided that it makes sense for us to wrap our Postgres migrations in transactions. This came out of discussions after a migration that worked in review apps was promoted but broke in `dev`, requiring a restore/roll back.

This checks all newly added migrations are wrapped in a Postgres transaction in `pre-commit` and checks any migrations in a PR with DangerJS to make sure nothing slipped by.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3108

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -391,7 +391,7 @@ jobs:
       fail-fast: true
       matrix:
         # run copies of the current job in parallel
-        containers: [1, 2, 3, 4, 5, 6]
+        containers: [1, 2, 3, 4, 5, 6, 7, 8 ]
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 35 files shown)


haworku (approved):
LGTM!  🚀 I added a quick ADR. Want us to have ADRS more often when we add to our toolchain.

Interesting seeing the difference between the bash script and the TS code here.

mojotalantikite:
So I've been playing around with the `github-lock` script to make it so that a `pull_request` run and a `push` run don't cause conflicts with each other, but I couldn't get the action to lock even though it seemed like the code was fine. Well, turns out there's an ongoing issue with the json return from that API filter: https://github.com/orgs/community/discussions/53266

Going to check back in on it on Monday 🤷🏻‍♂️

PR #1612, opened 2023-04-13
Feature flag "packages with shared rates" dropdown

haworku:
## Summary

 Bring back feature flag to packages with shared rates dropdown so we have ability to remove the dorpdown from the form on demand if we want to. 

Notes: 
- Decided to leave the content displaying on submission summary, since we say that any data the states have filled out we want to display 
- Added `FeatureFlagSettings` type to use instead of hand typing the places we need this. 
-  Renamed the types a bit in flags file, I was getting confused when to expect the capitalized const enum we refer to versus the lowercased const from Launch Darkly. Open to renaming things back.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3214

#### Test cases covered
- no new tests added 

## QA guidance
- nothing should be changed about the packages with shared rates behavior with flag on.
- flag off should be an option

### DIFF
--- a/tests/cypress/support/launchDarklyCommands.ts
+++ b/tests/cypress/support/launchDarklyCommands.ts
@@ -1,10 +1,6 @@
 import '@testing-library/cypress/add-commands'
-import {
-    featureFlags,
-    featureFlagEnums,
-    FeatureFlagTypes,
-    FlagValueTypes,
-} from '../../../services/app-web/src/common-code/featureFlags'
+
+import { FeatureFlagLDConstant, FeatureFlagSettings, featureFlags, FlagValue, featureFlagKeys } from '../../../services/app-web/src/common-code/featureFlags/flags'
 
 /**
  * The code below was taken from this blog post and modified a bit for our use of Types in feature flags.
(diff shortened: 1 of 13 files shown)


tests/cypress/support/launchDarklyCommands.ts
```diff
@@ -75,26 +70,27 @@ Cypress.Commands.add('stubFeatureFlags', () => {
...
     /**
-     * Setting default for chip-only-form to true, currently there is no way to sync API flags with cypress. So
-     * defaulting to true in order for CI to pass.
-     *
-     * Otherwise, this will default feature flags to values set in common-code featureFlags
+     * Setting default values for flags for Cypress E2E tests only. Adding an entry here is not required. 
+     * Useful if you want default feature flags for tests that are different than default values set in common-code featureFlags
      **/
     cy.interceptFeatureFlags({'chip-only-form': true})
+    cy.interceptFeatureFlags({ 'packages-with-shared-rates': true })
```
JasonLin0991, line 77: Not a big deal, but you could combine all the flags into one object and then just call `interceptFeatureFlags` once.

JasonLin0991: Actually, this may be a big deal. It looks like calling it twice might have overwritten the previous call to set `chip-only-form` to true. The cypress tests are failing because `chip-only-form` related form fields are not enabled and when submitting the backend will check for this field. We currently do not have a way to sync flags between app and api in cypress.

I can see two paths we can take:
1. Combine both flags into one object and call the intercept once. 
   - Easiest solution.
   - `interceptFeatureFlags` sets the flag to what is passed into it and sets other flags to default values. Passing in an empty object will default all flags. This is not documented. We should add that to the `launchDarklyCommands.ts` file.
2. Rewrite `interceptFeatureFlags` to not overwrite values in the `featureFlagStore.json`.
   - The values we save in the JSON file could probably only be what flags were set using `interceptFeatureFlags` instead of all the flags.
   - This means `stubFeatureFlags` will also need code to set `featureFlagStore.json` to default values before setting flags globally. `stubFeatureFlags` is used before each test to clear out flags settings so they don't carry over to new tests.
   - Future uses of `interceptFeatureFlags` in tests will add to what is currently set in the flag store, instead of overwriting, until `stubFeatureFlags` is called.
   
I would choose option 1, but option 2 seems like something we can do in the future as it involves much more of a refactor.

JasonLin0991 (changes requested):
Cypress failing is due to the way we set flags in `stubFeatureFlags` command. I left a detailed comment about it.

JasonLin0991 (approved):

PR #1631, opened 2023-04-21
Run unit tests direct in CI

mojotalantikite:
## Summary

Running our unit tests via the `dev` tool is fine locally, but produces hard to read output when in CI (e.g. lines get wrapped making failures harder to read back). This moves them to be setup with their appropriate `lerna` commands and run direct via `yarn`.

I also noticed a couple places I missed updating the CodeClimate keys when writing this change, so that fix is here too.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -151,6 +151,10 @@ jobs:
         run: lerna run generate --scope=app-proto
 
       - name: Generate Prisma
+        env:
+          NODE_OPTIONS: --max_old_space_size=6000
+          REACT_APP_AUTH_MODE: AWS_COGNITO
+          DATABASE_URL: postgresql://postgres:shhhsecret@localhost:5432/postgres?schema=public&connection_limit=5 #pragma: allowlist secret
         working-directory: services/app-api
         run: |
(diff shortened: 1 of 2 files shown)


mojotalantikite:
Open question: do we care to make this change in `promote` as well? I left that alone, because I felt the place we're usually trying to diagnose flakes/failures is in PR branches. If we want to also do it for `promote` that's fine too, I can add it.

rswerve (approved):

haworku:
@mojotalantikite I think its fine to leave this as is for now. We are just trying to make debugging easier because of increased issues with flakes in PRs rn.

@macrael FYI so you see this. I know you care about the dev tool and staying consistent between local and deployed app.  But right now that tooling is making log outputs harder to read and we also are starting to experience some pain with unit test. I encouraged Mojo's idea to just make the change to directly run yarn

macrael:
I think this makes sense to me. I don't love that we're defining the dependencies between the unit tests and the generated GQL/Prisma/Proto code twice now between ./dev and deploy.yml but it's worth getting useable debugging logs out. What would y'all think about adding a --headless option or something that doesn't do the prefix but we can still keep that dependency encoded there? I'm open to either.

PR #1636, opened 2023-04-24
Add more containers for Cypress

haworku:
## Summary
This is recommended in the Cypress Dashboard interface based on number of tests we are running and current times 

## QA guidance
Tests should still run.


.github/workflows/deploy.yml
```diff
@@ -366,7 +366,7 @@ jobs:
       fail-fast: true
       matrix:
         # run copies of the current job in parallel
-        containers: [1, 2, 3, 4, 5, 6]
+        containers: [1, 2, 3, 4, 5, 6, 7, 8 ]
```
rswerve, line 369: a bit farther down in the same file, we collect the cypress coverage reports.  there are lines like
```
cp ./cypress-json-coverage-6/coverage-final.json ./cypress_temp/from-cypress-6.json
```
I think you'll need to add a couple more there, for the new containers.

mojotalantikite (approved):

PR #1638, opened 2023-04-24
MR-3155 Update MO programs and document

haworku:
## Summary
Update Missouri program and write documentation about how the process is working. 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3155

## QA guidance
The linked ticket shows what programs should change and the CSV programs list that we draw from.


JasonLin0991 (approved):
Thanks for updating the documentation!

PR #1642, opened 2023-04-25
MR-3177 Add another provision to CHIP excluded

haworku:
## Summary
New change coming from stakeholder meeting this week. There's another provision to exclude for CHIP. Also found a small bug around "missing field" logic for SubmissionsSummaryContractDetails related to provisions and fixed that as well.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3177

#### Test cases covered
Adjusted existing tests since logic around provisions changed by 1 provision. 

## QA guidance

- The 'Other financial, payment, incentive or related contractual provisions' should not appear for CHIP provisions.


JasonLin0991 (approved):
- ✅The 'Other financial, payment, incentive or related contractual provisions' should not appear for CHIP provisions.
- ✅./dev test web works without crashing your computer
- ✅yarn clean still works and also clear jest cache in app web and app api

LGTM!

dev_tool/src/test/web.ts
```diff
@@ -17,17 +16,11 @@ export async function runWebTestsWatch(jestArgs: string[]) {
 
     await installWebDepsOnce(runner)
 
-    // because we are inheriting stdio for this process,
-    // we need to not run spawnSync or else all the output
-    // for the graphql compiler will be swallowed.
-    const proc = spawn('yarn', ['test'].concat(jestArgs), {
```
macrael, line 23: I think this is good because it fixes the crash, but I want to call out that we lose the interactiveness of jest. You can't hit f to re-run failing tests without getting stdio.

haworku: Oh I didn't notice that. That shouldn't be happening, lerna should support jest running with watch. I peeked at things more and docs are saying our jest setup with lerna may needs improvement. similar to work @mojotalantikite did with eslint, we should have a shared jest config at the root.  Rather than walk down that path I pulled all the tooling changes off this PR can make separate PR next to troubleshoot this stuff.

macrael: The issue isn't the `lerna` part, it's the `spawn` part. runCommandAndOutput doesn't take input from the terminal, it just runs the process and prints the output. By calling spawn with `stdio: 'inherit'` then it properly accepts terminal input and sends it to the running process. And it's certainly something about that that is causing the kernel panic.

macrael: I'm very happy with ./dev calling lerna commands, providing a CLI interface with `--help` and the like. The one thing besides that that ./dev is doing that I don't know how to do with lerna is running dependencies correctly. When you run `./dev test web` it's not just calling `yarn test` but also running `lerna run gqlgen:watch`, and `lerna run generate:watch --scope=app-proto`

PR #1646, opened 2023-04-27
bugfix: Correct modified provisions MLR reference for non CHIP submission

haworku:
## Summary
The MLR references copy updates should be conditional (only apply to provisions when population covered = CHIP). This is from product design review.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3177

#### Screenshots
Look at this [figma](https://www.figma.com/file/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?t=mbHghBniks1zDfro-0) - left side is non-CHIP behavior and right side is CHIP amendments

#### Test cases covered
Updated tests again to cover expected behavior.

 Also added new tests for MLR copy specifically
- renders amended provisions with MLR references for a Medicaid contract
- renders amended provisions with correct MLR references for CHIP contract

## QA guidance
Create contract amendment submission. When you go between population covered = CHIP and other population covered options the MLR references will change subtly.

### DIFF
--- a/services/app-web/src/common-code/healthPlanFormDataType/ModifiedProvisions.ts
+++ b/services/app-web/src/common-code/healthPlanFormDataType/ModifiedProvisions.ts
@@ -1,7 +1,3 @@
-type ModifiedProvisions = {
-    [K in (typeof modifiedProvisionKeys)[number]]: boolean
-} // constructs a type that requires all of the below keys
-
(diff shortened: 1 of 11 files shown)


rswerve (approved):
I see the appropriate changes when switching to CHIP-only and back.

PR #1660, opened 2023-05-05
Refactor postgres_migrate return values and add OTEL.

mojotalantikite:
## Summary
This fixes up our `postgres_migrate` lambda to return errors consistently to the client. It also adds OTEL to record exceptions.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-2692

### DIFF
--- a/services/app-api/src/handlers/postgres_migrate.ts
+++ b/services/app-api/src/handlers/postgres_migrate.ts
@@ -138,7 +138,7 @@ export const main: Handler = async (): Promise<APIGatewayProxyResultV2> => {
             `${process.execPath}`,
             [
                 '/opt/nodejs/protoMigrator/migrate_protos.js',
-                'db',
+                'break-it',
                 '/opt/nodejs/protoMigrator/healthPlanFormDataMigrations',
             ],
             {
--- a/services/app-api/src/handlers/postgres_migrate.ts
+++ b/services/app-api/src/handlers/postgres_migrate.ts
@@ -1,10 +1,14 @@
-import { Handler } from 'aws-lambda'
+import { Handler, Context } from 'aws-lambda'
 import { RDSClient, CreateDBClusterSnapshotCommand } from '@aws-sdk/client-rds'
 import { spawnSync } from 'child_process'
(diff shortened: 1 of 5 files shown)


services/app-api/src/handlers/postgres_migrate.ts
```diff
@@ -79,51 +109,68 @@ export const main: APIGatewayProxyHandler = async () => {
...
+            return fmtMigrateError(errMsg)
         }
     }
 
     // run the data migration. this will run any data changes to the protobufs stored in postgres
     try {
-        execSync(
-            `${process.execPath} /opt/nodejs/protoMigrator/migrate_protos.js db '/opt/nodejs/protoMigrator/healthPlanFormDataMigrations'`,
+        const connectTimeout = process.env.CONNECT_TIMEOUT ?? '60'
+        const migrateProtosResult = spawnSync(
```
haworku, line 125: should this be `await spawnSync`

services/app-api/src/handlers/postgres_migrate.ts
```diff
@@ -1,72 +1,102 @@
...
+            'stdout',
+            prismaResult.stdout && prismaResult.stdout.toString()
+        )
+        if (prismaResult.status !== 0) {
+            const errMsg = `Could not run prisma migrate deploy: ${prismaResult.stderr.toString()}`
+            recordException(errMsg, serviceName, 'prisma migrate deploy')
+            return fmtMigrateError(errMsg)
         }
+    } catch (err) {
+        const errMsg = `Could not migrate the database schema: ${err}`
```
haworku, line 87:
```suggestion
        const errMsg = `Could not migrate the prisma database schema: ${err}`
```

haworku (approved):

services/app-api/src/handlers/postgres_migrate.ts
mojotalantikite: `spawnSync` doesn't return until the process returns, so it's not async like other types of `spawn` would be. I double checked though and we don't `await` it anywhere in our code base and my editor suggests removing it when you try to `await` it.

PR #1672, opened 2023-05-10
MR-3117: Attempt to update unmodifiable data

rswerve:
## Summary

Following up on @JasonLin0991 's suspicion that this bug was actually caused by failing to disable the Continue and Back buttons in time (he was right) I worked to debounce...something.  Initially, I debounced in the UI, but then thought that it would be nice to have a way to debounce resolver calls more generally (the front-end useDebouncer will live on in git history if we ever want to resurrect it).

With the useDebouncedMutation hook, we can debounce any mutation simply by wrapping it in the hook.

~Our best guess as to what's causing this bug is that in some cases, when the server call after `Continue` takes longer than usual, and the user navigates back to already-edited page before the call returns, and _then_ tries to continue again, we're trying to update with stale cache data.~

~Ultimately, this bug is possible because we don't handle potential concurrent editing.  If two people are simultaneously working on one submission, one shouldn't constantly see stale data warnings.~

~After chatting with @macrael, we decided that at some point we'll have to deal with the concurrent editing problem, but that's far out of scope for this ticket and bug (which I can't even reproduce, after much trying).  So we're going to update the error message to give the user some guidance about how to proceed, and we'll put [a ticket](https://qmacbis.atlassian.net/browse/MR-3353) in the backlog for concurrent editing.~

#### QA Guidance

You should not be able to trigger extra update calls, no matter how fast you click Continue.

#### Test cases covered

Once I finally figured out why our spies weren't working (exporting via index.ts can cause jest spies to not work), I still had a tough time getting the kind of test I really wanted, which was to click many times and see only one update.  My best guess is that the way Jest triggers clicks doesn't allow multiple fast clicks (I tried loops, doubleClick events, and several other strategies.)  The best I could do was confirm that the debouncer was called when update was called.


haworku:
@rswerve Did we attempt to reproduce in a deployed environment at all or only locally?

JasonLin0991:
@maz that `Continue` button callout got me curious, and I was able to reproduce this error on the review app by quickly clicking the continue button. Possibly our way of disabling the continue button is too slow or do we need to also prevent the update mutation from being fired more than once?

Screenshot 2023-05-10 at 10 35 06 AM

services/app-web/src/hooks/useDebouncedMutation.ts
```diff
@@ -0,0 +1,61 @@
...
+    MutationHookOptions,
+    MutationTuple,
+    TypedDocumentNode,
+    useMutation,
+} from '@apollo/client'
+
+import { debounce } from '../common-code/debounce/debounce'
+
+// mostly from https://stackoverflow.com/a/74109414/5351022
+export const useDebouncedMutation = <
```
JasonLin0991, line 15: This hook fixed the issue. Should we put a comment here about why we are debouncing mutations to not cause confusion? Not sure if it's just me, but I usually associate debounces with inputs and not mutations.

JasonLin0991 (approved):
Looks good to me

`useDebouncedMutation` hook fixed the issue. Should we put a comment at `useDebouncedMutation` about why we are debouncing mutations to not cause confusion? Not sure if it's just me, but I usually associate debounces with inputs and not mutations.

haworku (approved):

services/app-web/src/pages/StateSubmission/StateSubmissionForm.test.tsx
```diff
@@ -566,4 +571,78 @@ describe('StateSubmissionForm', () => {
...
+            void userEvent.type(textarea, 'updated something')
+
+            continueButton.click()
+            // looking for Step 2 of 6, which is a way to make sure the update has gone through and we're on the next page
+            await waitFor(() => {
+                expect(screen.getByText('2')).toBeInTheDocument()
+            })
+            expect(debouncedSpy).toHaveBeenCalled()
+            expect(mockUpdateSpy).toHaveBeenCalled()
+            jest.clearAllMocks()
```
haworku, line 645: Not blocking but something I was looking at -- The test I would like to see is one where the user clicks multiple times and  we expect `mockUpdateSpy.toHaveBeenCalledTimes(1)`.

I messed around a bit with writing the test and using jest timers but then I couldn't get it failing when expected (e.g. I removed the hook and test still passed lol).

rswerve: This is exactly what I was talking about in the summary.  I spent a full day trying to get that test to work.  No luck.

PR #1675, opened 2023-05-10
Rate Refactor: Add contracts and rates tables

macrael:
## Summary

This PR adds new Contract and Rate tables with their associated revisions and DB functions for updating them both and fetching a coherent list of revisions for each. 

Things that are not done yet:
* Errors are all generic, not an easy way to tell the difference between a random prisma error and passing an invalid ID into one of the db functions
* DB actions aren't in transactions yet, so it is possible to get the db in a bad state by doing things on top of one another
* The findContract/findRate functions chunk revisions based on timestamps. In reality each of those timestamps represent a single revision being submitted. Right now if you submit two revisions at too close to the same time it there isn't enough precision to tell them apart in those functions. That should be fixable but it's where it's at right now. This is why the tests I have here have sleeps in them between submissions. 
* Right now Drafts aren't really findable, they are omitted from findContract/findRate

#### Related issues

https://qmacbis.atlassian.net/browse/MR-3234

#### Test cases covered

There are two tests files for now mostly just testing happy path scenarios. 

## QA guidance

Please check this out and add a test! Try different combinations of submitting and updating and unlocking rates and contracts and make sure that the list of revisions spat out by findContract and findRate make sense.


JasonLin0991 (commented):
So far looks good.

One operation I was curious about was finding all contracts by state. Seems like we still need DB schema and functions work for that?

I also left some comments on the [contract and rates doc](https://docs.google.com/document/d/1R6_t3U4xyKcdlAew7GLwbPhPDzXBnFdsIW8qtqbyomA/edit) for other helpful sections.

docs/technical-design/database-diagram.md
```diff
@@ -76,3 +76,83 @@ Question ||--o{ QuestionReponse: questionID
...
+   String name
+}
+
+RateRevisionsOnContractRevisionsTable {
+   String rateRevisionID
+   String contractRevisionID
+
+   DateTime validAfter
+   DateTime validUntil
+
```
haworku, line 124: Make sure `isRemoval` goes here - or else we not add this now at all and instead add when we publish docs

services/app-api/src/postgres/contractAndRates/contractType.ts
```diff
@@ -0,0 +1,27 @@
+import { UpdateInfoType } from "../../domain-models"
+import { RateRevision } from "./rateType"
+
+interface Contract {
+    id: string
+    revisions: ContractRevision[]
+}
+
+// there needs to be a new contract revision after every set, even though the revision is the same?
```
haworku, line 9: Can we please organize comments a bit more before we merge.  Right now I can't tell if these ?? question??? statements are 1. open questions that need to be answered 2. outdated dev notes 3. more philosophical wondering or marginalia that I can ignore working int his file. 

Let's try to clean this up- make it more clear what needs further action or consideration in this file and what doesn't

services/app-api/src/postgres/contractAndRates/findContract.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    return {
+        updatedAt: info.updatedAt,
+        updatedBy: info.updatedByID,
+        updatedReason: info.updatedReason,
+    }
+}
+
+// this is for the internal building of individual revisions
```
haworku, line 24:
```suggestion
// this is for the internal building of individual revisions for the change history feature
```

haworku (commented):
I'm fine with merging this - my main concern is the comments that looks like questions would like those cleaned up or else grouped more clearer with a note about "these are open questions" "Still to be resolved"

services/app-api/prisma/migrations/20230612222743_remove_invalided_by/migration.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN;
+/*
+  Warnings:
+
+  - You are about to drop the column `invalidatedByContractRevisionID` on the `RateRevisionsOnContractRevisionsTable` table. All the data in the column will be lost.
+  - You are about to drop the column `invalidatedByRateRevisionID` on the `RateRevisionsOnContractRevisionsTable` table. All the data in the column will be lost.
```
JasonLin0991, line 6: Should we ignore these warnings?

JasonLin0991 (approved):
Looks good! We still have some documentation things, but that shouldn't hold up this from merging.

services/app-api/prisma/migrations/20230612222743_remove_invalided_by/migration.sql
macrael: These are real warnings but dropping those columns is what we want. They were used in my first implementation of the revision finds

PR #1689, opened 2023-05-19
MR-3233 Add provisions to base contract

haworku:
## Summary

- Base contracts now take provisions. There are different lists of accepted provisions for base contracts versus amendments, versus CHIP only submission.
- A new provision field was added called "in lieu service and settings"
- 🐛 found and fixed bug with federal authorities validation in formik for CHIP only
   
#### Related issues
https://qmacbis.atlassian.net/browse/MR-3261 
https://qmacbis.atlassian.net/browse/MR-3233

#### Tech notes
-  Even though provisions now are relevant for base contracts we decided to still store on `contractAmendmentInfo.modifiedProvisions` in the proto since rates refactor is on the horizon.
- Helper functions are now in `healthPlanSubmissionHelpers/provisions.ts`. I was seeing a lot of provisions related helper functions that were inside React components, pulling these into an external file makes it easier to see logic in one place and unit test.  
- Previously we had a `isLockedHealthPlanFormData` that was a problematic because it was also doing validation checks on proto unwrap. It started breaking in a lot of places with my work, since this ticket changes what is a valid base contract going forward (it will now requires provisions so any previously stored subs would break if this logic was unchanged).
   -  Rather than loosen checks for what is a valid contract (since that logic seems important) I decided to separate concerns into two functions. Now`isLockedHealthPlanFormData` now only checks submission status. There is also a new function `isValidAndCurrentLockedHealthPlanFormData` which adds on validation checks relevant for the latest version of the form.

#### Test cases covered
- added tests for base contracts use case everywhere we test provisions
- added unit tests for the new `generateApplicableProvisionsList`

## QA guidance
- See AC on the associated tickets for best guidance here.


services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.ts
```diff
@@ -77,24 +76,19 @@ function submit(
...
     }
 
+    // move this check into isValidContract when feature flag is removed
     const validPopulationCovered = featureFlags?.['chip-only-form']
         ? hasValidPopulationCoverage(
               maybeStateSubmission as LockedHealthPlanFormDataType
           )
         : true
 
-    const validRateCertAssurance = hasValidRateCertAssurance(
```
haworku, line 86: ^this check is now part of `isValidContract`. I think we should avoid standalone form field checks as much as possible in resolvers, try to lean on bigger functions like isValidContract or isValidRates that check entire sets of fields.  Ideally those bigger functions could also handle feature flags too (but that seems like a bigger refactor so I held off).

services/app-web/src/common-code/healthPlanSubmissionHelpers/provisions.ts
```diff
@@ -0,0 +1,148 @@
...
+    3. For non CHIP contract amendment
+
+    See also ModifiedProvisions.ts
+*/
+
+// Returns the list of provision keys that apply for given submission variant
+const generateApplicableProvisionsList = (
+    draftSubmission: HealthPlanFormDataType
+):
+    | CHIPProvisionType[]
```
macrael, line 43: I wonder if this type could be used where we're using GeneralizedProvisionType? 

That's my only thought about this PR, the `as unknown as ...` bits don't seem great and I wonder if there's a better way to do them. I'll try and mess with it a little tomorrow and see.

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
```diff
@@ -206,22 +167,24 @@ export const ContractDetailsSummarySection = ({
...
                         >
-                            {amendmentProvisionsUnanswered ? null : (
+                            {provisionsAreInvalid ? null : (
                                 <DataDetailCheckboxList
                                     list={modifiedProvisions}
                                     dict={
-                                        isCHIPOnly
-                                            ? CHIPModifiedProvisionsRecord
-                                            : ModifiedProvisionsRecord
+                                        isCHIPOnly(submission)
```
macrael, line 183: Would a helper function that gets the Record for a given submission be used multiple places?

macrael (approved):
Looks excellent

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
haworku: Ha, I wrote one ( `getProvisionDictionary` ) looks like I forgot to use it. Will fix this.

PR #1691, opened 2023-05-19
Fix CI Destroy Errors

mojotalantikite:
## Summary

This refactors the function `getStacksFromStage`, so that some edge errors don't get swallowed and to make it a little dumber for readability. We also will always alert if we're in CI and `destroy` is invoked, yet has nothing to destroy (another edge case).

One last thing is we remove some old stackPrefixes which were old services that no longer are a part of our application (renamed or deleted).

#### Related issues

https://qmacbis.atlassian.net/browse/MR-3304


mojotalantikite:
something is up with this branch name for deploys, recreating

PR #1704, opened 2023-05-30
MR-2461:  populate groups of submissions in cypress

JasonLin0991:
## Summary
[MR-2461](https://qmacbis.atlassian.net/browse/MR-2461)

We are quickly populating submissions in Cypress by making direct API requests to create, update, and submit a new health plan package.

I have also included a first pass **README** for Cypress that could use eyes on. It has a bit more detail on how all this was set up.

#### Setup
To set up Cypress for making direct API requests, I loosely mimicked our setup in `app-web` by adding the following to `services/cypress`
- `aws-amplify`: This is to configure our requests with cognito auth in CI.
- `@apollo/client`: We'll be using `Apollo` to make our `GraphQL` API requests
- `graphql`:  For making the API requests
- `axios`: We need this because I copied `fetchResponseFromAxios()` and `fakeAmplifyFetch` to configure the `Apollo` client.
- `gen` folder: Our generated `graphql` and `protobuf` files will also be in `services/cypress`. These files are used for configuring the `Apollo` client, making the `graphql` requests, and decode/encode `protobufs`.
- The Cypress preprocessor has been changed from the default `webpack` bundler to `esbuild`. This is because importing large packages like `aws-amplify` added 30-40 seconds to loading test files. Here is the [github issue](https://github.com/cypress-io/cypress/issues/25533#issuecomment-1430716924) that goes into the same issue.

#### Configuration
- Configuring `Amplify` and the `Apollo` client are done in `cypress/utils/apollo-test-utils.ts`. In this util file, the `apolloClientWrapper` function handles authorization for API requests and creates the `Apollo` client, which is then passed into a `callback` function. This `callback` function is where all the `apollo` requests should be made. 

- `Amplify.configure` uses `cognito` auth variables from CI for its configuration. Local deployments, this configuration does nothing unless you provide it with actual `cognito` variables and a review app `url`. Then you can run Cypress locally to test against a deployed review app.

#### Making API requests.
- With the configuration all wrapped into the `apolloClientWrapper`, we can make the functions that perform the requests and pass them into `apolloClientWrapper`.

- Take a look in `cypress/support.apiCommands.ts`; there are two commands, `apiCreateAndSubmitContractOnlySubmission` and `apiAssignDivisionToCMSUser`, for making API requests. 
   - They both call `cy.task<DocumentNode>('readGraphQLSchema')` and pass a `schema` to the `apolloClientWrapper`. This task, located in `cypress.config.ts`, reads the `schema.graphql` from our `gen` folder and converts it to `gql` for our `Apollo` client configuration in `apolloClientWrapper`. 
   - The `createAndSubmitContractOnlyPackage` is pass into `apolloClientWrapper` as the callback function. You can see all that is in that callback function is just performing the API requests with the passed in `apolloClient`, and protobuf things, but there are no configurations.
   
- Now, looking in `questionResponse.spec.ts` you can see how we are using both commands to assign a division to a CMS user and create, update, and submit a health plan only with API calls. Also, notice how we can use `.then` on these commands to wait for the API requests to finish before running the rest of the tests. This is because the return type of both commands is `Cypress.Chainable`.

#### Screenshots
Some average run times I had collected
Screenshot 2023-06-06 at 11 36 07 AM


services/cypress/README.md
```diff
@@ -0,0 +1,284 @@
+# Cypress
+This service deploys our Cypress app for e2e testing locally and CI. An important note is that the actual Cypress package is installed in the root `node_modules` this was the only way we could get Cypress commands to work in CI. The reason is still unknown, but we suspect it because some Cypress commands need a yarn.lock in the same directory where the package is installed and in our monorepo we only have a yarn.lock at the root directory.
+
+### Deploying locally
+  - Run local Cypress app and test against the local app deployment.
+    - `./dev test browser`
+  - We can also run the local Cypress app against a review app deployment.
+    - ```
+      REACT_APP_AUTH_MODE='AWS_COGNITO' TEST_USERS_PASS='<Test user pass here>' API_URL='<api url here>' COGNITO_USER_POOL_ID='<user pool id here>' COGNITO_REGION='<region here>' COGNITO_IDENTITY_POOL_ID='<identity pool id here>' COGNITO_USER_POOL_WEB_CLIENT_ID='<web client id here>' ./dev test browser -- open --config 'baseUrl=<insert your branch deployment url here>'
```
macrael, line 9: How do we get all this config when running `./dev local web --hybrid`? Maybe we could have a `./dev test browser --review-app` command?

services/cypress/cypress.config.ts
```diff
@@ -35,6 +49,5 @@ const defineConfig = ({
         runMode: 2,
         openMode: 0,
     },
+    chromeWebSecurity: false
 })
-
-module.exports = defineConfig
```
macrael, line 40: why move this?

services/cypress/support/apiCommands.ts
```diff
@@ -0,0 +1,141 @@
...
+    apolloClient: ApolloClient<NormalizedCacheObject>,
+    cmsUser: CMSUserType,
+    division: DivisionType
+): Promise<void> => {
+    // get all users query
+    const result = await apolloClient.query({
+        query: IndexUsersDocument,
+    })
+
+    // find zuko
```
macrael, line 84: not just zuko anymore.

services/cypress/support/apiCommands.ts
```diff
@@ -0,0 +1,141 @@
...
+        variables: {
+            input: {
+                cmsUserID: user.id,
+                divisionAssignment: division,
+            },
+        },
+    })
+}
+
+const seedUserIntoDB = async (
```
macrael, line 108: nice!

services/cypress/utils/apollo-test-utils.ts
```diff
@@ -0,0 +1,312 @@
...
+            modifiedPaymentsForMentalDiseaseInstitutions: false,
+            modifiedNonRiskPaymentArrangements: false,
+        },
+    },
+    managedCareEntities: ['MCO'],
+    federalAuthorities: ['STATE_PLAN'],
+    rateInfos: [],
+}
+
+const newSubmissionInput = {
```
macrael, line 80: I usually end up making this mock data into functions that return this info so that there's no chance of different tests modifying the global

macrael (approved):
So great. I'm still sad that there's not a way to get the real mutations out of Apollo but I haven't been able to find them so who knows

services/cypress/cypress.config.ts
JasonLin0991: 🤷🏻‍♂️ Just saw cypress had it that way.

services/cypress/README.md
JasonLin0991: This was something I thought about and had some questions. Let's talk about this at our next 1:1?

PR #1705, opened 2023-05-30
Cypress incremental improvements 

haworku:
## Summary
This is follow on work to #1689 - had trouble merging that work due to cypress flakes

Covered in this PR:
- Simplify login and logout in tests by using the helper only
- Remove all `cy.wait` from tests - [we should rely on `timeout` or aliased routes instead as a best practice](https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting)
- Increased timeouts specifically on tests I can see in Cypress dashboard have been flaking regularly this past month. Went through all "medium" severity flakes.
   - In particular `questionResponse.spec` and  `unlockResubmit.spec` cause issues
   - Every single medium severity flake is caused by login/logout or file upload. Made sure those helpers have long timeouts.
- Deleted tests that are redundant 
  -  Duplicate test scenarios or tests that aren't covering feature requirements 
- Adjusted Cypress config
  - Reduced retries to one time since we are increasing timeouts so much. Want to return more quickly for valid failures
  - Add run by workflow feature for local runs
  - Reduce crashes locally by using cypress config settings suggested by error messages on OOM crahses 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3410

#### Test cases covered

No new tests. Deleted a few tests entirely (1. does dashboard load - we check that implicitly through login; 2. do we set dates properly - we check that through other rates tests). Simplified remaining tests.

## QA guidance

Cypress should pass, welcome any other comments on the code change.


haworku:
I have tests green locally just trying to get runs through to check that I've totally fixed CI. struggling to get stuff to run

macrael (approved):
really good cleanup

services/cypress/cypress.config.ts
```diff
@@ -15,6 +15,7 @@ module.exports = defineConfig({
         videosFolder: 'videos',
         viewportHeight: 1080,
         viewportWidth: 1440,
+        experimentalRunAllSpecs: true,
```
JasonLin0991, line 18: This doesn't seem to be working for me. When I run all tests, only `documents.spec.ts` shows up, but never runs.

JasonLin0991 (approved):
Teamwork!

services/cypress/cypress.config.ts
haworku: @JasonLin0991 Yes -  the overall test runs are broken (its still in experimental stage, I think that's why I was reading github issues a bit). However, with this flag you can now run by individual folder (so just the cms workflow tests or just the state workflow tests). I thought that was still an incremental improvement from behavior with the flag. Thoughts?

JasonLin0991: Sounds good to me!

PR #1711, opened 2023-06-02
Move aws region into a var, remove from secrets

mojotalantikite:
## Summary

We hit an issue in a branch where Jason was attempting to output variables to use in our Cypress configuration, but because the value contains the string `us-east-1`, GitHub Actions masks that output value. There's no reason why `us-east-1` should be a secret, except I vaguely recall we had inherited that config from the quickstart.

This moves that variable to use `vars` instead of `secrets`.


JasonLin0991 (approved):
This fixed the issue!

PR #1715, opened 2023-06-05
MR-3117: Fix multiple concurrent mutations

rswerve:
## Summary

Let's try this one more time.  Debouncing didn't solve the problem, because it was still possible to sneak a mutation through between the debounce interval and new page load.  This PR takes a different approach, and thanks to @haworku for pointing out how Formik was working "behind the scenes".

By not passing a `continueOnClick` handler to some components, we were allowing Formik to silently handle the browser form submission, and by allowing that, we were giving up our ability to disable a mutation under whatever conditions we chose (in this case, when one had just been sent).  In the PR, you'll see logic moved from `onSubmit` (where we couldn't control how it was called) to `onContinueClick`, where we can.

#### Test cases covered

No new tests.  Slightly fixed up RateDetails.test.tsx to handle an issue of state bleeding across tests.


haworku (approved):
Do you feel like we should just add this to SubmissionType and Contacts page as well now? Then we just cover the whole form and are done here (documents page already has the pattern). 

Everything I was testing seemed to look good I was looking especially at the validations on continue click still working as expected as well as multiple clicks at once

rswerve:
> 

🤔 I was thinking of passing `continueOnClick` into PageActions as the real fix here, and those pages already do that.  Were you thinking that we'd also make `onSubmit` a no-op in those pages?

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.tsx
```diff
@@ -466,14 +466,8 @@ export const ContractDetails = ({
...
             initialValues={contractDetailsInitialValues}
-            onSubmit={(values, { setSubmitting }) => {
-                return handleFormSubmit(values, setSubmitting, {
-                    shouldValidateDocuments: true,
-                    redirectPath:
-                        draftSubmission.submissionType === 'CONTRACT_ONLY'
-                            ? `../contacts`
-                            : `../rate-details`,
-                })
+            onSubmit={() => {
```
macrael, line 469: what happens if you leave this function out?

macrael (approved):
This makes sense, we can do this in all of our forms that use this button.

I would also like to change the buttons themselves, I think it's a mismatch for us to control wether we call onClick but not control wether we call onSubmit. I think simply setting the onClick function to call preventDefault() when we don't want anything to happen will make them behave the way we expect. Happy to pair to show you what I mean.

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.tsx
rswerve: formik type error

haworku: Its a react-uswds props types error. We would need to update that external dep (which we aren't even up to date on to fix).

Honestly I would get around this by making a `MCRForm` that's just our internal form for use in this app, copy react-uswds form code which is very simple and just adjust the required props.

macrael (approved):
Yay, this looks good to me

PR #1736, opened 2023-06-12
Increase timeout for updateHealthPlanFormDataMutation.

JasonLin0991:
## Summary

Looking at the [cypress failures](https://cloud.cypress.io/projects/tt5hbz/runs/bae7be7b-918e-4ec5-b88f-f5ee7b28a879/test-results/226d784d-8eb8-4d47-bd44-2f407e61e05b) for the dependabot [PR#1735](https://github.com/Enterprise-CMCS/managed-care-review/pull/1735). 

Increasing the timeout for uploading files may have let servers go to sleep. Then hitting `Continue`, the wait for `updateHealthPlanFormDataMutation` times out. Increasing the default timeout for `cy.wait('@updateHealthPlanFormDataMutation')` should allow the tests to wait longer for server response.

Also, increasing the timeout for the `wait` in `questionResponseCommands.ts` since they also wait on graphql request right after document upload.


rswerve (approved):
sure, why not

PR #1739, opened 2023-06-14
MR-3452: Add a content security policy

rswerve:
## Summary

Our content security policy included a deprecated directive (prefetch-src), which seemed to be blocking the analytics package from CMS.  I set a CSP explicitly that uses standard directives and should allow their script.  We can't really know until it's in prod and they try it.  But @haworku is there a way to see if Tealium and Google Analytics are unaffected before promoting to prod?

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -13,6 +13,7 @@ export CT_AWS_ROLE='ct-ado-managedcare-developer-admin'
 export CT_IDMS='2'
 
 # values formerly in .env (required)
+export INLINE_RUNTIME_CHUNK=false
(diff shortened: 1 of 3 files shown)


haworku:
@rswerve No unfortunately we can't test GA and the actual analytics data flow -  its always been a test in prod situation we only look at the console and make the tealium load script doesnt throw errors (it will if that script just doesn't load is what I remember)

Right now I think the app isn't even loading...

rswerve:
Closing.  Based on our latest meeting with the Blast/Tealium folks, this problem wasn't on our end.

PR #1742, opened 2023-06-14
Upgrade Lerna to 7

mojotalantikite:
## Summary

We've been seeing some issues in CI where Lerna will complain about a configuration option that is no longer available in Lerna 7 (which is true), and sometimes be totally fine with our Lerna 6 config. Looks like GHA will sometimes put us on an Ubuntu server that has Lerna 7 installed, and sometimes it will put us on one with Lerna 6 installed.

This change upgrades us to Lerna 7 and then uses `npx` to call it, using our yarn installed local version, rather than the system version in the GHA environment.

https://qmacbis.atlassian.net/browse/MR-3458

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -99,7 +99,7 @@ jobs:
         run: npx lerna run gqlgen
 
       - name: Generate Protos
-        run: npx lerna run generate --scope=@managed-care-review/app-proto
+        run: npx lerna run generate --scope=app-proto
 
       - name: Web Unit Tests
         env:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -59,8 +59,7 @@ jobs:
 
       - name: build scripts
         shell: bash
-        run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/deploy.yml
```diff
@@ -248,7 +249,7 @@ jobs:
 
   deploy-infra:
     needs: [begin-deployment]
-    uses: Enterprise-CMCS/managed-care-review/.github/workflows/deploy-infra-to-env.yml@main
+    uses: Enterprise-CMCS/managed-care-review/.github/workflows/deploy-infra-to-env.yml@mt-fix-lerna-7
```
rswerve, line 252: assuming this is deliberate and will get changed back to main in the future...

rswerve (approved):

.github/workflows/deploy.yml
macrael: This is something we have to do every time you make a change to one of the sub-workflows. Once this PR merges it will be followed up with a PR that switches things back to main since this will now be on main.

PR #1748, opened 2023-06-16
Fix dev tool.

JasonLin0991:
## Summary

Running the dev tool was not working because of the Lerna upgrade. Just needed to add `npx` whenever the tool uses Lerna.


rswerve (approved):
thanks for fixing this

PR #1762, opened 2023-06-23
MR-3281: handle missing values

rswerve:
## Summary

This is a follow-up to [previous work](https://github.com/Enterprise-CMCS/managed-care-review/pull/1656).  We didn't anticipate that some document records would be missing a corresponding file in S3, but we ran into that in Val, so here we skip those records rather than blowing up the whole migration.  

In addition, a record with missing formData was breaking the reports endpoint, so I put in a similar fix there.

In both places, we've added a `recordException` call to otel to track when these errors occur.


services/app-api/src/handlers/add_sha.ts
```diff
@@ -34,62 +39,77 @@ export const calculateSHA256 = async (s3URL: string): Promise<string> => {
...
+            )
             formData.contractDocuments = await updateDocumentsSHA256(
-                formData.contractDocuments
+                formData.contractDocuments,
+                serviceName
             )
             for (const rateInfo of formData.rateInfos) {
                 rateInfo.rateDocuments = await updateDocumentsSHA256(
-                    rateInfo.rateDocuments
+                    rateInfo.rateDocuments,
```
JasonLin0991, line 111: Just to confirm, we will want this PR to merge in before [#1763](https://github.com/Enterprise-CMCS/managed-care-review/pull/1763)?

Otherwise, we would need to add `rateInfo.supportingDocuments` in here.

JasonLin0991 (approved):
Looks good!

PR #1763, opened 2023-06-23
MR-3310: Rate migration

rswerve:
## Summary

Please see [the ticket](https://qmacbis.atlassian.net/browse/MR-3310) for all the details.  In short, we're changing where we store supporting documents for rates.  We handle the general case and a specific case, as detailed in the ticket.

This will be run once per environment.

#### Test cases covered

Three tests to go with the migrator: the general case, the specific case, and whether it can skip a revision that contains errors.


services/app-api/src/handlers/migrate_rate_documents.ts
```diff
@@ -0,0 +1,152 @@
...
+    return result
+}
+
+export const main: Handler = async (event, context) => {
+    const store = await getDatabaseConnection()
+
+    const revisions = await getRevisions(store)
+
+    // Get the pkgID from the first revision in the list
+    const pkgID = revisions[0].pkgID
```
JasonLin0991, line 143: Curious why we only check the first revision and not all of them.

services/app-api/src/handlers/migrate_rate_documents.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (!pkgID) {
```
JasonLin0991, line 144: Curious why we only check the first revision and not all of them.

services/app-api/src/handlers/migrate_rate_documents.ts
```diff
@@ -0,0 +1,152 @@
...
+                    )
+                    throw new Error('Error updating revision')
+                }
+            } catch (err) {
+                console.error(`Error updating revision ${revision.id}: ${err}`)
+                throw err
+            }
+        } else {
+            const error = `Error decoding formDataProto for revision ${revision.id} in rate migration: ${decodedFormDataProto}`
+            console.error(error)
```
JasonLin0991, line 91: Could we include the `pkgID` in the error message too? If there was an error, it would be difficult to find which package this revision was from using the revision ID.

JasonLin0991 (approved):
Nice work. The tests look great too!

We should probably take a quick peek at the prod submissions again since new submissions came in. Just incase there is another edge case.

services/app-api/src/handlers/migrate_rate_documents.ts
rswerve: The pkgID is stable across revisions, so this is basically a check that we have well-formed data.  https://github.com/Enterprise-CMCS/managed-care-review/blob/d696f81ef6dc9ce7163124d1a35d4c018d48b220/services/app-api/prisma/schema.prisma#L34

services/app-api/src/handlers/migrate_rate_documents.ts
rswerve: good call.  done

PR #1781, opened 2023-06-30
MR-3404: email address change

rswerve:
## Summary

Made a custom hook to serve up strings, because this way we can use LaunchDarkly in a central spot to control a string throughout the app.

#### Screenshots

image
image

### DIFF
--- a/services/app-web/src/components/Header/UserLoginInfo/UserLoginInfo.tsx
+++ b/services/app-web/src/components/Header/UserLoginInfo/UserLoginInfo.tsx
@@ -8,7 +8,7 @@ import { idmRedirectURL } from '../../../pages/Auth/cognitoAuth'
 import { AuthModeType } from '../../../common-code/config'
 
 import styles from '../Header.module.scss'
-import { useString } from '../../../hooks/useString'
+import { useStringConstants } from '../../../hooks/useStringConstants'
 
 type LogoutHandlerT = (
     e: React.MouseEvent<HTMLButtonElement, MouseEvent>
@@ -18,7 +18,7 @@ const LoggedInUserInfo = (
     user: User,
     logout: LogoutHandlerT
 ): React.ReactElement => {
-    const stringConstants = useString()
+    const stringConstants = useStringConstants()
     const MAIL_TO_SUPPORT = stringConstants.MAIL_TO_SUPPORT
     return (
         <div className={styles.userInfo}>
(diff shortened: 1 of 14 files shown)


services/app-web/src/hooks/useString.ts
```diff
@@ -0,0 +1,15 @@
+import { useLDClient } from 'launchdarkly-react-client-sdk'
+import { featureFlags } from '../common-code/featureFlags'
+
+export const useString = () => {
+    const ldClient = useLDClient()
```
JasonLin0991, line 5: Do we plan on using this string long-term? Or will it be removed once this flag is gone?

services/app-web/src/hooks/useString.ts
```diff
@@ -0,0 +1,15 @@
+import { useLDClient } from 'launchdarkly-react-client-sdk'
+import { featureFlags } from '../common-code/featureFlags'
+
+export const useString = () => {
```
JasonLin0991, line 4: Do we plan on using this string long-term? Or will it be removed once this flag is gone?

If we are keeping it for the long run, I would like this hook to be named something like `useStringConstants`.
```suggestion
export const useStringConstants = () => {
```

services/app-web/src/components/Header/UserLoginInfo/UserLoginInfo.tsx
```diff
@@ -17,15 +18,18 @@ const LoggedInUserInfo = (
...
     logout: LogoutHandlerT
 ): React.ReactElement => {
+    const stringConstants = useString()
+    const MAIL_TO_SUPPORT = stringConstants.MAIL_TO_SUPPORT
     return (
         <div className={styles.userInfo}>
+            <span>Contact </span>
             <a
-                href="mailto: mc-review@cms.hhs.gov, mc-review-team@truss.works"
+                href={`mailto: ${MAIL_TO_SUPPORT}, mc-review-team@truss.works`}
```
JasonLin0991, line 27: This mailto link and the one in the footer open up a new tab with these console errors. This might be my browser because I tried opening it in incognito mode, and it worked fine.

If you can confirm this doesn't happen to you, then we can ignore this.

image

JasonLin0991 (approved):
Looks good.

I think cypress timed out trying to rerun all the specs in one run. This happens sometimes if we try to rerun failed cypress tests, instead we have to rerun all jobs.

services/app-web/src/components/Header/UserLoginInfo/UserLoginInfo.tsx
rswerve: Works for me.  Probably an extension on your browser, like you said.

services/app-web/src/hooks/useString.ts
rswerve: I'd like to keep it around, at least until we have a real solution for strings in the app, so I'll rename it.

PR #1791, opened 2023-07-05
add helpdesk user with same perms as admin

macrael:
## Summary

We want our helpdesk to have helpdesk user types in IDM, this adds them to the app with the same permissions as admin users.

#### Test cases covered

Only one basic test was added so far for the parsing. 

## QA guidance

there should be a new helpdesk@example.com user you can log in with on the review app and confirm that you can see the same stuff an admin can see


services/app-api/src/resolvers/healthPlanPackage/fetchHealthPlanPackage.ts
```diff
@@ -58,7 +59,11 @@ export function fetchHealthPlanPackageResolver(
                     'user not authorized to fetch data from a different state'
                 )
             }
-        } else if (isCMSUser(context.user) || isAdminUser(context.user)) {
+        } else if (
+            isCMSUser(context.user) ||
+            isAdminUser(context.user) ||
+            isHelpdeskUser(context.user)
```
JasonLin0991, line 65: With this new type of user who cannot fetch a draft, we should change the error message `'CMS user not authorized to fetch a draft'` to something more generic.

services/app-web/src/localAuth/LocalLogin.tsx
```diff
@@ -72,6 +73,13 @@ const localUsers: LocalUserType[] = [
         familyName: 'Coldstart',
         role: 'ADMIN_USER',
     },
+    {
+        id: 'user5',
```
JasonLin0991, line 77: We already have a user with the id `user5`
```suggestion
        id: 'user7',
```

services/app-web/src/pages/SubmissionSideNav/SubmissionSideNav.tsx
```diff
@@ -103,15 +103,19 @@ export const SubmissionSideNav = () => {
 
     const isCMSUser = loggedInUser?.role === 'CMS_USER'
     const isAdminUser = loggedInUser?.role === 'ADMIN_USER'
+    const isHelpdeskUser = loggedInUser?.role === 'HELPDESK_USER'
 
     // CMS Users can't see DRAFT, it's an error
-    if (submissionStatus === 'DRAFT' && (isCMSUser || isAdminUser)) {
+    if (
+        submissionStatus === 'DRAFT' &&
+        (isCMSUser || isAdminUser || isHelpdeskUser)
```
JasonLin0991, line 111: Curious if a help desk user would ever be asked for support on a draft submission.

JasonLin0991 (approved):
Looks good!

The only thing that would need to be fixed is the user id in `LocalLogin.tsx`

services/app-api/src/resolvers/healthPlanPackage/fetchHealthPlanPackage.ts
JasonLin0991: Unless help desk users are considered CMS users?

services/app-web/src/pages/SubmissionSideNav/SubmissionSideNav.tsx
macrael: Yes, they would, I'm going to file a ticket to track that. We should also make sure they can't pose questions etc. The goal with this PR is to get some access to our HD users quickly.

PR #1807, opened 2023-07-12
MR-3561: Validate prisma to domain model functions

JasonLin0991:
## Summary
[MR-3561](https://qmacbis.atlassian.net/browse/MR-3561)

This was a follow-up ticket to [MR-3428](https://qmacbis.atlassian.net/browse/MR-3428) adding in:
- Zod validations in `parseDomainData.ts`
- Contract domain types generated from Zod schema and no longer using types from `contractTypes.ts`
- Clean up work from MR-3428

#### Test cases covered
- `parseDomainData.test.ts`
   - `'parseDraftContract'`
      - `'can parse valid draft domain data with no errors'`
      - `'parseDraftContract returns an error when draft contract data is invalid'`
   - `parseDraftContractRevision`
      - `parseDraftContractRevision returns an error when draft contract data is invalid`
   - `parseContractWithHistory`
      - `'can parse valid contract domain data with no errors'`
      - `parseContractWithHistory returns an error when contract data is invalid`
- `prismaToDomainModel.test.ts`
   - `'contractFormDataToDomainModel'`
      - 'correctly adds document categories to each document'
   - `'getContractStatus'`
      - `'correctly gets contract status from unordered revisions'`


haworku:
@macrael I will take a pass at this tomorrow AM, would be great if you could review or 👍 so I know you saw this. Recognize you have thought quite a bit about zod design here.

services/app-api/src/domain-models/contractAndRates/parseDomainData.ts
```diff
@@ -0,0 +1,86 @@
...
+    revision: DraftContractRevisionTableWithRelations
+): ContractRevision | Error {
+    const draftContractRevision = draftContractRevToDomainModel(revision)
+    const parseDraft = contractRevisionZodSchema.safeParse(
+        draftContractRevision
+    )
+
+    if (!parseDraft.success) {
+        console.warn(
+            `ERROR: attempting to parse prisma draft contract revision failed: ${parseDraft.error}`
```
haworku, line 29: issue: Can you quickly throw in some OTEL here,  not just console.warn. Same with other errors in this file. I'm assuming we wouldn't call this function in prod unless the flag is on so not worried about flooding logs. Also check that the error also returns the top level  contract id or the rate id clearly.

haworku (approved):
I feel good merging this as long as we add the OTEL stuff. That will be useful to just build with OTEL in mind, we can always standardize the actual messages later.  I appreciate the type inference going here and how it clarifies our domain models.

services/app-api/src/postgres/contractAndRates/prismaToDomainModel.ts
```diff
@@ -92,19 +135,158 @@ function contractFormDataToDomainModel(
...
-        updatedBy: info.updatedBy.email,
-        updatedReason: info.updatedReason,
+        id: contract.id,
+        status: getContractStatus(contract.revisions),
+        stateCode: contract.stateCode,
+        stateNumber: contract.stateNumber,
+        revisions,
+    }
+}
+
```
macrael, line 201: Can you add a comment here and on the rate one? Suggested text:

contractWithHistoryToDomainModel constructs a history for this particular contract including changes to all of its revisions and all related rate revisions, including added and removed rates

services/app-web/src/common-code/proto/healthPlanFormDataProto/zodSchemas.ts
```diff
@@ -7,12 +7,12 @@
 
 import { z } from 'zod'
 
-const submissionTypeSchema = z.union([
+export const submissionTypeSchema = z.union([
```
macrael, line 10: Can you export these at the bottom of the file? That's our default style.

macrael (approved):
Looks great!

services/app-api/src/domain-models/contractAndRates/parseDomainData.ts
JasonLin0991: My thought was to log OTEL errors at the resolver level. 

I think if we wanted to log here, we could, but we would have to pass down the `span` from `context` into the `postgresStore`? If that is the case I would want to do that in a separate ticket because it would be some refactors in other parts of the code.

PR #1810, opened 2023-07-12
MR-3212: Create a lambda to run migrations

rswerve:
## Summary

Might have to adjust this ticket or write a new one, but basically we want a lambda that will run during each CI run.  The script to actually pull from and manipulate the database is the next phase (MR-3211).

This is a lambda that runs each time we run CI.  Currently, all it does is access the database and log a little snippet.  A hello world of lambdas.

Tagging @macrael to make sure this sounds like what we discussed
Tagging @mojotalantikite because I changed your scripts a bit


mojotalantikite (approved):
There's one small thing in that we call `invoke-migrate-lambda.sh` in [`deploy.yml` as well](https://github.com/Enterprise-CMCS/managed-care-review/pull/1810/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3L340), so we'll probably want to change this invocation to take into account your changes as well. Although, I'm not even sure that that invocation is relevant these days, we might be able to drop it, but we can figure that out in the future.

Otherwise LGTM!

rswerve:
> There's one small thing in that we call `invoke-migrate-lambda.sh` in [`deploy.yml` as well](https://github.com/Enterprise-CMCS/managed-care-review/pull/1810/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3L340), so we'll probably want to change this invocation to take into account your changes as well. Although, I'm not even sure that that invocation is relevant these days, we might be able to drop it, but we can figure that out in the future.
> 
> Otherwise LGTM!

Everything I know about bash I have to google, but I thought the changes in invoke-migrate-lambda set default values for variables 2 and 3, so passing in nothing should be ok?  Let me know @mojotalantikite

PR #1812, opened 2023-07-13
MR03434 Move addtl actuary to rate revision, remove from contract

haworku:
## Summary
Following up from DTBM, we discussed moving addtl actuary off contract and onto rates since this is rates data.

**This PR:**
- removes  `addtlActuaryContacts`  and `addtlActuaryCommunicationPreference` entirely from Contracts.
- adds `addtlActuaryContract` to rates and adds more precise/distinct field names.

**Assumptions (please 👀  these):**
- both addtl and certifying actuary contacts only need one actuary communication preference. We don't need to keep tracking preferences for each set of contacts (like we have been thus far).
-  we can rename the former Rates `actuaryContacts` to `certifyingActuaryContacts` without much confusion. That's what this field stores and it makes it easier to tell difference in intent between the two actuary fields on rates.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3434

#### Test cases covered
- No new tests, we haven't done rates work much yet
- Updated existing tests with contacts that expected actuary contacts on that data

## QA guidance
- let me know if there's another way we want the prisma schema structured. I took a best guess here.


JasonLin0991 (approved):
✅

PR #1813, opened 2023-07-14
MR-3429: createHealthPlanPackage resolver returns data inserted insertContract

JasonLin0991:
## Summary
[MR-3429](https://qmacbis.atlassian.net/jira/software/projects/MR/boards/234?selectedIssue=MR-3429)

This work switches `createHealthPlanPackage` resolver to use the new `insertContract` Prisma function when creating a new package when the feature flag `rates-db-refactor` is on.
- In `createHealthPlanPackage`, it will also convert the data from the new Prisma table into the `HealthPlanPackageType` with proto data of `UnlockedHealthPlanFormDataType`. This is done using a new function `convertContractRevisionToHealthPlanRevision`.
- `convertContractRevisionToHealthPlanRevision` uses existing functions `toDomain`, `toProtoBuffer` to do most of the validation of the data. Before our new `insertContract` Prisma function returns data, it is validated using `parseDraftContract`. So test coverage for this is handled elsewhere. **But let me know if you think we need any tests for this**, and I will note that getting invalid data into this function and out is pretty difficult.

#### Test cases covered

- `createHealthPlanPackage`
   - `'inserts contract to new tables and returns a health plan package'`

## QA guidance

**NOTES**
- The work here will not work in the app. We still need to get [MR-3446](https://qmacbis.atlassian.net/browse/MR-3446) ticket done before we can test it in the app. For now, the unit tests verify that it returns correct data to the frontend.
- Turning this feature flag in `dev` will cause the cypress tests to fail, since we cannot toggle backend flags based on each test as we do with frontend tests.


services/app-api/src/domain-models/healthPlanPackage.ts
```diff
@@ -60,9 +69,144 @@ function packageSubmitters(pkg: HealthPlanPackageType): string[] {
...
+        )
+        return err
+    }
+}
+
+function convertContractRevisionToHealthPlanRevision(
+    contract: ContractType
+): HealthPlanRevisionType[] {
+    if (contract.status !== 'DRAFT') {
+        throw new Error(
```
macrael, line 97: please return errors rather than throwing them, is there something else in this function that can throw? I always prefer us to box up dependencies that throw

services/app-api/src/resolvers/healthPlanPackage/createHealthPlanPackage.ts
```diff
@@ -71,24 +79,63 @@ export function createHealthPlanPackageResolver(
...
-            setErrorAttributesOnActiveSpan(errMessage, span)
-            throw new GraphQLError(errMessage, {
-                extensions: {
-                    code: 'INTERNAL_SERVER_ERROR',
-                    cause: 'DB_ERROR',
-                },
-            })
-        }
+        //Here is where we flag the insert
+        if (ratesDatabaseRefactor) {
```
macrael, line 83: yay! This is very satisfying to see

services/app-api/src/resolvers/healthPlanPackage/createHealthPlanPackage.test.ts
```diff
@@ -3,6 +3,7 @@ import CREATE_HEALTH_PLAN_PACKAGE from '../../../../app-graphql/src/mutations/cr
 import { constructTestPostgresServer } from '../../testHelpers/gqlHelpers'
 import { latestFormData } from '../../testHelpers/healthPlanPackageHelpers'
 import { testCMSUser } from '../../testHelpers/userHelpers'
+import { testLDService } from '../../testHelpers/launchDarklyHelpers'
 
 describe('createHealthPlanPackage', () => {
     it('returns package with unlocked form data', async () => {
```
macrael, line 9: Is it possible for us to reuse these tests? Run them twice, once with the feature flag and once without?

macrael (approved):
This looks great, why won't this work yet if the API is returning the right thing?

macrael:
Oh I get it, find will not find the data we created. But we should be able to have a successful create from the app which is cool!

services/app-api/src/resolvers/healthPlanPackage/createHealthPlanPackage.ts
```diff
@@ -71,24 +79,63 @@ export function createHealthPlanPackageResolver(
...
+                convertContractToUnlockedHealthPlanPackage(contractResult)
+
+            if (pkg instanceof Error) {
+                const errMessage = `Error converting draft contract. Message: ${pkg.message}`
+                logError('createHealthPlanPackage', errMessage)
+                setErrorAttributesOnActiveSpan(errMessage, span)
+                throw new GraphQLError(errMessage, {
+                    extensions: {
+                        code: 'INTERNAL_SERVER_ERROR',
+                        cause: 'PROTO_DECODE_ERROR',
```
haworku, line 108: Isn't this a `ZOD_PARSE_ERROR`

services/app-api/src/postgres/healthPlanPackage/insertHealthPlanPackage.ts
```diff
@@ -18,9 +18,9 @@ import { PopulationCoveredType } from '../../gen/gqlServer'
 
 export type InsertHealthPlanPackageArgsType = {
     stateCode: string
-    populationCovered?: PopulationCoveredType
+    populationCovered: PopulationCoveredType
     programIDs: string[]
```
haworku, line 22: I think these fields are still optional until we remove that CHIP_ONLY_FORM feature flag

services/app-api/src/resolvers/healthPlanPackage/createHealthPlanPackage.test.ts
haworku: I think we need to figure out how to, because thats what we have said is our definition of done.

services/app-api/src/resolvers/healthPlanPackage/createHealthPlanPackage.ts
JasonLin0991: At this point, the zod parse error would happen in `toDomain` so it would be when decoding. Maybe the error checking above this one would need more work

```typescript
const contractResult = await store.insertDraftContract(insertArgs)
if (contractResult instanceof Error) {
    const errMessage = `Error creating a draft contract. Message: ${contractResult.message}`
    logError('createHealthPlanPackage', errMessage)
    throw new GraphQLError(errMessage, {
        extensions: {
            cause: 'DB_ERROR',
        },
    })
}
```

That store function could return a prisma error or a parsing error.

services/app-api/src/resolvers/healthPlanPackage/createHealthPlanPackage.test.ts
JasonLin0991: I'll let someone else take over figuring out to rerun these tests since I'm out all next week. The rest should be good to go.

JasonLin0991: @macrael @haworku Hello, I read the jest docs, and `describe.each` seems to do what we wanted. Probably could use some more work to make it reusable. Feel free to modify it, or if you have a better way you can replace it!

PR #1846, opened 2023-08-01
MR-3569: Prod: User reported error in downloading zip files

JasonLin0991:
## Summary
[MR-3569](https://qmacbis.atlassian.net/browse/MR-3569)

Changed the expiration for the document URLs to 1 hour. I was still able to download files past 15 minutes.

This is the error after an hour (Using a different file)

## QA guidance
If you keep opening the same file, it will work past 1 hour. Not sure why, maybe it's cached? But below are instructions on how to correctly test the expiration time.

For testing this issue:

1. Open a submission, but do not click on any file.
2. Wait 45 minutes or right before an hour.
3. Click on any file to download, and it should still work
4. Wait past 1 hour and click on **A DIFFERENT file** and it should open a new tab to an error.


mojotalantikite (approved):
Confirmed -- download link works fine before an hour, access denied after an hour.

PR #1865, opened 2023-08-08
Insert Rate

macrael:
## Summary

I started from scratch and copied over rate work from Hana's branch to get insertRate working. 

There's still a decent amount of type cleanup to work through but this is working


services/app-api/src/postgres/contractAndRates/insertRate.ts
```diff
@@ -59,28 +125,30 @@ async function insertDraftRate(
...
                             updatedAt: contractRevision.updatedAt,
                             formData:
                                 contractFormDataToDomainModel(contractRevision),
-                            rateRevisions: contractRevision.rateRevisions.map(
-                                (rr) => ({
-                                    id: rr.rateRevisionID,
-                                    revisionFormData: rr.rateRevision.name,
-                                })
-                            ),
+                            rateRevisions: [],
```
JasonLin0991, line 146: Is `rateRevisions` still needed here? It could be confusing that these `contractRevisions` have an empty array of  `rateRevisions`. This empty string could also mean no relationships in the DB. I rather we are explicit and not include it in `finalRate`.

services/app-api/src/postgres/contractAndRates/insertRate.ts
```diff
@@ -59,28 +125,30 @@ async function insertDraftRate(
                 },
             })
 
-            return {
+            const finalRate: RateType = {
                 id: rate.id,
+                status: 'DRAFT',
```
JasonLin0991, line 130: I think you mentioned using `getContractStatus` here instead of hard coding the status.

JasonLin0991 (approved):
Looks awesome!

PR #1867, opened 2023-08-08
MR-3405: Reference help desk email in system notification emails

JasonLin0991:
## Summary
[MR-3405](https://qmacbis.atlassian.net/browse/MR-3405)

- References the help desk email in **New submission** state emails. Notice when you reply to the email, the help desk email is now in the recipient instead of `mc-review-qa@truss.works`.
- Removed references to MC review dev team email
- Updated Email settings page support tab with help desk email.


macrael (approved):
Looks good, are we still using the devTeamHelpAddress anywhere?

JasonLin0991:
> Looks good, are we still using the devTeamHelpAddress anywhere?
Doesn't look like it. Would we ever use this email again? Should I remove it from the backend and the store?

haworku:
@JasonLin0991 oh! Yes please remove from everywhere.  dev team is no longer support/help at all. Please also update all docs that explain configuration `Configuration.md` and https://github.com/Enterprise-CMCS/managed-care-review/blob/main/docs/Configuration.md#email-configuration those were created to avoid confusion around what the variables are and why they exist.

haworku (approved):
Good work. thanks for catching the settings page aspect also.

PR #1869, opened 2023-08-09
Fix type export issue 

haworku:
## Summary
To address failed promote of #1865


macrael (approved):
Weird that this works in PR

haworku:
@macrael pushed another commit I think this would prevent this class of issue since its about export/import for types

macrael:
nice

PR #1878, opened 2023-08-15
Adjust FormEditable to use domain model types

haworku:
## Summary
Follow on to #1876 
- have form editable types use domain model types rather than Prisma types
- address inconsistency in typing between the database and domain model with contacts fields. Made contacts fields all optional to address this - we need to be able to partially save as draft. 
- mock out the tests in progress to make it more visible what functionality is still missing from updates


services/app-api/src/emailer/emails/unlockPackageStateEmail.ts
```diff
@@ -19,9 +19,10 @@ export const unlockPackageStateEmail = async (
     submitterEmails: string[]
 ): Promise<EmailData | Error> => {
     const isTestEnvironment = config.stage !== 'prod'
-    const stateContactEmails = formData.stateContacts.map(
-        (contact) => contact.email
-    )
+    const stateContactEmails: string[] = []
+    formData.stateContacts.forEach((contact) => {
```
macrael, line 23: If you wanted to keep using map semantics you could add a filter here instead. But this totally works.

services/app-api/src/postgres/contractAndRates/updateDraftRate.ts
```diff
@@ -1,24 +1,20 @@
...
-    RateDocument,
-    RateSupportingDocument,
-    RateType as PrismaRateType,
-} from '@prisma/client'
-import { findRateWithHistory } from './findRateWithHistory'
+    RateType as DomainRateType,
+    ActuaryCommunicationType,
+    ActuaryContact,
+    SubmissionDocument,
+} from 'app-web/src/common-code/healthPlanFormDataType'
```
macrael, line 9: Yay, thank you, I missed this. My rule for non-domain types is that they should never escape the package they are relevant to. So since the editable types are arguments that are going to be passed in from our resolvers, they should be based on domain types so that nothing outside of the postgres package ever has to import something from prisma

macrael (approved):

services/app-api/src/postgres/contractAndRates/updateDraftRate.ts
haworku: Yep! Makes sense. This was an error from my code editor autocomplete I think - because the two types have almost the exact same names I didn't notice the discrepancy until I started writing the field by field tests.

PR #1881, opened 2023-08-16
Fix document download warning banner margin for Q&A flag off UI

JasonLin0991:
## Summary
Fixed margins for warning banner when the Q&A UI is toggled off.


haworku (approved):
LGTM

PR #1885, opened 2023-08-21
Skip ahead to re-run only Cypress

mojotalantikite:
## Summary

There is a bug -- or really, a lack of functionality -- with cypress re-runs when using the GHA `re-run failed jobs` option, which is a [known issue](https://github.com/cypress-io/github-action/issues/531#issuecomment-1608902697). If you're using a parallelized container build in GHA CI (as we do) and a container or two has a failed spec, choosing to use the GHA `re-run failed jobs` feature will load _all_ of the Cypress tests onto that container that has failed and try to re-run everything there. This is unexpected, as I think people assume only the failed specs and other tests that were allocated to the failed container will re-run. Since we have a 25 minute timeout on Cypress runs, and we're running the entire test suite on a fraction of the containers in this scenario, we'll always hit that timeout and get a failed run.

I've added a bit of a hack-y workaround here until a re-run feature is taken on by the Cypress team. Basically, if the CI skipping script sees a commit message with `cypress re-run` in it, it'll skip past all the infra and app deploy stages and bring us right to the Cypress run stage. It's not as good as re-running only the failed test specs, but I guess it's better than having to wait for an entire CI deploy due to a flaky test.

This also changes the ID for cypress runs to help us correlate the run in Cypress dashboard to the GHA run a little easier.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3225

### DIFF
--- a/scripts/get-changed-services/index.ts
+++ b/scripts/get-changed-services/index.ts
@@ -33,6 +33,12 @@ async function main() {
         return
     }
 
+    // force just a cypress re-run
+    if (commitMessage.includes('cypress re-run')) {
+        core.setOutput('changed-services', [])
+        return
+    }
+
     // if we haven't had a run on this branch, we need to deploy everything
     if (allWorkflowRuns.data.total_count === 0) {
         core.setOutput('changed-services', deployAllServices)
--- a/scripts/get-changed-services/index.ts
+++ b/scripts/get-changed-services/index.ts
@@ -28,6 +28,8 @@ async function main() {
     // get the latest commit in the branch to see if we are forcing a run
     const latestCommit = await getLatestCommitSHA()
     const commitMessage = await getLatestCommitMessage(latestCommit)
+
+    // force a complete CI run
     if (commitMessage.includes('ci-force-run')) {
         core.setOutput('changed-services', deployAllServices)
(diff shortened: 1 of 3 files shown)


scripts/get-changed-services/index.ts
```diff
@@ -28,11 +28,19 @@ async function main() {
...
     const commitMessage = await getLatestCommitMessage(latestCommit)
+
+    // force a complete CI run
     if (commitMessage.includes('ci-force-run')) {
         core.setOutput('changed-services', deployAllServices)
         return
     }
 
+    // force just a cypress re-run
+    if (commitMessage.includes('cypress re-run')) {
```
macrael, line 39: I didn't know about ci-force-run, where would be a good place to document these CI incantations?

macrael (approved):
straightforward workaround

scripts/get-changed-services/index.ts
mojotalantikite: The `force-ci-run` is in the main Readme, but it looks like it's a little buried and hard to pick out. I just pushed up a change that documents the new `cypress re-run` and breaks those docs out into a section so it's easier to run across!

PR #1886, opened 2023-08-21
MR-3639: Modify update hpp mutation

JasonLin0991:
## Summary
[MR-3639](https://qmacbis.atlassian.net/browse/MR-3639)

- Modified updateHHP to update contract only submissions with the new DB.
- Updated `managedCareEntities` type to be string literals of the entities instead of a string array.
- Removed `documentCategories` from the `contractFormDataSchema`. Adding document categories in `convertContractRevisionToHealthPlanRevision`.
- Updating `stateContacts`, `supportingDocuments`, and `contractDocuments` is not implemented in this PR. We will need a reliable way to CRUD these relationships using a `UUID` or something. This work of adding a unique ID will be done in a different ticket.
- Just heads up, looks like our frontend doesn't let state contacts be partially filled out when `Save as draft`

#### Test cases covered

- `updateHealthPlanFormData.test.ts`
   - Used `describe.each` to test flag on and off for existing tests.
   - `'updates valid fields in the formData'`
      - Updated to test several form data fields.
   - `'updates documents and state contacts. Complete after documents and state contacts have uuids in proto'`
      - Added this `todo` so we don't forget to test these fields.

### DIFF
--- a/services/app-api/src/domain-models/contractAndRates/rateTypes.ts
+++ b/services/app-api/src/domain-models/contractAndRates/rateTypes.ts
@@ -2,18 +2,31 @@ import { z } from 'zod'
 import {
     actuaryCommunicationTypeSchema,
     actuaryContactSchema,
+    rateCapitationTypeSchema,
+    rateTypeSchema,
+    sharedRateCertDisplay,
+    submissionDocumentSchema,
     contractExecutionStatusSchema,
     contractTypeSchema,
     federalAuthoritySchema,
     populationCoveredSchema,
-    rateCapitationTypeSchema,
-    rateTypeSchema,
-    sharedRateCertDisplay,
     stateContactSchema,
-    submissionDocumentSchema,
     submissionTypeSchema,
 } from '../../../../app-web/src/common-code/proto/healthPlanFormDataProto/zodSchemas'
 import { updateInfoSchema } from './updateInfoType'
+
+const documentSchema = z.object({
+    name: z.string(),
+    s3URL: z.string(),
+    sha256: z.string().optional(),
+})
+
+const managedCareEntitiesSchema = z.union([
+    z.literal('MCO'),
+    z.literal('PIHP'),
+    z.literal('PAHP'),
+    z.literal('PCCM'),
+])
 
 // hacky workaround - importing from contract types is causing circular imports and this error
 // https://github.com/colinhacks/zod/issues/643#issuecomment-1015277287
@@ -26,13 +39,13 @@ const _contractFormDataSchema = z.object({
     riskBasedContract: z.boolean().optional(),
     submissionDescription: z.string(),
     stateContacts: z.array(stateContactSchema),
-    supportingDocuments: z.array(submissionDocumentSchema),
+    supportingDocuments: z.array(documentSchema),
     contractType: contractTypeSchema,
     contractExecutionStatus: contractExecutionStatusSchema.optional(),
-    contractDocuments: z.array(submissionDocumentSchema),
+    contractDocuments: z.array(documentSchema),
     contractDateStart: z.date().optional(),
     contractDateEnd: z.date().optional(),
-    managedCareEntities: z.array(z.string()),
+    managedCareEntities: z.array(managedCareEntitiesSchema),
     federalAuthorities: z.array(federalAuthoritySchema),
     inLieuServicesAndSettings: z.boolean().optional(),
     modifiedBenefitsProvided: z.boolean().optional(),
@@ -61,7 +74,6 @@ const _contractRevisionSchema = z.object({
     updatedAt: z.date(),
     formData: _contractFormDataSchema,
 })
-//
 
 // The rate form data  is the form filled out by state users submitting rates for review
 const rateFormDataSchema = z.object({
--- a/services/app-api/src/domain-models/contractAndRates/rateTypes.ts
+++ b/services/app-api/src/domain-models/contractAndRates/rateTypes.ts
@@ -1,115 +1,8 @@
 import { z } from 'zod'
 import {
-    actuaryCommunicationTypeSchema,
-    actuaryContactSchema,
-    rateCapitationTypeSchema,
-    rateTypeSchema,
-    sharedRateCertDisplay,
-    submissionDocumentSchema,
-    contractExecutionStatusSchema,
-    contractTypeSchema,
-    federalAuthoritySchema,
-    populationCoveredSchema,
-    stateContactSchema,
-    submissionTypeSchema,
-} from '../../../../app-web/src/common-code/proto/healthPlanFormDataProto/zodSchemas'
-import { updateInfoSchema } from './updateInfoType'
-
-const documentSchema = z.object({
-    name: z.string(),
-    s3URL: z.string(),
-    sha256: z.string().optional(),
-})
-
-const managedCareEntitiesSchema = z.union([
-    z.literal('MCO'),
-    z.literal('PIHP'),
-    z.literal('PAHP'),
-    z.literal('PCCM'),
-])
-
-// hacky workaround - importing from contract types is causing circular imports and this error
-// https://github.com/colinhacks/zod/issues/643#issuecomment-1015277287
-// we should resolve this issue to avoid duplicating efforts but hardcoding it for now to make progress
-
-const _contractFormDataSchema = z.object({
-    programIDs: z.array(z.string()),
-    populationCovered: populationCoveredSchema.optional(),
-    submissionType: submissionTypeSchema,
-    riskBasedContract: z.boolean().optional(),
-    submissionDescription: z.string(),
-    stateContacts: z.array(stateContactSchema),
-    supportingDocuments: z.array(documentSchema),
-    contractType: contractTypeSchema,
-    contractExecutionStatus: contractExecutionStatusSchema.optional(),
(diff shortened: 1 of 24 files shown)


services/app-api/src/domain-models/contractAndRates/rateTypes.ts
```diff
@@ -2,19 +2,32 @@ import { z } from 'zod'
...
+const managedCareEntitiesSchema = z.union([
+    z.literal('MCO'),
+    z.literal('PIHP'),
+    z.literal('PAHP'),
+    z.literal('PCCM'),
+])
+
 // hacky workaround - importing from contract types is causing circular imports and this error
 // https://github.com/colinhacks/zod/issues/643#issuecomment-1015277287
 // we should resolve this issue to avoid duplicating efforts but hardcoding it for now to make progress
```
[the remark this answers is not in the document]
JasonLin0991: @macrael and I synced on this, and I will be splitting our `formData` and `revision` schema and types into their own files. This removes duplicate schemas and fixes the circular reference errors.

services/app-api/src/resolvers/healthPlanPackage/updateHealthPlanFormData.ts
```diff
@@ -59,168 +114,277 @@ export function updateHealthPlanFormDataResolver(
...
+                    Date.UTC(
+                        contractWithHistory.draftRevision.updatedAt.getUTCFullYear(),
+                        contractWithHistory.draftRevision.updatedAt.getUTCMonth(),
+                        contractWithHistory.draftRevision.updatedAt.getUTCDate()
+                    )
+                ),
+                updatedAt: contractWithHistory.draftRevision.updatedAt,
+            }
+
+            const unfixedFields = validateProtectedFields(
```
haworku, line 202: Do we even need to handle this case for the feature flag on case? I would assume we just ignore even if the proto is adjusting fields that can't be edited - because we aren't actuall writing those fields to the data and are relying on prisma handler to update things reliably (or error handle is possible concurrent calls are happening)

haworku:
>Just heads up, looks like our frontend doesn't let state contacts be partially filled out when Save as draft

haworku:
This is an excellent PR description - I got all the info I needed there about state of the code. Thanks for also follow the `it.todo` pattern so we start mocking out future test plans. I'll let @macrael give final approval but everything I saw looked good just had a comment on the editing fixed fields stuff.

services/app-api/src/resolvers/healthPlanPackage/updateHealthPlanFormData.ts
JasonLin0991: My thought was in case anyone of these fields did drift, we wouldn't want to update the data. For, example when two users try to update the same submission at the same time, one of them would be working with stale data.

services/app-api/src/resolvers/healthPlanPackage/updateHealthPlanFormData.ts
```diff
@@ -59,168 +114,277 @@ export function updateHealthPlanFormDataResolver(
...
+                const errMessage = `Issue finding a contract with history with id ${input.pkgID}. Message: ${contractWithHistory.message}`
+                logError('fetchHealthPlanPackage', errMessage)
+                setErrorAttributesOnActiveSpan(errMessage, span)
 
-        const planPackage: HealthPlanPackageType = result
+                if (contractWithHistory instanceof NotFoundError) {
+                    throw new GraphQLError(errMessage, {
+                        extensions: {
+                            code: 'NOT_FOUND',
+                            cause: 'DB_ERROR',
```
macrael, line 134: does every error need a cause? I'm not sure it makes sense to me to call a not found a db error, what other causes do we use?

services/app-api/src/resolvers/healthPlanPackage/updateHealthPlanFormData.ts
```diff
@@ -59,168 +114,277 @@ export function updateHealthPlanFormDataResolver(
...
+                    'user not authorized to fetch data from a different state',
+                    span
+                )
+                throw new ForbiddenError(
+                    'user not authorized to fetch data from a different state'
+                )
+            }
+
+            // Can't update a submission that is locked or resubmitted
+            if (!['DRAFT', 'UNLOCKED'].includes(contractWithHistory.status)) {
```
macrael, line 164: I wonder if this check is really necessary anymore now that we have the separate draftRevision slot

services/app-api/src/resolvers/healthPlanPackage/updateHealthPlanFormData.ts
macrael: Yes I hope we can drop this, if we want to check on the updated at thing we can make that a separate error that indicates that concurrent editing is happening, but none of the rest of those fields should be settable from the API anymore.

macrael (approved):
Looks great! Fun to see that we are finally able to get rid of some stuff that has stuck around b/c of the protobuf stuff

services/app-api/src/resolvers/healthPlanPackage/updateHealthPlanFormData.ts
JasonLin0991: It's not required to have a `cause`. I've been using it to specify where this error happened. Some other causes: `PROTO_DECODE_ERROR`, `EMAIL_ERROR`, `INVALID_PACKAGE_STATUS`.

services/app-api/src/resolvers/healthPlanPackage/updateHealthPlanFormData.ts
JasonLin0991: Yes, that is true. However, I would opt to keep this in to have explicit errors for contracts that are not in draft and when they are in draft but have issues getting the draft revision.

services/app-api/src/resolvers/healthPlanPackage/updateHealthPlanFormData.ts
```diff
@@ -170,49 +170,28 @@ export function updateHealthPlanFormDataResolver(
...
-            if (unfixedFields.length !== 0) {
-                const errMessage = `Transient server error: attempted to modify un-modifiable field(s): ${unfixedFields.join(
-                    ','
-                )}.  Please refresh the page to continue.`
+            // If updatedAt does not match concurrent editing occurred.
+            if (
+                contractWithHistory.draftRevision.updatedAt !==
+                unlockedFormData.updatedAt
+            ) {
+                const errMessage = `Transient server error: Concurrent editing occurred. Please refresh the page to continue.`
```
JasonLin0991, line 191: @macrael @haworku Could use some input in the error message here. I asked chatgpt to come up with some options

Optimistic Concurrency Control (commonly used with versioning):
1 ."Update failed: Record has been modified by another user."
2 ."Concurrent update error: The data you are trying to modify has changed since you last retrieved it."
3 ."Update conflict: Your changes were not saved because someone else has already updated this record."
4. "Version mismatch: The record has been updated by another process. Please refresh and try again."

Pessimistic Concurrency Control (commonly used with locks):
5. "Update failed: Record is locked by another user."
6. "Concurrency error: This data is currently being edited by another user. Please try again later."
7. "Locking conflict: Another user currently holds a lock on this record."
8. "Resource locked: Please wait for the other transaction to complete and try again."

General Concurrency Issues:
9. "Database contention: Too many concurrent updates. Please try again in a few moments."
10. "Concurrency violation: Simultaneous updates detected."
11. "Transaction collision: Another process has modified the data you are trying to update."
12. "Concurrent update error: Your operation could not be completed due to simultaneous updates by other users."

I like 2 with the addition of `Please refresh the page to continue.`

services/app-api/src/resolvers/healthPlanPackage/updateHealthPlanFormData.ts
macrael: How is the 'DRAFT' status calculated? If it's calculated based on the presence of a draftRevision then this really feels redundant.

services/app-api/src/resolvers/healthPlanPackage/updateHealthPlanFormData.ts
JasonLin0991: We don't. We show the generic error banner.

macrael: Ok then this is fine being technical

services/app-api/src/resolvers/healthPlanPackage/updateHealthPlanFormData.ts
JasonLin0991: Sounds good. Should I move the error message to the check below this? Like so:

```ts
// Can't update a submission that is locked or resubmitted
if (!contractWithHistory.draftRevision) {
    const errMessage = `Package is not in editable state: ${input.pkgID} status: ${contractWithHistory.status}`
    logError('updateHealthPlanFormData', errMessage)
    throw new UserInputError(errMessage, {
        argumentName: 'pkg',
    })
}
```

Because `findContractWithHistory` returns `ContractType` where `draftRevisions` can be undefined. We still need the check for now.

macrael: yes that makes sense to me! I can imagine our future draft/submitted types making this check feel more secure

PR #1889, opened 2023-08-23
Improve updateDraftRate/Contract to handle documents and contacts

haworku:
## Summary
Ensure that linked documents (both primary and supporting) and contacts (actuary or state) can be updated and deleted entirely.

**Other changes**
- `sha256` required everywhere
- pass the database ids through to the domain models 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3648

#### Test cases covered
- New tests in updateDraft/Contract postgres handlers around three rounds of updates. Feels a bit like overkill  now that we just delete the entire group of linked records and create again. However,  I kept the tests since originally I wrote them thinking we were doing true updates of these tables where some would change and some wouldn't - maybe we would go back to that sometime?

## QA guidance

Eventually tests pass.


services/app-api/src/postgres/contractAndRates/updateDraftContract.ts
```diff
@@ -66,62 +68,82 @@ async function updateDraftContract(
...
-                populationCovered: populationCovered,
-                programIDs: programIDs,
-                riskBasedContract: riskBasedContract,
-                submissionType: submissionType,
-                submissionDescription: submissionDescription,
-                contractType: contractType,
-                contractExecutionStatus,
-                contractDocuments: {
-                    create: contractDocuments,
+        await client.$transaction([
```
macrael, line 71: I assume this works, client.transaction with an array of client.whatever commands. When I did it before I got a transaction handle that I then called

macrael (approved):
I think this looks great. I think I read that you could include multiple operations in a single nested call. Does something like>

```
data: {
    contractDocuments: {
        deleteMany: {},
        create: contractDocuments,
    }
}
```
work?

services/app-api/src/postgres/contractAndRates/updateDraftContract.ts
haworku: Yes this is [sequential cal.](https://www.prisma.io/docs/concepts/components/prisma-client/transactions#sequential-prisma-client-operations). I believe that's what we want when making simple sequential requests like this. The syntax I saw in your work looked like the [interactive transation syntax](https://www.prisma.io/docs/concepts/components/prisma-client/transactions#interactive-transactions). Don't need that here.

JasonLin0991: Should we delete all the records or [disconnect](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#disconnect) them? <strike>I would think deleting would cause some issues with revisions. If an old revision still had these relationships, I don't think deleting would work; if it did, those revisions would then have a broken relationship. I think testing around this would be good. </strike>

Ah nvm. Looks like in `unlockContract.ts`, we don't have it implemented fully to copy over all the data to the new revision. Also, in the Prisma schema, documents and revisions have a one-to-many relationship, so a document could not be related to many contract revisions. So new revisions just create a whole new set of documents?

JasonLin0991: I wonder if v2 of document and contact updating could be something like this:

1.  Programmatically compare updated data with db data and split data into three buckets: create, disconnect, and update.
2. Then we use these buckets in their own Prisma operation in the sequential operation: 
   - createMany, 
   - disconnectMany ( this one can be in `contractRevisionTable.update`) or deleteMany
   - updateMany
3. This would have to be repeated for every document and contact table 🤡. 

Just deleting and recreating seems wrong, I know we don't use the `updatedAt` or `createdAt` fields, but doing this would make those fields a field of LIES. (Not a strong feeling, but a medium one)

services/app-api/src/postgres/contractAndRates/updateDraftContract.test.ts
```diff
@@ -40,10 +41,185 @@ describe('updateDraftContract', () => {
...
     })
 
-    it.todo(
-        'updates linked document tables as expected (via upsert, not create)'
-    )
-    it.todo('updates linked contact table as expected (via upsert, not create)')
+    it('updates linked documents as expected in multiple requests', async () => {
+        const client = await sharedTestPrismaClient()
+
+        const draftRateForm1: ContractFormEditable = {
```
JasonLin0991, line 47:
```suggestion
        const draftContractForm1: ContractFormEditable = {
```

It might be confusing in the future with this naming.

services/app-api/src/postgres/contractAndRates/updateDraftContract.test.ts
```diff
@@ -40,10 +41,185 @@ describe('updateDraftContract', () => {
         )
     })
 
-    it.todo(
-        'updates linked document tables as expected (via upsert, not create)'
-    )
-    it.todo('updates linked contact table as expected (via upsert, not create)')
+    it('updates linked documents as expected in multiple requests', async () => {
```
JasonLin0991, line 44: <strike>Let's add a test like `deletes linked documents and contacts without breaking old revisions.`</strike> Unlock contract is not fully implemented so I don't think we could test this anyways.

JasonLin0991 (approved):
Overall looks good. The only thing I would like to see is naming changes in `updateDraftContract.test.ts`

services/app-api/src/postgres/contractAndRates/updateDraftContract.ts
JasonLin0991: I wonder if v2 of document and contact updating could be something like this:

Programmatically compare updated data with db data and split data into three buckets: create, disconnect/delete, and update.
Then we use these buckets in their own Prisma operation in the sequential operation:
createMany,
disconnectMany ( this one can be in contractRevisionTable.update) or deleteMany
updateMany
This would have to be repeated for every document and contact table 🤡.
Just deleting and recreating seems wrong, I know we don't use the updatedAt or createdAt fields, but doing this would make those fields a field of LIES. (Not a strong feeling, but a medium one)

haworku: @JasonLin0991 Yep. I hear you. You are rehasing the conversation MacRae and I had yesterday. I had functions that covered each table. What you are proposing is slightly different but goes down a similar path of multi part update where we create (I was going to create an update in each table) and then disconnect/delete. I'm down to do this but I would like to discussion as a team first to all get on same page.

services/app-api/src/postgres/contractAndRates/updateDraftContract.test.ts
haworku: @JasonLin0991 I will add a todo for this its a good test to have later. I don't think this solution runs that risk because udpate only happens on new drafts and it delete things only related to that specific

JasonLin0991: > Also, in the Prisma schema, documents and revisions have a one-to-many relationship, so a document could not be related to many contract revisions. So new revisions just create a whole new set of documents?

You are correct with the way we set up the relationship, a document can never be related to more than one revision. So deleting a document that can only be related to one revision would not encounter an error where there the document has another relationship, and it would either cascade the delete to other related records or return an error because there were other relationships.

There could be a discussion on whether creating new records on unlock is something we want to continue. Since technically they are not "new" documents.

haworku: @JasonLin0991 Interesting, tis feels like an architectural discussion that goes beyond even the scope of the issues you are raising about how updating resources should work elsewhere in this PR,- the premise of our audit log has been that we create a distinct and new record even in protos. I have been assuming we would copy over all of our scalar fields similarly with contract and rates revisions and that we would also copy over the documents/contracts.

services/app-api/src/postgres/contractAndRates/updateDraftContract.ts
haworku: @JasonLin0991 I'm gonna open on a follow on PR to allow us to discuss this  - we can continue on shipping other work but use that PR to iterate on a v2 if we want. 

 Will merge this work as is.

PR #1890, opened 2023-08-24
MR-3186: migrate indexhpp resolver

JasonLin0991:
## Summary
[MR-3449](https://qmacbis.atlassian.net/browse/MR-3449)

- Migrate indexHPP resolver
   - **NOTE**: Without submit and unlock resolvers migrated, we cannot re-run existing tests without refactoring every test. For now I just placed a `todo` at the top of the tests.
   - added `validateContractsAndConvert` function to validate contract results returned from Prisma
      - Contracts that failed parsing or converting will be filtered out and logged to Otel.
   - Minimal amount of work needed in the resolver; we just need to get the contracts into the correct type using the `validateContractsAndConvert`function.
- `findAllContractsWithHistoryBySubmitInfo` new Prisma function to get all contracts that have submitted info. This will grab all contracts that are/have been submitted.
- `findAllContractsWithHistoryByState` new Prisma function to get all contracts by `stateCode`
- Updated contract and rate tests data to correctly store `programIDs` as the state program id and not the name. The frontend uses the id to find the name. Before this all contracts on the submission dashboard would have unknown programs.

#### Test cases covered

`findAllContractsWithHistoryBySubmitInfo.test.ts`
- `'returns only contracts that have been submitted or unlocked'`

`indexHealthPlanPackage.test.ts`
- `'correctly filters and log contracts that failed parsing or converting'`

## QA guidance

- The only way to verify the application is showing the contracts is through the state dashboard with draft contracts only. CMS only shows submitted and unlocked submissions.


services/app-api/src/domain-models/healthPlanPackage.ts
```diff
@@ -72,8 +72,6 @@ function packageSubmitters(pkg: HealthPlanPackageType): string[] {
 function convertContractToUnlockedHealthPlanPackage(
     contract: ContractType
 ): HealthPlanPackageType | Error {
-    console.info('Attempting to convert contract to health plan package')
-
```
haworku, line 76: Nit: file organization - this function takes parameter from the rates refactor (`ContractType`) whereas everything else in this file is about the domain model HPP and takes HPP related arguments. Should be separate file in domain models.

services/app-api/src/testHelpers/contractAndRates/contractHelpers.ts
```diff
@@ -9,20 +9,30 @@ import type {
 } from '../../postgres/contractAndRates/prismaSubmittedContractHelpers'
 import type { StateCodeType } from 'app-web/src/common-code/healthPlanFormDataType'
 import type { ContractFormDataType } from '../../domain-models/contractAndRates'
+import type { ProgramType } from '../../domain-models'
+import statePrograms from 'app-web/src/common-code/data/statePrograms.json'
+
+function getProgramsFromState(stateCode: StateCodeType): ProgramType[] {
```
haworku, line 15: Good idea for reusable function - I see this code all over. 
 
 Nit: file organization -  this looks like a stateHelper not a contractHelper.

JasonLin0991: This is your reusable function! I just took it and put it in here. I will extract them all out and put them in a `stateHelpers.ts` file.

services/app-api/src/domain-models/healthPlanPackage.ts
JasonLin0991: Agreed, I could see this being in `domain-models/contractAndRates` in some file that contains all the converter functions we will have. I think though I want to punt this to your ticket. I think you have to make converter functions for locked HPP.

PR #1891, opened 2023-08-25
Refactor prisma further

haworku:
## Summary
Two proposed changes to new rates refactor db that have come up via Slack conversations and in person pairing. 

1. Supporting documents fields by using named relation, rather than standalone tables
2. Add optional `display_seq` field.

### DIFF
--- a/services/app-api/prisma/schema.prisma
+++ b/services/app-api/prisma/schema.prisma
@@ -187,6 +187,7 @@ model ActuaryContact {
   rateWithAddtlActuaryID      String?
   rateActuaryCertifying       RateRevisionTable? @relation(name: "CertifyingActuaryOnRateRevision", fields: [rateWithCertifyingActuaryID], references: [id])
   rateActuaryAddtl            RateRevisionTable? @relation(name: "AddtlActuaryOnRateRevision", fields: [rateWithAddtlActuaryID], references: [id])
+  display_seq                 Int?
 }
 
 model ContractDocument {
@@ -200,6 +201,7 @@ model ContractDocument {
   supportingDocumentRevisionID       String
   contractDocumentContractRevision   ContractRevisionTable? @relation(name: "ContractDocumentOnContractRevision", fields: [contractDocumentRevisionID], references: [id])
   supportingDocumentContractRevision ContractRevisionTable? @relation(name: "SupportingDocumentOnContractRevision", fields: [supportingDocumentRevisionID], references: [id])
+  display_seq                        Int?
 }
 
 model RateDocument {
@@ -213,6 +215,7 @@ model RateDocument {
   supportingDocumentRevisionID       String
   ratetDocumentContractRevision      RateRevisionTable? @relation(name: "RateDocumentOnRateRevision", fields: [rateDocumentRevisionID], references: [id])
   supportingDocumentContractRevision RateRevisionTable? @relation(name: "SupportingDocumentOnRateRevision", fields: [supportingDocumentRevisionID], references: [id])
+  display_seq                        Int?
 }
 
 model StateContact {
@@ -224,6 +227,7 @@ model StateContact {
   email              String?
   contractRevisionID String
   contractRevision   ContractRevisionTable @relation(fields: [contractRevisionID], references: [id])
+  display_seq        Int?
 }
 
 model State {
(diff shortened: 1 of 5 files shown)


services/app-api/prisma/schema.prisma
```diff
@@ -175,62 +175,47 @@ model SharedRateCertifications {
...
+  createdAt                          DateTime               @default(now())
+  updatedAt                          DateTime               @default(now()) @updatedAt
+  name                               String
+  s3URL                              String
+  sha256                             String
+  contractDocumentRevisionID         String
+  supportingDocumentRevisionID       String
+  contractDocumentContractRevision   ContractRevisionTable? @relation(name: "ContractDocumentOnContractRevision", fields: [contractDocumentRevisionID], references: [id])
+  supportingDocumentContractRevision ContractRevisionTable? @relation(name: "SupportingDocumentOnContractRevision", fields: [supportingDocumentRevisionID], references: [id])
+  display_seq                        Int?
```
macrael, line 204: Seems a bit dodgy to have this be optional, when would we not want to set it? I think all of these things need to have a visual order every time

services/app-api/prisma/schema.prisma
```diff
@@ -175,62 +175,47 @@ model SharedRateCertifications {
...
-  contractRevision   ContractRevisionTable @relation(fields: [contractRevisionID], references: [id])
+  id                                 String                 @id @default(uuid())
+  createdAt                          DateTime               @default(now())
+  updatedAt                          DateTime               @default(now()) @updatedAt
+  name                               String
+  s3URL                              String
+  sha256                             String
+  contractDocumentRevisionID         String
+  supportingDocumentRevisionID       String
+  contractDocumentContractRevision   ContractRevisionTable? @relation(name: "ContractDocumentOnContractRevision", fields: [contractDocumentRevisionID], references: [id])
```
macrael, line 202: If we're going to do it this way, why not have a single documents table? 

The reason I suggested we have 4 different tables here was that then the primary key could be required in the database, ensuring that we don't accidentally orphan a document at some point. But I don't think that's that big a deal, it would be nice if we could add a db constraint that made sure that one of the four primary keys were set so you can't make a document that doesn't belong to anything.

services/app-api/src/testHelpers/contractAndRates/rateHelpers.ts
```diff
@@ -110,32 +109,38 @@ const createRateRevision = (
     supportingDocuments: [
         {
             id: uuidv4(),
-            rateRevisionID: 'rateRevisionID',
+            rateDocumentRevisionID: 'rateDocRevisionID',
```
macrael, line 112: shouldn't a document only ever have one parent ID?

macrael (approved):
I'm fine with this change. Having a single documents table seems like a win. I put in some notes.

services/app-api/prisma/schema.prisma
haworku: @macrael The reason i didn't set this is we haven't discussed when and how it gets set. I was just adding it in to be used in the future since we discussed it in slack.  Is it set in handlers I'm guessing - so on create or update prisma actions?  And then we start to `orderBy` that in the queries? Let me know if that's right --  I will implement!

services/app-api/prisma/schema.prisma
haworku: @macrael It could be a single table but in that case we lost a document having to have a relationship with either a contract or a rate revision would not be a requirement. Right now because they are separate a RateDocument has a required relationship to its rate revision, similar with contract docs and contract. Do we want to lose that granularity?  I believe the revisionID relationship has to then be optional if its all one

services/app-api/src/testHelpers/contractAndRates/rateHelpers.ts
haworku: This is an artificial of having the named relationship. Here are the [docs](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#disambiguating-relations). What do you think of this? Let me know your thoughts.

haworku:
Closing this branch since I have changes in flights that will rework these migrations based on feedback - need to start a fresh branch.

PR #1903, opened 2023-09-01
MCR-2359 submit + unlock resolver

haworku:
## Summary

This PR includes getting submit and unlock to work. 

I'd like to get all the tests passing so if you see some that aren't running against the new feature flag lmk and I'll enable them. We should be able to run all our API tests now that all the resolvers have been updated. 

There's some noisy changes unfortunatley. Making sha256 required touched a lot of files, as well as getting rid of some of our .d.ts files b/c I hoped it would fix our build issue (and in fact revealed some weird typing issues that were hidden!)

Things left to do:
- [ ] get tests passing 

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-2359
https://qmacbis.atlassian.net/browse/MCR-2360


services/app-api/src/postgres/contractAndRates/submitContract.ts
```diff
@@ -66,16 +64,16 @@ async function submitContract(
                 data: {
                     submitInfo: {
                         create: {
-                            updatedAt: groupTime,
+                            updatedAt: currentDateTime,
```
mojotalantikite, line 67: Is the motivation here that we want to have the `updatedAt` at precisely the same timestamp across the transaction? I feel like even if we're inside a transaction the timestamp from just using the Prisma built in [updatedAt](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#updatedat) would only differ by milliseconds, and we might actually want to know when it actually persisted in the db. Display to users would still be the same date when made human readable.

mojotalantikite (approved):

services/app-api/src/postgres/contractAndRates/submitContract.ts
haworku: I think this was my change and I think we should revert it.  My change  wasn't based on any strong opinion or reason other than I was thinking they are all getting updated at the same tune --   I think we should go back to the db time if there's any reason we would use it which sounds like there is or could be @mojotalantikite

services/app-api/src/postgres/contractAndRates/prismaSubmittedRateHelpers.ts
```diff
@@ -5,6 +5,18 @@ import {
     includeRateFormData,
 } from './prismaSharedContractRateHelpers'
 
+const includeFirstSubmittedRateRev = {
```
JasonLin0991, line 8: nit: Should this be `includeLatestSubmittedRateRev`? From the first look I thought it was getting the first ever submitted revision, but looks like its looking at the most recent created revision that is submitted.

services/app-api/src/postgres/contractAndRates/prismaSubmittedContractHelpers.ts
```diff
@@ -7,6 +7,18 @@ import {
 
 // Generated Types
 
+const includeFirstSubmittedContractRev = {
```
JasonLin0991, line 10: Same as rate. Should this be `includeLatestSubmittedContractRev`

services/app-api/src/postgres/contractAndRates/updateDraftContractWithRates.ts
```diff
@@ -123,6 +142,14 @@ async function updateDraftContractWithRates(
 
     try {
         return await client.$transaction(async (tx) => {
+            const foundBeforeContract = await findContractWithHistory(
+                tx,
+                contractID
+            )
```
JasonLin0991, line 148: Curious if this is just to make sure the contract exists or was the another purpose? The resolver queries for the contractID before calling this handler.

services/app-api/src/resolvers/healthPlanPackage/unlockHealthPlanPackage.ts
```diff
@@ -63,89 +76,242 @@ export function unlockHealthPlanPackageResolver(
...
-                    code: 'INTERNAL_SERVER_ERROR',
-                    cause: 'PROTO_DECODE_ERROR',
-                },
+            // unlock all the revisions, then unlock the contract, in a transaction.
+            const currentRateRevIDs = contract.revisions[0].rateRevisions.map(
+                (rr) => rr.id
+            )
+            const unlockRatePromises = []
+            for (const rateRevisionID of currentRateRevIDs) {
+                const resPromise = store.unlockRate({
```
JasonLin0991, line 123: Do we want to abort all rate unlocks if one fails, then return an error to client? If so, I don't think this will do it. Each one of the unlocks would have to be in the same transaction.

JasonLin0991 (commented):
First pass! Looking good. Thanks for the refactoring to address null and undefined in the Prisma queries.

services/app-api/src/postgres/contractAndRates/updateDraftContractWithRates.ts
macrael: great catch, this was for some debugging. It's gone now.

services/app-api/src/postgres/contractAndRates/prismaSubmittedRateHelpers.ts
macrael: I like it. more clear

services/app-api/src/resolvers/healthPlanPackage/unlockHealthPlanPackage.ts
macrael: you are quite right. I am going to make a ticket for the cleanup to figure this out. Right now we are only doing transactions inside of postgres functions. We need to decide on a design here. Either we should copy your pattern for update where we have a combined "unlockContractAndRates" function that we call from the handler, or we need to make it so that a transaction can be created in a resolver and passed into multiple postgres functions. I'm not sure how to do the latter yet. 

WE'll also need to clean up Submit in the same manner.

services/app-api/src/postgres/contractAndRates/updateDraftContractWithRates.ts
```diff
@@ -1,18 +1,37 @@
...
 import { includeDraftRates } from './prismaDraftContractHelpers'
 import { rateRevisionToDomainModel } from './prismaSharedContractRateHelpers'
 import type { RateFormEditable } from './updateDraftRate'
 import { isEqualData } from '../../resolvers/healthPlanPackage/contractAndRates/resolverHelpers'
 
+// since prisma needs nulls to indicate "remove this field" instead of "ignore this field"
+// this function translates undefineds into nulls
+function nullify<T>(field: T | undefined): T | null {
+    if (field === undefined) {
+        return null
```
haworku, line 20: We should have a generic db helpers place to put this kind of stuff

services/app-api/src/postgres/contractAndRates/updateDraftContractWithRates.ts
```diff
@@ -215,16 +234,44 @@ async function updateDraftContractWithRates(
...
+                                                    })
+                                                ),
                                         },
                                         addtlActuaryContacts: {
-                                            create: rateRevision.addtlActuaryContacts,
+                                            create:
+                                                rateRevision.addtlActuaryContacts &&
+                                                rateRevision.addtlActuaryContacts.map(
+                                                    (c, idx) => ({
+                                                        position: idx,
```
haworku, line 271: @macrael I think we should add this handling to `insertContract` and `insertRate` also -  feels like that could be a place that introduces inconsistency. Otherwise looks good!

PR #1912, opened 2023-09-07
Migrate existing protos to Prisma tables

mojotalantikite:
## Summary

This goes through and migrates us from the `HealthPlanPackage` with the embedded form data in protobuf to the new Prisma (postgres) layout. After each HPP is migrated we fetch the migrated contract back using `findContractWithHistory()`, which goes through domain model validators using `parseContractWithHistory()`. 

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3211


services/app-api/src/postgres/contractAndRates/proto_to_db_UpdateInfoTable.ts
```diff
@@ -0,0 +1,40 @@
...
+            return error
+        }
+
+        console.info(`DEBUG: ${JSON.stringify(user)}`)
+        console.info(`DEBUG: ${JSON.stringify(formData.updatedAt)}`)
+        const updateInfo = await client.updateInfoTable.create({
+            data: {
+                updatedAt: formData.updatedAt,
+                updatedByID: user.id,
+                updatedReason: 'Migrated from previous system',
```
haworku, line 27: We display the update info in the user facing UI. These update reasons have up until now always been text entered by our users. Is the plan to filter these entries out from display?

services/app-api/src/postgres/contractAndRates/proto_to_db_ContractRevisions.ts
```diff
@@ -0,0 +1,108 @@
...
+            contractExecutionStatus: formData.contractExecutionStatus ?? null,
+            contractDateStart: formData.contractDateStart ?? null,
+            contractDateEnd: formData.contractDateEnd ?? null,
+            managedCareEntities:
+                formData.managedCareEntities as ManagedCareEntity[],
+            federalAuthorities: formData.federalAuthorities,
+            modifiedBenefitsProvided: null,
+            modifiedGeoAreaServed: null,
+            modifiedMedicaidBeneficiaries: null,
+            modifiedRiskSharingStrategy: null,
```
haworku, line 79: Why are these being assigned to null should these be form data fields?

haworku (commented):
I don't see contacts being handled at all. We have state contacts on the contract and then two types of contacts on the rates. Am I missing something? Let me know how to help.

Also a possible gap I'm seeing for testing is how will we check every single field is migrated (e.g. check for completeness). I wonder if we should make a way to specifically test a submitted health plan package (which is complete, has all fields no undefined for required fields) works as expected. Anyone have an idea how we could test for this?

services/app-api/src/postgres/contractAndRates/proto_to_db_Documents.ts
```diff
@@ -0,0 +1,184 @@
...
+
+                    if (!existingRateSupportDoc) {
+                        const rateSupportDocument: RateSupportingDocument = {
+                            id: uuid(),
+                            createdAt: revision.createdAt,
+                            updatedAt: new Date(),
+                            name: supportRateDoc.name,
+                            s3URL: supportRateDoc.s3URL,
+                            sha256: supportRateDoc.sha256 ?? '',
+                            rateRevisionID: rateRevision.id,
```
haworku, line 159: I believe this field is now required.

mojotalantikite: Yeah I think you're right, but it looks like common-code's `healthPlanFormData` types still have it as an optional, for example:

```typescript
type SubmissionDocument = {
    id?: string
    name: string
    s3URL: string
    sha256?: string
    documentCategories: DocumentCategoryType[]
}
```

Is there any danger in updating that to required?

services/app-api/src/postgres/contractAndRates/proto_to_db_ContractRevisions.ts
mojotalantikite: That was from Maz, I'm not sure the motivation there, but I didn't see those fields on the `HealthPlanFormDataType` which is being used to populate them from the proto. Should those be in there or can I get the data from somewhere else to fill it in?

services/app-api/src/postgres/contractAndRates/proto_to_db_UpdateInfoTable.ts
mojotalantikite: Good call out -- I think it's something we probably want to decide on as a team. Do we want to have a default like this? Do we want to filter it out on the FE? I don't have a strong opinion here.

mojotalantikite:
> Also a possible gap I'm seeing for testing is how will we check every single field is migrated (e.g. check for completeness). I wonder if we should make a way to specifically test a submitted health plan package (which is complete, has all fields no undefined for required fields) works as expected. Anyone have an idea how we could test for this?

Yeah, I was thinking about this a bit. Do we want to do some sort of field by field comparison, sort of like you would on a test with `expect()`, to validate everything migrated as expected? After talking to MacRae I had just gone ahead and did a `findContractWithHistory()` at the end of each migrate so that we can read it through the zod validators. I don't know how detailed we want to get, but I'd lean towards extra paranoia rather than less for something like this. Happy to hear everyone's comments on it here.

JasonLin0991:
>  I don't know how detailed we want to get, but I'd lean towards extra paranoia rather than less for something like this. Happy to hear everyone's comments on it here.

If we use `findContractWithHistory()` to pull the data out then use `convertContractToUnlockedHealthPlanPackage` we can then compare it using expect().toEqual() for a deep equality check. This should at least verify that we did not lose any data when migrating into the DB.

services/app-api/src/postgres/contractAndRates/proto_to_db_RateRevisions.ts
```diff
@@ -0,0 +1,144 @@
...
+        if (formData.status === 'SUBMITTED' && revision.submittedBy) {
+            const user = await client.user.findFirst({
+                where: { email: revision.submittedBy },
+            })
+
+            if (user) {
+                const existingUpdateInfo =
+                    await client.updateInfoTable.findFirst({
+                        where: { updatedByID: user.id },
+                    })
```
JasonLin0991, line 37: Same as `migrateContractRevision`.

services/app-api/src/postgres/contractAndRates/proto_to_db_RateRevisions.ts
```diff
@@ -0,0 +1,144 @@
...
+                actuaryCommunicationPreference:
+                    rateInfo.actuaryCommunicationPreference ?? null,
+                rateType: rateInfo.rateType ?? null,
+                rateCapitationType: rateInfo.rateCapitationType ?? null,
+                rateDateStart: rateInfo.rateDateStart ?? null,
+                rateDateEnd: rateInfo.rateDateEnd ?? null,
+                rateDateCertified: rateInfo.rateDateCertified ?? null,
+                rateProgramIDs: rateInfo.rateProgramIDs ?? [],
+                rateCertificationName: rateInfo.rateCertificationName ?? null,
+            }
```
JasonLin0991, line 122: This is missing `addtlActuaryContacts`. For this value, we need to get it from `formData.addtlActuaryContacts`. The `rateInfo` in `formData.rateInfos` does not have `addtlActuaryContacts` at all, but our new model does, so we just need to take it from the top level `formData`.

services/app-api/src/postgres/contractAndRates/proto_to_db_ContractRevisions.ts
JasonLin0991: These values should be in `formData.ContractAmendmentInfo.modifiedProvisions`.

services/app-api/src/postgres/contractAndRates/proto_to_db_Documents.ts
JasonLin0991: > Is there any danger in updating that to required?
I don't think so since we migrated existing docs to have the sha and any new documents will generate a sha on upload. The only thing that worries me is if `calculateSHA256(file)` fails, we just set `sha256` to a empty string. Also possibility of a lot of fixing the code for the type change.

When I'm converting HPP rate type to our new domain rate type in `convertHealthPlanPackageRatesToDomain`, I'm generating a sha if there isn't one. This doesn't solve the empty string possibility though.

services/app-api/src/postgres/contractAndRates/proto_to_db_RateRevisions.ts
JasonLin0991: There is also shared rates work that has not been done yet. So we should remember to address that once [MR-3568](https://qmacbis.atlassian.net/browse/MR-3568) is finished.

services/app-api/src/postgres/contractAndRates/proto_to_db_JoinTable.ts
```diff
@@ -0,0 +1,54 @@
+import type { PrismaClient } from '@prisma/client'
+
+async function migrateAssociations(
+    client: PrismaClient
+): Promise<void | Error> {
+    try {
+        // Retrieve all submitted ContractRevisions.
```
haworku, line 7: From RSP: Look at where migrating relationships for drafts is happening if at all

services/app-api/src/postgres/contractAndRates/proto_to_db_Documents.ts
mojotalantikite: Opened a ticket for this as follow on work: https://qmacbis.atlassian.net/browse/MCR-3527

services/app-api/src/postgres/contractAndRates/proto_to_db_RateRevisions.ts
```diff
@@ -0,0 +1,179 @@
...
+        rateRevisionData.push(dataToCopy)
+    }
+
+    const insertRateData: Prisma.RateTableCreateInput = {
+        state: {
+            connect: {
+                stateCode: stateCode,
+            },
+        },
+        stateNumber: stateNumber,
```
JasonLin0991, line 163: This will have to be a new number. Rates have their own state number count in the `State` table `latestStateRateCertNumber`. We should add code to increment `latestStateRateCertNumber` every time a new rate is inserted or have a total to update at once if we are inside a transaction.

JasonLin0991 (commented):
@macrael Brought up stateNumbers and I totally missed this one.

macrael (approved):

PR #1910, opened 2023-09-06
[MR-3339] update content on the landing page

pearl-truss:
## Summary

This PR updates the text copy on the landing page. It also adds a link to the help page that should
 1) open in the current tab 
2) link to the documents description table
3) and is accessible to unauthenticated users

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3339

#### Screenshots
Screenshot 2023-09-07 at 2 43 47 PM

PR #1913, opened 2023-09-08
[MR-3339] ensure that logged in and non logged in users can access the help page

pearl-truss:
## Summary
This PR fixes an issue that was introduced in https://github.com/Enterprise-CMCS/managed-care-review/pull/1910 by ensuring that logged in users can access the help page

[Relevant slack thread](https://cmsgov.slack.com/archives/C014CRU197U/p1694207335614259)

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3339

## QA guidance

As a state user, make sure the links to the help page are accessible

### DIFF
--- a/services/app-web/src/pages/App/AppRoutes.tsx
+++ b/services/app-web/src/pages/App/AppRoutes.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React, { Fragment, useEffect, useState } from 'react'
 import { useNavigate, useLocation } from 'react-router'
 import { Route, Routes } from 'react-router-dom'
 import { useLDClient } from 'launchdarkly-react-client-sdk'
@@ -48,6 +48,14 @@ function componentForAuthMode(
             assertNever(authMode)
     }
 }
+
+// Routes that are agnostic to user login and authentication
+// Should be available on all defined routes lists
+const UniversalRoutes = (
+    <Fragment>
+        <Route path={RoutesRecord.HELP} element={<Help />} />
+    </Fragment>
+)
 
 const StateUserRoutes = ({
     authMode,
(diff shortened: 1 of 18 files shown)


services/app-web/src/pages/App/AppRoutes.tsx
```diff
@@ -69,6 +69,8 @@ const StateUserRoutes = ({
                     element={<StateDashboard />}
                 />
 
+                <Route path={RoutesRecord.HELP} element={<Help />} />
```
macrael, line 72: Looks like we didn't have this route for CMS users as well, can you add it there? Or is there somewhere we can put it in this file that will apply everywhere rather than being repeated in all 3 log in states?

macrael (approved):
Nice and easy

services/app-web/src/pages/App/AppRoutes.tsx
haworku: Im looking at this yes I think we need a constant here for the routes that are user type agnostic. Also I think we should get some tests around routes.
 I'll push up something before I leave today for your review @macrael

macrael (approved):
looks great, do we not have any other routes that cut across all users? I'm a little suprised we are repeating the help route in our 3 different user blocks

haworku:
@macrael Right, whoops I didn't make that change got distracted by figuring out tests. Just pushed another commit. To answer your question no we don't have another place where we directly share routes we explicitly map out the difference between cms and state user routes. Its a tradeoff here with being succinct versus being clear and explicit. It gets pretty confusing with routing sometimes if everything is a constant. I agree that that in this case though (a route which every app visitor can see, logged in or logged out) there should be a contant though. this is out first case of that.

macrael:
sounds good

PR #1923, opened 2023-09-14
Mr 3576 add required and optional labels to fields

pearl-truss:


haworku:
@pearl-truss Is it possible to just use the input's aria-required to determine the label? I'm wondering if the `showRequiredOptionalLabel` could be redundant and if we could use the html input required field attribute (or in the this case `aria-required` to figure out once if the input is required --- then display text accordingly?

pearl-truss:
> @pearl-truss Is it possible to just use the input's required attribute to determine the label? I'm wondering if the `showRequiredOptionalLabel` could be redundant - what are the use cases where we don't want to show the label?. I thought we were just moving towards labeling each field input in general across the forms in the application.

Maybe I'm mistaken (I'll clarify with design) but what I saw on [figma](https://www.figma.com/file/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?type=design&node-id=3686-104662&mode=design&t=eY7cDMTBCjFzDC5Q-0) showed the labels just on the contract and rate pages. And I'm agreed that if it is for every label than `showRequiredOptionalLabel` would be redundant

haworku:
Looks good I found one set of fields on rate details that doesn't have this handling yet  - its the components in `PackagesWithSharedRates`
Screenshot 2023-09-15 at 2 29 12 PM

pearl-truss:
> 🛑 I found one set of fields on rate details that doesn't have this handling yet - its the components in `PackagesWithSharedRates` theres a radio and then a thing called PackageSelect rendered Screenshot 2023-09-15 at 2 29 12 PM
> 
> Otherwise looks v good!

great catch! I just pushed an update

haworku (approved):

PR #1942, opened 2023-09-25
[MCR-2625] Limit State Users to One File Upload for Rate Certs

pearl-truss:
#### Related issues
https://qmacbis.atlassian.net/browse/MCR-2625

#### To Do

- [x] Copy change for error message on Rate details page

#### Screenshots
Screenshot 2023-09-27 at 11 03 07 AM

Screenshot 2023-09-27 at 11 13 29 AM

#### Test cases covered

- `FileUpload` only allows 1 upload when its `multiple` prop is set to false
- Rate details page displays an error if the submission was unlocked and had multiple files uploaded
- Submit Review page displays an error if there are multiple rate cert files uploaded

## QA guidance

With the current code you can test locally that only one rate cert file can be uploaded. If you try to upload a second file after the first was added you should see that the first file was removed and replaced with the most recent addition 

To test that error messages appear on the Review Submit page and the Rate Details page (for an unlocked submission)

- Set `allowMultipleUploads` to true in `SingleRateCert.ts`
- As a state user add multiple files on the Rate details page for the rate cert field
- Set `allowMultipleUploads` back to false in `SingleRateCert.ts`
- On the summary page ensure that the previously uploaded files still appear
- Sign in as a CMS user and unlock the submission
- Sign as a state user and see the summary page, ensure the "Only one document allowed" error message appears above the file list
- Click edit to go the the unlocked Rate details page and ensure that if you click Continue without making any changes you see an "Only one document.." error message

### DIFF
--- a/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
+++ b/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
@@ -368,6 +368,7 @@ export const RateDetailsSummarySection = ({
                                         documentDateLookupTable={
                                             documentDateLookupTable
                                         }
+                                        multipleDocumentsAllowed={false}
                                         caption="Rate certification"
                                         documentCategory="Rate certification"
                                         isEditing={isEditing}
--- a/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
+++ b/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
@@ -372,6 +372,7 @@ export const RateDetailsSummarySection = ({
                                         caption="Rate certification"
                                         documentCategory="Rate certification"
                                         isEditing={isEditing}
+                                        isSubmitted={isSubmitted}
                                     />
                                 ) : (
                                     <span className="srOnly">'LOADING...'</span>
--- a/services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.test.tsx
+++ b/services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.test.tsx
@@ -244,7 +244,7 @@ describe('RateDetails', () => {
             continueButton.click()
             await waitFor(() => {
                 expect(
-                    screen.getAllByText('You must upload at least one document')
+                    screen.getAllByText('You must upload a rate certification')
                 ).toHaveLength(2)
                 expect(continueButton).toHaveAttribute('aria-disabled', 'true')
             })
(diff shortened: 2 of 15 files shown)


services/app-web/src/pages/StateSubmission/RateDetails/SingleRateCert/SingleRateCert.tsx
```diff
@@ -164,6 +164,7 @@ export const SingleRateCert = ({
                     label="Upload one rate certification document"
                     renderMode="list"
                     aria-required
+                    allowMultipleUploads={false}
                     error={showFieldErrors('rateDocuments')}
```
JasonLin0991, line 168: I'm getting the error `You must upload a rate certification` when viewing a unlocked submission with multiple rate cert files.

Reproduction steps:
- deploy app using `main` branch.
- create a submission with multiple rate cert files and submit.
- shut down the app, except for the Postgres instance.
- switch to this branch and only deploy all but Postgres using this command `./dev local --api --web --s3 --otel`
- you may have to manually start the Postgres instance in docker.
- unlock the submission we made using `main`. (This simulates unlocking an existing submission)
- as a state user view the rate details page and hit continue. 
- see error message not display correct text.

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.test.tsx
```diff
@@ -1266,7 +1314,37 @@ describe('RateDetails', () => {
...
+            const continueButton = screen.getByRole('button', {
+                name: 'Continue',
+            })
+            expect(continueButton).not.toHaveAttribute('aria-disabled')
+
+            continueButton.click()
+
+            await waitFor(() => {
+                expect(
+                    screen.getAllByText('You must upload a rate certification')
```
JasonLin0991, line 1347: I think this should be:

```suggestion
                    screen.getAllByText('Only one document is allowed for a rate certification. You must remove documents before continuing.')
```

services/app-web/src/testHelpers/apolloMocks/healthPlanFormDataMock.ts
```diff
@@ -307,6 +307,118 @@ function mockStateSubmission(): LockedHealthPlanFormDataType {
     }
 }
 
+function mockContractAndRatesDraftWithDocuments(
```
JasonLin0991, line 310: I don't think we need this since we can pass in partial package data into `mockContractAndRatesDraft`.

JasonLin0991 (changes requested):
Found some things and a 🪳

This ticket, on the surface, seemed easy, but after taking a look, it is really complex! Great job so far with this one.

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/RateDetailsSummarySection.tsx
```diff
@@ -368,6 +368,7 @@ export const RateDetailsSummarySection = ({
                                         documentDateLookupTable={
                                             documentDateLookupTable
                                         }
+                                        multipleDocumentsAllowed={false}
```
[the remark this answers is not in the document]
pearl-truss: thanks for catching this!

JasonLin0991 (approved):
Thanks for the changes. Looks great!

PR #1945, opened 2023-09-27
MR-2364: Dealt with packages with shared rates in the current UI

JasonLin0991:
## Summary
[MR-2364](https://qmacbis.atlassian.net/browse/MCR-2364)

- Changed the Prisma schema of `packagesWithSharedRateCerts` to `contractsWithSharedRateRevision`.
   - The field is now a many-to-many between `ContractTable` and `RateRevisionTable`.
   - Removed `SharedRateCertifications` join table.
   - Other changes are just formatting.
- API Work
   - Take a peek at `includeDraftRates` and `subincludeUpdateInfo` in the prisma helpers. Added `contractsWithSharedRateRevision` that includes only the latest contract revision.
      - The idea is we take the latest revision by `createdAt` with the assumption that we don't care if its draft or submitted.
      - Is that correct?
   - `updateDraftContractWithRates` just get an array of the contract ids and `connect` when creating a rate and `set` when updating. using set here because we are only updating the relationship, no other complexity here unlike updating the rate revisions.
   - `rateFormDataToDomainModel` function now converts `contractsWithSharedRateRevision` from our DB to the domain model `packagesWithSharedRateCerts`.
      - It loops through `contractsWithSharedRateRevision` and generates the `packageName` for each item.
      - It will throw an error if even one of the packages fails to generate a name.
   - `convertHealthPlanPackageRatesToDomain` will filter out any `packagesWithSharedRateCerts` that is missing a `packageId`. I don't think it will ever happen, but it's here to satisfy the type and in case we overlooked anything.
   - Removed `getProgramsFromState` used in tests. This is basically duplicate of `findStatePrograms` but instead of returning an error, it will return an empty string. Not good for tests, because we should not expect to find no programs for a valid state.
- Frontend Work
   - `PackagesWithSharedRates.tsx` now uses the package level id instead of the form data level. 
      - This does not affect how old-world shared rate certs work.
      - This does fix our new world where the from data level id is the revision id and not the contract id.
   - Several page components that use `useParams` was causing an eslint error. Updated the types to get rid of the error.
   - `packageName` has been refactored to take in arguments needed, instead of the entire package, for generating a name. Now, this can be used with new DB model contracts to generate the same name. 
   - `generateRateName` Move defaulting to package programs into this function instead of `packageName`.


services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
```diff
@@ -60,6 +63,16 @@ const includeRateFormData = {
     supportingDocuments: true,
     certifyingActuaryContacts: true,
     addtlActuaryContacts: true,
+    contractsWithSharedRateRevision: {
+        include: {
+            revisions: {
+                take: 1,
```
haworku, line 69: This looks fine to me based on the prisma table -  to answer your question from the PR description. I think returning just latest rev for the linked contract makes sense.

 I think it _could_ be possible then we could display too early  the updated programs from a rate that is unlocked and being edited by the state ...but I don't see a way around that since its a contract - rate revision relationship and not a contract revision - rate revision level relationship right now.

haworku (approved):
LGTM - I think its a clever way to keep supporting this feature- watch out for interactions with #1903

macrael (approved):
I don't love how much extra error handling we had to introduce here b/c we had to look up the package names. that will be nice to get into the db and no longer have to check in app space. 

Nice work!

PR #1947, opened 2023-09-29
MR-1816: Add and update app permissions for mc review business owner

JasonLin0991:
## Summary
[MCR-1816](https://qmacbis.atlassian.net/browse/MCR-1816)

- Add `BUSINESSOWNER_USER` type to the app that maps to the idm role `macmcrrs-bo`
- Set the local login user cards to be width of 200px


services/app-api/prisma/migrations/20230929194337_add_business_owner_user_role/migration.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+-- AlterEnum
+ALTER TYPE "Role" ADD VALUE 'BUSINESSOWNER_USER';
+
+-- AlterTable
+ALTER TABLE "UserAudit" ALTER COLUMN "updatedAt" DROP DEFAULT;
```
JasonLin0991, line 6: @macrael I coped this line from your merged [PR](https://github.com/Enterprise-CMCS/managed-care-review/pull/1791) that added the helpdesk user. I'm guessing this is still needed here?

services/app-graphql/src/schema.graphql
```diff
@@ -462,6 +462,16 @@ type HelpdeskUser {
     familyName: String!
 }
 
+"BusinessOwnerUser is a user that ..."
```
JasonLin0991, line 465: Halp! What exactly do we want to be the description here?

JasonLin0991 (commented):
Couple of things I had questions on.

services/app-graphql/src/schema.graphql
haworku, line 465:
```suggestion
"BusinessOwnerUser is a CMS user that manages CMS and program level settings in the app."
```

services/app-web/src/localAuth/LocalLogin.tsx
```diff
@@ -90,6 +99,7 @@ const userAvatars: { [key: string]: string } = {
     'roku@example.com': rokuAvatar,
     'izumi@example.com': izumiAvatar,
     'appa@example.com': appaAvatar,
+    'shi-tong@example.com': shiAvatar,
```
haworku, line 102: approved 👍

haworku (approved):
Noticing that there seems to be a group of user roles that have similar permissions -   e.g, they can access settings panel (logic like `isAdminUser || isHelpdeskUser || isBusinessOwnerUser`). Can we encode that in domain model helper `canViewSettings` or something like that.

services/app-api/prisma/migrations/20230929194337_add_business_owner_user_role/migration.sql
macrael: That default should already be dropped so I don't think it makes sense to drop it again.

services/app-graphql/src/schema.graphql
macrael: The Business Owner is the root of our role hierarchy with the power to approve new admin accounts

JasonLin0991:
> Noticing that there seems to be a group of user roles that have similar permissions - e.g, they can access settings panel (logic like `isAdminUser || isHelpdeskUser || isBusinessOwnerUser`). Can we encode that in domain model helper `canViewSettings` or something like that.

How about this?

```ts
function hasAdminPermissions(user: UserType): boolean {
    const authorizedAdmins = [
        'ADMIN_USER',
        'HELPDESK_USER',
        'BUSINESSOWNER_USER',
    ]

    return authorizedAdmins.includes(user.role)
}
```

PR #1954, opened 2023-10-03
Fix additional actuary duplication and add tests.

JasonLin0991:
## Summary
Fixes additional actuaries getting copied multiple times based on the number of rates on a contract when converting our new domain contract data to HPP contract data.

Using `Set.add()` does not check for the equality of objects when adding them to the array. This caused every additional actuary contract from every rate to be copied to the `contract.addtlActuaryContacts` then when saving to the db, that `contract.addtlActuaryContacts` is then set to every `rate.addtlActuaryContacts`. Run this circle a few times, and it will duplicate the contacts exponentially. 

#### Test cases covered

`updateHealthPlanFormData.test.ts`
- `'creates, updates, and deletes rates in the contract'` - Update to test additional actuaries.


macrael (approved):
BOOOO I wish we could use a set for this

JasonLin0991:
> BOOOO I wish we could use a set for this

Maybe soon https://bugzilla.mozilla.org/show_bug.cgi?id=1658309

PR #1958, opened 2023-10-05
MCR-2583 Rates Dashboard epic - includes  indexRates and RateSummary 

haworku:
## Summary
Ended up combing my tickets into one PR to avoid conflicting with rates refactor. I would like to land this PR now once QA'ed, then do small follow on PRs for items in progress. 

**Notes:** 
- I couldn't link to the related contract without requerying the DB (I only had contract revision ids, no contract ids in the rate). Decided to just display the rate summary instead behind that url link in the rates reviews dashboard
- I started to remove some supporting docs code when I touched the `UploadedDocumentsTable`
    -  specifically this PR removes the asterik for marking supporting docs are both contract and rate related  (which is I believe unused)
- Still need to refactor to share code between `RateDetailsSummarySection` <> `RateSummary` as well as between `CMSDashboard` <> `RateReviewsDashboard` in the next sprints. This will be follow on work. 
- Still need to addf frontend tests for new UI (both Cypress and jest). This will be follow on work. 
- Still need to add backend tests for fetchRate. This will be follow on work. 

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-2578
https://qmacbis.atlassian.net/browse/MCR-2557
https://qmacbis.atlassian.net/browse/MCR-2558

#### Screenshots
Rates Summary page without related "Contract Actions" and without Change History.
Screenshot 2023-10-10 at 4 38 53 PM

#### Test cases covered
Backend tests covered
- Unit tests for `findAllRatesWithSubmitInfo`
- Integration tests for `indexRates` - started out by testing similar cases as `indexHealthPlanPackages`
- Fix for `unlockHealthPlanPackage`  - now we use the same test assertion regardless of feature flag state

## QA guidance
- check interaction between contracts and rates by creating contract and rate submissions (including multi-rate subs) and making sure the rates show up as expected in dashboard and summary


haworku:
If anyone peeks at this code I found some errors to fix with the nested resources (contacts and documents) and how those return through graphql. Working on that now.

haworku:
@pearl-truss Also adding you as reviewer since I know you are working on an API also :)

JasonLin0991 (changes requested):
First pass! Seems to work on local deployment.

services/app-web/src/pages/CMSDashboard/RateReviewsTable.tsx
```diff
@@ -0,0 +1,478 @@
...
+        .filter((item) => item.id === id)
+        .map((item) => ({ value: item.value, label: item.value }))
+    return filterValues as FilterOptionType[]
+}
+
+type TableVariantConfig = {
+    tableName: string
+    rowIDName: string
+}
+export const RateReviewsTable = ({
```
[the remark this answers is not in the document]
JasonLin0991: That sounds good to me.

services/app-web/src/pages/CMSDashboard/CMSDashboard.tsx
```diff
@@ -214,18 +217,96 @@ const SubmissionsDashboard = (): React.ReactElement => {
 }
 const RateReviewsDashboard = (): React.ReactElement => {
     const { loggedInUser } = useAuth()
-    if (!loggedInUser) {
+    const { data, loading, error } = useIndexRatesQuery({
```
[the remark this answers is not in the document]
haworku: Sure, good idea. I'll look at this in the next PR. I would like to try to land this to enable others to build off both the dashboard and the rate summary work next sprint

JasonLin0991 (approved):

services/app-graphql/src/schema.graphql
```diff
@@ -689,3 +711,144 @@ type FetchEmailSettingsPayload {
...
+    stateCode: String!
+    "Fuller state data for the submitting state"
+    state: State!
+    stateNumber: Int!
+    """
+    Array of revisions for this rate. Each revision represents a single submission
+    for the rate and contains the full data from when the rate cert was submitted
+    """
+    revisions: [RateRevision!]!
+    contractRevisions:[RelatedContractRevisions!]!
```
macrael, line 836: I'm not sure what contractRevisions means for a Rate. We have that linkage in the DB but only because of the contractRevision.draftRates linkage. I don't know what the front end will use Rate.contractRevisions for

services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
```diff
@@ -137,13 +148,41 @@ const columnHash = atomWithHash('filters', [] as ColumnFiltersState, {
...
 })
 
+/* transform react-table's ColumnFilterState (stringified, formatted, and stored in the URL) to react-select's FilterOptionType
+    and return only the items matching the FilterSelect component that's calling the function*/
+const getSelectedFiltersFromUrl = (
+    columnFilters: ColumnFiltersState,
+    id: string
+) => {
+    type TempRecord = { value: string; label: string; id: string }
+    const valuesFromUrl = [] as TempRecord[]
```
macrael, line 158: can you not write `const valuesFromUrl: TempRecord[] = []` and avoid the "as"?

macrael: Oh I didn't realize this was just getting moved from somewhere else. Don't sweat it

services/app-web/src/pages/CMSDashboard/CMSDashboard.tsx
```diff
@@ -214,18 +217,96 @@ const SubmissionsDashboard = (): React.ReactElement => {
...
+                    previousRevision?.unlockInfo?.updatedAt,
+                ])
+            }
+
+            // Type guards - graphql has loose types with form data, let's narrow in now
+            if (
+                !displayRateFormData ||
+                !displayRateFormData.rateProgramIDs ||
+                !displayRateFormData.rateType ||
+                !displayRateFormData.rateDateEnd ||
```
macrael, line 256: once we have a rate dashboard for states it should be possible to display rates that are incomplete

services/app-web/src/pages/CMSDashboard/RateReviewsTable.tsx
```diff
@@ -0,0 +1,478 @@
...
+}
+
+const columnHash = atomWithHash('filters', [] as ColumnFiltersState, {
+    serialize: fromColumnFiltersToReadableUrl,
+    deserialize: fromReadableUrlToColumnFilters,
+})
+
+/* transform react-table's ColumnFilterState (stringified, formatted, and stored in the URL) to react-select's FilterOptionType
+    and return only the items matching the FilterSelect component that's calling the function*/
+const getSelectedFiltersFromUrl = (
```
macrael, line 114: is this duplicated from above?

macrael (approved):
Looks great. Since responding to these comments now is going to delay us being able to announce this feature I think they can be addressed in a future PR

services/app-graphql/src/schema.graphql
```diff
@@ -318,7 +340,7 @@ input CreateQuestionInput {
     "The ID of the package for which to create a question"
     pkgID: ID!
     "A list of documents to attach to the question"
-    documents: [DocumentInput!]!
+    documents: [DocumentInput!]! # should be Generic Document with sha256
```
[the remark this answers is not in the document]
haworku: Yes I agree.  was thinking  everywhere in graphql could use my new `GenericDocument` eventually. @macrael Does that sound fine?

  I did start to just make change monday, adding sha2556 to Q&A stuff, but quickly ran into types issues because there were more places using that type in small utility functions and other places than I anticipated. I backtracked then and just left the comment.

services/app-web/src/pages/CMSDashboard/CMSDashboard.tsx
haworku: @macrael Are you saying you want our dashboard to display `???` or something like that if the field is missing? We don't have that built in right now - we assume all fields listed here exist on dashboard these are required types for a submitted package. Since we dont have a locked type I believe we need to manually check for missing fields here.

macrael: Maybe this won't play out that way but it depends on how we end up creating rates. But I can imagine it being possible to create a rate and not finish filling it out and expect it to show up on the index page

haworku: This already threw an error so anyway shouldn't do the full page takeover. Will stick with logging and check this out more next week.  Have no idea what data was missing since I thought all those fields would be there. Moving to fallback text

services/app-api/src/postgres/contractAndRates/findAllRatesWithHistoryBySubmitInfo.ts
```diff
@@ -0,0 +1,49 @@
...
+                    not: 'AS', // exclude test state as per ADR 019
+                },
+            },
+            include: {
+                ...includeFullRate,
+                state: true,
+            },
+        })
+
+        const parsedRatesOrErrors: RateOrErrorArrayType = rates.map((rate) => ({
```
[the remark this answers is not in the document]
haworku: The type here is not `Rate | Errors` it is

```
{
    rateID: string
    rate: RateType | Error
}[]
```
and the reason its returned (I grabbed this from findAllContracts code) makes sense to me - looks like we want to return the keys for all possible records so that we can clearly send back of list of which records errors and which didn't for display in the table. 

@macrael is there a change you want here with the types or the approach - I'm happy to make it in both places.

PR #1957, opened 2023-10-05
MR-2369: turn on rates db refactor flag for cypress

JasonLin0991:
## Summary
[MR-2369](https://qmacbis.atlassian.net/browse/MCR-2369)

- Turns on `rates-db-refactor` flag for Cypress tests
- Turns on `supporting-docs-by-rate` flag for Cypress tests
   - Since `supporting-docs-by-rate` is dependent on `rates-db-refactor`, the UI changes behind it were never activated in `dev`. By activating `rates-db-refactor` in dev we now see the UI in dev Cypress tests.
   - All tests have been updated for the UI changes.
- Skip `questionsResponse.spec` tests because we have not migrated the Q&A API endpoint yet.
- **NOTE** Currently, the change history is incorrect and our cypress tests does not accurately test the change history and its order.
   - This test is in `unlockResubmit.spec`
   - It still passes
   - I will have a follow-up ticket to address this once we fix the change history.  See the [rates db refactor notes](https://docs.google.com/document/d/1E1kcXBZYfM6aGjFyrWR7clX9lflHwnfr6Ijxyob_4_M/edit#heading=h.1gehl816ra0k) doc for more info on the investigation of the bug.

PR #1955, opened 2023-10-05
Enable remaining API tests

macrael:
## Summary
Enabling the tests we had not enabled until all the APIs were converted to the rates refactor. 

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-2370

PR #1959, opened 2023-10-05
MR-3545, MR-2370: Fix change history, add cypress tests, and turn on remaining API Unit tests

JasonLin0991:
## Summary
[MR-3545](https://qmacbis.atlassian.net/browse/MCR-3545)
[MR-2370](https://qmacbis.atlassian.net/browse/MCR-2370)

The cause was duplicate contract revisions and Apollo Client Caching. See Apollo Client [storing normalized docs](https://www.apollographql.com/docs/react/caching/overview/#4-store-normalized-objects). The duplicate contract revisions are made in contractWithHistoryToDomainModel when we construct a revision history for each change in rate revisions by creating a new contract revision with the same ID as the contract revision the rate change belongs to. When this data is returned to the frontend, Apollo cache sees duplicate contract revision IDs and merges the data.

See [notes](https://docs.google.com/document/d/1E1kcXBZYfM6aGjFyrWR7clX9lflHwnfr6Ijxyob_4_M/edit) for more details on the investigation.

Changes:
- `contractWithHistoryToDomainModel` removed creating revision history for rate changes.
- Updated `unlockResubmits.spec` to test the order and count of the revision history.
- Updated API unit tests that were testing rate revision history in the contract revisions.
   - Fixing the change history bug also fixed CI cypress test failing.
   - Previous [PR](https://github.com/Enterprise-CMCS/managed-care-review/pull/1957) to fix all these tests missed this one, because I forgot to turn the flag back on to run CI 🫠
- Pulled in all work from this [PR](https://github.com/Enterprise-CMCS/managed-care-review/pull/1955) to turn on remaining API unit tests.


services/app-api/src/postgres/contractAndRates/findContractWithHistory.test.ts
```diff
@@ -124,7 +124,9 @@ describe('findContract', () => {
         if (threeContract instanceof Error) {
             throw threeContract
         }
-        expect(threeContract.revisions).toHaveLength(4)
```
macrael, line 127: I'm very hesitant to lose the specific values in this test. This test was written to check that we get the expected number of revisions everywhere. I'd rather mark this test as skip or something than adapt it to work with the interim solution.

macrael (approved):
I want to be sure we can switch back to the old accounting easily on those long findContractWithHistory tests but otherwise this seems like a good interim fix

services/app-api/src/postgres/contractAndRates/findContractWithHistory.test.ts
JasonLin0991: Let's mark as skip.

PR #1960, opened 2023-10-06
[MCR-2614] backend changes for mccrsid

pearl-truss:
## Summary

This PR creates a new `updateContract` endpoint. This code is separate from the `formData` and is used to update the `mccrsID` field 

#### Related issues

https://qmacbis.atlassian.net/browse/MCR-2614

#### To Do

- [x] Fix the existing test for `updatingHealthPlanFormData`  that's failing (this may be an issue with the db refactor)
- [x] Add additional test for the `updateContract` postgres function
- [x] Error handling in the `updateContractResolver` 

#### Test cases covered

- Updating contracts is restricted to CMS users and the contract must be submitted 

## QA guidance

This can be tested with the graphQL Explorer http://localhost:3000/dev/graphql-explorer
- Add the `updateContract` mutation
```
mutation UpdateContract($input: UpdateContractInput!) {
  updateContract(input: $input) {
    pkg {
      id
      mccrsID
    }
  }
}
```

- Under variables supply an `id` that can be found from the network tab on the review submission page. You should also supply a `mccrsID` 
- You will need to be signed in as a CMS user
- Once run, you should see the `mccrsID` in the response


services/app-api/src/postgres/contractAndRates/updateMCCRSID.ts
```diff
@@ -0,0 +1,50 @@
...
+import type { PrismaClient } from '@prisma/client'
+import type { ContractType } from '../../domain-models/contractAndRates'
+
+type UpdateMCCRSIDArgsType = {
+    contractID: string
+    mccrsID?: string
+}
+
+// Update the MCCRS record number
+async function updateMCCRSID(
```
pearl-truss, line 12: I went back and forth on if this should be `updateContract` or `updateMCCRSID`?

services/app-api/src/resolvers/contractResolvers/updateContractResolver.ts
```diff
@@ -0,0 +1,110 @@
...
+                    extensions: {
+                        code: 'INTERNAL_SERVER_ERROR',
+                        cause: 'PROTO_DECODE_ERROR',
+                    },
+                })
+            }
+            return {
+                pkg: convertedPkg,
+            }
+        } else {
```
pearl-truss, line 88: If the `RatesDatabaseRefactor` flag is off, I originally wanted to return `pkg: null` but that resulted in a type error. So I passed the entire converted package but I'm not sure if it makes more sense to return something else?

services/app-api/src/resolvers/contractResolvers/updateContractResolver.ts
```diff
@@ -0,0 +1,110 @@
...
+                    'user not authorized to update contract'
+                )
+            }
+
+            const contract = await store.findContractWithHistory(input.id)
+            if (contract instanceof Error) {
+                throw contract
+            }
+
+            const hasSubmitInfo =
```
pearl-truss, line 49: It was called out to check that a revision has submitInfo but I was wondering if a status check for `SUBMITTED` would work instead?

services/app-api/src/resolvers/contractResolvers/updateContractResolver.ts
```diff
@@ -0,0 +1,110 @@
...
+                contract.revisions.filter((rev) => rev.submitInfo).length > 0
+
+            if (!hasSubmitInfo) {
+                const errMessage = `Can not update a contract has not been submitted. Fails for contract with ID: ${contract.id}`
+                logError('updateContract', errMessage)
+                setErrorAttributesOnActiveSpan(errMessage, span)
+                throw new UserInputError(errMessage, {
```
pearl-truss, line 56: In my mind this resolver encountering a contract in a non submitted status would be a 409 HTTP conflict error. I'm not sure if `UserInputError` is the correct error to throw in this case?

services/app-api/src/domain-models/contractAndRates/convertContractWithRatesToHPP.ts
```diff
@@ -116,13 +116,7 @@ function convertContractWithRatesToFormData(
...
 
             for (const additionalActuary of addtlActuaryContacts) {
-                if (
-                    !pkgAdditionalCertifyingActuaries.find((actuary) =>
-                        isEqualData(actuary, additionalActuary)
-                    )
-                ) {
-                    pkgAdditionalCertifyingActuaries.push(additionalActuary)
-                }
+                pkgAdditionalCertifyingActuaries.add(additionalActuary)
```
JasonLin0991, line 119: Using `Set.add()` was causing duplicate additional actuaries. It might be good to pull in main again.

services/app-api/src/postgres/contractAndRates/updateMCCRSID.ts
haworku: I think its good to leave this as a function that just updates the MCCRSID like you have it. Right now `updateContract` at least at postgres level updates the contract revision. When we eventually build that into graphql I imagine it would update the contract revision form data for state user and mayybe update

services/app-api/src/resolvers/contractResolvers/updateContractResolver.ts
JasonLin0991: Is this feature only for our new DB model? If so, we should hide the UI and return an error from this endpoint if `ratesDatabaseRefactor` flag is off.

services/app-api/src/postgres/contractAndRates/updateContract.ts
```diff
@@ -0,0 +1,53 @@
...
+async function updateContract(
+    client: PrismaClient,
+    args: UpdateContractNoFormArgsType
+): Promise<ContractType | NotFoundError | Error> {
+    const { contractID, mccrsID } = args
+
+    try {
+        return await client.$transaction(async (tx) => {
+            // Get the Contract associated with this given contract ID
+            const currentContract = await tx.contractTable.findFirst({
```
pearl-truss, line 22: This is returning `null` and is one of the things I'll need to look more into

services/app-api/src/postgres/contractAndRates/updateMCCRSID.ts
pearl-truss: Oh sorry, I posted this question and updated it to `updateContract` before I saw this response. I will change it back

services/app-api/src/resolvers/contractResolvers/updateContractResolver.ts
macrael: Right now we're using UserInputError for basically all 4XX errors. If it's an error where the user made a mistake that's a UserInputError. If it's an unexpected server error to an otherwise valid request, that's an Error()

haworku:
@JasonLin0991 LGTM but I wanted to try updating it on the deployed app but seeing a web test unit failure `› handles rates across submissions › correctly checks shared rate certification radios and selects shared packages` which seems odd 😮‍💨 . Could you take a look at this?

JasonLin0991:
> @JasonLin0991 LGTM but I wanted to try updating it on the deployed app but seeing a web test unit failure `› handles rates across submissions › correctly checks shared rate certification radios and selects shared packages` which seems odd 😮‍💨 . Could you take a look at this?

Working on it as you posted! It should be fixed now. It was just our test data missing `mccrsID`. 

Also `IndexHealthPlanPackages` query included `mccrsID`, which caused creating a new submission to error out. This is because we are manually updating the `IndexHealthPlanPackages` Apollo cache when this new submission is created, but since we do not have a `mccrsID` yet, updating the cache returned an error. The quickest solution was to add `mccrsID` to `createHealthPlanPackage` query schema.

haworku (approved):

PR #1963, opened 2023-10-09
update migrators handling of submittedAt and unlocked rates

macrael:
## Summary

Found some issues when running this migration on dev.

* submittedAt was not being set correctly
* unlockedAt was not getting set at all on an unlocked unsubmitted package
* unlocked packages were not setting draftRates correctly so their rates were coming back empty


mojotalantikite (approved):

macrael:
Merged into Jasons' PR

PR #1964, opened 2023-10-09
Add test for 'undefined' values on migrator

mojotalantikite:
## Summary
When we ran the protobuf migrator I noticed that settings in a 'DRAFT' submission that were previously unset had come back with a value of `false`. We should be preserving these values as unset.


mojotalantikite:
after standup it looks like this wasn't a bug, it was an issue with the test data we were looking at. Closing.

PR #1965, opened 2023-10-10
Reorder migrator to migrate rates after contracts to prevent linking errors

macrael:
## Summary

We encountered an error migrating dev where a rate revision was not able to be created because the contract it linked to did not exist yet. This PR reorganizes the migrator to migrate all the contracts and then all the rates.


mojotalantikite (approved):
LGTM!

PR #1977, opened 2023-10-13
MCR-3557: Cms sees link to submission on rate summary page and shared rate submissions

JasonLin0991:
## Summary
[MCR-3557](https://qmacbis.atlassian.net/browse/MCR-3557)

- `Submission this rate was submitted with` section
   - This will show the submission this rate was submitted with.
- The shared rates submission link is now on the documents section like the current summary pages.
- Prisma query changes
   - `includeContractFormData`
      - `contractID` was not enough to generate a package name, so I'm including the contract data the revision belongs to
      - This is mainly needed when querying rates and including its `contractRevisions`. Since querying contracts we have the data at the top level.
- GraphQL Schema and Domain Type changes
   - Domain Types
      - For `RateRevisionType`, the `contractRevisions` on this type only the `contractID` field; this was not enough data to generate a package name.
      - I could add `packageName` to the `ContractRevisionType` or add the contract data to the `ContractRevisionType` and generate the package name on the front end.
      - I decided to add contract data because it was less refactoring of our parsing functions, which would have involved finding programs and more error handling. On the front end, we have access to the state programs, so package name generation was less complicated without all the parsing functions in the API.
      - `packagesWithSharedRateCerts` now has a new `packageStatus` field to determine if we should display a link to the submission or just the name. Currently, with protobufs we have to make another `indexHealthPlanPackage` query and call functions to parse through it to find out if our shared rate packages are drafts. With the new DB we no longer have to do that.
      - GraphQL schema
         - Added types for contract data.
         - Updated rate queries
         - Added `packageStatus` to `packagesWithSharedRateCerts`

### DIFF
--- a/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/SingleRateSummarySection.tsx
+++ b/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/SingleRateSummarySection.tsx
@@ -12,6 +12,7 @@ import {
     Rate,
     RateFormData,
     RateRevision,
+    RelatedContractRevisions,
 } from '../../../gen/gqlClient'
 import { UploadedDocumentsTable } from '../UploadedDocumentsTable'
 import type { DocumentDateLookupTableType } from '../../../documentHelpers/makeDocumentDateLookupTable'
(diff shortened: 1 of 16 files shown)


services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/SingleRateSummarySection.tsx
```diff
@@ -91,10 +119,9 @@ export const SingleRateSummarySection = ({
     isSubmitted: boolean
     statePrograms: Program[]
 }): React.ReactElement | null => {
-    const formData: RateFormData =
-        rate?.status === 'UNLOCKED'
-            ? rate?.revisions[1].formData
-            : rate?.revisions[0].formData
+    const rateRevision =
+        rate?.status === 'UNLOCKED' ? rate?.revisions[1] : rate?.revisions[0]
```
haworku, line 123: This was causing a bug - forgot that draft revisions don't come down on the rate revisions anymore. This should always be the first revision in the list.

haworku (commented):
I like the choice to add full contract. Can you please add frontend tests for this UI too? Now that we have gotten enough work merged to demo we need to go back to adding tests to ever new feature as we ship.

PR #1991, opened 2023-10-18
MCR-3570 fix: Previous contract submissions contained incorrect rate data

haworku:
## Summary
[MCR-3570](https://qmacbis.atlassian.net/browse/MCR-3570)

**Bug**
 Previous contract submissions would contain incorrect rate data.

**Cause**
`contractWithHistoryToDomainModel` builds the contract revision history for a contract. Each contract revision record from the **database** will include rate revisions after the contract revision has been submitted, but `contractWithHistoryToDomainModel` did not account for these rate revisions. So they would get added to the contract revision history and remove the previous version of that rate revision. This results in the history not displaying the correct rate revision when the contract was submitted.

**Reproduction steps**
- Create and submit a contract and rate submission.
- Unlock submission and edit rate.
- Resubmit the submission.
- View the previous submission; it will have the updates you made.

**Fix**
- The fix was to ignore all rate revisions that were submitted after the contract revision submission.
- There is also a fix to preserve rate order contract revision. This was done by adding some `rate` data to the `RateRevisionType`. Then, in `ratesRevisionsToDomainModel` function, we sort by `rate.createdAt` instead of the revision createdAt.

**Additional Work**
`submitHealthPlanPackage.ts` was using the first revision of the contract history for submission instead of `draftRevision`. Changed to use the `draftRevision`

#### Test cases covered
- fixed a few tests that were not passing the strict rate id checks
-  add `submitHealthPlanPackage` test that checks revision data for rates is correct after unlock, resubmit when editing rates in place and adding new rates
- add `fetchRate` tests that check that revisions data is correct

### DIFF
--- a/docs/technical-design/contract-rate-change-history.md
+++ b/docs/technical-design/contract-rate-change-history.md
@@ -29,9 +29,23 @@ At the Postgres Table level, draft revisions and submitted revisions live in the
 
 The list of revisions returned from prisma is run through [Zod](https://zod.dev/) to return [domain mode types](../../services/app-api/src/domain-models/contractAndRates). This is initiated by the `*WithHistory` database functions. See [parseContractWithHistory](../../services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts) and [parseRateWithHistory](../../services/app-api/src/postgres/contractAndRates/parseRateWithHistory.ts).
 
+#### Contract History
+- `parseContractWithHistory` takes our prisma contract data and parses into our domain `ContractType`. In `ContractType` the `revisions` is an array of **contract** **revisions**; this is the contract history.
+- `revisions` differs from `draftRevision` in the `ContractType`. The `draftRevision` is a singular revision that is not submitted and this data has no historical significance until it is submitted. Most of the data in this revision can be updated.
+- Each **contract revision** in `revisions` is submitted and retains data at the time of the submission. These revision's data will never be updated to retain its historical integrity.
+- An important note about the `rateRevisions` field in each contract revision in `revision`.
+   - Like contracts, rates also have **rate revisions** which are used to construct a rate history through submissions, but the purpose of `rateRevisions` on a contract revision is not for rate history.
+   - The purpose of `rateRevisions` is to retain the data of a rate linked to this contract revision at the time of submission.
+   - For that we need the single rate revision that was submitted at the time this contract revision was submitted.
+   - Here are some guidelines for each rate revision in `rateRevisions` of a contract revision.
+      - Each rate revision in `rateRevisions` is unique by rate id, meaning there will never be two rate revisions with the same rate id in `rateRevisions`
+      - Each rate revision is the latest submitted up till the contract revision submitted time.
+      - Like contract revision, rate revision is read only and cannot be updated to retain its historical integrity.
+
+
 *Dev Note*: If the `draftRevision` field has a value and the `revisions` field is an empty array, we know the Contract or Rate we are looking at is an initial draft that has never been submitted.
 
-## The link between contract and rates is versioned. That link is solidified on submit.
+### The link between contract and rates is versioned. That link is solidified on submit.
 
 It's possible to tell if a link between a contract and rate has become outdated by refencing the `valid After` and `validUntil` fields on the join table between contract and rate revisions. The `validFrom` is set when a link is created (when a contract is submitted with a link to rates). At the point of creation, the `validUntil` is still null.
 
--- a/services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
+++ b/services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
@@ -203,9 +203,7 @@ function contractWithHistoryToDomainModel(
                 )
             }
 
-            // Make sure this rate revision was not submitted after this contract revision, and it was not removed. We
-            // do not want to include updated rate data from after this submission and we don't want to include revisions
-            // marked isRemoval.
+            // Make sure this rate revision was not submitted after this contract revision, and it was not removed.
             if (
                 rateRev.rateRevision.submitInfo.updatedAt <=
                     contractRev.submitInfo.updatedAt &&
@@ -216,12 +214,9 @@ function contractWithHistoryToDomainModel(
                     (rr) => rr.rateID === rateRev.rateRevision.rateID
                 )
 
-                // This conditional is to retain order of the rate entries by createdAt. contractRev.rateRevisions are
-                // queried from the DB in ASC order of createdAt.
                 if (existingIndex >= 0) {
                     // If rate exists, replace the existing rate entry with the current rate of the loop. We only want
-                    // the latest rate revision that was submitted with this contract revision, any earlier rate revisions
-                    // with the same rateID was from an earlier submission.
+                    // the latest rate revision which was submitted with this contract revision and retain order.
                     initialEntry.rateRevisions.splice(
                         existingIndex,
                         1,
--- a/services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
+++ b/services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
@@ -194,7 +194,8 @@ function contractWithHistoryToDomainModel(
         //     }
         // }
 
-        // Basically the same as above, except we do not create new contract revisions for rate changes.
+        // Basically the same as above, except we do not create new contract revisions for rate changes and non-draft
+        //  contract revisions do not inlcude rate revisions that where submitted after the contract revision was submitted.
         for (const rateRev of contractRev.rateRevisions) {
             if (!rateRev.rateRevision.submitInfo) {
                 return new Error(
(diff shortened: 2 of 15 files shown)


services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.test.ts
```diff
@@ -135,6 +139,172 @@ describe.each(flagValueTestParameters)(
...
+            })
+
+            // First, create new submissions
+            const submittedEditedRates =
+                await createAndSubmitTestHealthPlanPackage(server, {
+                    rateInfos: [initialRateInfos()],
+                })
+            const submittedNewRates =
+                await createAndSubmitTestHealthPlanPackage(server, {
+                    rateInfos: [initialRateInfos()],
```
macrael, line 191: This is creating rates in two submissions with the same ID, is that intentional? That's not something we expect to have ever happened in our app

services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.test.ts
```diff
@@ -135,6 +139,172 @@ describe.each(flagValueTestParameters)(
...
+            expect(previousFormData(newRatesPackage).rateInfos).toHaveLength(1)
+            expect(
+                previousFormData(newRatesPackage).rateInfos[0].rateDateStart
+            ).toMatchObject(initialRateInfos().rateDateStart)
+            expect(
+                previousFormData(newRatesPackage).rateInfos[0].rateDateEnd
+            ).toMatchObject(initialRateInfos().rateDateEnd)
+            expect(
+                newRatesPackage.revisions[1].node.submitInfo?.updatedReason
+            ).toBe('Initial submission')
```
macrael, line 305: This is a great test, can we also test that indexRates pulls the right number of rates down after? Proving that we edited one rate and created another?

services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.ts
```diff
@@ -367,9 +367,9 @@ export function submitHealthPlanPackageResolver(
             }
 
             // If there are rates, submit those first
-            if (contractWithHistory.revisions[0].rateRevisions.length > 0) {
+            if (contractWithHistory.draftRevision.rateRevisions.length > 0) {
```
macrael, line 370: I think this wants to be contractWithHistory.draftRevision.draftRates ?

services/app-api/src/testHelpers/gqlHelpers.ts
```diff
@@ -178,6 +178,40 @@ const updateTestHealthPlanFormData = async (
     return updateResult.data.updateHealthPlanFormData.pkg
 }
 
+const updateTestHealthPlanPackage = async (
```
macrael, line 181: I'm confused, this function seems like it already exists on main?

services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
```diff
@@ -194,29 +194,26 @@ function contractWithHistoryToDomainModel(
         //     }
         // }
 
-        // Basically the same as above, except we do not create new contract revisions for rate changes.
+        // Basically the same as above, except we do not create new contract revisions for rate changes and non-draft
+        //  contract revisions do not inlcude rate revisions that where submitted after the contract revision was submitted.
         for (const rateRev of contractRev.rateRevisions) {
```
macrael, line 199: I think this could be simpler. 

What we're saying right now is that we are only going to return the rates that existed when the contract was submitted, and ignore all rates that came afterwards. 

So I think this whole for loop becomes

```
initialEntry.rateRevisons = contractRev.rateRevisions.filter((rr) => rr.rateRevision.submitInfo.updatedAt <=
                contractRev.submitInfo.updatedAt)
```

macrael: and I think we should be making the same change in parseRateWithHistory b/c that's what turns into the rate dashboard.

services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.ts
macrael: NO I think I'm wrong, at this point it's been turned into our domain model which doesn't have draftRates

macrael (approved):
I think it makes sense why this was busted, but we shoudl make the same change for rates.

services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.test.ts
haworku: @macrael Yes! I'm actively working on `indexRates` tests like this in that test file. I think indexRates tests should be much more robust.  It will be a separate PR though.

services/app-web/src/components/DataDetail/DataDetailContactField/DataDetailContactField.tsx
```diff
@@ -5,9 +5,9 @@ import {
 } from '../../../common-code/healthPlanFormDataType'
 import { getActuaryFirm } from '../../SubmissionSummarySection'
 import { DataDetailMissingField } from '../DataDetailMissingField'
+import { ActuaryContact as GQLActuaryContact } from '../../../gen/gqlClient'
 
-type Contact = ActuaryContact | StateContact
-
+type Contact = ActuaryContact | StateContact | GQLActuaryContact
```
haworku, line 10: @macrael this was another issue where the graphql types and common code types don't translate well `Type 'Maybe<string> | undefined' is not assignable to type 'string | undefined'.`  

There are many ways to work around this using null coalescing checks, adding to the types (like I'm doing here), manually asserting its a same type but ideally I wanted to map these two types together using the `codegen.yml.` similar to what we do with Rate Revision. Couldn't get this to actually fixing the issue though. Any suggestions for the future? This kind of types issue conflict will keep happening on the frontend

services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.test.ts
haworku: I don't think that's what is happening this is a function call and its using uuid util - I would think they have different ids

services/app-api/src/testHelpers/gqlHelpers.ts
haworku: @macrael  I made because the function on main  `updateTestHealthPlanFormData` takes in complete form data, proto-izes it and returns it. 

I made a `updateTestHealthPlanPackage` to make an update call that is more similar to the other test gql helpers at this point in our code base -  it takes in partial updates and conditionally applies them but otherwise self generates its own faked data.

services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.test.ts
JasonLin0991: > I don't think that's what is happening this is a function call and its using uuid util - I would think they have different ids

Yes!

services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
JasonLin0991:
```ts
initialEntry.rateRevisons = contractRev.rateRevisions.filter((rr) => rr.rateRevision.submitInfo.updatedAt <=
                contractRev.submitInfo.updatedAt)
```

This wouldn't remove the rates that had been removed. Also, using the array method, we would have to check the `submitInfo` on both revisions for `undefined` again.

I did simplify the loop a bit more.

JasonLin0991: > and I think we should be making the same change in parseRateWithHistory b/c that's what turns into the rate dashboard.

This one is slightly different from contract revisions since the rates are submitted before the contract using the `submitHealthPlanPackage` resolver.

docs/technical-design/contract-rate-change-history.md
```diff
@@ -29,9 +29,23 @@ At the Postgres Table level, draft revisions and submitted revisions live in the
...
+- Each **contract revision** in `revisions` is submitted and retains data at the time of the submission. These revision's data will never be updated to retain its historical integrity.
+- An important note about the `rateRevisions` field in each contract revision in `revision`.
+   - Like contracts, rates also have **rate revisions** which are used to construct a rate history through submissions, but the purpose of `rateRevisions` on a contract revision is not for rate history.
+   - The purpose of `rateRevisions` is to retain the data of a rate linked to this contract revision at the time of submission.
+   - For that we need the single rate revision that was submitted at the time this contract revision was submitted.
+   - Here are some guidelines for each rate revision in `rateRevisions` of a contract revision.
+      - Each rate revision in `rateRevisions` is unique by rate id, meaning there will never be two rate revisions with the same rate id in `rateRevisions`
+      - Each rate revision is the latest submitted up till the contract revision submitted time.
+      - Like contract revision, rate revision is read only and cannot be updated to retain its historical integrity.
```
haworku, line 43: Thank you!

services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
```diff
@@ -248,7 +250,7 @@ function ratesRevisionsToDomainModel(
     }
 
     domainRevisions.sort(
-        (a, b) => a.createdAt.getTime() - b.createdAt.getTime()
+        (a, b) => a.rate.createdAt.getTime() - b.rate.createdAt.getTime()
```
JasonLin0991, line 253: This preserves the rate order by createdAt in each revision. I noticed in the UI the rates were being sorted by the latest updated rate.

services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.test.ts
macrael: right!

services/app-web/src/components/DataDetail/DataDetailContactField/DataDetailContactField.tsx
macrael: I think that GQL using Maybe is something that can be configured in codegen.yml so that's probably worth looking into

PR #1992, opened 2023-10-18
Get stricter about rateinfo ids

macrael:
## Summary
We're seeing duplicate entries in the rates table for updated rates.

This was due to the fact that we were not actually perpetuating the UUID for rateInfos correctly. The front end didn't send the ID in its updateHPFD api call, so it was getting reset on every call to that API from the RateDetails page. This PR will stop the bleeding, future unlocked submissions will no longer create duplicate rates. A following PR will clean up the existing rates. I expect there to be duplicate rates for both migrated and unmigrated data. 

#### Related issues
[MCR-3566: Previous versions of rates are displayed on the rates dashboard](https://qmacbis.atlassian.net/browse/MCR-3566)

#### Test cases covered

We didn't effectively test this because it's a failure in between app-web and app-api. I'm not sure how to test this yet outside of the yet to be written RatesDashboard cypress test

### DIFF
--- a/services/app-api/src/handlers/proto_to_db.test.ts
+++ b/services/app-api/src/handlers/proto_to_db.test.ts
@@ -10,6 +10,7 @@ import {
     unlockTestHealthPlanPackage,
     updateTestHealthPlanFormData,
 } from '../testHelpers/gqlHelpers'
+import { v4 as uuidv4 } from 'uuid'
 import { latestFormData } from '../testHelpers/healthPlanPackageHelpers'
 import { testLDService } from '../testHelpers/launchDarklyHelpers'
 import { testCMSUser } from '../testHelpers/userHelpers'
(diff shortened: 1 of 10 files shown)


JasonLin0991:
> We didn't effectively test this because it's a failure in between app-web and app-api. I'm not sure how to test this yet outside of the yet to be written RatesDashboard cypress test

I have some rate dashboard tests merged in.

haworku (approved):
LGTM. That protobuf id code is something we all [discussed ](https://github.com/Enterprise-CMCS/managed-care-review/pull/1911#discussion_r1323801864 )before, its confusing.

I think this is an issue of having clarity on where ids and unique indentifiers should be created and required. Usually its on the backend -- but in this case (as well as the sha256) its on the frontend.  

Thank you so much for catching this I was struggling to see what was causing this last Friday ⭐️

PR #1997, opened 2023-10-20
MCR-3569 & MCR-3567: CHIP only submissions crash rates dashboard

JasonLin0991:
## Summary
[MCR-3569](https://qmacbis.atlassian.net/browse/MCR-3569)

In `submitHealthPlanPacakge` resolver, after we update the contract, we still use `contractWithHistory` to submit rates. `contractWithHistory` is now old data, so those disconnected rates are getting submitted when a user changes to a contract-only submission either by manually selecting `contract-only` or switching to a Chip only submission.

The fix was to use the updated contract data to submit the rates. If this were a contract-only submission, the updated contract data would have removed them.

**Steps to reproduce**:
1. Unlock and edit a contract and rates submission
2. Resubmit
3. Look at revision history

[MCR-3567](https://qmacbis.atlassian.net/browse/MCR-3567)
This was also the cause for MCR-3567: Contract only submissions are displaying as rows on the rates dashboard. 

**This only prevents the issue from occurring again; there still needs to be migration work to fix existing rates**
 
**Steps to reproduce**:
1. create a new `contract-and-rates` draft
2. on the rates details page, add rates but leave all fields blank and click save as draft
3. turn the submission to a `contract-only` submission and submit
4. on the rates dashboard, you should see blank entries.

#### Test cases covered

- `submitHealthPlanPackage.test.ts`
   - 'removes any rate data from CONTRACT_ONLY submissionType and submits successfully'
      - Added tests to make sure disconnected rates where not submitted along with the contact when flag is on.

### DIFF
--- a/services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.test.ts
+++ b/services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.test.ts
@@ -1,4 +1,5 @@
 import SUBMIT_HEALTH_PLAN_PACKAGE from '../../../../app-graphql/src/mutations/submitHealthPlanPackage.graphql'
+import FETCH_RATE from '../../../../app-graphql/src/queries/fetchRate.graphql'
 import {
     constructTestPostgresServer,
     createAndUpdateTestHealthPlanPackage,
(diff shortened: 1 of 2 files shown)


macrael (approved):
Makes sense!

PR #2002, opened 2023-10-24
Fix rate history contract revisions

JasonLin0991:
## Summary

Fixes rate history revisions contract data.

This fix refactors `rateWithHistoryToDomainModel` to only include the earliest submitted contract after the rate revision submission `submitInfo.updatedAt`. Any proceeding contract revision with the same `contractID` will be skipped.

There was a discussion in [DTBM 10.24.23](https://docs.google.com/document/d/1uVi5I_7KUPskeUQpkdWeWg4qP4ejwNJAiF1GJlfQFM4/edit) around the `Submission this rate was submitted with` and if it should link to the latest submission summary or the submission summary that this rate was submitted with. Specifically, when a rate was removed, the summary is still linked to the latest submission summary.

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/findRateWithHistory.test.ts
+++ b/services/app-api/src/postgres/contractAndRates/findRateWithHistory.test.ts
@@ -376,7 +376,8 @@ describe('findRate', () => {
         )
     })
 
-    it('finds a full rate', async () => {
+    // This test mimics the way rates are created, updated, and disconnected using the app today.
+    it('finds a full rate with history', async () => {
         const client = await sharedTestPrismaClient()
 
         const stateUser = await client.user.create({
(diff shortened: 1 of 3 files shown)


pearl-truss (approved):
code changes LGTM

haworku (approved):

services/app-api/src/postgres/contractAndRates/parseRateWithHistory.ts
```diff
@@ -165,42 +166,58 @@ function rateWithHistoryToDomainModel(
                 )
             }
 
-            // if it's from before this rate was submitted, it's there at the beginning.
+            // Finding the single earliest contract rev submitted right after rate rev submission date time.
```
haworku, line 169: This comment confuses me. Isn't this code finding multiple revs not just a single contract rev? Seems like we adding any contract rev that was submitted after rate rev to `initialEntry.contractRevs` with an array push.

Also why are keeping all this commented old code along? I think we can delete - its in version history.

services/app-api/src/postgres/contractAndRates/findRateWithHistory.test.ts
```diff
@@ -400,275 +401,343 @@ describe('findRate', () => {
...
+            await submitRate(client, {
+                rateID: rateIDOne,
+                submitReason: 'resubmit without contract',
+                submittedByUserID: stateUser.id,
+            })
         )
 
-        // Each Revision needs a Reason, one of the contracts or revisions associated with it should have changed and why.
+        // Expect no draft revision
+        expect(latestRateTwoResubmit.draftRevision).toBeUndefined()
```
haworku, line 714: Good to test this

services/app-api/src/postgres/contractAndRates/parseRateWithHistory.ts
JasonLin0991: This is building the rate history using the `rate.revisions` in the type `RateType`. 
- Since `rate.revisions` are all submitted rates, we need to attach the single contract revision that was submitted along with this rate to preserve the same submission data at that time.
-  We don't want future submission revisions of the same contract attached to it otherwise, this data would have future edits trickle down the revision history.

Let me know if that makes sense.

PR #2016, opened 2023-10-27
Clear out erroneous rates, reparenting rate revisions to logical "rates"

macrael:
## Summary

We were accidentally creating different "rates" every time we called updateHPP b/c we were setting a new UUID every time. That bug was fixed in #1992 this data migration fixes it in previous data. 

It's a little complex because there's no guarantee for how to determine if a resubmitted rate is actually the same "rate" as was submitted before. It's always possible that the user hit delete and re-added a similar rate meaning it to have been a distinct rate. 

The heuristic I used is that if two rates on separate submissions of the same contract share the same set of programs, then I consider them submissions of the same "rate"

I've run this migration locally against prod data and it seems to have done well, getting rid of duplicate rates I looked at. 

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-3566

#### Test cases covered
The tests I was using to write this don't really make sense anymore. More could be done to make tests that actually pass and ensure they are run correctly.

I downloaded all of prod data and ran this on it and checked it locally. I plan to do the same on this PR review app.


haworku (approved):
We talked through this in person but next steps:
- Make a ticket to investigate implications of cascade deletes
- Make a ticket for cypress test that would explicitly check for duplicate
 rates being made on edits
 - Add documentation for how migrator works and when to use this

PR #2021, opened 2023-10-30
MCR-2618 & MCR-3604: filter rates by rating period

JasonLin0991:
## Summary
[MCR-3604](https://qmacbis.atlassian.net/browse/MCR-3604)
- Copy the `DatePicker` from `react-uswds`.
- `DatePicker` has a new prop, `inputRef`.
   - returns ref to the `DatePicker` input element.
   - ref has a `clearInput()` function to clear input values.

[MCR-2618](https://qmacbis.atlassian.net/browse/MCR-2618)

- Added `FilterDateRange` component to the filter accordion and set up new `react-table` filters for rating period start ranges.
   - A recreation of `DateRangePicker` with error messages on each date input instead of the entire component and input validations.
   - A general `dateRangeFilter` filter function for `react-table`. There is a built-in `getFacetedMinMaxValues` that only works with numbers. We could probably get this to work with dates by converting all dates to numbers before passing them into the table. I chose not to do this because:
      - Wanted to keep all the dates in the same string format everywhere, only converting it when needed for date comparison in the filter functions.
      - `getFacetedMinMaxValues` unsure what refactor work would be needed for this.
      - Already had a date filter function working, refactoring that was less work to me.
   - These validations also do not update the URL filters unless on a valid input. Because of how we implemented the URL filters, any updates re-renders the component and refocuses the date range input. This caused some issues with validations, so my best option was to not update the URL until a valid date was inputted or the input was unfocused. Let me know if you have a better way,
- Fixed a table flickering issue on page load with filters in the URL. 
   - Created another `useAtom` with `loadable`, which doesn't seem to return with a setter. `loadable` does return a loading state with the filters from the URL. Until the data is parsed, it will be loading. Then, once done loading, we can set a state with the filters in a `useLayoutEffect`, which will then not be set again once it holds a value. These values are used for the default date range values.

Other

- Added cypress/downloads to gitignore

#### Test cases covered
`RateReviewsTable.test.ts`
- `'can filter on date ranges'`

`FilterDateRange.test.ts`
- `'returns inputted start and end date'`
- `'can clear inputs with forward refs'`
- `'shows error on invalid date input'`


services/app-web/src/components/DatePicker/Day.stories.tsx
```diff
@@ -0,0 +1,151 @@
+import React from 'react'
+
+import { Day } from './Day'
+
+/*
+// THIS STORY FOR INTERNAL DEVELOPMENT ONLY
```
haworku, line 6: Can you also drop a README in this DatePicker component folder about why we are doing this, when, and under what condition we could go back to using react-uswds version?

services/app-web/src/components/FilterAccordion/FilterDateRange/FilterDateRange.module.scss
```diff
@@ -0,0 +1,21 @@
...
+    width: 100%;
+  }
+  display: grid;
+  grid-template-columns: repeat(2, 1fr);
+  column-gap: 32px;
+}
+
+[class='usa-fieldset'] {
+  legend {
+    font-weight: normal;
```
haworku, line 19: Good use of targeted scss here.

Design nit: Please close the vertical spacing on the label - in the [designs](https://www.figma.com/file/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?type=design&node-id=138-9433&mode=design&t=ThvwlqwoeXAVG1zJ-0) the date range label is closer to the hint text which gives it more space between the inputs above
Screenshot 2023-11-07 at 10 02 51 AM

haworku: Design nit: Feels like we should have
Screenshot 2023-11-07 at 10 09 08 AM
 both inputs slide down when there is an error on either side. Keep them in alignment.

services/app-web/src/components/FilterAccordion/FilterDateRange/FilterDateRange.tsx
```diff
@@ -0,0 +1,310 @@
...
+ *
+ * We will no loner need to a modified DatePicker when:
+ * - DatePicker is updated to surface a ref to the input
+ * - DateRangePicker adds the option to display errors on each DatePicker input
+ */
+
+export type FilterDateRangePropType = {
+    startDateLabel?: string
+    startDateHint?: string
+    startDatePickerProps: Omit<DatePickerProps, 'rangeDate'>
```
haworku, line 30: I would add a comment here for both start date and end date picker props. I think there is basically a required `onChange` for each input in this component. Its intended to be used with an `onChange` that relates to behavior in a parent.  `// expect an onChange to be passed in for updating filter UI and URL params`

services/app-web/src/pages/CMSDashboard/RateReviewsDashboard/RateReviewsTable.test.tsx
```diff
@@ -185,4 +189,91 @@ describe('RateReviewsTable', () => {
...
+        // expect to only show rates with end dates on or before 11/30/2024
+        const thirdFilterRows = await screen.findAllByRole('row')
+        expect(thirdFilterRows).toHaveLength(3)
+        expect(
+            within(thirdFilterRows[1]).getByText('rate-2-certification-name')
+        ).toBeInTheDocument()
+        expect(
+            within(thirdFilterRows[2]).getByText('rate-1-certification-name')
+        ).toBeInTheDocument()
+        expect(screen.getByText('Displaying 2 of 3 rates')).toBeInTheDocument()
```
haworku, line 277: Great test and very needed to have unit test for this

haworku (approved):
LGTM.

I see things working and FYI I wasn't able to see any issue with the validations (besides that we don't clearly warn when the manually entered start date is after the end date, but I noticed that wasn't part of the AC).  Whats happening with the URL seemed fine as well

I do think we need to prioritize refactoring the tables soon.  A lot going on in these dashboard files now. That can be a separate effort.

services/app-web/src/pages/CMSDashboard/RateReviewsDashboard/RateReviewsTable.tsx
```diff
@@ -256,6 +311,7 @@ export const RateReviewsTable = ({
                 meta: {
                     dataTestID: `${tableConfig.rowIDName}-date`,
                 },
+                filterFn: 'endDateFilter',
```
haworku, line 314: 🐛 This is the bug - as discussed in zoom. This end date filter is not a filter for the rate end date column. It should be applied along side the the start date filter for the rating start date column (we are filtering that a rating start date is within/between a range)

services/app-web/src/pages/CMSDashboard/RateReviewsDashboard/RateReviewsTable.tsx
```diff
@@ -134,6 +144,42 @@ const getSelectedFiltersFromUrl = (
...
+        (filter) => (filterLookup[filter.id] = filter.value as string[])
+    )
+
+    if (filterLookup[id]) {
+        return filterLookup[id][0]
+    }
+    return undefined
+}
+
+// Even though both date filters are similar we don't want to combine into one function because we don't always know
```
haworku, line 162: I feel like these should be one function because they are filtering the same columnn and we pass in its id

services/app-web/src/pages/CMSDashboard/RateReviewsDashboard/RateReviewsTable.tsx
```diff
@@ -396,22 +396,37 @@ export const RateReviewsTable = ({
...
-        } else {
-            filterColumn.setFilterValue([date])
-        }
+        filterColumn.setFilterValue(
+            (value: RatingPeriodFilterType): RatingPeriodFilterType => {
+                const prevDates = value ?? ['', '']
+                const toDate =
+                    date[0] !== undefined
+                        ? date[0]
+                        : prevDates[0] !== undefined
```
haworku, line 412: Suggestion (not blocking):  simplify the logic here and adding comments. Why are we preserving previous dates in the filter here?  I believe `const toDate = date[0] !== undefined ? date[0] : prevDates[0] || ''` is still valid

haworku (approved):
LGTM I see the range filters working now on the rating period start date. Also like that we are leveraging react-table built-ins now.  🎉

services/app-web/src/pages/CMSDashboard/RateReviewsDashboard/RateReviewsTable.tsx
JasonLin0991: This handles the existing `from` or `to` input date when updating the opposite input. Because this is a single field, the `rating period start`, that we are filtering on, the state has to hold both values in the filter state `[from, to]`. When updating, we need to set either the `to` or `from` in that array while preserving the opposite input value so as to not clear it out. 

It is a little complicated because I kept it as one update function instead of two, one for `to` and one for `from`.

JasonLin0991: I will add a comment.

haworku: Yes - I like that its one function especially because its acting on the same single data column. A little comment clears this right up about why we preserve the prev values (since inputs are updated one at a time not simultaneously, so unchanged one needs to stay the same).

PR #2027, opened 2023-11-01
Connect QA to Contracts

macrael:
## Summary

This PR changes Questions to be related to the ContractTable rather than the HPPTable and migrates all existing Questions over to the new table (these only exist in dev/val)

When we add questions to Rates it will add complexity but this PR is pretty simple. 

Joy: almost all the code changes were driven by compiler errors. 

#### Related issues

https://qmacbis.atlassian.net/browse/MCR-2640

## QA guidance

Should be able to create questions in environments with the db refactor (which is all of them now)

### DIFF
--- a/services/app-api/prisma/migrations/20231031204348_migrate_questions_to_contracts/migration.sql
+++ b/services/app-api/prisma/migrations/20231031204348_migrate_questions_to_contracts/migration.sql
@@ -1,0 +1,17 @@
+BEGIN;
+
+ALTER TABLE "Question" ADD COLUMN "contractID" TEXT;
+
+-- AddForeignKey
+ALTER TABLE "Question" ADD CONSTRAINT "Question_contractID_fkey" FOREIGN KEY ("contractID") REFERENCES "ContractTable"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+
+-- migrate data over
+UPDATE "Question" SET "contractID"="pkgID";
+
+-- make not nullable, drop pkgID
+ALTER TABLE "Question" 
+  ALTER COLUMN "contractID" SET NOT NULL,
+  DROP CONSTRAINT "Question_pkgID_fkey",
+  DROP COLUMN "pkgID";
+
+COMMIT;
--- a/services/app-api/prisma/schema.prisma
+++ b/services/app-api/prisma/schema.prisma
@@ -13,7 +13,6 @@ model HealthPlanPackageTable {
   stateCode String
   state     State                     @relation(fields: [stateCode], references: [stateCode])
   revisions HealthPlanRevisionTable[]
-  questions Question[]
 }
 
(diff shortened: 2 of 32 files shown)


services/app-api/prisma/schema.prisma
```diff
@@ -291,12 +292,12 @@ model UserAudit {
 }
 
 model Question {
-  id            String                 @id @default(uuid())
-  createdAt     DateTime               @default(now())
-  updatedAt     DateTime               @default(now()) @updatedAt
-  pkgID         String
-  pkg           HealthPlanPackageTable @relation(fields: [pkgID], references: [id])
```
JasonLin0991, line 298: With these changes to `Question` and `HealthPlanPackageTable`, would we want to make a migration or drop the Q&A tables? I think this only applies to `Val` since it is the only environment that has existing Q&A records.

macrael: Take a closer look at that migration! All the Questions that were associated with HPP should now be associated with Contracts

services/app-api/prisma/migrations/20231031204348_migrate_questions_to_contracts/migration.sql
```diff
@@ -0,0 +1,17 @@
+BEGIN;
+
+ALTER TABLE "Question" ADD COLUMN "contractID" TEXT;
+
+-- AddForeignKey
+ALTER TABLE "Question" ADD CONSTRAINT "Question_contractID_fkey" FOREIGN KEY ("contractID") REFERENCES "ContractTable"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+
+-- migrate data over
+UPDATE "Question" SET "contractID"="pkgID";
```
macrael, line 9: This is where we copy all the pkgIDs into contractIDs

The way this works is that we don't add the NOT NULL constraint until after we copy this over. 
then we can safely drop the pkgID column

services/app-api/prisma/schema.prisma
```diff
@@ -130,21 +131,21 @@ model RateRevisionTable {
...
+  rateType                        RateType?
+  rateCapitationType              RateCapitationType?
+  rateDocuments                   RateDocument[]
+  supportingDocuments             RateSupportingDocument[]
+  rateDateStart                   DateTime?                @db.Date
+  rateDateEnd                     DateTime?                @db.Date
+  rateDateCertified               DateTime?                @db.Date
+  amendmentEffectiveDateStart     DateTime?                @db.Date
+  amendmentEffectiveDateEnd       DateTime?                @db.Date
+  rateProgramIDs                  String[]
```
haworku, line 143: Hmm maybe we need to build in prisma formatting into our precommit scripts. I've been noticing these kind of formatting difff on our prisma PRs over the last few months

PR #2030, opened 2023-11-03
[MCR-3568] fix 2 A11y issues for screen readers

pearl-truss:
## Summary

This PR fixes 2 bugs

**Bug 1:**

1. Go to https://assistivlabs.com/ and login with your credentials
2. Select JAWS from the tool panel
3. Once in the remote session, open a browser and go to MC-Review 
4. Login 
5. Using the TAB key, tab through the CMS dashboard page until you reach the views, Submissions and Rate Reviews
6. Tab back and forth from these views and listen to the screen reader tool

**Expected behavior in this PR:** JAWS will read out which tab is active. The narrator does also say "press enter to activate" but design [says](https://qmacbis.atlassian.net/browse/MCR-3568?focusedCommentId=125394) that's okay

**Bug 2:**

1. Go to https://assistivlabs.com/ and login with your credentials
2. Select JAWS from the tool panel
3. Once in the remote session, open a browser and go to MC-Review 
4. Login 
5. Navigate to the Rate Review dashboard view and select a rate review
6. Using the ARROW keys, navigate through the Rate Summary page until you've reached the Rate Documents table

**Expected behavior in this PR:** JAWS will only announce the "Rate Credentials" and "Rate Supporting Documents" labels once

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-3568

#### Screenshots
**Bug 1 fix**

https://github.com/Enterprise-CMCS/managed-care-review/assets/67110378/69a7c41d-53e1-41e8-9bea-51b1f653da2d

**Bug 2 fix**

https://github.com/Enterprise-CMCS/managed-care-review/assets/67110378/00ddaba8-3d5b-4bc5-a19d-8a581ced84d8

## QA guidance

- Both issues are testable as a CMS user
- Bug 1 is on the dashboard page
- Bug 2 is on the rate review summary page


services/app-web/src/components/Tabs/Tabs.tsx
```diff
@@ -39,12 +39,17 @@ export const Tabs = ({
                             role="tab"
                             aria-selected={activeTab === tab.name}
                         >
-                            <Heading className={styles['easi-tabs__tab-btn']}>
+                            <Heading
+                                className={styles['easi-tabs__tab-btn']}
+                                aria-description={
+                                    activeTab === tab.name ? 'active' : ''
```
haworku, line 45: A couple things here:
1. `aria-description` is [usually used for long description](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description). I don't think this is proper usage.
2. It feels like the bug here is really something with the `aria-selected` behavior. I would look into why that selected* aria isn't working properly instead - that's what is supposed to announce what tab is in use in semantic HTML. I'm confused if this was always broken or maybe I broke it when I hooked up tabs with the URL routing. 
-  Here's some documentation on how tabs should work (I helped write this and can answer any questions) - or see [WCAG ](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/)or MDN docs.

haworku: in case its helpful - I just did a recording myself of the simple tabs from WCAG in assitive labs. This is using no aria describe or aria labels - the word "selected" comes from that aria-selected field being set. 

https://github.com/Enterprise-CMCS/managed-care-review/assets/10750442/40d13f87-0d3f-49f4-b4be-0a5134b95b89

pearl-truss: Thanks for the links, Hana! I'll look into this some more. `aria-selected` on the tab was definitely the first thing I tried but I haven't been able to get it to announce selected yet

pearl-truss: I was able to get `aria-selected` read out with changes to the HTML. However, when a tab is first activated, JAWS doesn't finish announcing the tab before the page URL changes. I looked into trying to trying `e.preventDefault` and passing in `replace: true` to `navigate()`. However the only thing that worked was adding a timeout after the active tab is selected but I really don't think we actually want to do that. 

https://github.com/Enterprise-CMCS/managed-care-review/assets/67110378/2d01842c-6794-49da-8a8b-c84889f1d57e

haworku: Yes that makes sense. I don't think this is a bug since our tabs also are doing entire page loads (and acting more link nav links). To me, that behavior looks right to me for tabs hacked to behave as links. We are losing focus and its cutting off because of the reroute.

haworku (approved):

PR #2036, opened 2023-11-07
[MCR-3616] update state programs

pearl-truss:

### DIFF
--- a/services/app-web/src/common-code/data/statePrograms.json
+++ b/services/app-web/src/common-code/data/statePrograms.json
@@ -138,16 +138,6 @@
                     "name": "AHF"
                 },
                 {
-                    "id": "53a6a9ce-e2f9-47af-9795-12901e13a986",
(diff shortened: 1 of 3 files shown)


services/app-web/src/common-code/data/statePrograms.json
```diff
@@ -832,6 +812,21 @@
...
                     "fullName": "SoonerCare Choice",
                     "name": "SoonerCare Choice"
+                },
+                {
+                    "id": "",
+                    "fullName": "SoonerSelect Dental",
+                    "name": "SoonerSelect Dental"
+                },
+                {
+                    "id": "",
```
haworku, line 822: Annoying gotcha here -  we need to also manually add the `id` field when adding new programs to the list. Looks like we forgot to include this in the howto-update-state-programs doc as well, could you add it there please  as well?

haworku (approved):

PR #2039, opened 2023-11-08
Add Kentucky state icon to header.

JasonLin0991:

PR #2040, opened 2023-11-08
Add Kentucky to the list of state code enums in protobuf

macrael:
## Summary
Amanda discovered that attempting to create a new submission in Kentucky errored. 

We managed to leave KY out of the list of enums in our protobuf definition.

I've added it and compared it to list here https://en.wikipedia.org/wiki/List_of_U.S._state_and_territory_abbreviations it is the only one that was missing besides some territories we don't need to support.


macrael:
Jason added this in #2039

PR #2053, opened 2023-11-14
[MCR-3625] MCCRS Record Number page a11y fixes

pearl-truss:
## Summary

This PR resolves 2 A11y issues in the MCCRS record number flow

**Bug 1**

- Located on the submission summary page, after a MCCRS record number has been added for a submission
- If you click on the MCCRS record number (that is a link to the mccrs system) you should now hear a clear label for this link instead of silence
- If you click on the "Edit MCCRS record number" link you should now hear a clear label for this link instead of silence

**Bug 2**

- Located on the Add MCCRS record number page
- Tabbing through the breadcrumbs, the middle breadcrumb for the submission page should no longer be read out as current page
- Added in this PR is the ability to tab to the third breadcrumb and hear that the mccrs record number page is the current page

#### Related issues

https://qmacbis.atlassian.net/browse/MCR-3625

#### Screenshots

**Bug 1**

https://github.com/Enterprise-CMCS/managed-care-review/assets/67110378/d7fcfd4e-5115-45d2-b347-198397921b66

Bug 2

https://github.com/Enterprise-CMCS/managed-care-review/assets/67110378/06c87cc7-35b6-4ae2-ada4-764cfd14c917

### DIFF
--- a/services/app-web/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/services/app-web/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -4,7 +4,7 @@ import styles from './Breadcrumbs.module.scss'
 
 type BreadcrumbItem = {
     text: string
-    link?: string
+    link: string
 }
 export type BreadcrumbsProps = {
     items: BreadcrumbItem[]
@@ -12,21 +12,13 @@ export type BreadcrumbsProps = {
 
 const Crumb = (crumb: BreadcrumbItem) => {
     const { link, text } = crumb
-    if (link) {
-        return (
-            <Breadcrumb>
-                <Link asCustom={NavLink} to={link} end>
-                    <span>{text}</span>
-                </Link>
-            </Breadcrumb>
-        )
-    } else {
-        return (
-            <Breadcrumb>
+    return (
+        <Breadcrumb>
+            <Link asCustom={NavLink} to={link} end>
                 <span>{text}</span>
-            </Breadcrumb>
-        )
-    }
+            </Link>
+        </Breadcrumb>
+    )
 }
 
 const Breadcrumbs = ({ items }: BreadcrumbsProps) => {
--- a/services/app-web/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/services/app-web/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -15,7 +15,7 @@ const Crumb = (crumb: BreadcrumbItem) => {
     if (link) {
         return (
             <Breadcrumb>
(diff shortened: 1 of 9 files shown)


services/app-web/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -15,7 +15,7 @@ const Crumb = (crumb: BreadcrumbItem) => {
     if (link) {
         return (
             <Breadcrumb>
-                <Link asCustom={NavLink} to={link}>
+                <Link asCustom={NavLink} to={link} end>
```
pearl-truss, line 18: This resolves an issue with [react router dom](https://reactrouter.com/en/main/components/nav-link#end) where the submission page was being read out as the current page due to the url on that page being contained in the `submissions/:submissionID/mccrs-record-number` url which resulted in "current page" being read out for both breadcrumbs

services/app-web/src/pages/MccrsId/MccrsId.tsx
```diff
@@ -176,7 +176,10 @@ export const MccrsId = (): React.ReactElement => {
                                     link: `/submissions/${id}`,
                                     text: pkgName || '',
                                 },
-                                { text: 'MC-CRS record number' },
+                                {
+                                    text: 'MC-CRS record number',
+                                    link: `/submissions/${id}/mccrs-record-number`,
```
pearl-truss, line 181: The ticket called out that  "I need the breadcrumb title to clearly call out the page I am on,  for example “MC-CRS record number - Current Page”, so I easily understand where I am on the page.", so in order for this to be tab-able and read out by JAWS I turned it into a link. If there's a better solution please let me know

haworku: Cool. Let's go ahead and make `link` required then on the Breadcrumbs props itself.  Sounds like there is an a11y impact when we don't ensure every item as a link. That way this doesn't reoccur

haworku (approved):

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -179,6 +180,11 @@ export const SubmissionSummary = (): React.ReactElement => {
                                     className={
                                         pkg.mccrsID ? styles.editLink : ''
                                     }
+                                    aria-label={
+                                        pkg.mccrsID
```
haworku, line 184: nit: I try to use a shared variable when I'm trying to align display text with aria-label text so that these cannot accidentallynot diverge

PR #2054, opened 2023-11-14
Add jumpbox clone command to dev tool

macrael:
## Summary

This PR adds a subcommand to ./dev to allow cloning down a db from dev/val/prod to your local machine

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-2633


dev_tool/src/jumpbox.ts
```diff
@@ -0,0 +1,323 @@
...
+        // running state code
+        console.info('Jumpbox is running')
+        return jumpboxStartInstance
+    }
+
+    console.info('Jumpbox is not running', jumpboxStartInstance.State?.Name)
+
+    if (jumpboxStartState !== 80) {
+        console.info('Jumpbox is not stopped yet. waiting to start it')
+        // wait for it to be running
```
mojotalantikite, line 101: I think we're waiting for it to be stopped here? Probably should just fix that comment.

dev_tool/src/jumpbox.ts
```diff
@@ -0,0 +1,323 @@
...
+    })
+    if (startedInstance instanceof Error) {
+        console.error('error fetching restarted jumpbox', startedInstance)
+        return startedInstance
+    }
+
+    return startedInstance
+}
+
+async function ensureWhitelistIP(
```
mojotalantikite, line 139: Not to be that sjw, but I think people like to use "allowlist"/"blocklist" more recently.

dev_tool/src/jumpbox.ts
```diff
@@ -0,0 +1,323 @@
...
+        console.info('Connecting to ', jumpboxIP)
+        const ssh = new NodeSSH()
+
+        // If the jumpbox just started we often have the connection refused for ~10 seconds until it's really up.
+        const connectionResult = await retry(async () => {
+            try {
+                await ssh.connect({
+                    host: jumpboxIP,
+                    username: 'ubuntu',
+                    privateKeyPath: '/Users/macrae/.ssh/wml_jumpbox',
```
mojotalantikite, line 245: Hmm, we're going to need to use each individual's private key here. I looked at the lib and it seems to require a path. There's [another lib](https://github.com/mcavage/node-ssh-agent) that can read from ssh-agent, but we should probably think about what to do here.

services/postgres/serverless.yml
```diff
@@ -181,7 +184,11 @@ resources:
         UserData:
           Fn::Base64: !Sub |
             #!/bin/bash
-            apt update && apt install unzip postgresql postgresql-contrib -y
+            # get apt data for postgres-14
+            sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
```
mojotalantikite, line 188: Don't really need a sudo here as everything runs as root. 

One thing though is that the user data scripts only get run once and it's at time of creation, so we should make sure we cycle the machines so this gets applied on dev/val/prod VMs.

mojotalantikite (approved):

services/postgres/serverless.yml
macrael: I've already run this on the existing jump boxes, should we be cycling these automatically?

haworku:
YAY! A few comments and TODOs from running through things
- [ ] Add docs for this dev command and note when it should be used
- [ ] Address `Error: Region is missing` AWS config error on initial run. You said this might be something to add to the script itself. 
- [ ] Address `failed to connect to jumpbox over ssh AssertionError [ERR_ASSERTION]: config.privateKeyPath does not exist at given fs path`
- [ ] Have the script print out where the file was copied to locally which path and what filename? I have a few versions of the file so having the filename would be important.
- [ ] Have the script put the files that are created into a specific folder that is `.gitignore`d for this repo. Don't want things accidentally committed when this is run in our repo root. 
- [ ] After merging check that we addressed `error fetching jumpbox Error: Did not find one and only one instance` when trying to run on VAL and above

PR #2055, opened 2023-11-14
MCR-3571: Sort state filter options in alphabetical order.

JasonLin0991:
## Summary
[MCR-3751](https://qmacbis.atlassian.net/browse/MCR-3751)

- Sort state filter options.


macrael (approved):
what would it take to have these filters be searchable by state code as well as state name?

JasonLin0991:
> what would it take to have these filters be searchable by state code as well as state name?

Custom filter logic for react-select. I think it would be at least a 2 pointer.

PR #2059, opened 2023-11-16
Incremental ./dev improvements 

haworku:
## Summary
- Get `./dev clean` to cover more cases
- Make `./dev rebuild` for use after clean 
- Remove storybook from `./dev local`so developers don't have to keep typing `no-storybook`. Storybook is something that can be spun up on demand when needed in separate tab, its not essential for local dev.
- Make the db exit 1 when we have a db migration fail.


haworku:
This PR has been open for sometime, I'm going to go ahead and merge cc: @macrael for visibility since you gave feedback on zoom

PR #2068, opened 2023-11-20
[MCR-2538] email notifications to states

pearl-truss:


services/app-api/src/emailer/emailer.ts
```diff
@@ -186,6 +195,27 @@ function newSESEmailer(config: EmailConfiguration): Emailer {
                 return await this.sendEmail(emailData)
             }
         },
+        sendQuestionsStateEmail: async function (
```
macrael, line 198: as you are working on these four emails, I wonder if we could refactor the SESEmailer and the localEmailer code to be a little less repetitive

services/app-api/src/emailer/emails/newQuestionStateEmail.test.ts
```diff
@@ -0,0 +1,270 @@
...
+        defaultSubmitters,
+        cmsUser,
+        testEmailConfig(),
+        defaultStatePrograms,
+        dateAsked
+    )
+
+    if (template instanceof Error) {
+        console.error(template)
+        return
```
macrael, line 45: I would throw this error if it's an error, that will fail the test. I don't think returning will fail the test but maybe it will b/c there haven't been any assertions? In general, if there's some kind of error in a test I like to throw it to fail. 

Lots of our test helpers throw (see `createAndSubmitTestHealthPlanPackage` for instance)

This is a big exception to our general rule of "return Errors, don't throw" but it's basically the API of jest so it makes sense to me

services/app-api/src/resolvers/questionResponse/createQuestion.ts
```diff
@@ -77,6 +83,47 @@ export function createQuestionResolver(
             throw new UserInputError(errMessage)
         }
 
+        const conversionResult = convertContractWithRatesToFormData(
```
macrael, line 86: Can we move away from converting the contract to the old HealthPlanPackage and HealthPlanFormData types? The rates refactor was the first step in deprecating those types. We're still converting to them in order to return them from our handlers, but when we convert the front end to be able to support rates across submissions the plan is to try and root those types out all together. 

I think the only real change this would require would be for the sendQuestionsStateEmail function to take in a ContractWithRatesType instead of a HPFD type, but it should have all the data you need to compose the email in it, then.

pearl-truss:
@macrael thanks for the review! This is ready for another pass when you get a chance

services/app-api/src/emailer/emails/sendQuestionStateEmail.test.ts
```diff
@@ -0,0 +1,320 @@
...
+            },
+            {
+                name: 'test1',
+                titleRole: 'Foo1',
+                email: 'test1@example.com',
+            },
+        ],
+    }
+    const sub: ContractRevisionWithRatesType = {
+        ...mockContractRev({ formData: formDataWithDuplicateStateContacts }),
```
macrael, line 159: what's up with expanding mockContractRev here instead of just using it by itself?

services/app-api/src/emailer/emails/sendQuestionStateEmail.test.ts
```diff
@@ -0,0 +1,320 @@
...
+        sub,
+        defaultSubmitters,
+        cmsUser,
+        testEmailConfig(),
+        defaultStatePrograms,
+        dateAsked
+    )
+
+    if (template instanceof Error) {
+        console.error(template)
```
macrael, line 129: there's a bunch more of these instances where you're returning an error instead of throwing them, can you fix those too?

services/app-api/src/resolvers/questionResponse/createQuestion.ts
```diff
@@ -86,6 +102,30 @@ export function createQuestionResolver(
             throw new Error(errMessage)
         }
 
+        const dateAsked = new Date()
+        const sendQuestionsStateEmailResult =
```
macrael, line 106: yay, I love how much less prep there is here now.

services/app-api/src/emailer/emailer.ts
```diff
@@ -180,181 +236,117 @@ function newSESEmailer(config: EmailConfiguration): Emailer {
...
+                    return emailData
+                } else {
+                    return await this.sendEmail(emailData)
+                }
             }
         },
     }
 }
 
+function newSESEmailer(config: EmailConfiguration): Emailer {
```
macrael, line 343: I'm very glad you maintained these entry points. I think we can simplify things a bit more. Looking at it I think the main weirdness is that sendEmail is not like the other functions. And I forgot that we also duplicate this structure in our testing helper too. I think we could break out the sendTestEmail, sendSESEmail, sendLocalEmail functions into something that you pass into a single "emailer". I can sketch it out or we can meet and talk it through.

macrael (approved):
This looks great, so glad to see the HPP stuff gone. I think more refactoring can come in further PRs or in this PR up to you. Also I'm happy to meet and try and work on the refactor together, I think I finally see something in how we could pull the sendEmail function out of the emailer thing

PR #2069, opened 2023-11-20
Unify colors and add styles docs

haworku:
## Summary
- Shared color and container tokens for use across the application. Use `$mcr-*` colors everywhere
- Prefer `@use` to `@import` in module scss files and move towards namespaced imports
- MCR Colors storybook component to make colors in use more visible
- Add documentation how on to use styles and module scss files
- Add stylelint and enforce no hex codes rule

**Note:** There is a new major storybook version ([v7](https://storybook.js.org/releases/7.0)) from early 2023 that we should move to eventually. Seemed out of scope for this effort. I imagine we would port the Color storybook over tp e [color palette format](https://storybook.js.org/docs/api/doc-block-colorpalette) at that time. For now just copied spaceforce code for our `Colors.stories.tsx`

## Related Issue
https://qmacbis.atlassian.net/browse/MCR-3847 
#### Screenshots
Screenshot 2023-11-20 at 5 21 46 PM

## QA guidance

- [x] Read the new doc 
- [x] Take a look at Colors storybook 
- [x] Run through the application, check each page of the application and compare with whats on VAL - ensure there are no regressions with spacing or colors specifically.[ Jira ticket](https://qmacbis.atlassian.net/browse/MCR-3847) has more detailed instructions. 
    - If you find style regressions, please take side by side screenshots and comment them. Thank you!


haworku:
I have a known issue where custom stylesheets is loaded multiple times on certain pages because of the way I import. I'm still figuring out the best option there  -- but everything else in terms of code change and the styles in the application is ready for review.

haworku:
@macrael @pearl-truss @JasonLin0991 Still waiting for reviews on this from an engineer. Design has taken a look. There are QA steps that I think make this manageable to review. Would like to land this early next week.

services/app-web/src/components/ActionButton/ActionButton.module.scss
```diff
@@ -1,5 +1,5 @@
-@import '../../styles/uswdsImports.scss';
-@import '../../styles/custom.scss';
+@use '../../styles/custom.scss' as custom;
+@use '../../styles/uswdsImports.scss' as uswds;
```
pearl-truss, line 2: Question: is the uswds import necessary? I know it's not newly added in this PR but it doesn't seem to be used in this file

services/app-web/src/components/Banner/Banner.module.scss
```diff
@@ -1,5 +1,5 @@
-@import '../../styles/uswdsImports.scss';
-@import '../../styles/custom.scss';
+@use '../../styles/custom.scss' as custom;
+@use '../../styles/uswdsImports.scss' as uswds;
```
pearl-truss, line 2: Could these imports be removed since they aren't used in this file?

services/app-web/src/components/Breadcrumbs/Breadcrumbs.module.scss
```diff
@@ -1,4 +1,5 @@
-@import '../../styles/uswdsImports.scss';
+@use '../../styles/custom.scss' as custom;
+@use '../../styles/uswdsImports.scss' as uswds;
```
pearl-truss, line 2: Same as above, I don't believe the uswds import is necessary here

services/app-web/src/components/Colors/Colors.module.scss
pearl-truss, line 1: Praise: I love how succinct and clear this is

services/app-web/src/components/DoubleColumnGrid/DoubleColumnGrid.module.scss
```diff
@@ -1,7 +1,8 @@
-@import '../../styles/uswdsImports.scss';
-@import '../../styles/custom.scss';
+@use '../../styles/custom.scss' as custom;
```
pearl-truss, line 1: I don't think the custom import is needed here

services/app-web/src/components/DownloadButton/DownloadButton.module.scss
```diff
@@ -1,5 +1,5 @@
-@import '../../styles/uswdsImports.scss';
-@import '../../styles/custom.scss';
+@use '../../styles/custom.scss' as custom;
```
pearl-truss, line 1: I don't think either of the imports are needed here

services/app-web/src/components/Header/Header.module.scss
```diff
@@ -1,54 +1,42 @@
-@import '../../styles/uswdsImports.scss';
-@import '../../styles/custom.scss';
+@use '../../styles/custom.scss' as custom;
+@use '../../styles/uswdsImports.scss' as uswds;
 
 .userInfo {
-    font-weight: font-weight('light');
-    padding: 0 units(1);
-    color: $cms-color-white;
+    font-weight: uswds.font-weight('light');
```
pearl-truss, line 5: I've noticed others files in our codebase use `font-weight: normal/bold` instead of the way that's used here. I'm not sure but it's probably beyond the scope of the PR to unify this?

services/app-web/src/components/SubmissionSummarySection/UploadedDocumentsTable/UploadedDocumentsTable.module.scss
```diff
@@ -1,26 +1,29 @@
...
     td {
         text-align: left;
-        padding: units(1);
+        padding: uswds.units(1);
     }
 
     tbody {
         tr:nth-child(even) {
-            background: $theme-color-base-lightest;
+            background: custom.$mcr-gray-lightest
```
pearl-truss, line 20: I think this needs a `;`

services/app-web/src/styles/custom.scss
```diff
@@ -1,100 +1,69 @@
...
-    text-align: center;
-
-    ol {
-        justify-content: center;
-    }
-}
-
-[class^='usa-step-indicator__header'] {
-    display: inline;
+/* ACCESSIBILITY */
```
pearl-truss, line 51: I think this comment is for line 61

services/app-web/src/styles/custom.scss
```diff
@@ -1,100 +1,69 @@
...
+// Re-useable classname for hiding UI visually but still having it be read by screenreaders
```
pearl-truss, line 52: I think this comment is for line 61

pearl-truss (approved):
I left and then deleted comments related to files having `custom` and or `uswds` imports that might not be necessary because it's probably beyond the scope of this PR

I also left some comments on the documentation but overall this looks really good! I really like the use of namespaces for imports and it's much clearer to me what is coming from where

services/app-web/.stylelintrc
```diff
@@ -0,0 +1,14 @@
+{
```
pearl-truss, line 1: I haven't been able to get this to work. I enabled vscode-stylelint to test this, are you using something different?

haworku:
> I left and then deleted comments related to files having custom and or uswds imports that might not be necessary because it's probably beyond the scope of this PR 

Appreciate that note though. I did just use across every file as you noticed with find and replace - not really caring if its actually used or not. I can clean up that up once this lands in between sprint work.

docs/technical-design/howto-style-css-scss.md
```diff
@@ -0,0 +1,99 @@
...
+    1. We are dealing with container elements that repeat throughout the application outside of a specific component (divs, cards, display tables)
+    2. We need override a behavior in USWDS that we can't control via settings throughout the application
+    3. We want to hide an element from the screen using our global screenreader only class
+
+
+Prefer encapsulating into components with their own module.scss.
+
+### React components may use conditional styles and compose together styles using the `classnames` package
+
+- Sometimes React components have complex logic related to styles. To assist with this use [`classnames`](https://github.com/JedWatson/classnames) to compose together styles from different places.
```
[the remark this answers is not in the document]
haworku: Standardized on `classnames` to follow @trussworks/react-uswds. Wanted css patterns to be folks be well trodden path for folks less comfortable with it.

I don't recommend template literals unless wrapped in some kind of shared helper function that would check what comes out looks correct. Particularly when you have cases with multiple conditionals (4+)  - file by file template literals get messy. If we wanted to swap out classnames packages (very open to that) would rather make our own simple function pattern and use it everywhere in one refactor rather than file by file handling. 

@pearl-truss Would you want to see a sentence on this in the docs? Or else something like a quick basic ADR on this - could write easily.

services/app-web/.stylelintrc
haworku: I was running stylelint in app-web manually `npx styelint **/*.scss`-  let me surface that as a yarn command also and document. It should be hooked into lint-staged as well (precommit script) I'll double check that.

docs/technical-design/howto-style-css-scss.md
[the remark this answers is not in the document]
pearl-truss: Gotcha, thanks for laying that out! I think adding a blurb to the docs here would be enough

services/app-web/src/components/FilterAccordion/FilterDateRange/_DatePicker/DatePicker.test.tsx
haworku, line 1: Diff in this file is due to addressing a skipped test that failed lint. I deleted them since they aren;'t needed. Then prettier ran so things shifted around abit

pearl-truss:
Thanks for making those changes! And I see the stylelint working on precommit now 🎉

PR #2096, opened 2023-12-04
[MCR-2539] CMS notified when a state is sent questions

pearl-truss:
## Summary

This PR creates a way for CMS users to be notified via email when someone from the same CMS division sends a question to a state. The state analysts for a state will be notified regardless of which CMS division the question came from.

#### Related issues

https://qmacbis.atlassian.net/browse/MCR-2539

#### Screenshots

Screenshot 2023-12-05 at 4 31 44 PM

#### Test cases covered

-  the to address list includes state analysts regardless of which CMS division submitted the question
-  the to address list includes OACT emails when someone from OACT submitted the question
-  the to address list includes DMCP emails when someone from DMCP submitted the question
-  subject line is correct
-  the email itself includes relevant info (like link to a QA, name, email and division of the CMS user who submitted the question)
-  renders overall email

## QA guidance

- Login to the review app as a CMS user
- Find a submitted submission and click on the QA tab, upload a question
- Check the email, state analysts should be notified and the group DMCP is a DMCP user submitted the question, same for OACT


services/app-api/src/emailer/emails/sendQuestionCMSEmail.ts
```diff
@@ -0,0 +1,78 @@
...
+    config: EmailConfiguration,
+    statePrograms: ProgramType[],
+    dateAsked: Date,
+    roundNumber: number
+): Promise<EmailData | Error> => {
+    let receiverEmails = [...stateAnalystsEmails, ...config.devReviewTeamEmails]
+    if (cmsRequestor.divisionAssignment === 'DMCP') {
+        receiverEmails.push(...config.dmcpEmails)
+    } else if (cmsRequestor.divisionAssignment === 'OACT') {
+        receiverEmails.push(...config.oactEmails)
```
JasonLin0991, line 27: I think we are missing `DMCO` primary inbox `config.dmcoEmails` here.

AC 1
```
If a CMS user from ANY division (DMCO, DMCP, or OACT) submits a round of questions for a submission 
Then the CMS DMCO users assigned to the state receives an automated confirmation email.
```

services/app-api/src/emailer/emails/sendQuestionCMSEmail.ts
```diff
@@ -0,0 +1,78 @@
...
+import type { ContractRevisionWithRatesType } from '../../domain-models/contractAndRates'
+
+export const sendQuestionCMSEmail = async (
+    contractRev: ContractRevisionWithRatesType,
+    stateAnalystsEmails: StateAnalystsEmails,
+    cmsRequestor: CMSUserType,
+    config: EmailConfiguration,
+    statePrograms: ProgramType[],
+    dateAsked: Date,
+    roundNumber: number
```
JasonLin0991, line 21: Not sure if we wanted to do this, but if we passed in the `question` data from the resolver, we could reduce the parameters by 2. The `question` data will have `addedBy` and `createdAt` for the CMS requestor email and the date the question was asked. Also, we then won't need to generate `dateAsked` using `new Date()` in the resolver

cc: @macrael

services/app-api/src/emailer/emails/sendQuestionStateEmail.ts
```diff
@@ -43,14 +43,14 @@ export const sendQuestionStateEmail = async (
...
-    const packageURL = submissionSummaryURL(
+    const questionResponseURL = submissionQuestionResponseURL(
         contractRev.contract.id,
         config.baseUrl
     )
 
     const data = {
         packageName,
-        submissionURL: packageURL,
+        submissionURL: questionResponseURL,
```
JasonLin0991, line 53: Let's change this field to `questionResponseURL` and update the `eta` template to match the CMS version.

services/app-api/src/resolvers/questionResponse/createQuestion.ts
```diff
@@ -126,6 +128,40 @@ export function createQuestionResolver(
...
+            dateAsked,
+            roundNumber
+        )
+
+        if (sendQuestionsCMSEmailResult instanceof Error) {
+            logError(
+                'sendQuestionsCMSEmail - CMS email failed',
+                sendQuestionsCMSEmailResult
+            )
+            setErrorAttributesOnActiveSpan('CMS email failed', span)
```
JasonLin0991, line 157: Let's add the contract ID and Question ID to this email error message for debugging if this error was ever logged.

JasonLin0991 (approved):
🙌 Looks great! 

Some minor things we could touch up.

services/app-api/src/emailer/emails/sendQuestionCMSEmail.ts
pearl-truss: Would `config.dmcoEmails` go to all DMCO users? My understanding of that AC was to only send to the state analysts for the particular state the question is about

services/app-api/src/emailer/emails/sendQuestionCMSEmail.ts
macrael: I feel like these functions have gone back and forth on what they accept over time. In general I'm down for stuff like Question and Contract being passed directly and the emailer being responsible for pulling out what matters for the email, rather than the resolvers pulling that data out themselves. But I don't care that much either way. It used to be inconsistent between email handlers which is probably not ideal. I do think part of why this went this way was b/c of the complication of parsing protobuf stuff which we now hopefully are avoiding

pearl-truss:
@JasonLin0991 thanks for review! I made all the updates except the one to include the DMCO inbox. I'll reach out to design for clarity since I'm not sure if they want that or not

services/app-api/src/emailer/emails/__snapshots__/sendQuestionCMSEmail.test.ts.snap
```diff
@@ -0,0 +1,12 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`renders overall email for a new question as expected 1`] = `
+"DMCO sent questions to the state for submission MCR-MN-0003-SNBC<br />
+<b>Sent by:</b> Ronald McDonald (DMCO) <a href="cms@email.com">cms@email.com</a>
+<br />
+<b>Round:</b> 1<br />
+<b>Date:</b> 01/01/2024<br />
+<br />
+<br />
```
JasonLin0991, line 10: Oops forgot this one. There is an extra `<br />`.

JasonLin0991:
> @JasonLin0991 thanks for review! I made all the updates except the one to include the DMCO inbox. I'll reach out to design for clarity since I'm not sure if they want that or not

I just took a peek at the AWS parameter store configuration in PROD for division emails. `/configuration/email/dmcp` does not contain `DMCPrates@cms.hhs.gov`, just `MMCratesetting@cms.hhs.gov`. From this [dovetail](https://macbis.dovetailapp.com/data/Q-A-email-notification-DMCP-16hKz3dyA6uQIoLe43RVod) doc, `MMCratesetting@cms.hhs.gov` should not be getting these Q&A emails. `OACT` in the parameter store contains the correct email for Q&A. So, the current CMS receiver emails are incorrect.

I think the way to go here is to make distinct parameter stores for each division's Q&A email. Maybe something like  `/configuration/email/[divison]/questionAnswer/email`

pearl-truss:
> I just took a peek at the AWS parameter store configuration in PROD for division emails. `/configuration/email/dmcp` does not contain `DMCPrates@cms.hhs.gov`, just `MMCratesetting@cms.hhs.gov`. From this [dovetail](https://macbis.dovetailapp.com/data/Q-A-email-notification-DMCP-16hKz3dyA6uQIoLe43RVod) doc, `MMCratesetting@cms.hhs.gov` should not be getting these Q&A emails. `OACT` in the parameter store contains the correct email for Q&A. So, the current Q&A CMS receiver emails are incorrect.
> 
> I think the way to go here is to make distinct parameter stores for each division's Q&A email. Maybe something like `/configuration/email/[divison]/questionAnswer/email`

Great catch @JasonLin0991! Yeah I think it makes sense to make separate stores for Q&A and I can go ahead and do that

pearl-truss:
> @JasonLin0991 thanks for review! I made all the updates except the one to include the DMCO inbox. I'll reach out to design for clarity since I'm not sure if they want that or not

Got [confirmation](https://qmacbis.atlassian.net/browse/MCR-2539?focusedCommentId=126042) to not send emails to the general dmco inbox

pearl-truss:
> /configuration/email/[divison]/questionAnswer/email

@macrael do you have thoughts on if we should create new parameter stores for QA emails?

JasonLin0991 (approved):
Great work! Thanks for verifying recipients. 

We will need to add the new parameter stores `/configuration/email/dmcpReview` and `/configuration/email/dmcpSubmission` to AWS for dev to pass CI. 

I would start with the `DEV` environment, then add them to `Val` and `Prod` when you merge, but don't delete the old parameter `/configuration/email/dmcp` store until  promote finishes for Prod.

Also please make sure to update the docs `docs/Configuration.md`

services/app-api/src/emailer/emails/sendQuestionCMSEmail.ts
```diff
@@ -0,0 +1,77 @@
...
+        contractRev.contract.stateNumber,
+        contractRev.formData.programIDs,
+        packagePrograms
+    )
+
+    const questionResponseURL = submissionQuestionResponseURL(
+        contractRev.contract.id,
+        config.baseUrl
+    )
+    const roundNumber = question.responses.length + 1
```
pearl-truss, line 46: Screenshot 2023-12-05 at 11 31 00 AM

Based on this UI, it appears that a "round" is based on the number of questions and not number of responses. It looks like `HealthPlanPackageType` has a `questions` field, should a contract have the same?

pearl-truss, line 46: Screenshot 2023-12-05 at 11 31 00 AM

Based on this UI, it appears that a "round" is based on the number of questions and not number of responses. It looks like `HealthPlanPackageType` has a `questions` field, should a contract have the same? cc @macrael @JasonLin0991

JasonLin0991: Good catch; rounds are based on the number of questions, not the responses. 

As for the `questions` field, I think eventually the contract will have the same, but since the frontend is still using strictly `HealthPlanPackageType`, we won't need to add it yet. What do you think @macrael?

pearl-truss: Okay cool, and I see there's an existing postgres function to get all questions for a contract that I can use. So I agree with nothing needing to change yet

PR #2101, opened 2023-12-07
Remove a bunch of Feature Flags

macrael:
## Summary

WIP Removing a raft of feature flags


services/app-web/src/components/FileUpload/FileProcessor/FileProcessor.tsx
```diff
@@ -223,7 +223,6 @@ export const FileProcessor = ({
             hasRecoverableError={hasRecoverableError}
             handleDelete={handleDelete}
             handleRetry={handleRetry}
```
haworku, line 225: Are we removing the table mode fully or not? If we are this is fine but then I would expect everything associated with the table `renderMode` on the top level fileupload to go. Otherwise I would think this checkbox stuff stays and just the feature flag is removed from the code consuming this component.

JasonLin0991 (approved):
Looks good. Just a couple things from me and the comment about `renderMode` from @haworku

JasonLin0991:
One more I missed is in `cypress/support/stateSubmissionFormCommands.ts`. The command `waitForDocumentsToLoad` has a switch for `tableView`. We probably won't need that anymore.

JasonLin0991 (commented):
One more I missed is in `cypress/support/stateSubmissionFormCommands.ts`. The command `waitForDocumentsToLoad` has a switch for `tableView`. We probably won't need that anymore.

services/app-web/src/components/FileUpload/FileProcessor/FileProcessor.tsx
macrael: I forgot to, I'll add it in my pr response commit

services/app-web/src/common-code/healthPlanFormDataType/healthPlanFormData.test.ts
```diff
@@ -797,261 +794,4 @@ describe('submission type assertions', () => {
...
-                    },
-                ],
-                rateInfos: [],
-            },
-        },
-    ]
-    test.each(contractOnlyWithValidRateData)(
-        'Remove rates data on CONTRACT_ONLY submission: $testDescription',
-        ({ submission, expectedResult }) => {
-            expect(removeRatesData(submission)).toEqual(
```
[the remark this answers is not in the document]
macrael: these tests only covered getting document categories right which I believe is now obsolete. I'm going to remove document categories all together in another PR

PR #2102, opened 2023-12-07
MCR-2541: CMS notified when a state uploads response

JasonLin0991:
## Summary
[MCR-2541](https://qmacbis.atlassian.net/browse/MCR-2541)

Send CMS email notifications for new responses to questions.

- Add email notification documentation on recipients for each notification
- API Work
   - `sendQuestionResponseCMSEmail` emailer function and tests
   - `getQuestionRound` helper function and tests.
   - Refactored Q&A Postgres functions to reduce duplicate code
   - The `insertQuestionResponse` Postgres function uses nested create on the `question` table to create the response instead of creating it from the `questionResponse` table and then returning the complete question data with responses. This serves a couple of purposes:
      - One less query to get question data that's needed in the resolver
      - The return type matches with the create question handler, so the front only has to deal with 2 types, `Questions` and `IndexQuestionsPayload.`
   - Add sending email notifications in `createQuestionResponse` resolver
   - Add `emailerHelpers` helper function `mockQuestionAndResponses` to generate test Q&A data.
- GraphQL Work
   - Update `CreateQuestionResponsePayload` schema type in `schema.graphql`
   - Update `createQuestionResponse` mutation return data
- Frontend work
   - Update `createResponseWrapper` manual cache update for the `graphql.schema` change.
   - Add a comment explaining why we are manually updating the Apollo cache for Q&A mutations

#### Test cases covered

- `sendQuestionResponseCMSEmail.test.ts`
   - `'OACT Q&A response email contains correct recipients'`
   - `'DMCP Q&A response email contains correct recipients'`
   - `'DMCO Q&A response email contains correct recipients'`
   - `'renders overall CMS email for a new state response as expected'`

- `createQuestionResponse.test.ts`
   - `'sends CMS email'`

- `templateHelpers.test.ts`
   - `getQuestionRound`


JasonLin0991:
There is an outstanding question in [Jira](https://qmacbis.atlassian.net/browse/MCR-2541?focusedCommentId=126127) for an edge case where the state user is in the process of submitting a response when a CMS user unlocks the submission. This would mean the response is being created on an unlocked submission. In the UI, the response will be complete, and the state user will be taken to the unlocked submission at the Submission type page of the form.

haworku (approved):
I'm approving this since everything other than the unlocked issue looks good. Ping me directly in slack if you want re-review or another round of QA

PR #2105, opened 2023-12-09
Remove document categories

macrael:
## Summary

This PR cleans up the document list component and goes further to remove documentCategories from the entire application since they are no longer needed now that we have removed the rate-supporting-documents flag.


services/app-web/src/components/FileUpload/FileProcessor/FileProcessor.tsx
```diff
@@ -97,32 +82,19 @@ type FileProcessorProps = {
...
     retryItem,
-    renderMode,
-    handleCheckboxClick,
-    isContractOnly,
-    shouldValidate,
 }: FileProcessorProps): React.ReactElement => {
     const { name, status, file } = item
-    const isRateSupporting = item.documentCategories.includes('RATES_RELATED')
-    const isContractSupporting =
-        item.documentCategories.includes('CONTRACT_RELATED')
```
haworku, line 117: This is good. Thank you for looking carefully at this file.  I was pretty frustrated when we had to add this because we were leaking all this domain data into what was supposed to be a multipurpose re-useable components (e.g. it could be user avatar image upload, it could be any random upload). Now removing this we are back to a truly re-useable domain agnostic component

haworku (approved):
Would like to see Cypress pass full but looks good.

PR #2114, opened 2023-12-13
MCR-2642: Document resolver data type interfaces

JasonLin0991:
## Summary
[MCR-2642](https://qmacbis.atlassian.net/browse/MCR-2642)

Through thinking about the data flow from Postgres to the resolvers, I found it was just a tiny piece of an overall design for our resolvers. I ended up adding more about the resolver and some diagrams
- Overview about resolvers
- Diagram of the resolver design
- Filled out the section about `Postgres Database Operations`


docs/technical-design/resovler-design.md
```diff
@@ -0,0 +1,146 @@
...
+- Third party API calls
+   - SES Email
+   - LaunchDarkly Feature Flags
+- Error handling
+
+The resolvers often do not directly contain the code to perform these tasks, instead they call functions specific for each task. Importantly we use types defined in our [domain models](design-patterns.md#domain-models) as the internal communication interface for different services as an attempt to decrease the coupling of our code.
+
+Errors from these tasks propagate up to the resolver where it will be handled and a response sent to the client. See docs about [error handling](error-handling.md) for details.
+
+## Diagram
```
macrael, line 18: I think this diagram could be made simpler. You are getting at the heart of things but it might be too complex for mermaid

My notes:
* With the arrows coming in at the center of the box it's unclear that the top level resolvers are represented by the big box on the left. 
* I think it would be cool to show the individual modules as being in separate boxes with the dependencies they wrap, rather than all of our wrappers being in a box together
* Right now, our handlers are an exception where we are directly not wrapping the interface with Apollo Server, but that is another dependency we have that probably should be called out. The inputs / return values (including the thrown exceptions) from our handlers are in Apollo Server's language.
* maybe call the "hppToDomainType" stuff our "proto" module because those functions are intentionally boxed up in there to wrap that dependency. 

We can do some sketching together if you like

macrael (approved):
I love this, great to see all the concrete examples of wrapped dependencies

haworku (commented):
f its easy can you make the text labeling the bolded boxes more readable? Right now because they are overlapping I cannot easily read the labels on the bold box
Screenshot 2023-12-14 at 4 27 59 PM

docs/technical-design/resovler-design.md
```diff
@@ -0,0 +1,146 @@
+# Resolver design
+
+## Background
+Resolver functions are responsible for handling requests made to the MC-Review GraphQL API and populating data for the response. Between receiving the request data and returning a response, the resolvers perform various tasks for generating the response.
```
haworku, line 4:
```suggestion
Resolver functions are responsible for handling requests made to the MC-Review GraphQL API and populating data for the response on the backend. Between receiving the request data and returning a response, the `app-api` resolvers perform various tasks for generating the response.
```

To me its always good to clarify backend or frontend when we dealing with something that crosses the boundary. Graphql as our api protocol has helpers/ data transformation/ utilities on both sides of the stack.

docs/technical-design/resovler-design.md
```diff
@@ -0,0 +1,146 @@
...
+
+These tasks include:
+- Business logic, data transformation, and validations
+- Database operations
+- Third party API calls
+   - SES Email
+   - LaunchDarkly Feature Flags
+- Error handling
+
+The resolvers often do not directly contain the code to perform these tasks, instead they call functions specific for each task. Importantly we use types defined in our [domain models](design-patterns.md#domain-models) as the internal communication interface for different services as an attempt to decrease the coupling of our code.
```
haworku, line 14:
```suggestion
The resolvers often do not directly contain the code to perform these tasks, instead they call functions specific for each task. Importantly we use types defined in our [domain models](design-patterns.md#domain-models) as the internal communication interface on the backend for different services as an attempt to decrease the coupling of our code.
```

docs/technical-design/resovler-design.md
haworku: I agree with this in terms of maybe too complex. Def recommend taking more time here or working with others. This is important but also challenging task!

It if is difficult to know how to adjust the diagram -- I  recommend running by each member of the team (especially @pearl-truss as the newest team member and least indoctrinated into how we all talk about the code) and ask them what they notice or what they take way from it.

docs/technical-design/resovler-design.md
```diff
@@ -0,0 +1,146 @@
...
+
+    callInsertContract -- Domain Model --> insertContractHandler
+    parseContractWithHistory -- Domain Model OR Error --> callInsertContract
+
+    prismaOrm -- Operation --> dbContainer
+    dbContainer -- Data --> prismaOrm
+
+    style resolver stroke-width:4px,fill:#1f1f1f;
+    style postgresHandlers stroke-width:4px,fill:#1f1f1f;
+```
```
haworku, line 129: Really like the granularity of this diagram. Feels right sized. I can digest it. Also like that its showing something that is a bit hard to see in the code without opening multiple files.

docs/technical-design/resovler-design.md
JasonLin0991: > might be too complex for mermaid

This diagram was a pain and was too complex for a mermaid. I'm thinking of just sketching this on the Miro board, adding a screenshot and linking to the docs.

> I recommend running by each member of the team (especially @pearl-truss as the newest team member and least indoctrinated into how we all talk about the code) and ask them what they notice or what they take way from it.

This is a great idea. I will fix the diagram, and maybe @pearl-truss can give me feedback either here or a Zoom meeting when she is back from OOO.

docs/technical-design/resovler-design.md
```diff
@@ -0,0 +1,146 @@
...
+## General Guidance
+### Postgres Database Operations
+[Postgres](https://www.postgresql.org/docs/) database operations are contained in functions that we call **postgres handlers**. You can find all the handlers in [`services/app-api/src/postgres/`](../../services/app-api/src/postgres). In these handler functions the database operations are performed using [Prisma ORM](https://www.prisma.io/docs/orm) which itself uses Prisma generated types and returns data from the database as these types. So in most of the handlers, data from the operation must be converted to a domain model type before returning to the resolver and errors occurring in the handler functions should be returned to the resolver to handle.
+
+It's important that data from the DB is converted to the domain model before returning to the resolver to adhere to our strategy of using the domain model as the internal communication protocol for different parts of our app. 
+
+Resolvers are passed Postgres handlers via [dependency injection](design-patterns.md#dependency-injection). The configuration for the Postgres handler dependency and many others are done in [apollo_gql.ts](../../services/app-api/src/handlers).
```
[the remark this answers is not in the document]
JasonLin0991: > I'd call out that dependencies are also wired up in every single test with constructTestPostgresServer

I put this in the `Background` section.

docs/technical-design/resovler-design.md
JasonLin0991: > Right now, our handlers are an exception where we are directly not wrapping the interface with Apollo Server, but that is another dependency we have that probably should be called out. The inputs / return values (including the thrown exceptions) from our handlers are in Apollo Server's language.

Could you explain this more?

.images/resolver-design-diagram.png
pearl-truss, line 1: Question: Overall this diagram makes sense and is presented well. The part I have a question about is, in the "Resolver" box, what's the significance of having the "Error handling" and "validation" boxes nested inside? Is there an order of operations reason for the resolver -> error handling -> validation structure? If it's just because resolvers handle the error handling and validation, it might be simpler to leave out these nesting boxes (since it's mentioned in the docs) but that's just my opinion

docs/technical-design/resovler-design.md
```diff
@@ -0,0 +1,56 @@
...
+
+The diagram below is the data flow diagram for `createHealthPlanPackage` resolver.
+
+[Miro link](https://miro.com/app/board/o9J_lS5oLDk=/?moveToWidget=3458764573517610448&cot=14)
+![createHealthPlanPackage-diagram](../../.images/createHealthPlanPackage-diagram.png)
+
+Form the diagram above, you can see that `createHealthPlanPackage` resolver calls `insertDraftContract` Postgres handler function to create a new draft contract. Within the `insertDraftContract` after performing the database operation using Prisma we pass the Prisma model data into `parseContractWithHistory`, which is a function that coverts our Prisma model to domain model before returning data to the resolver.
+
+### Data transformation
```
pearl-truss, line 42: Are these intentionally left blank for now?

pearl-truss:
Great job Jason! I just left a few comments, mostly about the diagrams

docs/technical-design/resovler-design.md
macrael: > Could you explain this more?

So the general rule that this diagram is demonstrating is that all of our 3rd party code dependencies are getting wrapped in a module of our own design. That way we control the interface (using domain models) and contain the dependency's ways of doing things to the one module. The one exception to that is Apollo Server. That's a 3rd party dependency that we have not wrapped up in our own module. We have instead implemented our core business logic directly in Apollo's handlers. So, for instance, that's the only place in our code where we are throwing errors instead of returning them because that is how Apollo handles errors. 

I'm not sure exactly how to get that into this diagram but since this is showing off all our dependencies I think it would be good to call this out as well

docs/technical-design/resovler-design.md
JasonLin0991: Yes. I left them blank for now. I think I'll tackle them in a separate ticket.

.images/createHealthPlanPackage-diagram.png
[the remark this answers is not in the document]
JasonLin0991: I don't think so because we pass the data from `Postgres` -> `Prisma ORM` -> `parseContractWithHistory` to parse and convert into domain model type. The data is one-way, and nothing is passed back to `Prisma ORM` after parsing. It is returned to the resolver at that point.

I think the diagram is confusing. What do you think of this update?

Screenshot 2023-12-18 at 3 25 18 PM

docs/technical-design/resovler-design.md
JasonLin0991: Lets pair tomorrow at our 1:1 on this?

.images/createHealthPlanPackage-diagram.png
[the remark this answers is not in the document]
pearl-truss: @JasonLin0991 I think this is much clearer! Thanks for making the updates

JasonLin0991:
> Great job Jason! I just left a few comments, mainly about the diagrams

Just made some more changes. Please take a look and let me know what you think.

.images/resolver-design-diagram.png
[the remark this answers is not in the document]
JasonLin0991: > Is both encoding and decoding happening unidirectionally from the proto module to the protobuf?

Yes. This part may need to be fleshed out in its section, but the plan is to remove protobuf from the app altogether.

> And outside of that this LGTM! I'm going to go ahead and approve

Thanks for the great feedback!

PR #2138, opened 2023-12-21
Add createAPIToken endpoint and jwtLib service

macrael:
## Summary

This is step one for adding a new page that can generate JWTs. 

I created a new GraphQL mutation for creating a new API Token. 

I created a new service module called `jwt` that wraps what seemed like overwhelmingly the most popular JWT JS library. 

@pearl-truss I think you will want to use the JWTLib I added here to do the validation of the JWT coming in to the lambda authorizer. Feel free to modify it at will, I made the API just return a user ID from the token, that might not be all we need? I think it's got what I need for the creation page is what I was focused on. 

#### Related issues

https://qmacbis.atlassian.net/browse/MCR-3833

### DIFF
--- a/docs/technical-design/api-jwt-security.md
+++ b/docs/technical-design/api-jwt-security.md
@@ -15,4 +15,4 @@ The JWT will have 3 claims in it:
 
 These are all standard fields and are explicitly accounted for in the JWT library we are using
 
-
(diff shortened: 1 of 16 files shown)


pearl-truss (approved):
This LGTM and so far with what I have and what I've seen online, just the user ID is enough. The lambda authorizer will return it in `APIGatewayAuthorizerResult`

PR #2146, opened 2023-12-27
MCR-3834 : API Auth: Create Lambda Authorizer to validate incoming JWTs

pearl-truss:
## Summary

This PR creates the lambda authorizer that will be used for 3rd party API access. It checks that the JWT is valid with `jwtLib.userIDFromToken(authToken)` which calls `verify` from the `jsonwebtoken` library which will return an error if the JWT is expired 

`jwthealth` is a copy of the `health_check` endpoint with the authorizer attached to temporarily use for testing the authorizer 

#### Related issues

https://qmacbis.atlassian.net/browse/MCR-3834


services/app-api/serverless.yml
```diff
@@ -160,6 +160,21 @@ functions:
...
+      handler: src/handlers/health_check.main
+      events:
+        - http:
+            path: health_check
+            method: get
+            cors: true
+            authorizer:
+              name: third_party_api_authorizer
+              identitySource: method.request.header.Authorization 
+              resultTtlInSeconds: 3600 # cache the result for 1 hour
```
pearl-truss, line 176: This is an optional field, not sure if we actually want to cache results now

services/app-api/src/handlers/third_party_API_authorizer.ts
```diff
@@ -0,0 +1,59 @@
+import type {
+    APIGatewayAuthorizerResult,
+    PolicyDocument,
+    Handler,
+} from 'aws-lambda'
+import { newJWTLib } from '../jwt'
+import { AuthenticationError } from 'apollo-server-lambda'
+
+// Hard coding this for now, next job is to run this config to this app.
```
pearl-truss, line 9: This is a duplication from what was temporarily added in `apollo_gql.ts`. But necessary here to call its `userIDFromToken` method

macrael (approved):
Nice job hooking all this up! It's cool how I could test this with a `curl`

I'd like you to change to returning the expected return values instead of throwing here but otherwise this looks good. 

I'm trying to write a little test harness for this so we can be sure of what we're returning on different error cases but it's taking longer to setup than I hoped. I'll give it to you when I've got it working.

services/app-api/serverless.yml
pearl-truss: Yeah it seems like it's based on the authorizer token/identity source. I came across [this](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html#http-api-lambda-authorizer.caching)
https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html#http-api-lambda-authorizer.caching

I don't fully understand it so I'll remove but if we want to add it back in we can

pearl-truss:
@macrael thanks for the review! I believe I caught everything in the latest push

services/app-api/src/handlers/third_party_API_authorizer.ts
```diff
@@ -0,0 +1,68 @@
...
+
+        console.info({
+            message: 'third_party_API_authorizer succeeded',
+            operation: 'third_party_API_authorizer',
+            status: 'SUCCESS',
+        })
+
+        return generatePolicy(userId, event)
+    } catch (err) {
+        console.error('Invalid auth token. err => ', err)
```
macrael, line 39: Can you change this error message to be more like "unexpected exception" or something? an invalid token should just be handed with the error on line 24, I think if we are catching here it's because something very weird happened.

services/app-api/src/handlers/third_party_API_authorizer.ts
```diff
@@ -0,0 +1,68 @@
...
+): APIGatewayAuthorizerResult {
+    // If the JWT is verified as valid, send an Allow policy
+    // this will allow the request to go through
+    // otherwise a Deny policy is returned which restricts access
+    const policyDocument: PolicyDocument = {
+        Version: '2012-10-17', // current version of the policy language
+        Statement: [
+            {
+                Action: 'execute-api:Invoke',
+                Effect: userId ? 'Allow' : 'Deny',
```
macrael, line 56: Rather than relying on the truthiness of empty string, how about you make userID `string | undefined`? I like the explicitness of undefined for control flow.

macrael (approved):
Nice! looks great

PR #2152, opened 2024-01-05
MCR-3835: Attach lambda authorizer to api gateway endpoint

pearl-truss:
## Summary

New `graphql_external` endpoint created. This is a copy of the `graphql` endpoint with the following changes:
- ~~Only has a get method. I believe we said we would limit 3rd parties to queries for now.~~ I added in the post method because of the language in https://qmacbis.atlassian.net/browse/MCR-3836 seems like it's something we will eventually implement
- URL path to invoke is `v1/graphql/external` per eng notes on the ticket
- 3rd party authorizer is attached

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-3835

## QA guidance

I tested locally with curl

```
curl \
  --request GET \
  --header "Accept: text/javascript" \
  --header "Authorization: Bearer {token}" \
  --data '{"query": "query { indexHealthPlanPackage { totalCount edges{} } }"}' \
http://127.0.0.1:3030/local/v1/graphql/external
```

**With a valid token an apollo error is returned**

`Context creation failed: Log: placing user in gql context failed","    at ApolloServer.context`. 

I believe this error will resolve after https://qmacbis.atlassian.net/browse/MCR-3836 is completed

**With an invalid token a permission error is returned**

`"error":"Forbidden","message":"No principalId set on the Response"`

### DIFF
--- a/services/app-api/serverless.yml
+++ b/services/app-api/serverless.yml
@@ -197,18 +197,6 @@ functions:
           method: get
           cors: true
           authorizer: aws_iam
-    timeout: 60 # aurora cold start can be long
-    vpc:
-      securityGroupIds:
-        - ${self:custom.sgId}
-      subnetIds: ${self:custom.privateSubnets}
-    layers:
-      - !Ref PrismaClientEngineLambdaLayer
-      - arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-nodejs-amd64-ver-1-18-1:1
-
-  graphql_external:
-    handler: src/handlers/apollo_gql.graphqlHandler
-    events:
       - http:
           path: v1/graphql/external
           method: post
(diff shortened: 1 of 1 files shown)


services/app-api/serverless.yml
```diff
@@ -217,6 +206,31 @@ functions:
       - !Ref PrismaClientEngineLambdaLayer
       - arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-nodejs-amd64-ver-1-18-1:1
 
+  graphql_external:
+    handler: src/handlers/apollo_gql.graphqlHandler
+    events:
+      - http:
```
macrael, line 212: Could these events be inside the regular graphql lambda or do they need to be their own lambda?

macrael (approved):
I didn't think about this when doing the separate jwthealth lambda, but if we can just add these as separate events using a different authorizer to our existing lambda rather than having to clone the whole lambda that would be nice

services/app-api/serverless.yml
pearl-truss: I confirmed it works with combining with the regular graphql lambda. I think this falls under "[lambda proxy integration](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html)" which is the default used with [serverless](https://www.serverless.com/framework/docs/providers/aws/events/apigateway#lambda-proxy-integration)

PR #2155, opened 2024-01-08
MCR-2641: Revision history diagram

JasonLin0991:
## Summary
[MCR-2641](https://qmacbis.atlassian.net/browse/MCR-2641)

- Added diagram of the revision history for the new data model.
- Added diagram for examples of what the high-level view of the data looks like when fetching.
- Most of the detailed implementation docs where existing.


haworku:
I really like that the columns have circles around them in this version.
 
I did notice I was initially  confused about the `--draft revision--` text because its used in places I wouldn't expect. Here's an [example](https://miro.com/app/board/o9J_lS5oLDk=/?moveToWidget=3458764574966677737&cot=14) of what I would expect -  where draft revision is only included in cells for the data model that has new draft already associated.  I put "unchanged" for places where nothing is shifting with the history. 

🤷‍♀️ Is this any different? Not sure. I am just sharing what my initial confusion was. I figured it out quickly. Would be happy to ship this as is.

.images/contract-and-rate-history-diagram.png
pearl-truss, line 1: This looks great! Easy to follow and understand

.images/read-contract-and-rate-history-diagram.png
pearl-truss, line 1: Thoughts: I think the orange and blue arrows and lines coming off the contract history column are making this diagram harder to read, at least initially. I think the sub headings under Example 1 and Example 2 are clear enough without the lines what part of the history each example covers

Nit: the timeline line is going straight through the dates making it a bit harder to read. Could it be updated to look more like the line used in the other diagram?

pearl-truss (approved):
This looks great @JasonLin0991, I left a few minor comments but nothing blocking

JasonLin0991:
> I really like that the columns have circles around them in this version.
> 
> I did notice I was initially confused about the `--draft revision--` text because its used in places I wouldn't expect. Here's an [example](https://miro.com/app/board/o9J_lS5oLDk=/?moveToWidget=3458764574966677737&cot=14) of what I would expect - where draft revision is only included in cells for the data model that has new draft already associated. I put "unchanged" for places where nothing is shifting with the history.
> 
> 🤷‍♀️ Is this any different? Not sure. I am just sharing what my initial confusion was. I figured it out quickly. Would be happy to ship this as is.

That's a good call. What about removing `--draft revision--` from the historical revision columns altogether and putting `--no revision--` instead? That way, it clearly shows that drafts do not belong in history.

[Miro link](https://miro.com/app/board/o9J_lS5oLDk=/?moveToWidget=3458764574799194467&cot=14)

.images/contract-and-rate-history-diagram.png
macrael: The "no revision" and "draft revision" lines are a bit confusing to me. for instance under the 1/10/2024 unlock there's a draft for rate A but no draft for Contract or Rate B so I don't get why all three say "Draft revision"

It might make things more clear to put the draft revisions in line with the submitted revisions rather than being in their own column? Right the draft column and the other three columns all alternate wether they have information in them.

.images/read-contract-and-rate-history-diagram.png
JasonLin0991: > Nit: the timeline line is going straight through the dates making it a bit harder to read. Could it be updated to look more like the line used in the other diagram?

👍

> Thoughts: I think the orange and blue arrows and lines coming off the contract history column are making this diagram harder to read, at least initially. I think the sub headings under Example 1 and Example 2 are clear enough without the lines what part of the history each example covers

Yeah, I agree it makes it harder to read, but the tradeoff here is that someone with no knowledge of how the data is structured will know how to read the diagram. Let's keep it for now, and we can make more changes in the future if it becomes more of a hindrance.

[Miro link](https://miro.com/app/board/o9J_lS5oLDk=/?moveToWidget=3458764574799194467&cot=14)

.images/contract-and-rate-history-diagram.png
JasonLin0991: Like this [miro link](https://miro.com/app/board/o9J_lS5oLDk=/?moveToWidget=3458764574794469060&cot=14)?

PR #2159, opened 2024-01-08
MCR-3849 Fix: check auth for session expiration after a failed file upload

haworku:
## Summary
Automatic logout when file upload fails due to session expiration

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-3849

#### Test cases covered

This requires manual testing right now.

## QA guidance

**Steps to Reproduce**
- Sign into MC-Review _with the Toph CMS user_ 
- Leave the tab open on a form page with file upload. 
- Let tab and browser go inactive for 45 min
- Return to the MC-Review tab
- Immediately try to add a file on that same page 
- File upload fails
- App logs out and kicks you back to homepage (or auth page if its a review app) with session alert telling you what happened

**Other things to test**
- make sure logout is still working in general and session expiration modal still shows when app is active

### DIFF
--- a/services/app-web/src/components/FileUpload/FileUpload.tsx
+++ b/services/app-web/src/components/FileUpload/FileUpload.tsx
@@ -180,8 +180,8 @@ export const FileUpload = ({
     }
     // Upload to S3 and update file items in component state with the async loading status
     // This includes moving from pending/loading UI to display success or errors
-    const asyncS3Upload = (files: File[] | File) => {
-        const upload = async (file: File) => {
+    const handleS3Upload = (files: File[] | File) => {
+        const asyncUploadAndScan = async (file: File) => {
             const sha = (await calculateSHA256(file)) || ''
             uploadFile(file)
                 .then((data) => {
@@ -291,10 +291,10 @@ export const FileUpload = ({
 
         if (!(files instanceof File)) {
             files.forEach((file) => {
-                upload(file).catch((e) => console.error(e))
+                asyncUploadAndScan(file).catch((e) => console.error(e))
             })
         } else {
-            upload(files).catch((e) => console.error(e))
+            asyncUploadAndScan(files).catch((e) => console.error(e))
         }
     }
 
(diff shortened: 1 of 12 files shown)


haworku:
What I'm seeing so far on the review apps is 1. [the original bug](https://qmacbis.atlassian.net/browse/MCR-3849) not happening 2.  the forced session timeout not happening either even when I leave the computer asleep overnight and coming back. It looks like the session is always active, which is off. 

 I am going to check if the DEV environments have different cognito settings than prod.

haworku:
I _think_ I fixed this. It's hard to see in  review app which behaves so differently from IDM based login. Can folks test this. At least if we are sure it doesn't break anything I think it's worth merging through to test on dev itself.

services/app-web/src/components/FileUpload/FileUpload.tsx
```diff
@@ -270,13 +272,28 @@ export const FileUpload = ({
...
+                        `UPLOAD_ERROR: ${JSON.stringify(
+                            _e
+                        )}. Files added: ${JSON.stringify(files)}`
+                    )
+                    recordJSException(error)
+
+                    // file upload failing could be due to session timeout that did not hit the modal
+                    // double check the user still has their session, if not, logout to update the React state with their login status
+                    try {
+                        await checkAuth()
```
JasonLin0991, line 286: I'm still trying to test this out in the app,  but I was curious about why we check the session after a failed upload instead of checking before starting the upload.

haworku: @JasonLin0991 To avoid delaying the start of file upload with another request is the reason I had.

haworku:
Update here - I'm going to make further code changes, probably next week since is low priority compared to feature work. 

 Last update is Jason wasn't able to verify its working but also wasn't able to replicate the original bug where the bad file upload happens... the modal kept popping up first. That is **not** the expected experience with my change

haworku:
closing to free up resources

PR #2165, opened 2024-01-10
MCR-3813: Made rate summary route available to state user

ruizajtruss:
## Summary
Added a new route for state users to be able to view rate summary's when accessing the `/rates/:id` url, the dashboard link has also been changed to route to the standard dashboard for state users but still routes to the rates dashboard for CMS users.

#### Related issues
[MCR-3813](https://qmacbis.atlassian.net/browse/MCR-3813)

#### Screenshots
image

#### Test cases covered

Wrote two unit tests:

- State user has access to `/rates/:id`
- Expected error page is rendered when rate ID is invalid

## QA guidance

visit the `/rates/:id` url as a state user using a valid rate id

### DIFF
--- a/services/app-web/src/pages/App/AppRoutes.tsx
+++ b/services/app-web/src/pages/App/AppRoutes.tsx
@@ -92,7 +92,6 @@ const StateUserRoutes = ({
                     path={RoutesRecord.DASHBOARD_SUBMISSIONS}
                     element={<StateDashboard />}
                 />
-
                 <Route
                     path={RoutesRecord.SUBMISSIONS}
                     element={<StateDashboard />}
@@ -100,6 +99,10 @@ const StateUserRoutes = ({
                 <Route
                     path={RoutesRecord.SUBMISSIONS_NEW}
                     element={<NewStateSubmissionForm />}
+                />
+                <Route
+                    path={RoutesRecord.RATES_SUMMARY}
+                    element={<RateSummary />}
                 />
                 <Route element={<SubmissionSideNav />}>
                     {showQuestionResponse && (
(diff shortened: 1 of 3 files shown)


macrael (approved):
First Post!

services/app-web/src/pages/App/AppRoutes.tsx
```diff
@@ -101,6 +100,10 @@ const StateUserRoutes = ({
                     path={RoutesRecord.SUBMISSIONS_NEW}
                     element={<NewStateSubmissionForm />}
                 />
+                <Route
```
haworku, line 103: Please feature flag this new route for state users. There is an edit unlock flag. Thank you!

JasonLin0991:
We should check in with design on if `Back to dashboard` link:
- should it take the state user back to the contract dashboard?
- should it be removed?

Screenshot 2024-01-16 at 12 38 25 PM

macrael:
do we have another ticket to make the rate dashboard available to state users?

ruizajtruss:
> do we have another ticket to make the rate dashboard available to state users?

There is a ticket to make that dashboard available in a later epic from what I understand.
I figure since it's all tucked behind feature flags and the dashboard will come up down the line it's probably fine for now? Unless we want to have that button redirect differently in the interim

services/app-web/src/pages/App/AppRoutes.tsx
```diff
@@ -101,6 +109,12 @@ const StateUserRoutes = ({
                     path={RoutesRecord.SUBMISSIONS_NEW}
                     element={<NewStateSubmissionForm />}
                 />
+                {showRateSummaryPage && (
+                    <Route
+                        path={RoutesRecord.RATES_SUMMARY}
+                        element={<RateSummary loggedInUser={loggedInUser}/>}
```
JasonLin0991, line 115: We could also use `useAuth` hook in the `RateSummary` page component instead. Following the same pattern of not passing in props in the routes.

JasonLin0991 (approved):

services/app-web/src/pages/App/AppRoutes.tsx
ruizajtruss: Hadn't spotted that, I'll make the change, feels a lot better than passing props

PR #2166, opened 2024-01-10
MCR-3767 CMS has unlock rate button

haworku:
## Summary
Add rate unlock button

Reminder: no banners, no modal yet. All that happens is that when CMS user clicks the button, we hit the `unlockRate`. The button starts showing up disabled while the rate is unlocked.

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-3767

#### Screenshots
Screenshot 2024-01-11 at 8 07 20 PM

- **(from design review)**  We talked about how the rate name was was not breaking properly. There's empty space to the right of the first line. Looks like css `break-word` or overflow does not work as expected because our name also includes numbers. I found a solution using HTML entities in [cf47ef1](https://github.com/Enterprise-CMCS/managed-care-review/pull/2166/commits/cf47ef17c6ffcc513d7a7a805bc10c4d103f59b4) but pulled it off this PR to refine it more and ship as it's own work. FYI @kelseylistrom still working on this. 

#### Test cases covered
`app-web`
- renders the unlock button to CMS users
- disables the unlock button for CMS users when rate already unlocked
- does not render the unlock button to state users

`app-api`
- changes rate status to UNLOCKED and creates a new draft revision with unlock info
- returns status error if rate is actively being edited in draft
- 'returns unauthorized error for state user

## QA guidance
- Confirm feature flag on/off state
- Rate should unlock properly

### DIFF
--- a/services/app-graphql/src/schema.graphql
+++ b/services/app-graphql/src/schema.graphql
@@ -67,7 +67,7 @@ type Query {
 
     Errors:
     - ForbiddenError: User must be a CMS or Admin type user
-    - NotFoundError: No rates with at least one submitted revision were found 
+    - NotFoundError: No rates with at least one submitted revision were found
     """
     indexRates: IndexRatesPayload!
     """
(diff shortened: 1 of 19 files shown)


services/app-graphql/src/schema.graphql
```diff
@@ -1131,6 +1157,13 @@ type Rate {
     This value is used to generate the rateName
     """
     stateNumber: Int!
+     """
+    Information about who, when, and why this revision was unlocked.
+    Will be blank on the initial revision.
+    """
+    unlockInfo: UpdateInformation
+    "Information on who, when, and why this revision was submitted."
+    submitInfo: UpdateInformation
```
JasonLin0991, line 1166: I don't think we want `unlockInfo` and `submitInfo` at the `Rate` level. Was there something new you were trying to do here? Otherwise the `RateRevision` already has both.

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/SingleRateSummarySection.tsx
```diff
@@ -182,23 +197,60 @@ export const SingleRateSummarySection = ({
...
+                variables: {
+                    input: {
+                        rateID: rate.id,
+                        unlockedReason: '',
+                    },
+                },
+            })
+
+            if (data?.unlockRate.rate) {
+                // navigate('') TODO
```
JasonLin0991, line 214: nit: Is there a redirect here? From the AC, there is no mention of one, and if this follows the same pattern as contracts, it does not redirect the CMS user after unlocking. It might be confusing to leave a comment here suggesting a TODO for navigating elsewhere.

JasonLin0991 (approved):

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/SingleRateSummarySection.tsx
haworku: Ah yes, I'll remove this!

PR #2167, opened 2024-01-11
MCR-3879: Refine matching rate revision to contract revision

JasonLin0991:
## Summary
[MCR-3879](https://qmacbis.atlassian.net/browse/MCR-3879)

This change is to allow `contractWithHistoryToDomainModel` to still find the correct rate revision with the [Edit & Unlock Rate](https://qmacbis.atlassian.net/browse/MCR-3777) feature work. 

- Update `contractWithHistoryToDomainModel` to find the rate revision for the contract revision when rates can be independently submitted and unlocked.
- Update `rateWithHistoryToDomainModel` to find the contract revision for the rate revision when rates can be independently submitted and unlocked.
- I decided not to pull out the part of `contractWithHistoryToDomainModel` that finds the rate revision into its own function.
   - Unit testing this function with mock data would have been difficult. Since it is embedded into the `findContractWithHistory` and `findRateWithHistory` hanlders, tests for matching revisions are done there.
- Updated docs. 
- Fix `submitRate` and `unlockRate` so `findFirst` starts in descending order.

#### Test cases covered

`findContractWithHistory.test.ts`
- `'matches correct rate revisions to contract revision with independent rate unlocks and submits'`

`findRateWithHistory.test.ts`
- `'matches contract revision to rate revision with independent rate submit and unlocks'`

## QA guidance

Some existing API tests do test for the correct rate revision, and they all passed, so I think the change will result will be the same, just expanded to accommodate independent rate unlock/resubmit. Just to make sure, I would manually unlock and resubmit submissions with rate changes and make sure latest submission and previous submissions display the correct rate data.


macrael (approved):
I think at least a single test showing this behavior has changed would be worthwhile. To test my understanding, if you submit a contract with rates, then unlock a rate and resubmit it, we would expect with the old behavior for findContractWithRates to return the out of date rate before, and to return the new rate now?

JasonLin0991:
> I think at least a single test showing this behavior has changed would be worthwhile. To test my understanding, if you submit a contract with rates, then unlock a rate and resubmit it, we would expect with the old behavior for findContractWithRates to return the out of date rate before, and to return the new rate now?

@macrael Good call on the tests! I forgot about updating the rate side for rate-only submissions and unlocks. I added tests for both `findRateWithHistory` and `findContractWithHistory`.

services/app-api/src/postgres/contractAndRates/submitRate.ts
```diff
@@ -52,6 +52,9 @@ async function submitRate(
                         include: includeLatestSubmittedRateRev,
                     },
                 },
+                orderBy: {
+                    createdAt: 'desc',
+                },
```
JasonLin0991, line 57: Interesting find here. When using the `rateID` to unlock a rate, `findFirst` did not find the first record in descending order. After forcing descending, it would then work correctly.

PR #2170, opened 2024-01-11
Add dev IPs to jump box security group

mojotalantikite:
## Summary

We received a security alert from CMS cloud that the security group rule that allows access on `0.0.0.0/0` for ssh (protected by pub key auth) is disallowed. That rule was automatically reverted by their scanning and we've been adding dev IPs to an allowlist for ssh. They've asked us to also remove that open rule from our automation, which is what this is.

This adds the dev IPs that were manually added to the security group to our cloud formation template for that security group.

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-3825

### DIFF
--- a/services/postgres/serverless.yml
+++ b/services/postgres/serverless.yml
@@ -103,6 +103,7 @@ resources:
       - !Equals ['${sls:stage}', 'main']
       - !Equals ['${sls:stage}', 'val']
       - !Equals ['${sls:stage}', 'prod']
+      - !Equals ['${sls:stage}', 'mtsechubsg']
 
   Resources:
     # VPC endpoint for rotation lambda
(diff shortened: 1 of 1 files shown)


macrael:
What does specifying these IPs in the cloud formation do for us? We can still manually add them (or I guess I care mostly that my script that adds them via the API will keep working)

macrael (approved):
Knock em down

mojotalantikite:
@macrael What CMS cloud was complaining about is the single rule we had in this security group that allowed ssh without restrictions based on IPs. We've already remediated that by adding things in manually, but their security hub scans still notice that old rule for open ssh. 

We still need a rule in that security group in order to provision it (and keep it provisioned), so in this PR I'm just copying the IPs we already have manually added. An alternative would maybe be to just add that single internal IP block, so that the group stays provisioned. That change will probably clear out the security group, so we'd just have to go back in and re-add all the IPs. I don't have strong opinions here.

Either way, I don't think this should affect your script that adds IPs there via the aws sdk.

PR #2171, opened 2024-01-11
MCR-3836 - Modify app-api to process request from the 3rd party authorizer

pearl-truss:
## Summary
Requests that are authorized through the `third_party_API_authorizer` will not have a cognito auth provider for looking up the user who's making the request. This PR updates the `app-api` code so that these requests are routed in both local and deployed environments to use the `userId` that's returned from the authorizer to lookup and fetch the user

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-3836

#### To Do

- [x] works locally but I need to update the code to work with review app/deployed environments

## QA guidance

The following cURL command can be used to **test locally**
```
curl -s http://127.0.0.1:3030/local/v1/graphql/external -X POST \
-H "Authorization: Bearer ${token}" \
-H "Content-Type: application/json" \
--data '{"query":"query IndexRates {\n  indexRates {\n    totalCount\n    edges {\n      node {\n       id\n      }\n    }\n  }\n}"}'
```

The following cURL command can be used to **test with review app**
```
curl -s https://79stqk83kb.execute-api.us-east-1.amazonaws.com/mcr3836modifyappap806bc/v1/graphql/external 
-X POST \
-H "Authorization: Bearer ${token}" \
-H "Content-Type: application/json" \
--data '{"query":"query IndexRates {\n  indexRates {\n    totalCount\n    edges {\n      node {\n       id\n      }\n    }\n  }\n}"}'
```


services/app-api/src/authn/thirdPartyAuthn.ts
```diff
@@ -0,0 +1,31 @@
...
+        const auroraUser = await lookupUserAurora(store, userId)
+        if (auroraUser instanceof Error) {
+            return err(auroraUser)
+        }
+
+        if (auroraUser instanceof Error) {
+            return err(auroraUser)
+        }
+
+        if (auroraUser === undefined) {
```
pearl-truss, line 22: @macrael if a user isn't stored in postgres we won't be able to use cognito to create and insert a new user into postgres like we're doing in `cognitoAuthn`, but I'm wondering if there's anything we should do or if erroring is okay?

macrael: Let's just error. Like you said we don't have the info to be able to create the user. This should be pretty impossible since someone will have had to login with cognito to generate a JWT so the user better be in the DB already.

services/app-api/src/handlers/apollo_gql.ts
```diff
@@ -81,8 +88,21 @@ function contextForRequestForFetcher(userFetcher: userFromAuthProvider): ({
...
 
                 const store = NewPostgresStore(pgResult)
-                const userResult = await userFetcher(authProvider, store)
+                const userId = event.requestContext.authorizer?.principalId
+
+                let userResult
+                if (authProvider && !fromThirdPartyAuthorizer) {
+                    userResult = await userFetcher(authProvider, store)
+                } else if (fromThirdPartyAuthorizer) {
+                    userResult = await userFromThirdPartyAuthorizer(
```
macrael, line 97: I'm confused how this compiles. This function looks like it takes a `string` for userID, but userID with the question mark in the chain should be possibly `undefined`

macrael (commented):
a few comments but this looks good. I think we should hold off merging it until the secrets are in place right, though.

services/app-api/src/authn/thirdPartyAuthn.ts
```diff
@@ -0,0 +1,27 @@
...
+            return err(auroraUser)
+        }
+
+        return ok(auroraUser)
+    } catch (e) {
+        console.error('ERROR: failed to look user in postgres')
```
haworku, line 24: I _think_ this should be `recordException` and going through OTEL. Assuming we would want to alert on this? Also noticing we don't have technical docs on OTEL besides the the initial ADR... so maybe you haven't seen this code yet @pearl-truss

services/app-api/src/handlers/apollo_gql.ts
macrael: oh duh, `&& userId` right above this line.

services/app-api/src/authn/thirdPartyAuthn.ts
pearl-truss: @haworku this error gets returned with the function and handled [here](https://github.com/Enterprise-CMCS/managed-care-review/blob/main/services/app-api/src/handlers/apollo_gql.ts#L86)  and is a part of the [tracing middleware](https://github.com/Enterprise-CMCS/managed-care-review/blob/main/services/app-api/src/handlers/apollo_gql.ts#L131C10-L131C27) that is using an otel tracer. I'm not particularly familiar with this area of the code base. Given the above, do you think `recordException` should be used here?

services/app-api/src/handlers/third_party_API_authorizer.ts
```diff
@@ -31,7 +31,7 @@ export const main: APIGatewayTokenAuthorizerHandler = async (
     const authToken = event.authorizationToken.replace('Bearer ', '')
     try {
         // authentication step for validating JWT token
-        const userId = jwtLib.userIDFromToken(authToken)
+        const userId = await jwtLib.userIDFromToken(authToken)
```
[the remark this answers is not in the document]
pearl-truss: 👍 this has been changed back, I added it when I was debugging earlier

services/app-api/src/authn/thirdPartyAuthn.ts
mojotalantikite: Yeah, unfortunately we need to do those `recordExceptions` anywhere we want to record a span to a trace. Unlike pretty much every other monitoring system I've used, OTEL doesn't do this stuff automatically for you (there is some auto-instrumentation, but most of the time you need to do the things yourself). We should write up some docs about these patterns and clean things up a bit, it's not all that intuitive!

services/app-api/src/authn/authn.ts
```diff
@@ -4,5 +4,6 @@ import type { Store } from '../postgres'
 
 export type userFromAuthProvider = (
     authProvider: string,
-    store?: Store
-) => Promise<Result<UserType, Error>>
+    store?: Store,
+    userId?: string
+) => Promise<Result<UserType, Error | undefined>>
```
macrael, line 9: We decided not to return undefined, right?

services/app-api/src/authn/thirdPartyAuthn.ts
```diff
@@ -0,0 +1,26 @@
+import type { Result } from 'neverthrow'
+import { ok, err } from 'neverthrow'
+import type { UserType } from '../domain-models/index'
+import type { Store } from '../postgres'
+import { lookupUserAurora } from './cognitoAuthn'
+
+export async function userFromThirdPartyAuthorizer(
```
macrael, line 7: you could probably have the type def of this function be explicitly the userFromAuthProvider type

services/app-api/src/authn/authn.ts
macrael: and, this is confusing b/c it's the only place in the codebase we're using Result, but I think the right way to write this would be `Result<UserType, Error, undefined>` as opposed to using the sum type

macrael (approved):
Looking good!

services/app-api/src/authn/thirdPartyAuthn.ts
pearl-truss: I didn't use that here because `authProvider` is required with userFromAuthProvider type and that wouldn't be used in this function.

services/app-api/src/authn/thirdPartyAuthn.ts
pearl-truss: I'm not 100% sure what needs to happen here but I took an educated guess and pushed that code up. Lmk if anything needs changing @haworku @mojotalantikite

pearl-truss: I'll merge since I'm seeing things working but if anything comes up for improving this code I can make those changes in https://github.com/Enterprise-CMCS/managed-care-review/pull/2195

PR #2189, opened 2024-01-22
Setup Secrets for JWT

macrael:
## Summary

JWT secrets are being stored in secret manager, generated via CF. They are passed into the app via env vars

apollo_gql and third_party_authorizer both get the JWT secret, needed for creating and validating JWTs, from an environment variable. 

That variable is set by serverless from an AWS Secret manager secret. That secret is provisioned on first deploy via CloudFormation. 

The CloudFormation secret generation code is really designed for passwords, but we ultimately want random bytes, so I worked around it like this: the password generator is configured to generate a password that contains only numbers and the letters a-f. This then should be the same as a random hex encoded string. That's what we pass in as an env var to the two lambdas. 

#### Related issues

https://qmacbis.atlassian.net/browse/MCR-3833

#### Test cases covered
Still no test for this really since we haven't written the page that generates codes


macrael:
@pearl-truss This PR is ready for review, and should unblock your PR that enables the API to actually work since now it will be protected by a real secret

services/app-api/src/jwt/jwt.ts
```diff
@@ -45,6 +47,8 @@ interface JWTLib {
 
 function newJWTLib(config: JWTConfig): JWTLib {
     return {
+        // this is an experiment, using `curry` here, It seems clean but I'm not sure
+        // exactly what it's getting us yet -wml
```
pearl-truss, line 51: My understanding is currying is good for higher order functions that are functions that return other functions

services/app-api/src/resolvers/APIKey/createAPIKey.test.ts
```diff
@@ -7,7 +7,7 @@ describe('createAPIKey', () => {
     it('creates a new API key', async () => {
         const jwt = newJWTLib({
             issuer: 'mctestiss',
-            signingKey: 'foo',
+            signingKey: Buffer.from('123af', 'hex'),
```
pearl-truss, line 10: Can you add a comment here for why `123af` is used here?

pearl-truss (commented):
@macrael Can you pull in main? There's a lot of scss file changes I don't think are a part of this PR and make it a bit harder to see what the changes are

macrael:
huh I just pulled in main and those sass changes are still there?? I'm very confused about that. @haworku do these changes look familiar to you? Is this automated linting?

services/app-api/src/resolvers/APIKey/createAPIKey.test.ts
macrael: I wanted a short dumb key for testing, and since we're passing in keys as hex encoded strings I wanted to do that the same way we are elsewhere too

pearl-truss (approved):
Nevermind on pulling in main, I was able to go through the files particular to the JWT secret, overall it LGTM!

PR #2195, opened 2024-01-23
MCR-3839 restrict 3rd party API access by IP address

pearl-truss:
## Summary

- `configuration/allowed_ip_addresses` created in the parameter store
- a default IP address is supplied in `.envr` for local
- the `thirdPartyAuthn` function compares the current IP address with those in the parameter store value. If the IP address isn't allowed it will return an error and not allow the request to go through

#### Related issues
https://jiraent.cms.gov/browse/MCR-3839

#### Screenshots
Screenshot 2024-02-01 at 12 24 10 PM

## QA guidance
 ```
curl -s https://t4rnz8nwfk.execute-api.us-east-1.amazonaws.com/mcr3839restrictaute0128/v1/graphql/external -X POST \
-H "Authorization: Bearer {jwt}" \
-H "Content-Type: application/json" \
--data '{"query":"query IndexRates {\n  indexRates {\n    totalCount\n    edges {\n      node {\n       id\n      }\n    }\n  }\n}"}'
```

The IP address that hits the review app changes daily. You can run the cURL command above and receive an error. Check the AWS logs for the outputted ip address that you can then add to the allowed_ip_addresses parameter store value. Once the correct IP address is added there you'll be able to hit the API

### CODE REVIEW

**Review by mojotalantikite — APPROVED** [review]

**Comment by pearl-truss** [comment]
@macrael did you want to review this as well? With the cypress test I saw a way to have a happy path, at least for local so that's what I implemented. There's also an expected fail case added for non local environments

File: services/app-api/src/authn/thirdPartyAuthn.ts [hunk 1]
```diff
@@ -17,8 +18,20 @@ export async function userFromThirdPartyAuthorizer(
...
     initTracer(serviceName, otelCollectorURL)
-
+    const allowedIpAddresses = process.env.ALLOWED_IP_ADDRESSES
+    if (allowedIpAddresses === undefined || allowedIpAddresses === '') {
+        throw new Error(
+            'Configuration Error: ALLOWED_IP_ADDRESSES is required to run app-api.'
+        )
+    }
     try {
+        const ipAddressIsValid = allowedIpAddresses.includes(ipAddress)
```
macrael, line 28: This is a little strange to me, using string matching against potentially an array of ip addresses, but I think it works fine. honestly it's nice that it's robust against spaces and other stuff being in the array.

services/app-api/src/authn/thirdPartyAuthn.ts
```diff
@@ -17,8 +18,20 @@ export async function userFromThirdPartyAuthorizer(
 
     const serviceName = 'third-party-authorizer'
     initTracer(serviceName, otelCollectorURL)
-
+    const allowedIpAddresses = process.env.ALLOWED_IP_ADDRESSES
```
macrael, line 21: Please don't read env vars outside of the initialization of one of our services. I try for us to do all that kind of checking and throwing at the apollo_gql.ts level and then have everything else take parameters. This makes things easier to test and ensures that everything is configured at startup, and gives us nicer interfaces to things with types ensuring that everything is present. Happy to talk about this more this is one of my big architecture goals for the app.

services/cypress/integration/thirdPartyAPIAccess/thirdPartyAPIAccess.spec.ts
```diff
@@ -69,7 +69,12 @@ describe('thirdPartyAPIAccess', () => {
                 body: '{"query":"query IndexRates { indexRates { totalCount edges { node {  id } } } }"}',
                 failOnStatusCode: false,
             }).then(res => {
-                expect(res.status).to.equal(200)
+                if (Cypress.config().baseUrl === 'http://localhost:3000') {
+                    expect(res.status).to.equal(200) // okay 
+                } else {
+                    expect(res.body.errors[0].message).to.contain('this IP address is not in the allowed list')
+                    expect(res.status).to.equal(500) // server error
```
macrael, line 76: I would not expect a 500 for a request that is coming from the wrong IP address. 500 should only be returned when an unexpected internal server error has occurred. An outlawed IP address should be a 403 forbidden, I think

services/app-api/serverless.yml
```diff
@@ -253,10 +255,16 @@ functions:
...
+  Conditions:
+    IsDevValProd: !Or
+      - !Equals ['${sls:stage}', 'main']
+      - !Equals ['${sls:stage}', 'val']
+      - !Equals ['${sls:stage}', 'prod']
   Resources:
     ApiGwWebAclAssociation:
       Type: AWS::WAFv2::WebACLAssociation
       DependsOn: 'ApiGatewayDeployment${sls:instanceId}'
+      Condition: IsDevValProd
```
macrael, line 267: why are you modifying the ACL in this PR? Since you're checking the IP in code I didn't think there needs to be any change here

macrael (changes requested):
I like the string matching strategy and I think that doing this is code should work great. 

I would like this check to be performed directly in the lambda authorizer instead of in the 3rd party auth function inside of apollo_gql, though. 

This would contain our auth logic to the authorizer, it would be checking the JWT header and verifying the IP address in one place, returning an error before our application code is even engaged in the case that something is not right. 

I believe you said that the authorizer gets the ip address plus port in the args so I think if you strip the port off you should still be able to check this there.

services/app-api/serverless.yml
pearl-truss: Sorry that's a holdover from this [convo ](https://cmsgov.slack.com/archives/C014CRU197U/p1706289589824939)

services/app-api/src/authn/thirdPartyAuthn.ts
pearl-truss: 👍 I would be interested in hearing more on this

pearl-truss:
> I like the string matching strategy and I think that doing this is code should work great.
> 
> I would like this check to be performed directly in the lambda authorizer instead of in the 3rd party auth function inside of apollo_gql, though.
> 
> This would contain our auth logic to the authorizer, it would be checking the JWT header and verifying the IP address in one place, returning an error before our application code is even engaged in the case that something is not right.
> 
> I believe you said that the authorizer gets the ip address plus port in the args so I think if you strip the port off you should still be able to check this there.

Maybe we can pair on this? I spent a lot of time trying to get this to work within the authorizer first. I wanted to parse out the IP address from `host` in the headers for the event. I saw that it worked locally but I kept getting back an error in review app because the headers were apparently undefined there

services/app-api/src/authn/authn.ts
```diff
@@ -4,5 +4,6 @@ import type { Store } from '../postgres'
 
 export type userFromAuthProvider = (
     authProvider: string,
-    store?: Store
-) => Promise<Result<UserType, Error>>
+    store?: Store,
+    userId?: string
+) => Promise<Result<UserType, Error | undefined>>
```
macrael, line 9: I think this can go back as well, no undefined being returned here

macrael (approved):
Looks great!

PR #2196, opened 2024-01-23
MCR-3842: API for edit submit rate

JasonLin0991:
## Summary
[MR-3842](https://qmacbis.atlassian.net/browse/MCR-3842)

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/updateDraftContractWithRates.ts
+++ b/services/app-api/src/postgres/contractAndRates/updateDraftContractWithRates.ts
@@ -176,31 +176,32 @@ async function updateDraftContractWithRates(
                 ratesFromDB.push(domainRateRevision)
             }
 
+            // Parsing rates from request for update or create
             const updateRates =
                 rateFormDatas && sortRatesForUpdate(ratesFromDB, rateFormDatas)
 
             if (updateRates) {
                 for (const rateFormData of updateRates.upsertRates) {
-                    // Check if the rate exists
-                    // - We don't know if the rate revision exists in the DB we just know it's not connected to the contract.
-                    // - toProtoBuffer gives every rate revision a UUID if there isn't one, so we cannot rely on revision id.
-                    // - We can use this revision id to check if a rate and revision exists.
+                    // Current rate with the latest revision
+                    let currentRate = undefined
 
-                    // Find the rate of the revision with only one draft revision
-                    const currentRate = await tx.rateTable.findFirst({
-                        where: {
-                            id: rateFormData.id,
-                        },
-                        include: {
-                            // include the single most recent revision that is not submitted
-                            revisions: {
-                                where: {
-                                    submitInfoID: null,
-                                },
-                                take: 1,
+                    // If no rate id is undefined we know this is a new rate that needs to be inserted into the DB.
+                    if (rateFormData.rateID) {
+                        currentRate = await tx.rateTable.findUnique({
+                            where: {
+                                id: rateFormData.id,
                             },
-                        },
-                    })
+                            include: {
+                                // include the single most recent revision that is not submitted
+                                revisions: {
+                                    where: {
+                                        submitInfoID: null,
+                                    },
+                                    take: 1,
+                                },
+                            },
+                        })
+                    }
 
                     const contractsWithSharedRates =
                         rateFormData.packagesWithSharedRateCerts?.map(
--- a/services/app-api/src/resolvers/rate/submitRate.ts
+++ b/services/app-api/src/resolvers/rate/submitRate.ts
@@ -1,0 +1,153 @@
+import type { Store } from '../../postgres'
+import type { MutationResolvers } from '../../gen/gqlServer'
+import {
+    setErrorAttributesOnActiveSpan,
+    setResolverDetailsOnActiveSpan,
+} from '../attributeHelper'
+import type { RateFormDataType } from '../../domain-models'
+import { isStateUser } from '../../domain-models'
+import { logError } from '../../logger'
+import { ForbiddenError, UserInputError } from 'apollo-server-lambda'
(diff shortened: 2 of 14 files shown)


services/app-api/src/resolvers/rate/submitRate.ts
```diff
@@ -0,0 +1,153 @@
...
+        const { user, span } = context
+        const { rateID, submitReason, formData } = input
+        const featureFlags = await launchDarkly.allFlags(context)
+
+        setResolverDetailsOnActiveSpan('submitRate', user, span)
+        span?.setAttribute('mcreview.rate_id', rateID)
+
+        // throw error if the feature flag is off
+        if (!featureFlags?.['rate-edit-unlock']) {
+            const errMessage = `Not authorized to submit rate, the feature is disabled`
```
haworku, line 29:
```suggestion
            const errMessage = `Not authorized to edit and submit a rate independently, the feature is disabled`
```

services/app-api/src/resolvers/rate/submitRate.ts
```diff
@@ -0,0 +1,153 @@
...
+                    cause: 'DB_ERROR',
+                },
+            })
+        }
+
+        const draftRateRevision = unsubmittedRate.draftRevision
+
+        if (!draftRateRevision) {
+            throw new Error(
+                'PROGRAMMING ERROR: Status should not be submittable without a draft rate revision'
```
haworku, line 71: No change needed since this is a pattern in other places but I find it odd we are having to check this kind of stuff in the resolver - would like this handled closer to the database. Would be nice if by the time we get to the resolver we trust our `status` field fully.

services/app-graphql/src/schema.graphql
```diff
@@ -1194,8 +1247,106 @@ type UnlockRatePayload {
...
+    """
+    packagesWithSharedRateCerts: [PackageWithSameRateInput!]!
+}
+
+input SubmitRateInput {
+    rateID: ID!
+    "User given submission description"
+    submitReason: String!
+    "Rate related form data to be updated with submission"
+    formData: RateFormDataInput!
```
haworku, line 1351: 🛑 Could we  just use `formData: RateFormData` and use the existing type here? Is there a difference between these types?

 I would really like compose together existing types rather than duplicate types between the `type` and the `input` when possible.

haworku (approved):
Only one comment blocking - if we can clean up this gql schema that would be nice

haworku:
@JasonLin0991 What is your manual QA guidance on this PR? Is it just testing user story or is there anything else you want me to check.

services/app-api/src/postgres/contractAndRates/updateDraftContractWithRates.ts
```diff
@@ -176,31 +176,32 @@ async function updateDraftContractWithRates(
...
-                                where: {
-                                    submitInfoID: null,
+                    // If no rate id is undefined we know this is a new rate that needs to be inserted into the DB.
+                    if (rateFormData.rateID) {
+                        currentRate = await tx.rateTable.findUnique({
+                            where: {
+                                id: rateFormData.id,
+                            },
+                            include: {
+                                // include the single most recent revision that is not submitted
```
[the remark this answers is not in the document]
JasonLin0991: I think that comment was left over. Order won't matter here, we are trying to find a unique ID and there should never be a duplicate ID in this table.

I will delete this comment  along with `take: 1` because `findUnique` implies only one record.

services/app-api/src/resolvers/rate/submitRate.ts
JasonLin0991: I believe this is a thing with the types.

```ts
const rateSchema = z.object({
    id: z.string().uuid(),
    createdAt: z.date(),
    updatedAt: z.date(),
    status: statusSchema,
    stateCode: z.string(),
    stateNumber: z.number().min(1),
    // If this rate is in a DRAFT or UNLOCKED status, there will be a draftRevision
    draftRevision: rateRevisionWithContractsSchema.optional(),
    // All revisions are submitted and in reverse chronological order
    revisions: z.array(rateRevisionWithContractsSchema),
})
```

We would need to update `draftRateSchema` like this:
```ts
const draftRateSchema = rateSchema.extend({
    status: z.literal('DRAFT'),
    draftRevision: rateRevisionWithContractsSchema,
    revisions: z.array(rateRevisionWithContractsSchema).min(1),
})
```
 Then export a `DraftRateType` to use in `findRateWithHistory` to return the type based on the status.

services/app-graphql/src/schema.graphql
JasonLin0991: I tried, but it gave errors. From my understanding, GraphQL `input` and `types` are different, and only `input` can be used as arguments for a mutation. I think this [StackOverflow](https://stackoverflow.com/questions/41743253/whats-the-point-of-input-type-in-graphql/55881719#55881719) answer gives a better explanation than I could. It's an old thread, but I think it is still correct.

haworku: Yes okay, I was looking at that too. Alright ✅

JasonLin0991:
> @JasonLin0991 What is your manual QA guidance on this PR? Is it just testing user story or is there anything else you want me to check.

I think you could test this by:
- unlocking through the app and seeing the unlock button is disabled.
- manually resubmitting a rate using the GraphQL Explorer tool.
- the rate summary should then show the unlock button as enabled.

This might be a better thing to test when we implement the submit button.

services/app-graphql/src/schema.graphql
haworku, line 1351: From slack conversation  -should not require formData to submit, should also work with `rateID` and `submitReason` only

haworku (approved):

PR #2202, opened 2024-01-24
Add clamdscan to uploads lambda layer

mojotalantikite:
## Summary

This brings in the `clamdscan` binary into our lambda layer. We need this version of the tool to eventually call a remote, always on `clamd` to do the scanning.

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-3819


.github/workflows/deploy.yml
```diff
@@ -249,9 +249,30 @@ jobs:
...
+      - uses: actions/upload-artifact@v3
+        with:
+          name: lambda-layers-clamav
+          path: ./services/uploads/src/avLayer/build/lambda_layer.zip
+
   deploy-infra:
-    needs: [begin-deployment]
-    uses: Enterprise-CMCS/managed-care-review/.github/workflows/deploy-infra-to-env.yml@main
+    needs: [begin-deployment, build-clamav-layer]
+    uses: Enterprise-CMCS/managed-care-review/.github/workflows/deploy-infra-to-env.yml@mt-av-layer
```
haworku, line 275: Assuming this is the pattern where we set up pointing at review app and then change back to main later?

haworku (approved):

.github/workflows/deploy.yml
mojotalantikite: yeah, I wish we had a better way of doing this, but I'll change it back once it merges!

PR #2207, opened 2024-01-25
MCR-3758: Update launch darkly and fix testing

JasonLin0991:
## Summary
[MCR-3758](https://qmacbis.atlassian.net/browse/MCR-3758)

- Replace `ldUseClientSpy` with `LDProvider` to configure feature flags for unit testing.
   - Using `LDProvider` to initialize our tests because `jest.spyOn` used in `ldUseClientSpy` is not working with `launchdarkly-react-client-sdk` versions 3.0.9 and above. I'm not exactly sure why it stopped working. I've tried to fully construct a `useLDClient` `mockImplemenation` in `ldUseClientSpy` with no success.
   - With `LDProvider`, we are initializing a mocked `ldClient` for our unit tests when using `renderWithProviders`. Since we are initializing the `ldClient` outside of `LDProvider,` it does not do it internally, which requires an API call to LaunchDarkly; we are isolating tests from flag values in LaunchDarkly.
   - With using `LDProvider,` follows the pattern we set in `renderWithProviders`, so I'm happy with that.
 - Updated the docs on the new method for unit testing.
 - Updated all tests with feature flags.
 - Fixed a React error.

## QA guidance

It's a pretty simple implementation, so I'm not sure what QA tests can be done here. There should be no side effects to the UI.

### DIFF
--- a/docs/technical-design/launch-darkly-testing-approach.md
+++ b/docs/technical-design/launch-darkly-testing-approach.md
@@ -23,60 +23,88 @@ export LD_SDK_KEY='Place Launch Darkly SDK key here'
 ## Feature flag unit testing
 
 ### Client side unit testing
-Client side unit testing utilizes `jest.spyOn()` to mock the LaunchDarkly `useLDClient` hook and return default flag values or values specified. This implementation is done in our jest helper function `ldUseClientSpy()` located in `app-web/src/testHelpers/jestHelpers.tsx`.
+Client side unit testing utilizes the `LDProvider` in our [renderWithProviders](../../services/app-web/src/testHelpers/jestHelpers.tsx) function to set up feature flags. We use this method for testing because the official documented [unit testing](https://docs.launchdarkly.com/guides/sdk/unit-tests/?q=unit+test) method by LaunchDarkly does not work with our implementation. Previously we had used `jest.spyOn` to intercept `useLDClient` and mock the `useLDClient.variation()` function with our defined feature flag values. With `launchdarkly-react-client-sdk@3.0.10` that method did not work anymore.
 
-`ldUseClientSpy` takes in an object of feature flags and values as an argument. You can configure multiple flags with the object passed into `ldUseClientSpy`.
+Currently, LaunchDarkly does not provide an actual mock `LDProvider` so if or when they do, then we could update this with that provider.
 
-```javascript
-ldUseClientSpy({
-  'rates-across-submissions': true,
-  'rate-cert-assurance': true,
-})
+#### Configuration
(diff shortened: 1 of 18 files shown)


docs/technical-design/launch-darkly-testing-approach.md
```diff
@@ -23,60 +23,98 @@ export LD_SDK_KEY='Place Launch Darkly SDK key here'
...
+  <ContractDetails
+          draftSubmission={medicaidAmendmentPackage}
+          updateDraft={jest.fn()}
+          previousDocuments={[]}
+  />,
+  {
+    apolloProvider: {
+      mocks: [fetchCurrentUserMock({ statusCode: 200 })],
+    },
+    featureFlags: {
```
macrael, line 111: I LOVE THIS YAY

services/app-web/src/testHelpers/jestHelpers.tsx
```diff
@@ -45,23 +74,44 @@ const renderWithProviders = (
...
-            <MemoryRouter initialEntries={[route || '']}>
-                <AuthProvider authMode={'AWS_COGNITO'} {...authProvider}>
-                    <S3Provider client={s3Client}>
-                        {location && <WithLocation setLocation={location} />}
-                        <PageProvider>{ui}</PageProvider>
-                    </S3Provider>
-                </AuthProvider>
-            </MemoryRouter>
-        </MockedProvider>
+        <LDProvider {...ldProviderConfig}>
```
macrael, line 101: Should we be using LDProvider in App.tsx? I'm a little confused how these get read correctly, but it seems like a major improvement in how LD works

macrael (approved):
I love this!

Honestly I never cared much for using jest spy/mocks whatnot. Seems like LD got their shit together and made a real way to do it! So glad.

services/app-web/src/testHelpers/jestHelpers.tsx
JasonLin0991: > Should we be using LDProvider in App.tsx?
We actually initialize LaunchDarkly in `app-web/src/index.tsx` using `LDProvider`.

The important part of this method is that instead of letting `LDProvider` initialize `ldClient` by making the API calls to get the flags, we initialize it ourselves with predefined flags and then pass it to `LDProvider`.

JasonLin0991:
> Honestly I never cared much for using jest spy/mocks whatnot. Seems like LD got their shit together and made a real way to do it! So glad.

This is still not the official way of doing it, since that way doesn't work for us. This was my hacky way of creating a "mock" `LDProdiver`.

macrael:
well I like this better than that, too

PR #2218, opened 2024-01-30
Add API Key Generation Page

macrael:
## Summary

This PR adds a route "dev/api-access" that allows any user to generate an API key that will allow access to the app for 90 days.

#### Related issues

https://jiraent.cms.gov/browse/MCR-3833

#### Screenshots

Screenshot 2024-01-29 at 4 35 10 PM

#### Test cases covered

Wrote a new cypress test that generates an API key and calls an API


services/app-web/src/pages/APIAccess/APIAccess.tsx
```diff
@@ -0,0 +1,108 @@
...
+                                </code>
+                            </li>
+                        </ul>
+                    </li>
+                </ul>
+
+                <h2>Resources</h2>
+                <ul>
+                    <li>
+                        The MC-Review{' '}
```
haworku, line 82: use `&nbsp;` not jsx with empty strings to ensure accurate spaces visually - this is in a few places

services/app-web/src/index.tsx
```diff
@@ -136,6 +141,7 @@ if (ldClientId === undefined) {
                     authMode={authMode}
                     apolloClient={apolloClient}
                     s3Client={s3Client}
+                    apiURL={apiURL}
```
haworku, line 144: Why string this variable through as a prop everywhere ? I think its not needed, it is not used across the application logic ( whereas`authMode` is )

I suggest we just access this process.env directly in the `AppAccess.tsx` component and remove the code changes in App/AppBody/AppRoutes entirely.

haworku (approved):

services/app-web/src/pages/APIAccess/APIAccess.tsx
```diff
@@ -0,0 +1,108 @@
...
+                <h2>Usage</h2>
+                <ul>
+                    <li>Make GraphQL requests to the URL: {apiURL}</li>
+                    <li>
+                        Include the JWT as a Bearer token in the Authorization
+                        header. Example:
+                        <ul>
+                            <li>
+                                <code>
+                                    Authorization: Bearer eyJhbGciOiJIU...
```
pearl-truss, line 72: I think it could be helpful to include a complete sample cURL request here

services/cypress/integration/thirdPartyAPIAccess/thirdPartyAPIAccess.spec.ts
```diff
@@ -0,0 +1,75 @@
+describe('thirdPartyAPIAccess', () => {
```
pearl-truss, line 1: It's great to have cypress test for this!

pearl-truss (approved):
This looks great! I left a few comments and a design suggestion would be to display both buttons side by side. When they're both on screen it looks a little off with the current placement to me

Screenshot 2024-01-30 at 4 26 49 PM

services/app-web/src/pages/APIAccess/APIAccess.tsx
macrael: huh those were all added by the linter

services/app-web/src/index.tsx
macrael: Alright, I can do that. Agreed this is annoying to prop drill so deep. Maybe we could have some kind of config context to read from? Anytime we read env vars out side of index.tsx (or apollo_gql.ts on the back end) it smells a little, making it a little harder to test things cleanly

PR #2227, opened 2024-02-01
Create rates testing utilities in app-api

haworku:
## Summary
This PR creates test helpers and mocks for rates API. It also migrate existing rates api tests. These apis are
- [x] `unlockRate`
- [x] `submitRate`
- [x] `indexRates`
- [x] `fetchRate`

In the progress of this PR fixed a small bug with submitRate
- `submittedReason` should be optional - handle adding default "initial submission" reason in the api layer (similar to how HPP submit api works). Also changed name from `submitReason` to `submittedReason` to match other unlock/resubmit code.

NOTES

- This PR created a few contract test helpers for graphQL. The largely rely on interacting with the database via prisma. They should be updated to use the resolvers once they are created. Same for the rate test helpers that are using prisma for resolvers that don't yet exist
- The submitRate test has one holder over test that uses packages because it's testing that submission has a contract that's updated with a rate. The current `updateContract` relies on `pkg` so I didn't change this taken given that `updateContract` will change in future refactors. I believe for linked rates?

#### Related issues
https://jiraent.cms.gov/browse/MCR-3881

#### Test cases covered

Test helpers used in the test for submitRate, fetchRate, indexRates and unlockRate use different helper functions but the spirit and assertions of each test is the same as what was there previously

## QA guidance

API tests are passing


macrael (approved):
This looks good. I think my only remaining question is about switching to using `must` on `server.executeOperation()` I'm not convinced that is doing anything b/c I don't think executeOperation returns an Error type on error. 

https://www.apollographql.com/docs/apollo-server/v3/testing/testing

It returns the same { data, errors } that we get from apollo client. So I'm pretty sure the must() is doing nothing

haworku (commented):
LGTM great to see these cleaned up and using new API

PR #2228, opened 2024-02-01
MCR-3813 Restricting rate access to each respective state

ruizajtruss:
…tate code

## Summary
A continuation to MCR-3793, the purpose of this PR is to deny access to rates from unrelated states (i.e. Minnesota should not be able to access Virginia's rates)

#### Related issues
[MCR-3813](https://jiraent.cms.gov/browse/MCR-3813)
#### Screenshots
Will display custom error message if provided:
image

Standard forbidden error message if no error message is provided:
image

## QA guidance

1. Grab a rate id of your choice
2. Log in as a state user that differs from the state attached to the rate
3. When visiting `/rates/:id` and `rates/:id/edit` you should be met with the Forbidden error page


haworku (approved):
LGTM

PR #2229, opened 2024-02-02
Revert "Send utag.view tag data on initial app load"

haworku:
Reverts Enterprise-CMCS/managed-care-review#2215

This is failing in higher environments due to `utag.view` not loading in the page yet. Prepping this PR in case we decide to revert. See failed promote https://github.com/Enterprise-CMCS/managed-care-review/actions/runs/7749469528.


macrael (approved):
If the other change passes muster then I don't think we need to revert

PR #2230, opened 2024-02-02
bugfix: make sure utag.view exists

haworku:
## Summary
This should address the issue the cause the failed promote referenced in #2229. Please take a look though, let me know your thoughts. This adds a second to the initial app load before tealium called. From what I can tell from those tags usually take around 35 ms to load. 

Also added guard clause so this won't break app compile when window.utag is not set yet. 

#### Related issues
#2215


macrael (approved):
missing await on line 98 caused a build error

PR #2253, opened 2024-02-14
MCR-3777 Design review feedback

haworku:
## Summary
Addressing a point from design review
- Rate Summary page buttons should say Cancel and Submit ... not Back and Continue

Also noticed  🐛s:
- the validation error messages for capitation rates and the actuary preference were not displaying the proper strings. This was due to `null` instead of `undefined` types passing through to yup validators
- banner styles not consistent for unlocked on review and submit versus summary page

#### Related issues
https://jiraent.cms.gov/browse/MCR-3771


macrael (approved):
sweet

PR #2243, opened 2024-02-07
Add always-on ClamAV server 

mojotalantikite:
## Summary

This adds a server that runs `clamavd` and `freshclam` with a configuration that allows it to accept incoming `clamdscan` requests from connections coming from within it's security group. This sets us up to allow our lambda antivirus scanning function to call an always-on `clamavd` instance that will return results without incurring the startup cost that we've been living with in the lambda only version.

#### Related issues
https://jiraent.cms.gov/browse/MCR-3820

PR #2256, opened 2024-02-14
Run av scanning from ClamD server

mojotalantikite:
## Summary

Note: this supersedes #2243, as it takes those changes and builds on top of them.

This adds a server that runs `clamavd` and `freshclam` with a configuration that allows it to accept incoming `clamdscan` requests from connections coming from within it's own security group. This allows our lambda antivirus scanning function to call an always-on `clamavd` instance that will return results without incurring the startup cost that we've been living with in the lambda only version. In testing this brings wait times from around 30s to less than 1s.

This also adds the configurations and code changes to the lambda to call `clamdscan` directly, which uses this new always on instance to do the actual scan.

We also add a private DNS server that allows us to lookup the location of the server with the hostname of `clamav.mc-review.local`. This was done so we don't have to rely on parameter store or any environment variables, we can just set the server to this hostname and in each of dev/val/prod it will resolve to the proper host.

Follow on work will be required to remove the current lambdas that do virus definition downloads and the associated s3 bucket that stores them. This new server runs `freshclam` every hour and updates the definitions locally, so we no longer have a need to run av definition updates in the old way.

#### Related issues
https://jiraent.cms.gov/browse/MCR-3820
https://jiraent.cms.gov/browse/MCR-3821
https://jiraent.cms.gov/browse/MCR-3822


services/uploads/serverless.yml
```diff
@@ -109,6 +107,9 @@ functions:
     layers:
       - !Ref ClamAvLambdaLayer
       - arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-nodejs-amd64-ver-1-18-1:1
+    vpc:
+      securityGroupIds: ${self:custom.sgId}
```
macrael, line 111: is the clamAV image inside our security group that can talk to the db for instance? I don't see why it would need to be?

services/uploads/serverless.yml
```diff
@@ -109,6 +107,9 @@ functions:
...
+      subnetIds: ${self:custom.privateSubnets}
```
macrael, line 112: I think you can remove the avDownloadDefinitions lambda, too

macrael (approved):
love it!

services/uploads/serverless.yml
mojotalantikite: I wrote a ticket to do this cleanup as follow on work: https://jiraent.cms.gov/browse/MCR-3954

I started deleting the now unneeded code paths and then realized that there's a decent amount in there and the tests are going to have to get fixed up a bit. Figured I'd land this and then go through and delete/cleanup the tests.

services/uploads/serverless.yml
mojotalantikite: You know, there were _reasons_ when I was setting this up, but I can't recall. We can probably drop this as the av scanning lambda doesn't need to be in the VPC with our other lambdas and DB. I'll fix it up before merging!

mojotalantikite: Oh actually, I remember now. If the lambda is not in this VPC with the server, then all the communicating traffic with the server is going to route over the public internet, which means we have to pay that egress cost. We can use VPC gateway endpoints, but we don't have access to create VPC related things in our account (CMS Cloud team has to do it for us). So I just dropped it in our VPC. 

I think we can drop the default security group though, which will keep it away from the Aurora instance.

PR #2265, opened 2024-02-16
Fix private DNS and clamd.conf for av scanning

mojotalantikite:
## Summary

The PR to add clamav scanning from a dedicated server (#2256) had an issue with name resolution in higher environments (val and prod). I was under the impression that Route53 private DNS would be isolated in each environment, so I configured it to have the same DNS name in each environment and in the config (`clamav.mc-review.local`). This seems to have been incorrect, as lookups for that name weren't working properly in val and prod.

This change moves to using the serverless stage in each environment to better differentiate the servers and avoid collisions (e.g. `clamav.mc-review-val.local`). To do this, I use webpack to alter the `clamd.conf` at deploy time instead of packaging it with the lambda layer.

#### Related issues
https://jiraent.cms.gov/browse/MCR-3960


macrael (approved):
SO WEIRD that this didn't work before but let's get it working.

PR #2270, opened 2024-02-20
Use single DNS name for clamav

mojotalantikite:
## Summary

It turns out the reason why DNS wasn't resolving properly in val and prod for `clamav.mc-review.local` in PR #2256 was because when CMS provisioned our VPCs, hostname resolution wasn't turned on in higher environments. Now that it's turned on, we should be fine going back to the simpler version here of using the same hostname across environments.

#### Related issues
https://jiraent.cms.gov/browse/MCR-3960


macrael (approved):
yay thanks for doing this straightaway

PR #2271, opened 2024-02-20
add updateDraftContractRates mutation to GQL

macrael:
## Summary

Adds the GQL for our updateDraftContractRates mutation to GQL

Unfortunately GQL does not support unions for inputs, so we kind of fake one here with validation on the server side. 

This supports Hana's work on the RateInfo page


haworku (approved):
TY! This works.

 weird graphql can have unions as output types and not input types.

PR #2274, opened 2024-02-21
update GQL schema names

macrael:
## Summary

in RSP today we made a few updates to the GQL schema. These are their stories.

### DIFF
--- a/services/app-graphql/src/schema.graphql
+++ b/services/app-graphql/src/schema.graphql
@@ -172,6 +172,7 @@ type Mutation {
     - UserInputError:
         - any of the linked or update rates' ids don't exist
         - attempts to update the data of a nibling rate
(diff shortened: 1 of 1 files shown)


macrael:
@Enterprise-CMCS/mc-review-admins I got this PR working by including a bunch more stuff in fetchRate (b/c Rate now includes draftContracts)

I suspect we'll need to do the same thing we did with our domain models and have explicit Contract and ContractWithRates types so that there isn't an infinite type loop. If you run into this using the Contract types let's try and fix it like we did there.

macrael:
OK I ended up getting nerd sniped by the draftContracts bit. We'll address the symmetry between Contract and Rate once we implement the history, I think. I couldn't get stuff working without a bunch of reworking rn

mojotalantikite (approved):

PR #2268, opened 2024-02-20
Removing FE cacheing from application

haworku:
## Summary
Remove caching on state submission form as per refactor plan for Contract & Rates Types. To recap...

_Benefits:_
-  prep for multiple APIs in use on frontend (HPP on some pages, contract and rates on other pages)
- reduce chance for regressions while we refactor page by page and also have feature flags in use (Linked Rates)
- ensure submission form, summary pages, dashboard and Q&A all fetch their own correct data

_Drawbacks:_
- During this interim period, we will have the loading indicator displaying more often when users change pages. 
- Cypress could get a little more flaky and times for cypress go up slightly

We can reduce the time period we have no caching by having a clear timeline to fully implement refactor 

#### Related issues
None, this is prep for Linked Rates and having two versions of Rate Details and Review/Submit in use in the form

#### Test cases covered
Tests should pass as normal

## QA guidance
- Manual checks that new or edited submissions appear at the top of the dashboard list with proper information
- Manual checks that going between submission form and Q&A pages behave as expected

PR #2275, opened 2024-02-21
MCR-3944: ReviewSubmit updated to use contract and rates

pearl-truss:
## Summary
**Primarily**

- Implemented V2 versions of the ReviewSubmit page and its sub components to consume graphql contract and rate types

**Bug fixes**

- Added attestation fields to the GQL contractFormData type
- Updated non unique GQL fragment names to resolve `Not all fragments have an unique name` on GQL generation
- Fixed `undefined fakesha` error on storybook for ContractDetails
- Updated the GQL `StateContact` type to use `titleRole` instead of `title` to match the other contact types like actuary
- Updated fetchContract to pass `updatedAt` and `createdAt` fields on the contract (included making updates to the related prisma functions and fetchContract resolver)

**Out of scope/not covered in this ticket**
- Navigating to the submission page from the Review Submit page
- Updating the unlock modal component to use contracts instead of HPP will be separated out into its own ticket
- Separate ticket to use `makeDocumentTableLookup` for ReviewSubmit will be created due to the function relying on a `revisionLookup` and HPP related types

**Remaining To Dos For this PR**

- [x] Fix server and unit test
- [x] Look into UI bug with contact detail's edit button
- [ ] Add test for ReviewSubmitV2

#### Related issues

https://jiraent.cms.gov/browse/MCR-3944

#### Screenshots
Screenshot 2024-03-05 at 12 33 29 PM
Screenshot 2024-03-05 at 12 33 38 PM
Screenshot 2024-03-06 at 4 10 38 PM

Screenshot 2024-03-05 at 12 34 09 PM
Screenshot 2024-03-05 at 12 34 00 PM

#### Test cases covered

- With the exception of the ReviewSubmitV2 page, all V2 sub components (RateDetails, ContactSummary, ContractSummary, SubmissionTypeSummary) have had their tests updated to work with contracts and rates

## QA guidance

- Use the app as a state user and create a new submission. Should be able to reach and view the Review Submit page (actual submission is out of scope)
- Cypress test with linked rates off passes
- server and unit tests are passing


services/app-web/src/testHelpers/apolloMocks/healthPlanFormDataMock.ts
```diff
@@ -124,6 +124,114 @@ function mockBaseContract(
     }
 }
 
+function mockContractAndRatesDraftV2(
```
pearl-truss, line 127: @haworku just a heads up that I added this mock

haworku:
Hey Pearl! 
I noticed a few things:
1. "Is this a risk-based contract?" radio value is not displaying. When I click "no" on Submission Type it saves but then the review and submit page still says this value is missing
2. Same thing for "Do you attest that this contract complies with all applicable statutory and regulatory requirements including those contained in Title 42 Part 438 and Part 457 of the Code of Federal Regulations (CFR)?" - radio value is not displaying.
3. The new API request is caching itself and this leads to broken UI. It means the new review and submit  (contract and rates) gets out of sync with the form pages (using HPP) as soon as the user clicks "edit" and makes change then comes back to review and submit.  Can you try adding `fetchPolicy: 'network-only' to your new request and see if that addressing the issue?

If it doesn't address things we can also override  [default settings](https://www.apollographql.com/docs/react/api/core/ApolloClient/#defaultoptions-watchquery) for ApolloClient to address this. I mainly scoped #2268 to removing caching on an existing page by page basis from the form and dashboards. But, now wondering if maybe we need a hammer and just turn it off everywhere all the time even outside the submission form. FYI The app wide config is found in the index page - `const apolloClient = new ApolloClient({` line.

haworku:
Also FYI I noticed you and I both have an issue with the header (which was build of HPP apis and useing submission side nav to update) not added the package name when a page is hard refreshed. The package name should be after the word "Virginia" in this photo. I can fix today in a follow- on PR, will wrap it up in the PR i'm making to fix error page displays.
Screenshot 2024-03-06 at 10 15 46 AM

pearl-truss:
@haworku thanks for the review! 1 and 2 have been addressed and for #3 (The new API request is caching itself and this leads to broken UI. ) a few things

- Switching to include `network-only` in query was enough to get the form data in sync. HOWEVER
- When going from a contract with rates to just a contract, `fetchContract` still returns rates. I don't think it's a caching issue necessarily because when I made the app wide apollo client config change I still see same behavior 

I'll continue to look into this but thought I would raise now

haworku (approved):
Approving! Things look good to me to merge (regardless of what you decide about rates stuff and where we fix that).

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/RateDetailsSummarySectionV2.test.tsx
```diff
@@ -0,0 +1,782 @@
...
+import { renderWithProviders } from '../../../../../testHelpers/jestHelpers'
+import { RateDetailsSummarySectionV2 as RateDetailsSummarySection } from './RateDetailsSummarySectionV2'
+import { Rate } from '../../../../../gen/gqlClient'
+import { testS3Client } from '../../../../../testHelpers/s3Helpers'
+
+describe('RateDetailsSummarySection', () => {
+    const draftContract = mockContractPackageDraft()
+    const submittedContract = mockContractPackageSubmitted()
+    const statePrograms = mockMNState().programs
+    const mockRateInfos: Rate[] = [
```
macrael, line 19: I strongly recommend making this a function that returns this array. As is, if any test were to modify their rates, it would impact all other tests connected to these mocks.

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/RateDetailsSummarySectionV2.test.tsx
```diff
@@ -0,0 +1,782 @@
...
+        },
+    ]
+
+    const apolloProvider = {
+        mocks: [
+            fetchCurrentUserMock({
+                statusCode: 200,
+                user: mockValidCMSUser(),
+            }),
+            indexHealthPlanPackagesMockSuccess(),
```
macrael, line 116: are these pages calling indexHealthPlan?

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/RateDetailsSummarySectionV2.tsx
```diff
@@ -0,0 +1,319 @@
...
+        const rateFormData = getRateFormData(rate)
+        if (rateFormData?.rateType === 'AMENDMENT') {
+            return 'Amendment to prior rate certification'
+        }
+        if (rateFormData?.rateType === 'NEW') {
+            return 'New rate certification'
+        }
+    }
+
+    const getRateFormData = (rate: Rate | RateRevision) => {
```
macrael, line 115: I'd love for you to explicitly define these functions's return types. The fact that they can return undefined leads to a bunch of ?'s later down in the code that might not be necessary.

macrael (approved):
This is a lot of work! 

Looking great. I have some notes about types and some refactoring. Excited to see this live. Are we tracking the work to fix the document table stuff? If I remember right that's used to set the "new" tags on edited submissions

services/app-web/src/common-code/ContractType.ts
```diff
@@ -0,0 +1,68 @@
+import { Contract } from '../gen/gqlClient'
+
+const isContractOnly = (contract: Contract): boolean => {
```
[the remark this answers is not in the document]
haworku: @macrael your comment here feels like it's raising a higher level data validation and types issue (rather than something about small helper functions). It seems to me a little out of scope for Linked Rate. Can we land this work and discuss in RSP?

>I might even suggest that you have that function throw an error if there's neither a draftRevision nor a packageSubmission. 

Want to flag that throwing errors from these types of helper functions means we make the lower level React components handling these helpers have to have error states. This introduces error handling complexity to small child components (right now this is just used in `FileRow`, what would that small chunk of UI display when theres an error for this?). This type of error show be handled way before any page mounts its children, imo at the API layer or right as we exit the API. It would be a higher level error if we had data lacking both a draft Revision AND a package submission... this helper shouldn't even be called.

services/app-web/src/index.tsx
```diff
@@ -57,17 +57,25 @@ Amplify.configure({
 
 const authMode = process.env.REACT_APP_AUTH_MODE
 assertIsAuthMode(authMode)
+const cache = new InMemoryCache()
+const defaultOptions: DefaultOptions = {
+    watchQuery: {
+      fetchPolicy: 'network-only',
+      errorPolicy: 'ignore',
```
haworku, line 64: Pearl - can we remove off any `errorPolicy` changes here and just scope this to `fetchPolicy`. Sorry I missed this. cc: @macrael

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/ContactsSummarySectionV2.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+export const ContactsSummarySection = ({
+    contract,
+    editNavigateTo,
+}: ContactsSummarySectionProps): React.ReactElement => {
+    const isSubmitted = contract.status === 'SUBMITTED'
+    const contractFormData =
+        contract.draftRevision?.formData ||
+        contract.packageSubmissions[0].contractRevision.formData
+    const rateRev = contract.draftRates
```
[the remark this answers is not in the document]
pearl-truss: From my understanding the UI that displays rate info for this component can be found in the most recent rate rev

services/app-web/src/common-code/ContractType.ts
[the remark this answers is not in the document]
pearl-truss: You both have good points. I just pushed up changes to abstract out the rev check to its own function without throwing any errors. Let me know if more changes are needed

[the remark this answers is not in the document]
macrael: pearl, I like your fix where we don't check if `contract.packageSubmissions[0]` exists we just return it. This implicitly does what I suggested! Now we'll get a '.submissionType does not exist on undefined' error if we ever fuck this up and return a malformed Contract. Still not quite as clean as it could be but I like that we aren't using null coalescing the same way in the body of the components. I think you could implement something similar for rates to not have to have that ! assertion. 

My issue with this was that instead we had silent error handling. All of our code for displaying bits of form data are using ? null coalescing to catch the error case of if we have a contract with no draft revision nor package submission and silently displaying nothing in that case. That means if we do manage to mess up a contract and send it like that to this component we will have no idea that we did, and we potentially introduce other errors down the line because we're allowing formData to be null when in truth this component really expects formData to exist.

pearl-truss:
> > This is a lot of work!
> > Looking great. I have some notes about types and some refactoring. Excited to see this live. Are we tracking the work to fix the document table stuff? If I remember right that's used to set the "new" tags on edited submissions
> 
> Thanks! I wrote up a [new ticket](https://jiraent.cms.gov/browse/MCR-3988) for the document table work. And I'm continuing to look into the possible caching issue. If I can't find a solution by this afternoon I'll carve that out into a separate ticket to track that work too

I started on a path of adding draftRates to the zod and prisma code but saw that was covered in a recent PR that was merged. With main pulled in, everything works

PR #2283, opened 2024-02-26
Add fetchContract resolver for contracts in DRAFT

mojotalantikite:
## Summary

This adds the `fetchContract` resolver for contracts that are in `DRAFT`. At this point we're only concerned with a contract that is in `DRAFT` and a rate that the contract is a 'parent' of. We'll add linked rates and such once that related code is completed.

Note that the ticket's AC mentioned three types of rates, however "submitted rates that have a different parent" implies a rate has been linked, which we don't have just yet.

#### Related issues
https://jiraent.cms.gov/browse/MCR-3939


haworku:
I had in my notes that this is going to be the `fetchContractWithRates` API and that there will be `updateContractWithRates` and then a `submitContractWithRates` down the road. I think we could rename this now unless something changed

macrael (approved):
This looks pretty great. I think we'll simplify things with the model changes that are coming but this looks like it works

macrael:
Oh, I forgot to reply to @haworku 's comment. I think this should remain called "fetchContract" since it is via the GraphQL system that rate data is returned. The core thing you are requesting is a contract and different queries could request just contract data or also choose to include the rate data. It's a little mismatchy because we have findContractWithRates on the db side, but I think it makes sense since the db including is happening in that function, it's not configurable by the caller.

PR #2294, opened 2024-03-02
add updateDraftContractRates mutation

macrael:
## Summary

This adds a mutation for updating and linking and creating rates to a draftContract. 

There's still some edge cases I think aren't checked yet, like trying to updated a linked rate that has a different parent. I'll work on those refinements when I get back. 

The big missing piece is ordering the draft rates. That will require more work and must come later.

#### Related issues

https://jiraent.cms.gov/browse/MCR-3941

#### Test cases covered

Added a bunch of tests covering the adds and removes.


services/app-api/src/resolvers/contract/updateDraftContractRates.test.ts
```diff
@@ -0,0 +1,686 @@
...
+        expect(nonexistantRateResult.errors).toBeDefined()
+        if (!nonexistantRateResult.errors) {
+            throw new Error('define errors')
+        }
+        expect(nonexistantRateResult.errors[0].extensions?.code).toBe(
+            'NOT_FOUND'
+        )
+    })
+
+    it.todo('doesnt allow updating a non-child')
```
haworku, line 684: Where are we testing the error path for trying to link or delink rates that do not exist or are not yet submitted?  That feels like a possible error if the frontend was reading rate IDs (for example passing revision id) or handling API data incorrectly

services/app-api/src/resolvers/contract/updateDraftContractRates.test.ts
```diff
@@ -0,0 +1,686 @@
...
+            result.data.updateDraftContractRates.contract.draftRates
+
+        const draftFormData = draftRates[0].draftRevision.formData
+
+        expect(draftFormData.rateType).toBe('AMENDMENT')
+        expect(draftFormData.rateDocuments).toHaveLength(1)
+        expect(draftFormData.rateDocuments[0].name).toBe('updatedratedoc1.doc')
+    })
+
+    it('doesnt allow updating a non-linked rate', async () => {
```
haworku, line 404:
```suggestion
    it('doesnt allow updating a non-child rate', async () => {
```
I think that is what we are testing here... right?

services/app-api/src/resolvers/contract/updateDraftContractRates.test.ts
haworku, line 1: Where are we testing trying to link or unlink a rate that does not exist (e.g. bad rate ID passed in)? Want to check that the error is clear for that case

services/app-api/src/postgres/contractAndRates/updateDraftContractRates.ts
```diff
@@ -0,0 +1,259 @@
...
+                stateCode: contract.stateCode,
+                stateNumber: thisRateNumber,
+                revisions: {
+                    create: prismaRateCreateFormDataFromDomain(rateFormDataem),
+                },
+            }
+        })
+
+        // to delete draft rates, we need to delete their revisions first
+        await client.rateRevisionTable.deleteMany({
```
mojotalantikite, line 198: Do we want to wrap this `deleteMany` and the `update` in a transaction? If an `update` on the contract revision fails we probably want to roll back the draft rate deletions?

mojotalantikite (approved):

PR #2295, opened 2024-03-04
Adding Link your rates component to storybook

ruizajtruss:
## Summary
This PR creates the `LinkYourRates` component and adds it to storybook

### DIFF
--- a/services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
+++ b/services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
@@ -1,0 +1,54 @@
+import React from 'react'
+import Select from 'react-select'
+import styles from '../../components/Select/Select.module.scss'
+import { useIndexRatesQuery } from '../../gen/gqlClient'
+import { Fieldset } from '@trussworks/react-uswds'
+
+export const LinkRateSelect = () => {
+    const { data, loading, error } = useIndexRatesQuery()
+
+    // const onFocus: AriaOnFocus<any> = ({
+    //   focused,
+    //   isDisabled,
+    // }): string => {
+    //   return `You are currently focused on option ${focused.label}${
+    //       isDisabled ? ', disabled' : ''
+    //   }`
+    // }
+
+    const noOptionsMessage = () => {
+        if (loading) {
+            return 'Loading rate certifications...'
+        }
+        if (error) {
+            return 'Could not load rate certifications. Please refresh your browser.'
+        }
+        if (!data) {
+            return 'No rate certifications found'
+        }
+    }
+
+    return (
+        <Fieldset
+            role="radiogroup"
+            aria-required
+            legend={'Which rate certification was it?'}
+        >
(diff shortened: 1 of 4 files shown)


services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export interface LinkRateOptionType {
+    readonly value: string
+    readonly label: string
+    readonly isFixed?: boolean
+    readonly isDisabled?: boolean
+}
+
+export const LinkRateSelect = () => {
+    const { data, loading, error } = useIndexRatesQuery()
```
macrael, line 15: consider using the function `wrapApolloResult` to get tighter types for this

services/app-web/src/pages/LinkYourRates/LinkYourRates.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+                    ],
+                },
+            }
+        )
+
+        screen
+            .getByLabelText(
+                'Yes, this rate certification is part of another submission'
+            )
+            .click()
```
macrael, line 107: In general you should use the [userEvent](https://testing-library.com/docs/user-event/intro) library instead of "click". I think it handles actions better. 

For an example of this, see APIAccess.test.ts

you can pull the `user` out of the return from `renderWithProviders` then call `user.click(screen.getByLabelText(...))`

Calling click() directly on stuff is supposedly less user like and I suspect can lead to weird behavior in the end

services/app-web/src/pages/LinkYourRates/LinkYourRates.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+
+import { Fieldset, FormGroup } from '@trussworks/react-uswds'
+
+import styles from '../StateSubmission/StateSubmissionForm.module.scss'
+import { FieldRadio } from '../../components'
+import { getIn, useFormikContext } from 'formik'
+import { LinkRateSelect } from './LinkRateSelect'
+
+// export type LinkYourRatesProps = {
```
macrael, line 10: delete

macrael (approved):
Cool! Nice work.

haworku:
I see a UI bug. The bottom of the dropdown has no border.
Screenshot 2024-03-07 at 11 17 53 AM

haworku (changes requested):
@ruizajtruss Good work. I would like to see this component take props. Specifically think we need the `fieldNamePrefix` in there (and probably other props also, but that's the main one I see missing).  We would use this in the `name` and `inputID` on the select component.  Right now as is, the component is not hooking into Formik fully. You can see this as soon as you add multi-rates. The component quickly loses track of the radio button being click yes or no and you click and its not highlighted.

 I can pair with you on this to get this in before merging!

haworku (approved):
Yep that works. I see that now formik is tracking state on multi-rates and I also see bottom border issue is fixed.

PR #2296, opened 2024-03-04
MCR-3940 Change RateDetails page to call fetchContract

haworku:
## Summary
Main work completed:
- Implement  `MULTI` RateDetails workflow
   - I saw some rendering issues after adding additional `useQuery` calls for the new APIs. I spent the most time on trying to investigate what is going on with re-renders.
- Change the API that is called to be conditional -  contract API if we are in MULTI, rate API if we are in SINGLE 

Other notes:
- noticed there's still something odd going on with `ProgramSelect` in terms of that UI triggering additional re-renders. However, addressing that seemed out of scope + needs more investigation. It is an existing issue.

#### Related issues
https://jiraent.cms.gov/browse/MCR-3940

#### Screenshots

UI with linked rate flag off (red box shows the packages with shared rates UI is present) 
Screenshot 2024-03-04 at 5 52 02 PM

UI with linked rate flag on (no red box - these fields are not visible and we should not validate on them)
Screenshot 2024-03-04 at 5 52 53 PM

#### Test cases covered
- I'm starting to write Jest tests. Will keep incrementally committing these.  Let's not hold up on the PR on tests though -   this is still featured flagged (in use on rate edit page and on new linked rate workflow). 
- Cypress must pass, that helps ensure I didn't change flag off behavior

## QA guidance
- **Open Network tab** - you should see `/graphql` request to fetch contract when the linked rates feature flag is on. 
- **Look at the UI** - There is no `PackagesWithSharedRates` UI shown when linked rates feature flag is on. Validations still work (e.g. no validation error for "You must select at least one submission" when you continue with no changes)
-  **Cypress** - Existing cypress tests for linked rates flag OFF still pass

### DIFF
--- a/services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
+++ b/services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
@@ -1,11 +1,9 @@
 import React from 'react'
-import { Form as UswdsForm } from '@trussworks/react-uswds'
+import { Button, Fieldset, Form as UswdsForm } from '@trussworks/react-uswds'
 import { FieldArray, FieldArrayRenderProps, Formik, FormikErrors } from 'formik'
-import { useNavigate, useParams } from 'react-router-dom'
+import { useNavigate } from 'react-router-dom'
 
 import styles from '../../StateSubmissionForm.module.scss'
-import { v4 as uuidv4 } from 'uuid'
-
 import {
     DynamicStepIndicator,
     ErrorSummary,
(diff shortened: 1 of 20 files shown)


services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
```diff
@@ -286,104 +359,212 @@ const RateDetailsV2 = ({
...
+                                                setSubmitting,
+                                                {
+                                                    type: 'CANCEL',
+                                                    redirectPath:
+                                                        'DASHBOARD_SUBMISSIONS',
+                                                }
+                                            )
+                                        } else {
+                                            navigate(
+                                                RoutesRecord.DASHBOARD_SUBMISSIONS
```
pearl-truss, line 537: I noticed that when I tried to click continue from the Rates Details page I was routed to the dashboard instead of the Contacts page which is what I was expecting

pearl-truss (approved):
Great job on this! I left one comment regarding navigation and also I noticed the linked rates flag isn't turned on yet for dev/review app so I tested this locally (wasn't sure if we wanted to turn that flag on in dev yet or not?)

PR #2299, opened 2024-03-05
Tweak tealium more - get it loading on page launch

haworku:
## Summary
I think my last fix  #2215 still didn't fully resolve this - let's try this. Moving the tealium call to be in the `.then` 

How to test: This can be tested in review apps by removing the environment specific guard clause (see below) and then putting a breakpoint or console logging where we call utag.view  

```// Do not add tealium for local dev or review apps
        if (process.env.REACT_APP_AUTH_MODE !== 'IDM') {
```

### DIFF
--- a/services/app-web/src/hooks/useTealium.ts
+++ b/services/app-web/src/hooks/useTealium.ts
@@ -14,6 +14,7 @@ import type {
     TealiumViewDataObject,
     TealiumEvent,
 } from '../constants/tealium'
+import { recordJSException } from '../otelHelpers'
 
 /*
 Tealium is the data layer for Google Analytics and other data tracking at CMS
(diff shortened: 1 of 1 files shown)


macrael (approved):
This feels weirder as time goes on. Googling a little I'm not seeing much about this not loading after a certain amount of time, but I am seeing that some content blockers can prevent it from loading. Not sure what behavior you're seeing but that's my quick look around. 

Anyway, if this fixes the issue I'm all for it

haworku:
@macrael Ah- I think you are missing the context from [this ticket ]( https://jira.cms.gov/browse/BLSTANALYT-7038). Take a look.

Basically, I saw it working _after_ the initial load when you switch pages.  That what I think no one else saw before. I used breakpoints on prod to determine that the script was called when expected. I also inspected loading time for the script in comparison to a breakpoint for the initial setup for Tealium. I could see there was a latency issue where the initial load page happened before the `utag.view` call. That meant we tried to log before the script fully loaded. That's why I believe the initial load continued to be missed and every time they looked in their tool it looked like nothing happening. The team testing things though doesn't have any login after all.  So all they could see really is the initial load.

macrael:
Thanks for the context!

PR #2301, opened 2024-03-06
Add V2 UnlockSubmitModal

haworku:
## Summary
Expands the type interface to handle either contract or rate unlock/submits. Includes TODO comments for future work needed in this file as we reimplement with contract and rates API

Background context: I had it on a branch locally from a few weeks ago when I was on the unlock + edit rate epic.  Although we delayed that epic in favor of Linked Rates, I decided to commit now so at least the file is up. We can build off it as we reimplement the unlock and submit modal to handle contract packages submit (linked rates epic). Then later we can come back around to rate package submit (unlock and edit rate).

#### Related issues
This was initially created during #2166 - I started to refactor the modal because I know that would be needed to land unlock and edit rates. 

#### Tests and QA
No tests. No QA needed. This code is all unused currently and will need to be tested. It will be used once we get to https://jiraent.cms.gov/browse/MCR-3974.


pearl-truss (approved):
thanks for getting this queued up

PR #2303, opened 2024-03-06
Fix: Error Page display (post sidebar) and Headings

haworku:
## Summary
- Fix design bugs from recent sprint work with error page displays and the page heading (the darker blue row in the header)

Other changes that enabled this:
- Deleted `PageHeading` component. It's not needed – we have another way we handle `h1` announcing now - which is via the setting focus on the Heading with `updateHeading`. That both updates the heading visually and resets the focus for a11y. 
- Add `contractName` to the contract revision data coming back from graphql. Did this to ensure this is always the same with contract name API rather than generating on the fly  (this leads to accessing state programs json all over the frontend code). 

#### Related issues
Design bugs came from recent work with #2296 and moving the sidebar to state submission form

#### Screenshots
**BEFORE**
Screenshot 2024-03-06 at 3 34 54 PM
Screenshot 2024-03-06 at 3 47 31 PM

**AFTER**
Screenshot 2024-03-06 at 3 35 48 PM
Screenshot 2024-03-06 at 3 46 03 PM

#### Test cases covered
Did not add tests

## QA guidance
- Trigger error pages in the app (not found or fetching draft submission as CMS user or fetching forbidden submission as state use)
- **Check that v2 Rate Details. Header should look correct and have the same contract name as that submission when its in the dashboard**


services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
```diff
@@ -196,8 +199,17 @@ const RateDetailsV2 = ({
...
     const initialRequestError = fetchContractError || fetchRateError
     const previousDocuments: string[] = []
+    const submissionIDForHeading =
+        fetchContractData?.fetchContract.contract?.draftRevision
+            ?.contractName ||
+        fetchRateData?.fetchRate.rate?.draftRevision?.formData
+            .rateCertificationName
+
+    useEffect(() => {
+        updateHeading({ customHeading: submissionIDForHeading ?? undefined })
```
haworku, line 209: FYI calling `updateHeading` is how I fixed the header issue @pearl-truss with the v2 pages. THe old pages did this in `useHealthPlanPackageForm`. 

 I can fix on review and submit v2 also once your work lands.

services/app-api/src/resolvers/contract/contractResolver.ts
```diff
@@ -32,7 +33,22 @@ export function contractResolver(store: Store): Resolvers['Contract'] {
         },
 
         draftRevision(parent) {
-            return parent.draftRevision || {}
+            const programsForContractState = statePrograms.states
+                .find((state) => state.code === parent.stateCode)
+                ?.programs.filter((program) => program !== undefined)
```
haworku, line 38: How do we feel about this? Making a backend change to fix design bug with headings (which are either package or rate names when we in the form). 

I was debating also just putting this directly on the `ContractFormData` but thats a bigger code change that touches more files. Thought I would throw this  up and get feedback

services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
pearl-truss: @haworku thanks! my work just landing and I'm reviewing this PR now

services/app-api/src/resolvers/contract/contractResolver.ts
pearl-truss: I think it's to approach it like this but I do think having it on `ContractFormData` would be nice so that it's similar to the rate name

pearl-truss (approved):
Great work!

macrael (approved):
I think that this does belong on the revision, since form data is all info that is being entered by users. We should move this to being it's own ContractRevisionResolver I think instead of adding it to the contract resolver, then it will work for revisions returned elsewhere instead of being special cased for draftRevision

haworku:
I am going to merge this as is because this fixes a design bug that right now is only on Rate Details page. I believe as is this PR makes an important incremental gain and makes linked rates work not look weird/sloppy and want to land it as is, particularly considering that the feedback suggest further refactors that could be done at any time. 

@pearl-truss suggests of the contract name and the rate name being calculated the same way and in the same place  - that makes sense to me. 

@macrael advises this could be in a revision resolver rather than just in draft revisions only that makes sens to me too. 

Seems to me we should have a ticket that does both of these things and maybe also delete rate certification name from the db. We could be unifying how we handle these names on the backend, reduce all the frontend tech debt around this all at the same time. I can make that

PR #2313, opened 2024-03-08
MCR-3900 Can continue or save as draft with new API

haworku:
## Summary

Happy path for Continue and Save As Draft - using the new  `updateDraftContractRates` mutation on Continue

- Change formik object to be called `rateForms`  instead of `rates`. I'm sorry for the code diff. It was getting very confusing to me in the code when we had `rates` as in GQL rates and when we had rate forms coming from Formik. Tried to make naming more distinct.  
- I moved `LinkYourRates` a level up so that I can control form data using the `FieldArray.replace` in next PR
- Fix route `navigate` call

#### Related issues
https://jiraent.cms.gov/browse/MCR-3900

#### Test cases covered
- Added tests for the new `generateUpdatedRates` utility
- Skip RateDetails tests again 

## QA guidance

- You can add and remove data from rate details V2 now with existing form data
**Note: Linked Rates radio and dropdown won't hook in yet - thats the next set of PRs**


haworku:
I think I'm seeing an issue with the API. It seems to be creating`draftRates` on the contract that are malformatted due to the `stateNumber`, I believe? Can investigate next week.  Surfacing for visibility @macrael 

Getting this error back
```
 Error from API fetch ApolloError: Issue finding a contract with history with id 2b18aa60-fec4-458e-82dd-b49d41690ef3. Message: [
  {
    "code": "too_small",
    "minimum": 1,
    "type": "number",
    "inclusive": true,
    "exact": false,
    "message": "Number must be greater than or equal to 1",
    "path": [
      "draftRevision",
      "rateRevisions",
      0,
      "rate",
      "stateNumber"
    ]
  },
  {
    "code": "too_small",
    "minimum": 1,
    "type": "number",
    "inclusive": true,
    "exact": false,
    "message": "Number must be greater than or equal to 1",
    "path": [
      "draftRates",
      0,
      "stateNumber"
    ]
  }
  ```

File: services/app-web/src/index.tsx [hunk 1]
```diff
@@ -62,7 +62,24 @@ Amplify.configure({
...
+            fields: {
+                formData: {
+                    merge: true,
+                },
+            },
+        },
+        RateRevision: {
+            fields: {
+                formData: {
+                    merge: true,
```
haworku, line 77: Getting GQL warnings (not blocking but they clutter the console) because form datas dont have `id` but are nested objects that are changing. 

https://www.apollographql.com/docs/react/caching/cache-field-behavior/#merging-non-normalized-objects 

Following docs, I believe this is the proper way to address but welcome other eyes.

macrael: interesting, funny because these aren't really stand alone objects, we just made formData to wrap the data up nicely. we could give them a fake ID that's the same as the enclosing revision if that helps. I will read more about the merging stuff

services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
```diff
@@ -199,70 +147,75 @@ const RateDetailsV2 = ({
...
-    */
-    const rates: Rate[] = React.useMemo(
+    const [updateDraftContractRates, { error: updateContractError }] =
+        useUpdateDraftContractRatesMutation()
+    const [submitRate, { error: submitRateError }] = useSubmitRateMutation()
+
+    // Set up data for form. Either based on contract API (for multi rate) or rates API (for edit and submit of standalone rate)
+    const ratesFromContract =
+        fetchContractData?.fetchContract.contract.draftRates // TODO WHEN WE IMPLEMENT UPDATE API, THIS SHOULD ALSO LOAD FROM ANY LINKED RATES
+    const initialRequestLoading = fetchContractLoading || fetchRateLoading
```
macrael, line 165: Ok, now I think I see where you are skipping requests. I'm a little surprised by this pattern, seems like a lot of overhead is required in order to use this component for both stand alone rate edit and inside of the RateDetails array. No need to change this now but I'd like to look at that after summit again.

services/app-web/src/pages/StateSubmission/RateDetails/V2/rateDetailsHelpers.ts
```diff
@@ -0,0 +1,115 @@
...
+            rateID: id,
+            type: isLinkedRate(status) ? 'LINK' : !id ? 'CREATE' : 'UPDATE',
+        }
+    })
+
+    return updatedRates
+}
+
+// convert from FormikRateForm to GQL RateFormData used in API
+const convertRateFormToGQLRateFormData = (
```
macrael, line 40: I like having these explicit converters

macrael (approved):
Looks good. I am a little surprised at the dual headed nature of this component but that's for another time

haworku:
> am a little surprised at the dual headed nature of this component but that's for another time
Yes hoping we remove this and that we get to actually do this right.  I still would like to see a SINGLE and MULTIPLE RateDetails v2 also with separation of concerns. 

However, I was finding keeping that interface clean and also figuring out how to encapsulate the API requests and not break the work I just did with single rate edit unlock wasmessy. Putting them together helped me see more clearly what's the same and whats different in one file.

pearl-truss (approved):
LGTM

services/app-web/src/pages/StateSubmission/RateDetails/V2/rateDetailsHelpers.ts
pearl-truss: Same, this is a great addition!

PR #2321, opened 2024-03-12
State can select from a list of previously submitted rates on rate details page

haworku:
## Summary

**Still to do for demo**:  
- [ ] Get `LinkRateSummary` to totally match Figma design specs
- [ ] Test "Continue"  when adding a combination of child rates and linked rates
- [ ] Test "Continue" when removing linked rates and / or child rates and then continuing 

**Not needed for demo but blockers for turning on Linked Rates for Production :**
- [ ] adjust single rate schema to ensure we never validate on linked rates (after all, this could prevent submission without the user knowing why is our submitted rates lack fields that are now required in the form validation
- [ ] add caching to indexRates call
- [ ] tests in place for linked rates both jest and cypress
_^ we need Jira tickets for this work or add to AC of remaining tickets_ 

**Notes on significant code changes:**
- adjusted `indexRates` to only return state user's state when a state user hits this API
- adjusted `LinkRateSelect`  to be a single select (not multi select)
- as soon as linked rate selected, override content in that  `rateForm` use the linked rate info instead. FieldArray `replace` will allow autofill of that linked rate. as soon as linked rate is cleared, clear the autofilled form data and go back to an empty `rateForm` 
- added `LinkRateSummary` to display instead of the form when there's a linked rate

#### Related issues
https://jiraent.cms.gov/browse/MCR-3900

#### Test cases covered
None yet.

## QA guidance
- I recommend testing this with a combination of adding and removing multiple rates, including  also adding both child rates and linked rates and removing each and hitting "Continue".

### DIFF
--- a/services/app-graphql/src/queries/indexRates.graphql
+++ b/services/app-graphql/src/queries/indexRates.graphql
@@ -75,6 +75,71 @@ query indexRates {
                             packageStatus
                         }
                     }
+                                     contractRevisions {
+                        id
+                        contract {
+                            id
+                            stateCode
+                            stateNumber
+                        }
+                        createdAt
+                        updatedAt
+                        submitInfo {
+                            updatedAt
+                            updatedBy
+                            updatedReason
+                        }
+                        unlockInfo {
+                            updatedAt
+                            updatedBy
+                            updatedReason
+                        }
+                        formData {
+                            programIDs
+                            populationCovered
+                            submissionType
+                            riskBasedContract
+                            submissionDescription
+                            stateContacts {
+                                name,
+                                titleRole
+                                email
+                            }
+                            supportingDocuments {
+                                name
+                                s3URL
+                                sha256
+                            }
+                            contractType
+                            contractExecutionStatus
+                            contractDocuments {
+                                name
+                                s3URL
+                                sha256
+                            }
+                            contractDateStart
+                            contractDateEnd
+                            managedCareEntities
+                            federalAuthorities
+                            inLieuServicesAndSettings
+                            modifiedBenefitsProvided
+                            modifiedGeoAreaServed
+                            modifiedMedicaidBeneficiaries
+                            modifiedRiskSharingStrategy
+                            modifiedIncentiveArrangements
+                            modifiedWitholdAgreements
+                            modifiedStateDirectedPayments
+                            modifiedPassThroughPayments
+                            modifiedPaymentsForMentalDiseaseInstitutions
+                            modifiedMedicalLossRatioStandards
+                            modifiedOtherFinancialPaymentIncentive
+                            modifiedEnrollmentProcess
+                            modifiedGrevienceAndAppeal
+                            modifiedNetworkAdequacyStandards
+                            modifiedLengthOfContract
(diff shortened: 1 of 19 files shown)


haworku:
@pearl-truss @ruizajtruss  @macrael As I mentioned in our little meeting a way I could see dividing up work here is one person works on the `LinkedRateSummary` page and making sure that matches Figma ---  while someone else works on making sure the linked rates to persist and can be linked/delinked on API mutation

services/app-graphql/src/queries/indexRates.graphql
```diff
@@ -74,6 +74,71 @@ query indexRates {
...
+                            modifiedStateDirectedPayments
+                            modifiedPassThroughPayments
+                            modifiedPaymentsForMentalDiseaseInstitutions
+                            modifiedMedicalLossRatioStandards
+                            modifiedOtherFinancialPaymentIncentive
+                            modifiedEnrollmentProcess
+                            modifiedGrevienceAndAppeal
+                            modifiedNetworkAdequacyStandards
+                            modifiedLengthOfContract
+                            modifiedNonRiskPaymentArrangements
```
pearl-truss, line 140: This seems to be missing new fields I added to contractFormData in a recentish PR for `statutoryRegulatoryAttestationDescription` and `statutoryRegulatoryAttestation`

services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
```diff
@@ -106,19 +112,51 @@ export const LinkRateSelect = ({
...
+            const linkedRateForm: FormikRateForm = convertGQLRateToRateForm(
+                getKey,
+                linkedRate
+            )
+            // put already selected fields back in place
+            linkedRateForm.ratePreviouslySubmitted = 'YES'
+            linkedRateForm.linkedRates = [
+                {
+                    rateId: linkedRateID,
+                    rateName: 'LOOK_FOR_ME', // linkedRateName,
```
pearl-truss, line 132: For now, are we intentionally keeping this for rateName?

pearl-truss (commented):
@macrael I left a few small comments but the main thing I'm seeing is that the No path doesn't work anymore. There's a 400 bad request coming through for that case. Otherwise everything else looks good

mojotalantikite:
I think I may have hit the same thing @pearl-truss did, which is that if I specify `No, this rate certification was not included with any other submissions" it works fine, but if I select `Yes, ...` and link a rate, but then switch it back to `No, ...` after an update, I get a system error.

services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
macrael: I'm a little confused on how this is used. linkedRateForm.linkedRates is not clicking for me. why does a rate have an array of linkedRates?

pearl-truss (approved):
Approving for the demo flow but I see an issue with clicking No -> going to the next page -> navigating back to Rate Details the 'No' selection isn't persisting. 

And also the 'Yes' selection (and all data inputted) gets cleared out when 'Save as draft' is selected

PR #2327, opened 2024-03-15
[MCR-3889] addressing notes from design review

ruizajtruss:
## Summary
This PR resolves the following:
- [x] Figure out why first option is scrolled out of view
- [x] Try to have 3 results minimum visible in the dropdown at open
- [x] Ensure its searchable
- [x] Have the dropdown expand by default
#### Related issues
https://jiraent.cms.gov/browse/MCR-3889

### DIFF
--- a/services/app-web/src/components/Select/Select.module.scss
+++ b/services/app-web/src/components/Select/Select.module.scss
@@ -4,6 +4,7 @@
 
 // react-select draws the chips in two parts, so we round the outer and square in the inner corners
 .multiSelect {
+    z-index: 4;
     padding-top: 12px;
     [class*='select__multi-value'] {
         border-radius: 99rem 99rem 99rem 99rem;
--- a/services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
+++ b/services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
@@ -30,39 +30,46 @@ export const LinkRateSelect = ({
     const user = loggedInUser as StateUser
     const statePrograms = user.state.programs
 
-    const rateNames = data?.indexRates.edges.map((edge) => {
-        const revision = edge.node.revisions[0]
-        return {
-            value: revision.id,
-            label: (
-                <>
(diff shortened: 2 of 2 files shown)


pearl-truss (commented):
@ruizajtruss when I tested this in the review app I'm seeing that it's saying there are no options even though there should be an eligible submission to link it to (I'll play around with this to see if I'm missing something in setting up the other submission). And I'm see that the drop down appears as soon as I click Yes. I think it was correct before when the dropdown only opened when you clicked into it

Screenshot 2024-03-18 at 11 24 31 AM

ruizajtruss:
> @ruizajtruss when I tested this in the review app I'm seeing that it's saying there are no options even though there should be an eligible submission to link it to (I'll play around with this to see if I'm missing something in setting up the other submission). And I'm seeing that the drop down appears as soon as I click Yes. I think it was correct before when the dropdown only opened when you clicked into it
> 
> Screenshot 2024-03-18 at 11 24 31 AM

It might be an issue with the setup, I'm not seeing anything code wise that might be an issue at the moment. And yes! I had the dropdown setup to only open if clicked but I was asked to have it open by default when `Yes` is selected so that's why I made the change

ruizajtruss:
I checked out the review app and I believe there are no options because no reviews exist? Do we know if there are pre-existing rates in the review app for the Toph user that we expect to appear? I tried submitting my own rate to check that but it didn't work.

ruizajtruss:
> @ruizajtruss when I tested this in the review app I'm seeing that it's saying there are no options even though there should be an eligible submission to link it to (I'll play around with this to see if I'm missing something in setting up the other submission). And I'm seeing that the drop down appears as soon as I click Yes. I think it was correct before when the dropdown only opened when you clicked into it
> 
> Screenshot 2024-03-18 at 11 24 31 AM

Was able to verify this just stems from Virginia not having any pre-existing rates.

haworku (approved):
Yes! see this working the search working and it saving as draft

PR #2332, opened 2024-03-19
[MCR-3996] link rates chooser default and filter values

ruizajtruss:
## Summary
This pr prepopulates the select field with existing linked rates

#### Related issues
[MCR-3996](https://jiraent.cms.gov/browse/MCR-3996)

## QA guidance

1. Navigate to the rate details page for a contract with an existing rate(s)
2. Select Yes
3. field should populate with existing rates if any

### DIFF
--- a/services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
+++ b/services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
@@ -30,39 +30,46 @@ export const LinkRateSelect = ({
     const user = loggedInUser as StateUser
     const statePrograms = user.state.programs
 
-    const rateNames = data?.indexRates.edges.map((edge) => {
-        const revision = edge.node.revisions[0]
-        return {
-            value: revision.id,
-            label: (
-                <>
-                    <strong>{revision.formData.rateCertificationName}</strong>
-                    <div style={{ lineHeight: '50%', fontSize: '14px' }}>
-                        <p>
-                            Programs:&nbsp;
-                            {programNames(
-                                statePrograms,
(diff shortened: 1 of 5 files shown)


pearl-truss:
@ruizajtruss I'm seeing the same issue with other linkable rates not appearing in the dropdown

services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
```diff
@@ -106,15 +115,23 @@ export const LinkRateSelect = ({
         }
     }
 
+    //Need this to make the label searchable since it's buried in a react element
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const filterOptions = ({ label }: any, input: string) =>
```
macrael, line 120: what does this do?

macrael (commented):
I made default value work in my PR, can you see if your work is necessary now?

services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
ruizajtruss: You can use `filterOption` to customize how the search filtration logic works for the dropdown menu. The dropdown menu search works on label and value _strings _ but the label is currently a react element. So I leveraged the filterOptions prop to dig into the react element so it can search by the rate name. I had to include `.toLowerCase()` to keep it case-insensitive as well.

ruizajtruss:
> I made default value work in my PR, can you see if your work is necessary now?

Went over this and paired with Hana as well, decided to scrap this PR 🎉

PR #2337, opened 2024-03-22
Fix another problemTealium load

haworku:
## Summary
My last fixed addressed what it should have (initial page loads work). However, it caused regression with subsequent page load functionality. Did more messing with the function calls to get both working again. 

The burden of making changes and waiting to production to verify is obnoxious. I couldn't think of a good enough reason why review apps can't also send analytics (sure it pollutes the dev analytics data... but we aren't checking that). By allowing review apps to send analytics it makes it easier for someone in code review to actually check it works and speed feedback loop.

#### Related issues
This is based on further follow up from the BlastAnalytics team in the CMS slack 

#### Test cases covered
- n/a

## QA guidance
- Add a breakpoint at the `utag.view` function call inside new `callUTagView` function in dev tools on the review app
- This breakpoint should be hit every time we change pages on the application.

### DIFF
--- a/services/app-web/src/hooks/useTealium.ts
+++ b/services/app-web/src/hooks/useTealium.ts
@@ -40,9 +40,9 @@ const useTealium = (): {
     // this effect should only fire on initial app load
     useEffect(() => {
         // Do not add tealium for local dev or review apps
-        if (process.env.REACT_APP_AUTH_MODE !== 'IDM') {
-            return
-        }
+        // if (process.env.REACT_APP_AUTH_MODE !== 'IDM') {
+        //     return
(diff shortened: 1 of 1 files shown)


haworku:
Ugh I'm seeing this firing multiple times for certain page loads now. For example clicking "start new submission" makes it get evaluated three times.  Going to try to figure out whats going on thats causing unnecessary re-renders thats a problem  (i bet its the page heading context). This is an issue for the app in general not just for analytics.

pearl-truss (approved):

mojotalantikite (approved):

PR #2340, opened 2024-03-25
fix ReviewSubmitV2: Hide shared rates across submission when linked rates flag is on 

haworku:
## Summary
We want to show the legacy UI (rates across submissions) only for the submission summary experience. This is for historical data reasons.

For unlocked contracts currently being edited - e.g. Review and Submit - let's hide the old SHARED tag and related submissions field in the documents table.

#### Related issues
product/design review on 
https://jiraent.cms.gov/browse/MCR-3886

#### Test cases covered
- renders shared rate certifications when submission is locked
- does not render shared rate cert info if submission is being unlocked and being edited

## QA guidance
- Look at legacy submission that has the old "rates across submissions" feature - such as[ MCR-MN-016](https://d25pfhkipqgt0p.cloudfront.net/submissions/4a1e13c9-1678-4a01-8d9c-fda00c694d68)
- See shared tag should be visible and linked submissions column is present
- Turn ON linked rates field
-  See shared tag should NOT be visible and linked submissions column will NOT be present on ReviewSubmitb2

### DIFF
--- a/services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/RateDetailsSummarySectionV2.test.tsx
+++ b/services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/RateDetailsSummarySectionV2.test.tsx
@@ -754,30 +754,25 @@ describe('RateDetailsSummarySection', () => {
         ).toBeNull()
     })
 
-    it('renders all necessary information for documents with shared rate certifications', async () => {
-        const draftContract = mockContractPackageDraft()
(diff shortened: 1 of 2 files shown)


ruizajtruss (approved):
LGTM!

macrael (approved):
makes sense. And so for future submitted rates that don't have this field, refreshPackagesWIthSharedRateCert will return undefined, right?

haworku:
@macrael yes! from reading the code that what is expected. we can easily confirm this once we have ability to submit new rates and submit unlocked rates

PR #2348, opened 2024-04-01
[MCR-3996] Filter our already linked rates from dropdown

ruizajtruss:
## Summary
This PR filters already selected/linked rates from the dropdown options list

### DIFF
--- a/services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
+++ b/services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
@@ -5,9 +5,13 @@ import { StateUser, useIndexRatesQuery } from '../../gen/gqlClient'
 import { useAuth } from '../../contexts/AuthContext'
 import { programNames } from '../../common-code/healthPlanFormDataType'
 import { formatCalendarDate } from '../../common-code/dateHelpers'
-import { FormikRateForm } from '../StateSubmission/RateDetails/V2/RateDetailsV2'
+import {
+    FormikRateForm,
+    RateDetailFormConfig,
+} from '../StateSubmission/RateDetails/V2/RateDetailsV2'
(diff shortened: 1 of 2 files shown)


services/app-web/src/pages/LinkYourRates/LinkYourRates.test.tsx
```diff
@@ -46,80 +52,322 @@ describe('LinkYourRates', () => {
...
-                <form>
-                    <LinkYourRates
-                        fieldNamePrefix="rateForms.1"
-                        index={1}
-                        autofill={jest.fn()}
-                    />
-                </form>
-            </Formik>,
+            <Routes>
+                <Route
```
haworku, line 109: As written, these are now RateDetailsV2 page level integration tests. They are not component level tests for `<LinkYourRates />` anymore.  Instead, logic here hinges `RateDetailsV2` continuing to be set up a certain way= and they will fail if that parent page file changed.  

I'm fine with that! However, please just move them up into `RateDetailsV2.test.tsx` then - put them together in their own nested describe block, e.g. `can link existing rate`

services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
```diff
@@ -135,13 +140,22 @@ export const LinkRateSelect = ({
             ?.toLowerCase()
             .includes(input.toLowerCase())
 
+    //We track rates that have already been selected to remove them from the dropdown
+    const selectedRates = values.rateForms.map((rate) => rate.id && rate.id)
```
haworku, line 144: @ruizajtruss I'm not seeing this code working in the review app. Down to  pair together - I could be missing something.

I started to wonder if it is just not r-rendering the options list when the values change

haworku (approved):

PR #2350, opened 2024-04-02
Submission Summary V2 with FetchContract

pearl-truss:
## Summary

This PR 
- creates a V2 SubmissionSummary page that uses the new `fetchContract` API
- Updates ReviewSubmitV2 to use `submitContract` so you should be able to test this without needing to switch feature flags on and off
- Added `mccrsID` to `fetchContract`

#### Related issues
https://jiraent.cms.gov/browse/MCR-4014

#### Screenshots

Screenshot 2024-04-15 at 3 42 43 PM
Screenshot 2024-04-15 at 3 42 50 PM
Screenshot 2024-04-15 at 3 42 55 PM
Screenshot 2024-04-15 at 3 43 02 PM


services/app-graphql/src/queries/fetchSubmittedContract.graphql
```diff
@@ -0,0 +1,299 @@
+query fetchSubmittedContract($input: FetchSubmittedContractInput!) {
```
pearl-truss, line 1: I'm very curious how people with more graphql experience feel about this? I created this query because `fetchContract` returns the `draftRevisions` and `draftRates` field that are null for submitted contracts . This was resulting in a `Cannot return null for a non nullable field` for `contract.draftRevision.id`

haworku: High level my questions are not about the graphql but the types pattern - to me this change could suggest a different mental model for types coming out of the API.  
- Would you envision then that we narrow the `status` field too - so only `SUBMITTED` or `RESUBMITTED` here?
-  What would be the API response then when you try to fetch a submitted contract that does exist by ID in DB, it is just not submitted (maybe its unlocked for example or it's an initial draft)? 
- If we went forward with this kind of type - should it be defined in `codegen.yml` as well? Not sure what defines what should go in that file.

As I write it out, these kinds of questions feel familiar. They have come up before. It makes me wonder are hitting a limitation in our current loose data model? I think we intentionally did not want to add back locked and unlocked types like we used to have...but is there a case for them coming up here?

haworku: More specific to the error -  @pearl-truss could you tell me how to replicate the ` Cannot return null for a non nullable field for contract.draftRevision.id` issue - just interested to see what you are talking about, it sounds kind of familiar.

pearl-truss: @haworku You can replicate it by replacing `useFetchSubmittedContractQuery` with `useFetchContractQuery` and using the V1 submit page to create a submission

And those are good questions! I think a major downside to what I have here is the implications for the ReviewSubmit page since that page is for both draft and submitted contracts. I don't yet have ideas on how to address that. And yeah I think it would make sense to narrow down the status field. If the status of the contract is draft or unlock, it could make sense to return an input error. And I'm not super familiar with codegen.yml but after taking a look it, I think it might make sense to include submittedContracts there but I need to look into it more because I'm not sure where this gets used

services/app-web/src/components/ChangeHistory/ChangeHistoryV2.tsx
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    const revisionHistory = flattenedRevisions()
+
+    const revisedItems: AccordionItemProps[] = revisionHistory.map(
+        (r, index) => {
+            const isInitialSubmission = r.updatedReason === 'Initial contract'
+            const isSubsequentSubmission = r.kind === 'submit'
+            // We want to know if this package has multiple contracts. To have multiple contracts, there must be minimum
+            // 3 revisions in revisionHistory the initial contract revision, unlock revision and recontract revision.
```
haworku, line 66: nit: this sentences confuses me. A package can only have one contract. It can have multiple contract submissions in its history, it can have multiple rates, it can have multiple related submissions (via linked rates) but I dont think it can have multiple contracts

services/app-web/src/components/Modal/V2/UnlockSubmitModalV2.tsx
```diff
@@ -31,9 +34,15 @@ const RATE_UNLOCK_SUBMIT_TYPES = [
...
+const CONTRACT_UNLOCK_SUBMIT_TYPES = [
+    'SUBMIT_CONTRACT',
+    'RESUBMIT_CONTRACT',
+    'UNLOCK_CONTRACT',
+] as const
 type PackageModalType = (typeof PACKAGE_UNLOCK_SUBMIT_TYPES)[number]
 type RateModalType = (typeof RATE_UNLOCK_SUBMIT_TYPES)[number]
-type SharedModalType = PackageModalType & RateModalType
+type ContractModalType = (typeof CONTRACT_UNLOCK_SUBMIT_TYPES)[number]
+type SharedModalType = PackageModalType & ContractModalType & RateModalType
```
haworku, line 45: We can fully delete the package modal types - this new v2 will be only contract or rate. I think I just didn't get that yet in the refactor.

services/app-web/src/gqlHelpers/mutationWrappersForUserFriendlyErrors.ts
```diff
@@ -154,6 +156,42 @@ export const submitMutationWrapper = async (
...
+        } else {
+            recordJSException(
+                `[UNEXPECTED]: Error attempting to submit, no data present but returning 200.`
+            )
+            return new Error(ERROR_MESSAGES.submit_error_generic)
+        }
+    } catch (error) {
+        return handleApolloErrorsAndAddUserFacingMessages(
+            error,
+            'SUBMIT_HEALTH_PLAN_PACKAGE'
```
haworku, line 190:
```suggestion
            'SUBMIT_CONTRACT'
```

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/SubmissionTypeSummarySectionV2.tsx
```diff
@@ -81,59 +80,70 @@ export const SubmissionTypeSummarySectionV2 = ({
...
-                        children={
-                            contractFormData.contractType
-                                ? ContractTypeRecord[
-                                      contractFormData.contractType
-                                  ]
-                                : ''
-                        }
-                    />
-                    {contractFormData.riskBasedContract !== null && (
+                    {contractFormData && (
```
haworku, line 83: Could we exit earlier if there is no `contractFormData`. Having these loose types means we create unexpected UI states. We shouldn't be rendering any part of this pagel if theres no `contractFormData` - if there's no contract functionally and we are in a full page error.

 Take a look at `ReviewSubmitV2` for a way to handle this -` if (!contractFormData) return <GenericErrorPage />` happens early.

services/app-web/src/common-code/ContractType.ts
```diff
@@ -1,36 +1,37 @@
...
-const getContractRev = (contract: Contract): ContractRevision => {
+const getContractRev = (contract: Contract): ContractRevision | undefined => {
     if (contract.draftRevision) {
         return contract.draftRevision
     } else {
-        return contract.packageSubmissions[0].contractRevision
+        return getLastContractSubmission(contract)?.contractRevision
     }
 }
 const isContractOnly = (contract: Contract): boolean => {
```
haworku, line 11: No change needed now just sharing thought when I see all these `?` introduced to our utility functions.  

Have all these optional checks around the contract revision's existence means the boolean because less sure - we could be getting `false` for mal-formatted data, not just for it not being a good contract. I like seeing utilities, especially on the frontend like this (which literally determining display state) be more strictly typed. We would have to handle data validation/ correctness checks much earlier in the React component, divert to an error state if theres any question about the data shape.  Also starting to wonder if helpers should take the `ContractFormData` not the `Contract` along those lines - maybe have the component itself figure out which form data to pass? Not sure.

services/app-web/src/components/ChangeHistory/ChangeHistoryV2.tsx
pearl-truss: Thanks, I could use extra eyes on this [change ](https://github.com/Enterprise-CMCS/managed-care-review/pull/2350/files#diff-1c7f359d5e6984ab5d140a1fd69d46a4490e742ab901468f6707274c00d1c244R67). I think we no longer want the cutoff to happen at >= 3?

services/app-web/src/components/ChangeHistory/ChangeHistoryV2.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    const revisionHistory = flattenedRevisions()
+
+    const revisedItems: AccordionItemProps[] = revisionHistory.map(
+        (r, index) => {
+            const isInitialSubmission = r.updatedReason === 'Initial contract'
+            const isSubsequentSubmission = r.kind === 'submit'
+            // We want to know if this contract has multiple submissions. To have multiple submissions, there must be minimum
+            // more than the initial contract revision.
+            const hasSubsequentSubmissions = revisionHistory.length > 1
```
pearl-truss, line 67: Previously this was set to >=3:

```
// We want to know if this package has multiple submissions. 
// To have multiple submissions, there must be minimum
            
// 3 revisions in revisionHistory the initial submission revision, 
// unlock revision and resubmission revision.
            
const hasSubsequentSubmissions = revisionHistory.length >= 3
```

services/app-web/src/components/ChangeHistory/ChangeHistoryV2.tsx
```diff
@@ -0,0 +1,135 @@
...
+}
+
+export const ChangeHistoryV2 = ({
+    contract,
+}: ChangeHistoryProps): React.ReactElement => {
+    const flattenedRevisions = (): flatRevisions[] => {
+        const result: flatRevisions[] = []
+        //Reverse revisions to order from earliest to latest revision. This is to correctly set version for each
+        // contract & recontract.
+        const reversedRevisions = [...contract.packageSubmissions].reverse()
```
macrael, line 24: Do we want to put package submissions in time order? I was matching revisions b/c calling contract.packageSubmissions[0] feels like a nice way to always get the most recent submission but could be convinced otherwise

services/app-web/src/components/ChangeHistory/ChangeHistoryV2.tsx
```diff
@@ -0,0 +1,135 @@
...
+                const newUnlock: flatRevisions = {} as flatRevisions
+                newUnlock.updatedAt = r.contractRevision.unlockInfo.updatedAt
+                newUnlock.updatedBy = r.contractRevision.unlockInfo.updatedBy
+                newUnlock.updatedReason =
+                    r.contractRevision.unlockInfo.updatedReason
+                newUnlock.kind = 'unlock'
+                //Use unshift to push the latest revision unlock info to the beginning of the array
+                result.unshift(newUnlock)
+            }
+            if (r.contractRevision.submitInfo) {
```
macrael, line 36: this is necessary b/c it is optional in the types, but just to reiterate every revision in packageSubmissions must have this set.

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/RateDetailsSummarySectionV2.tsx
```diff
@@ -77,8 +78,8 @@ export const RateDetailsSummarySectionV2 = ({
         contract.draftRevision?.formData ||
         getLastContractSubmission(contract)?.contractRevision.formData
     const rates =
-        contract.draftRates ||
```
macrael, line 80: little weird to me that we're just grabbing one or the other. 

I think we should be pulling from one or the other depending on isEditing. if isEditing then we are reading from draftRates, if not, then the latest submitted rates

services/app-web/src/pages/SubmissionSummary/V2/SubmissionSummaryV2.tsx
```diff
@@ -0,0 +1,299 @@
...
+        featureFlags.CMS_QUESTIONS.flag,
+        featureFlags.CMS_QUESTIONS.defaultValue
+    )
+
+    // API requests
+    const {
+        data: fetchContractData,
+        loading: fetchContractLoading,
+        error: fetchContractError,
+    } = useFetchContractQuery({
```
macrael, line 83: I know Hana isn't sold on it but I still think that the ApolloQueryWrapper is useful, makes some impossible cases impossible

macrael (approved):
looks great!

services/app-web/src/components/ChangeHistory/ChangeHistoryV2.tsx
haworku, line 24: 🔍 Is this code working properly with linked rates?  

 I am under the impression that we would only want contract related package submissions in our change right now. I _think_ at this point we only seek parity with current `ChangeHistory` display (e.g. even if there are linked rates we are not yet  adding in linked rates to change history). To achieve this, we would need to only map contract changes from packageSubmissions array to this `reversedRevisions` array. 
 
 Change history work with the full package submissions list is entirely different epic I believe
 
 @macrael can you weigh in?

macrael: I think that if you filtered out all packages that dont have reason: "CONTRACT_SUBMISSION" then you will have the same number of entries as before. That will also give us the correct unlock reasons, we should only ever show unlocks from CONTRACT_SUBMISSION, not from RATE* updates

I definitely want this code to be using packageSubmissions and get off of the old revisions code

pearl-truss: @macrael I think we do if we want to keep the order the same as what currently appears (unless I'm missing something?)
Screenshot 2024-04-18 at 2 50 16 PM

pearl-truss: @haworku that makes sense to me

services/app-web/src/pages/App/AppRoutes.tsx
```diff
@@ -216,7 +232,13 @@ const CMSUserRoutes = ({
                     )}
                     <Route
                         path={RoutesRecord.SUBMISSIONS_SUMMARY}
-                        element={<SubmissionSummary />}
+                        element={
```
pearl-truss, line 235: @haworku @macrael should the CMS version of the submissionSummary use the V2 page? They have the 'unlock contract' button that isn't currently covered

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/RateDetailsSummarySectionV2.tsx
haworku: @macrael I _believe_ it's because of types that we do this.Everything is optional, so you want to narrow it early to avoid that optionality messing up the UI.  If you do the current implementation of `isEditing` then either value could still be undefined which trickles down. 

Def think we could use type assertions or more precise tools eventually. Maybe we need a design ticket on that? However, right now coming out GraphQL we  don't have a union type that' s either a draft or a locked package, you have a type two optional subfields that determine what is being used, so you need to clear that early.

services/app-web/src/components/ChangeHistory/ChangeHistoryV2.tsx
haworku: I don't think so. Multiple pages use that `contract.packageSubmissions[0] ` logic already and its working well. This is edge case - I think it's literally just for that integer in the previous submission urls why we reverse in this order -  everywhere else assumes latest first.

haworku: @pearl-truss  sounds like @macrael agrees, we should be filtering only the contract_submission packages here - can you make that change and ignore rate stuff? Otherwise looks good.

haworku (approved):

services/app-web/src/pages/App/AppRoutes.tsx
haworku: Ugh! Scoop creep. Make it a new ticket. Yes it should because thats the only way CMS can see linked rates stuff. 

Maybe we could get away fetching with new API and doing the unlock with the old HPP API and the old unlock on this v2 page for this one case on the V2 page. Need to look

pearl-truss: I created the ticket [here](https://jiraent.cms.gov/browse/MCR-4052)

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/RateDetailsSummarySectionV2.tsx
macrael: WIth this change, getting draftRates after checking for last submission, as a state user you won't see the current version of the rates if you are editing a submission that has been submitted before. Looking at it again I think this is a bug

PR #2351, opened 2024-04-04
MCR-3905 CMS links from rate summary to parent submission to unlock a rate

haworku:
## Summary
This is interim UI for when Linked Rates is turned on but unlock and edit for standalone rates is still turned off. 

For that case, we render an unlock rate button but redirect to the first related contract submission so the rate can be unlocked with the contract submission unlock. 

#### Related issues
https://jiraent.cms.gov/browse/MCR-3905

#### Test cases covered
-  `renders unlock button that redirects to contract submission page when standalone rate edit and unlock is disabled`

## QA guidance
- get a submission in the dashboard
- log in as CMS user with linked rates feature flag ON but unlock edit rates flag OFF
- CMS user can click "Unlock rate" and they are redirected to the submission summary of the contract that made the rate

### DIFF
--- a/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/SingleRateSummarySection.tsx
+++ b/services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/SingleRateSummarySection.tsx
@@ -24,7 +24,7 @@ import { useS3 } from '../../../contexts/S3Context'
 import useDeepCompareEffect from 'use-deep-compare-effect'
 import { recordJSException } from '../../../otelHelpers'
 import { Link } from '@trussworks/react-uswds'
-import { NavLink } from 'react-router-dom'
+import { NavLink, useNavigate } from 'react-router-dom'
 import { packageName } from '../../../common-code/healthPlanFormDataType'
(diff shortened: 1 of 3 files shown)


services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/SingleRateSummarySection.tsx
```diff
@@ -225,7 +230,8 @@ export const SingleRateSummarySection = ({
             )
         }
     }
-
+    const parentContractSubmissionID =
```
haworku, line 233: @macrael Does this seem like a reasonable path to get the contract submission ID of the contract that originally was created with that rate? Or is there some other way to get to the parent contract

macrael: Shouldn't it be `rate.revisions[rate.revisions.length -1]...`. We should probably include it in the rate itself from the API rather than calculate it everywhere

haworku: Ah yes, that sounds more right. It's the first added to the array 😮‍💨  I will make that change. TY

pearl-truss (approved):

PR #2352, opened 2024-04-05
MCR-3988 Update the Dates in Documents Table to use Contract data

haworku:
## Summary
This was was needed so that Date Added and the "NEW" tag works in the `UploadedDocumentsTable` on V2 pages we have built with the Contract and Rate API. 

Changes:
- Add `dateAdded` field to backend on each `GenericDocument`. This required some type weaving and test fixing because we have a lot of strong assumptions about the form data and how it works in the application.
- Add `prismaContractRateAdaptors` to share the prisma operations used on .create and .update between database functions
- Create a new shared zod types for `ContractEditableType` and `RateEditableType` before these types were defined in random places in postgres and graphql code. 
- Remove `isSubmitting` prop on `<UploadedDocumentsTable />`, just rely on `isEditing`. Also make this component take in the `previousSubmissionDate` as a prop and accept documents that already know their own `dateAdded`.

#### Related issues
https://jiraent.cms.gov/browse/MCR-3988

#### Test cases covered
- Existing tests still pass including uploaded documents table

## QA guidance
- Submit a package
- Log in as a CMS user and check the submission summary page documents tables. Those tables should have a Date Added column with a date in place. 
- Date added is present  BOTH on the Contract Details summary and on the Rate Details summary  (one is a v2 page one is not)


pearl-truss (approved):
This was a lot of work! Code looks good and things are working as expected

services/app-api/src/postgres/contractAndRates/prismaContractRateAdaptors.ts
```diff
@@ -0,0 +1,294 @@
+import type {
+    ContractFormEditableType,
+    RateFormEditableType,
+} from '../../domain-models/contractAndRates'
+import type { GenericDocument } from '../../gen/gqlServer'
+import { emptify, nullify } from '../prismaDomainAdaptors'
+
+// ADD RETURN TYPES FROM PRISMA SEE "includes" "satifies"
+// contractTableFullPayload - contractTableCreateArgs / contract
```
haworku, line 9: FYI @macrael I did try adding return types here like  `Prisma.ContractRevisionTableCreateInput` or a partial of that type but it was resulting in lots of lint errors. Even when I resolved those (about 40 hours) I was seeing test failures. Did not feel confident at all in the change. Also looked at using `satisfies` keyword and Prisma.validator and got a bit confused. Maybe you could take a look? 

Either way can I get a 👍 that this is good to merge without the types in place for this spirnt

PR #2358, opened 2024-04-09
Prevent rates from being read as a linked rate just because its unlocked

pearl-truss:
#### Related issues

https://jiraent.cms.gov/browse/MCR-4000

#### Screenshots
Screenshot 2024-04-09 at 12 16 07 PM

## QA guidance

- Turn Linked rates feature flag off
- Make a contract with rates submission
- Sign in as a CMS user and unlock the submission
- Go to the submission as a state user and navigate to the Rate Details page. It should now display as No for the linked rates question and the other fields should be editable


haworku (approved):
Nice!

PR #2365, opened 2024-04-16
Move actuary contacts section to rate details page

ruizajtruss:


services/app-web/src/pages/StateSubmission/Contacts/Contacts.tsx
```diff
@@ -491,173 +470,6 @@ const Contacts = ({
                                         </FieldArray>
                                     </fieldset>
                                 </SectionCard>
-
```
haworku, line 494: Asked a question about this in Slack. The way this is implemented right now we would just be disappearing the additional actuary fields from the "flag off" state for LINK_ RATES.

That would be a regression on production, where LINK_RATES is not turned on. I wonder if this is related to the cypress issues you are seeing (not knowing what they are)

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
```diff
@@ -216,6 +241,17 @@ export const RateDetails = ({
 
         draftSubmission.rateInfos = cleanedRateInfos
 
+        const addtlActuaryContacts: ActuaryContact[] = []
+        rateInfos.forEach((rateInfo) => {
+            rateInfo.addtlActuaryContacts.forEach((contact) => {
+                addtlActuaryContacts.push(contact)
+            })
+        })
+        draftSubmission.addtlActuaryContacts = addtlActuaryContacts
```
JasonLin0991, line 250: There are more changes needed here. The way we saved `addtlActuaryContacts` using protobuffs will not work with the implementation of the UI here. The gist here is that, in the old way we took all `addtlActuaryContacts` from the package level and copied them to each rate. So now we need to do it correctly and save each `addtlActuaryContacts` to their respective rate in `toProtoBuffer`. 

Here is a list of things I found so far that need to be changed.

- Update `toProtoBuffer` to not save `addtlActuaryContacts` at the package level, instead `rateInfos` should save its own additional actuary contacts in the `addtlActuaryContacts` at the rate level.
- `RateDetails`, remove `addtlActuaryContacts` from the `draftSubmission`. (line 250)
- In the API remove `addtlActuaryContacts` from the HPP level and only use the rate level `addtlActuaryContacts`
    - In the API `resolverHelpers.ts`  update the `convertHealthPlanPackageRatesToDomain` function to not use the package level `addtlActuaryContacts`.
    - In `convertContractWithRatesToHPP.ts` the function `convertContractWithRatesToFormData` Remove `addtlActuaryContacts` from HPP level and add to rate level.

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
```diff
@@ -216,6 +241,17 @@ export const RateDetails = ({
...
+
+        draftSubmission.addtlActuaryCommunicationPreference =
```
JasonLin0991, line 252: Does the `addtlActuaryCommunicationPreference` need to be on each rate for flag off or does the single radio apply to all rates?

services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
```diff
@@ -539,6 +578,66 @@ const RateDetailsV2 = ({
                                             </>
                                         )}
                                     </FieldArray>
+                                    <SectionCard>
```
JasonLin0991, line 581: From the [figma](https://www.figma.com/file/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?type=design&node-id=12176-104410&mode=design&t=rf1bdhrjcAryfOlh-0) link, it looks like each rate has its own actuary com preference.

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
JasonLin0991: I think with the current code you will end up duplicating all the contacts across all the rates.

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/ReviewSubmitV2.tsx
```diff
@@ -76,7 +76,10 @@ export const ReviewSubmitV2 = (): React.ReactElement => {
     }
 
     const isStateUser = loggedInUser?.role === 'STATE_USER'
-    const contractFormData = getVisibleLatestContractFormData(contract, isStateUser)
+    const contractFormData = getVisibleLatestContractFormData(
+        contract,
```
JasonLin0991, line 80: You'll have to update `ReviewSubmit.tsx` to read the additional actuary contracts from each rate instead of the package level `addtlActuaryContacts`.

JasonLin0991 (changes requested):
Still some more things to clean up in the protos and the old API in regards to `addtlActuaryContacts`.

Actuary com preference when flag is on should appear in each rate instead of a single section. The question is if this applies to flag off (I think it should).

services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.tsx
JasonLin0991: I would look around more in the API to make sure I didn't miss a spot where this happens.

services/app-web/src/common-code/proto/healthPlanFormDataProto/toProtoBuffer.ts
```diff
@@ -239,7 +248,11 @@ const toProtoBuffer = (
                                   rateInfo.rateAmendmentInfo.effectiveDateEnd
                               ),
                           },
-                          actuaryContacts: rateInfo.actuaryContacts.map(
+                          /**
+                           * Not making more changes to the proto schema, instead additional actuaries are added to the
+                           * actuaryContacts array from index 1.
+                           */
```
haworku, line 254: Great thanks for all these comments in the `toProtobuffer` file thats helpful.

services/app-api/src/domain-models/contractAndRates/convertContractWithRatesToHPP.ts
```diff
@@ -225,6 +212,7 @@ function convertContractWithRatesToFormData(
                     contractRev.formData.modifiedNonRiskPaymentArrangements,
             },
         },
+        addtlActuaryContacts: [],
```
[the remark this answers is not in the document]
JasonLin0991: I'll add a comment. Type reasons and a bunch of code, mostly tests, would need to be fixed if we removed this.

haworku:
Left a comment in Slack about this - can we take a look again at the place these fields display on Review and Submit. Fields that are assigned by rate should be displayed by rate in the Summary page UI. 

 I didn't see any issues with Rate Details for either flag state  (besides small styles things that feel out of scope right now). That work looked good to me ⭐

JasonLin0991 (approved):

PR #2349, opened 2024-04-01
Contract + Rates shared history

macrael:
## Summary

Contract + Rates Submission

This PR implements submission and returns the package submission history for submitted contracts and rates. 

Submission submits the contract and any unsubmitted rates. We use the [design described in Notion](https://www.notion.so/Linked-Rates-History-DB-Design-b48817ceafb04d2b9a340f44864919bc) to write to the assorted tables on submit that then get read from in fetchContract.

#### Related issues
https://jiraent.cms.gov/browse/MCR-3942
https://jiraent.cms.gov/browse/MCR-3966

#### Test cases covered
MANY TESTS HAVE CHANGED

because I moved the logic for submitting rates into contracts, and because we have to keep track of parent contracts, the API no longer allows a rate to be submitted initially on its own. Many tests had to change to follow the rule that a rate is always initially submitted with a specific contract. 

## QA guidance
If you can make something break by submitting linked rates, please write a test case! I'm still working on a few more.

PR #1827, opened 2023-07-21
MR-3578 Fix Date Added document date bug

haworku:
## Summary
Use the `submitInfo.updatedAt` instead of the proto `updatedAt` to determine the date added for documents.

#### Related issues
https://qmacbis.atlassian.net/browse/MR-3578

#### Test cases covered
`UploadedDocumentsTable` has some new tests to reflect the new behavior on review and submit with new submissions (don't show any date/empty string)

## QA guidance
I ended up separating off the refactor so this change is pretty contained to uploaded documents behavior i would test that across submissions

PR #2369, opened 2024-04-18
fix: MCR-3988 document dates working again

haworku:
## Summary
This PR is related to #1827. That code only partially fixed the problem. 

Product review found there were still caseswhere `dateAdded` was missing in the UploadedDocumentsTable. It seemed to depend on if package was submitted once or resubmitted. 

I found that 
1. #2349 fixed one of the bugs. I was seeing an issue with linked rates flag on where the `packageSubmission` array was empty on an unlocked submission and only the draft revision was populated. However, after merging main today that bug seems gone
2. [8b8044b](https://github.com/Enterprise-CMCS/managed-care-review/pull/2369/commits/8b8044b12551a2bd44f931855cfdac08e2b4ce64) fixed the bug for linked rate flag off on a package that's only been submitted once and not unlocked I think this might have been an existing bug not related to recent work.  That file hadn't been touched in 9 months.

#### Related issues
https://jiraent.cms.gov/browse/MCR-3988

#### Test cases covered
I added one unit test. 

I also added a check to cypress that `N/A` is not appearing on submission summary pages  (that's the screen reader only experience when fields are missing).  

## QA guidance
   
Steps for full regression test of date added document table

1. Set up submissions with flag off in various states (draft, unlocked, submitted)
2. Test these cases BOTH with flag on and flag off

Submission summary
- date added on document tables for submitted packages
- date added on document tables for unlocked packages

Rate summary
- date added on document tables for submitted rates
- date added for document tables for unlocked rates

ReviewSubmit (aka state user view)
- date added on document tables on initial draft packages
-  date added on  document tables for submitted packages
- date added on  document tables for submitted packages 

QuestionResponse
- date added on question and  response tables for cms user(no need to test both cms and state user for this UI)


services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
```diff
@@ -309,7 +309,7 @@ export const ContractDetailsSummarySection = ({
                         ? new Date(
                               documentDateLookupTable.previousSubmissionDate
                           )
-                        : undefined
```
macrael, line 312: Curious why we had to switch from undefined to nulls here. Totally fine but I'm curious. in general I only expect to see nulls when dealing directly with Apollo GQL data

macrael (approved):
Looks good! Your change to the mock data may conflict with stuff on Pearl's branch

services/app-web/src/components/SubmissionSummarySection/ContractDetailsSummarySection/ContractDetailsSummarySection.tsx
haworku: Made this choice for React reasons, not GQL. Because this was previously an optional prop, the component is used in about 6 places, and the bug was that the date was present in some expected places and not in others it was very for me difficult to tell if data was coming back unexpected as undefined OR if it was just a case where the prop wasn't passed in at all intentionally.

I changed to a required prop. That means places using this component have to be clear when there is no previousSubmissionDate using explicit null

PR #2382, opened 2024-04-25
Migrate Old Submissions to New Contract+Rates relationship tables

macrael:
## Summary

Migration moves old contract + rates to new relationship tables. 

Used the old join table to copy into the new RelatedSubmissions and PackageSubmissionJoinTables

There is a test that you can run repeatedly locally if you un-skip it.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4016


haworku (approved):
@mojotalantikite can you please code review also.

PR #2384, opened 2024-04-25
Correct Previous Submission Link

macrael:
## Summary

We had put "contracts" in the url for previous submissions instead of "submissions" leading to 404s.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4067


haworku (approved):
yay

PR #2385, opened 2024-04-25
move document date processing to end of parseContractWithHistory

macrael:
## Summary

We are seeing broken date handling with the document dateAdded fields.

This is the first pass at fixing those, but is specific to fetchContractWithHistory. 

This same fix will need to be applied to fetchRateWithHistory

But I think the idea that this is an in-place change that happens after putting together the list of revisions feels like the right direction. (though thinking about it, linked rates might lose some of that... not every linked rate revision is going to be in packageSubmissions)

It's definitely a bit of a tricky fix, b/c I'm updating the revision object's dates in situ. But that means there are actually a bunch of copies of those revisions we need to update:

* contract.submissionPackage.contractRevision
* contract.submissionPackage.rateRevisions
^ These are the two this update fixes, and in a world where we are only use package submissions that should be enough

But there are many more:
* contract.revisions
* contract.revisions.rateRevisions
* rate.revisions
* rate.revisions.contractRevisions
* rate.submisisonPackages.contractRevision
* rate.submisisonPackages.rateRevisions

because the dates require looking through all the revisions to figure out when the doc first appeared, I'm not sure where else to do this, because we need an array of them to set it up.

I'm afraid this code has been broken for a minute, it was written to look for documents in a previous revision but I'm not seeing anywhere where the code is actually passed a previous revision to look at it. That might have gotten lost in the shuffle of the recent contract+rates PRs

The migration exacerbates things b/c it is changing the updateAt of all our revisions it looks like. 

Probably the long term fix is to stop duplicating documents across different tables and consolidating them into a single documents table with sha or name as the primary key.

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
+++ b/services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
@@ -3,6 +3,7 @@ import type {
     ContractRevisionWithRatesType,
     ContractRevisionType,
     RateType,
+    RateRevisionType,
 } from '../../domain-models/contractAndRates'
 import { contractSchema } from '../../domain-models/contractAndRates'
 import type { ContractPackageSubmissionType } from '../../domain-models/contractAndRates/packageSubmissions'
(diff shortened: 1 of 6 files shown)


haworku:
It seems like we don't need to do 
rate.submissionPackages.contractRevision
rate.submissionPackages.rateRevisions

Those aren't supported in the app currently (not used anywhere that I can see on frontend display).

haworku (approved):
I would be fine to merge this as is its an improvement. Review apps look good too for contracts at least.

PR #2392, opened 2024-04-29
Create SubmissionRevisionSummaryV2 to use Contract data

pearl-truss:
## Summary

- Update SubmissionRevisionSummary to use fetchContract
- Skipped test will be addressed in a follow on PR for this ticket

#### Screenshots

Screenshot 2024-04-30 at 10 46 01 AM
Screenshot 2024-04-30 at 10 46 13 AM

## QA guidance
- Create a linked rate submission
- Unlock and then resubmit the submission with a different rate
- Revision page should show the correct rate name for each revision reversion

### DIFF
--- a/services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/ContactsSummarySectionV2.tsx
+++ b/services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/ContactsSummarySectionV2.tsx
@@ -12,15 +12,22 @@ import {
     DataDetailContactField,
 } from '../../../../../components/DataDetail'
 import { SectionCard } from '../../../../../components/SectionCard'
-import { Contract, RateRevision } from '../../../../../gen/gqlClient'
+import {
+    Contract,
+    RateRevision,
+    ContractRevision,
+} from '../../../../../gen/gqlClient'
 import {
     getDraftRates,
     getLastContractSubmission,
+    getVisibleLatestContractFormData,
 } from '../../../../../gqlHelpers/contractsAndRates'
 
 export type ContactsSummarySectionProps = {
     contract: Contract
+    contractRev?: ContractRevision
     editNavigateTo?: string
+    isStateUser: boolean
 }
 
 export const getActuaryFirm = (actuaryContact: ActuaryContact): string => {
@@ -41,13 +48,17 @@ export const getActuaryFirm = (actuaryContact: ActuaryContact): string => {
 
 export const ContactsSummarySection = ({
     contract,
+    contractRev,
     editNavigateTo,
+    isStateUser,
 }: ContactsSummarySectionProps): React.ReactElement => {
     const isSubmitted = contract.status === 'SUBMITTED'
-    const contractFormData =
-        contract.draftRevision?.formData ||
-        getLastContractSubmission(contract)?.contractRevision.formData
+    const contractOrRev = contractRev ? contractRev : contract
 
+    const contractFormData = getVisibleLatestContractFormData(
+        contractOrRev,
+        isStateUser
+    )
     let rateRev: RateRevision | undefined = undefined
 
     if (contractFormData?.submissionType === 'CONTRACT_AND_RATES') {
--- a/services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummaryV2.test.tsx
+++ b/services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummaryV2.test.tsx
@@ -25,7 +25,7 @@ describe('SubmissionRevisionSummary', () => {
                 apolloProvider: {
                     mocks: [
                         fetchCurrentUserMock({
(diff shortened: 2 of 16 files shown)


services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/ContactsSummarySectionV2.tsx
```diff
@@ -12,15 +12,22 @@ import {
...
     getDraftRates,
     getLastContractSubmission,
+    getVisibleLatestContractFormData,
 } from '../../../../../gqlHelpers/contractsAndRates'
 
 export type ContactsSummarySectionProps = {
     contract: Contract
+    contractRev?: ContractRevision
     editNavigateTo?: string
+    isStateUser: boolean
```
haworku, line 30: nit **(no change needed now)**: To me, it feels preferable to just get the `loggedInUser` user from the `AuthContext` to access user info rather than adding more props. One of the benefits of react context is to avoid prop drilling. FYI if you ever use the auth context for the user and want to test it --  in tests you can pass in the cms or state user in the `getCurrentUserMock` to shift what user is stored there. Just make sure your test is async. 

This is a pattern I've been meaning to comment on before, wanted to share the thought forconsideration while it is fresh in my mind.

services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummaryV2.tsx
```diff
@@ -0,0 +1,158 @@
...
+            return (
+                <ErrorForbiddenPage
+                    errorMsg={fetchContractError.graphQLErrors[0].message}
+                />
+            )
+        } else if (
+            fetchContractError?.graphQLErrors[0]?.extensions?.code ===
+            'NOT_FOUND'
+        ) {
+            return <Error404 />
```
haworku, line 88: See `RateDetailsv2` for simpler way to handle this:
```
    // Display any full page interim state resulting from the initial fetch API requests
    if (initialRequestLoading) {
        return <ErrorOrLoadingPage state="LOADING" />
    }

    if (apiError) {
        return (
            <ErrorOrLoadingPage state={handleAndReturnErrorState(apiError)} />
        )
    }
  ```

File: services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummaryV2.test.tsx [hunk 3]
```diff
@@ -0,0 +1,101 @@
...
+        //     .tz('America/New_York')
+        //     .format('MM/DD/YY h:mma')} ET version`
+        // expect(await screen.findByText(submissionVersion)).toBeInTheDocument()
+        expect(
+            
+        await screen.findByTestId('previous-submission-banner')
+        ).toBeInTheDocument()
+            screen.debug()
+    })
+    // it('extracts the correct dates from the submission and displays them in tables', async () => {
```
haworku, line 60: I don't even think we need this test at this level now since we are using the shared components entirely.

services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummaryV2.tsx
```diff
@@ -0,0 +1,158 @@
...
+    const isCMSUser = loggedInUser?.role === 'CMS_USER'
+
+    const {
+        data: fetchContractData,
+        loading: fetchContractLoading,
+        error: fetchContractError,
+    } = useFetchContractQuery({
+        variables: {
+            input: {
+                contractID: id ?? 'unknown-contract',
```
haworku, line 47: I think we are missing some key logic here -- - assuming we are copying previous behavior the `id` on the route params here  - this URL param not the `id` of the contract it is a number that represents the number of this revisions  (e.g. a number like 1 for first ,2 for second ,3 for third , etc etc). 

 If the intent here is not clear - let's talk.  Its a weird thing we only do for urls for previous submissions and not where else - you would have to look at ChangeHistory and SubmissionRevisionSummary side by side to see how these URLs work.

pearl-truss: We can meet to talk about this, because I'm seeing that the route contains the id `/submissions/:id/revisions/:revisionVersion` and that the previous SubmissionRevisionSummary also pulled the id out to fetch the hpp

I think you're right that I was missing pulling out the `revisionNumber`. It's needed to find the index of the submission from packageSubmissions

haworku: Makes sense. Reading it this morning. my comment wasn't quite right 😮‍💨 😆 either. Yes the revision version matters also is the point.

haworku (approved):
Approving just in case there's partial work to ship here - I'm seeing all v2 files changes so its behind feature flag

haworku:
LGTM.

PR #2397, opened 2024-04-30
Fix unlock removed child rate crasher

macrael:
## Summary

If you removed a child rate, on unlock it would error saying you can't unlock an already unlocked rate

#### Related issues

https://jiraent.cms.gov/browse/MCR-4093

#### Test cases covered

made a new API test that reproduced the problem

## QA guidance

try the steps in the bug


pearl-truss (approved):
Good job! I didn't run into any errors following the steps

PR #2409, opened 2024-05-01
fix contract history index

macrael:
## Summary

off by one error computing index. We want the indices for history revisions to be 1 indexed, not 0 indexed

#### Related issues

https://jiraent.cms.gov/browse/MCR-4100

Should be able to look at past submissions in all their glory now


haworku (approved):
TY

PR #2415, opened 2024-05-02
Add unlock action to change history for drafts

pearl-truss:
## Summary

This PR fixes an issue with unlock events not appearing in the change history when the submission is a draft (state hadn't resubmitted yet)

#### Related issues
https://jiraent.cms.gov/browse/MCR-4103

#### Screenshots

Screenshot 2024-05-03 at 11 53 08 AM

#### Test cases covered

- submit and unlock events are displayed in the change history

## QA guidance

- Log in as a CMS user and unlock a submission
- Ensure that the unlock event gets logged in the change submission (without needing to the state user to first resubmit)

### DIFF
--- a/services/app-web/src/components/ChangeHistory/ChangeHistoryV2.test.tsx
+++ b/services/app-web/src/components/ChangeHistory/ChangeHistoryV2.test.tsx
@@ -1,0 +1,63 @@
+import { screen } from '@testing-library/react'
+import { renderWithProviders } from '../../testHelpers/jestHelpers'
+import { ChangeHistoryV2 as ChangeHistory } from './ChangeHistoryV2'
+import {
+    mockContractPackageSubmitted,
+    fetchCurrentUserMock,
+    mockValidStateUser,
+    mockValidCMSUser,
+    mockContractPackageUnlocked,
+} from '../../testHelpers/apolloMocks'
+
+describe('SubmissionTypeSummarySection', () => {
+    afterEach(() => {
+        jest.clearAllMocks()
+    })
+    const submittedContract = mockContractPackageSubmitted()
+
+    it('can render history for initial submission', () => {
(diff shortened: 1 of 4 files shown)


services/app-web/src/components/ChangeHistory/ChangeHistoryV2.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+    mockValidStateUser,
+    mockValidCMSUser,
+    mockContractPackageUnlocked,
+} from '../../testHelpers/apolloMocks'
+
+describe('SubmissionTypeSummarySection', () => {
+    afterEach(() => {
+        jest.clearAllMocks()
+    })
+    const submittedContract = mockContractPackageSubmitted()
```
macrael, line 16: why define this outside the test? In general it's best to call the mock functions every time you use one so that different tests don't smash each others' data

services/app-web/src/components/ChangeHistory/ChangeHistoryV2.tsx
```diff
@@ -28,35 +28,54 @@ export const ChangeHistoryV2 = ({
```
macrael, line 28: We're not performing the above filter in the SubmissionRevisionSummary component so our indices won't line up. I forget why we decided to filter these out in the first place? Either way the two should match

macrael (approved):
looks good

services/app-web/src/components/ChangeHistory/ChangeHistoryV2.tsx
pearl-truss: It came up in this PR comment that we don't want to pull unlock info from rates https://github.com/Enterprise-CMCS/managed-care-review/pull/2350#discussion_r1571208151. I updated the SubmissionRevisionSummary to also have this filter

PR #2416, opened 2024-05-02
display id and submitted at date in past revision summary

macrael:
## Summary

A couple things were missing from the submission revision summary page, this adds them back.

#### Related issues

https://jiraent.cms.gov/browse/MCR-4100

For a submission that is older than today, please confirm that the Submitted date on a previous revision shows correctly

### DIFF
--- a/services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/SubmissionTypeSummarySectionV2.tsx
+++ b/services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/SubmissionTypeSummarySectionV2.tsx
@@ -70,21 +70,22 @@ export const SubmissionTypeSummarySectionV2 = ({
                 {headerChildComponent && headerChildComponent}
             </SectionHeader>
             <dl>
-                {(isSubmitted || (!isStateUser && isUnlocked)) && (
-                    <DoubleColumnGrid>
-                        <DataDetail
-                            id="submitted"
-                            label="Submitted"
-                            children={
-                                <span>
-                                    {dayjs(initiallySubmittedAt).format(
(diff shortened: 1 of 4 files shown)


services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/SubmissionTypeSummarySectionV2.tsx
```diff
@@ -70,21 +70,22 @@ export const SubmissionTypeSummarySectionV2 = ({
                 {headerChildComponent && headerChildComponent}
             </SectionHeader>
             <dl>
-                {(isSubmitted || (!isStateUser && isUnlocked)) && (
```
haworku, line 73: Can we get a test please

haworku (approved):

services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummaryV2.tsx
```diff
@@ -54,7 +54,7 @@ export const SubmissionRevisionSummaryV2 = (): React.ReactElement => {
 
     const name =
         contract &&
-        contract?.packageSubmissions.length < Number(revisionVersion)
+        contract?.packageSubmissions.length > Number(revisionVersion)
```
haworku, line 57: We might want to revisit this log later. Why on earth would we be rendering this page if theres not enough submissions lol. I would expect us to show an error or something here right - not just show an empty name string

macrael: we do show an error below, but bc the name is rendered with the updateHeading effect we have to repeat the check here, I think

PR #2417, opened 2024-05-02
MCR-4105 fix: validations and inline errors for linked rates

haworku:
## Summary
Link rates error summary links not working and missing inline error

Also added:
- unifying styles around errors (we need that red border to the right of all error form fields)
- change aria label to say linked rate (singular) since you can only link one
- fixed how validations work for NO button click (we were still validating for the linked rate field if the user had started to select something there earlier but abandoned it) 

#### Related issues
https://jiraent.cms.gov/browse/MCR-4105

#### Test cases covered
error summary displays as expected 
- when rate previously submitted question is not answered
- when rate previously submitted question is answered with YES
- when rate previously submitted question is answered with NO

## QA guidance
Recheck all validation behavior for the multiple rate experience 

IF NOTHING IS SELECTED
we display errors about the previously submitted yes/no question being empty

IF YES IS SELECTED
we display errors about the missing the rate certification if rate dropdown is still empty

IF NOT IS SELECTED
we display errors for the rate form, we do not display any errors from rate dropdown

🐛 Also check bugs that were fixed here

- all links in the errors summary `<a />` links work and when you click jump you to form fields
- click yes, hit continue, hit no. Only the validations for form should appear.

### DIFF
--- a/services/app-web/src/pages/StateSubmission/RateDetails/RateDetailsSchema.ts
+++ b/services/app-web/src/pages/StateSubmission/RateDetails/RateDetailsSchema.ts
@@ -10,166 +10,177 @@ import {
 Yup.addMethod(Yup.date, 'validateDateFormat', validateDateFormat)
 
 const SingleRateCertSchema = (_activeFeatureFlags: FeatureFlagSettings) =>
-    Yup.object().when('.ratePreviouslySubmitted', {
-        // this first when is skipping all validations for a previously submitted rate.
-        is: 'YES',
-        then: Yup.object().shape({
-            linkedRatesYesNo: Yup.string().defined('You must select a rate certification'),
-        }),
-        otherwise: Yup.object().shape({
-            ratePreviouslySubmitted: _activeFeatureFlags['rate-edit-unlock'] || _activeFeatureFlags['link-rates']? Yup.string().defined(
-                "You must select yes or no "
-            ) : Yup.string(),
-            rateDocuments: validateFileItemsListSingleUpload({ required: true }),
-            supportingDocuments: validateFileItemsList({ required: false }),
-            hasSharedRateCert:  _activeFeatureFlags['rate-edit-unlock'] || _activeFeatureFlags['link-rates']?  Yup.string(): Yup.string().defined('You must select yes or no'),
-            packagesWithSharedRateCerts: _activeFeatureFlags['rate-edit-unlock'] || _activeFeatureFlags['link-rates']? Yup.array().optional():  Yup.array()
-            .when('hasSharedRateCert', {
-                is: 'YES',
-                then: Yup.array().min(
-                    1,
-                    'You must select at least one submission'
-                ),
-            })
-            .required(),
-            rateProgramIDs: Yup.array(
-            ).min(1, 'You must select a program'),
-            rateType: Yup.string().defined(
-                'You must choose a rate certification type'
-            ),
-            rateCapitationType: Yup.string().defined(
-                "You must select whether you're certifying rates or rate ranges"
-            ),
-            rateDateStart: Yup.date().when('rateType', (contractType) => {
-                if (contractType) {
-                    return (
-                        Yup.date()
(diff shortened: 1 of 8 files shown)


services/app-web/src/pages/StateSubmission/RateDetails/RateDetailsSchema.ts
```diff
@@ -10,166 +10,177 @@ import {
 Yup.addMethod(Yup.date, 'validateDateFormat', validateDateFormat)
 
 const SingleRateCertSchema = (_activeFeatureFlags: FeatureFlagSettings) =>
```
haworku, line 12: I changed this schema by 1. removing rate edit unlock feature flag logic from linked rates fields since seems like now this logic just hinges on linked rates 2. making it a three part validation instead of two part validation to err on side of being more explicit with the logic tree

services/app-web/src/pages/StateSubmission/RateDetails/RateDetailsSchema.ts
```diff
@@ -10,171 +10,176 @@ import {
...
-    Yup.object().when('.ratePreviouslySubmitted', {
-        // this first when is skipping all validations for a previously submitted rate.
-        is: 'YES',
-        then: Yup.object().shape({
-            linkRateSelect: Yup.string().defined('You must select a rate certification'),
```
JasonLin0991, line 17: We changed `linkRateSelect` to `linkedRateDropdown` there are a couple places we still need to make the same changes.

RateDetailsV2.tsx

```ts
export type FormikRateForm = {
    ...
    linkRateSelect?: string
}
```

rateDetailsHelpers.ts

```ts
const convertGQLRateToRateForm = (getKey: S3ClientT['getKey'], rate?: Rate, parentContractID?: string): FormikRateForm => {
    const handleAsLinkedRate = rate && rate.parentContractID !== parentContractID // TODO: Make this a more sophisticated check for child-rates
    const rateRev = handleAsLinkedRate ? rate?.revisions[0] : rate?.draftRevision
    const rateForm = rateRev?.formData
    return {
        ...
        linkRateSelect: handleAsLinkedRate && rate?.id ? 'true' : undefined
    }
}
```

services/app-web/src/pages/StateSubmission/RateDetails/RateDetailsSchema.ts
```diff
@@ -10,171 +10,176 @@ import {
...
+                    })
+                ),
+                actuaryCommunicationPreference: Yup.string().required(
+                    'You must select a communication preference'
+                )
+            })
     })
 const RateDetailsFormSchema = (activeFeatureFlags?: FeatureFlagSettings) => {
-    return activeFeatureFlags?.['rate-edit-unlock'] ||  activeFeatureFlags?.['link-rates'] ?
-    Yup.object().shape({
```
JasonLin0991, line 173: This function needs to return the `Yup.object`. 

Also, validations are not working because RateDetailsV2 expects `rateForms` and RateDetails(V1) expects `rateInfos`. Whichever one we want this to be, either V1 or V2 has to be refactored to use it.

JasonLin0991 (commented):
We chatted about the validations not working and some errors with tests. I was able to see what was going on and left some comments just to save you some time.

Otherwise, the schema changes you made are A++. Once the validations are working again I'll do more QA on the review app.

services/app-web/src/pages/StateSubmission/RateDetails/RateDetailsSchema.ts
JasonLin0991: Also, failing `rateDetailsSchema.test.ts` tests were due returning `rateInfo` instead of `rateForms`.

haworku:
FYI @JasonLin0991 Also been adding schema work for unlocked rates. I realized that page now is showing linked rates field which is shouldn't. Adding a test for that as well

JasonLin0991 (approved):
Looks great!

PR #2427, opened 2024-05-06
MCR-4113: Additional certifying actuary styling tweaks

JasonLin0991:
## Summary
[MCR-4113](https://jiraent.cms.gov/browse/MCR-4113)
[JIRA](https://www.figma.com/file/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?type=design&node-id=12474-67689&mode=design&t=RlVYLESkNaOpscTn-0)

- Update the `Add a certifying actuary` button styling to an outlined button.
- Fixed "Remove certifying actuary" and "Remove rate certification" button color.

### DIFF
--- a/services/app-web/src/pages/StateSubmission/RateDetails/V2/SingleRateFormFields.tsx
+++ b/services/app-web/src/pages/StateSubmission/RateDetails/V2/SingleRateFormFields.tsx
@@ -536,9 +536,9 @@ export const SingleRateFormFields = ({
                                         </div>
                                     )
                                 )}
-                            <Button
+                            <button
                                 type="button"
-                                unstyled
+                                className={`usa-button usa-button--outline ${styles.addRateBtn}`}
                                 onClick={() => {
                                     push(emptyActuaryContact)
                                     setFocusNewActuaryContact(true)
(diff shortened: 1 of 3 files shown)


services/app-web/src/pages/StateSubmission/RateDetails/V2/SingleRateFormFields.tsx
```diff
@@ -536,17 +536,17 @@ export const SingleRateFormFields = ({
                                         </div>
                                     )
                                 )}
-                            <Button
+                            <button
```
haworku, line 539: There is a react-uswds outline button too - not sure if that would work here but that would be first try to keep using the design system https://trussworks.github.io/react-uswds/?path=/story/components-button--outline

haworku (approved):

services/app-web/src/pages/StateSubmission/RateDetails/V2/SingleRateFormFields.tsx
JasonLin0991: Good catch!

PR #2430, opened 2024-05-07
Clear yes/no selection for ratePreviouslySubmitted field

pearl-truss:
## Summary

This PR is a small enhancement that makes the No selection behavior the same as the yes flow when a user selects an option for the yes/no Was this rate included in another submission' without filling in any additional info. The selection gets cleared for either selection

The implementation is in rateDetailsHelpers. I found that specific fields had to be omitted for checking if they were filled out or not because a few of the fields are auto populated on No selection. I initially was trying to add checks inside of RateDetails as a part of the save as draft handler flow. If there are other ideas  please let me know

#### Related issues

https://jiraent.cms.gov/browse/MCR-4108

## QA guidance

- As state user on RateDetails page select No to the 'Was this rate included in another submission'
- Without filling out any fields, click 'Save as draft'
- Returning to the page the previously selected No, is no longer selected


JasonLin0991 (approved):
Looks good. Just needed to add the certifying actuary field and add some parentheses.

PR #2432, opened 2024-05-08
MCR-40116 & MCR-4115: Fix linked rate select styling and accessbility

JasonLin0991:
## Summary
[MCR-4116](https://jiraent.cms.gov/browse/MCR-4116)
[MCR-4115](https://jiraent.cms.gov/browse/MCR-4115)

- Fix `Add another rate certification` and `Add a certifying actuary` focus.
   - Console errors from putting a `ref` in uswds `Button` component.
   - 
   - The fix was to revert back to regular `button` element.

- Fix menu list options sub-categories font size.
- Fix selected option display styling
   - When a rate is selected, the select box should only display the rate name cutting off overflowing portions.
   - When a rate is displayed in the select dropdown menu, then show the rate name and sub-categories.
   - [Figma](https://www.figma.com/file/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?type=design&node-id=9961%3A47335&mode=design&t=JCzr9HscGrtHJkCT-1)
- This code also had a side effect of fixing accessibility issues with the combo box. See [MCR-4115](https://jiraent.cms.gov/browse/MCR-4115). 

## QA guidance
- Selecting a rate from the linked rate dropdown list announces the rate name.
- With JAWS a user can remove a linked rate.

### DIFF
--- a/services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
+++ b/services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
@@ -549,12 +549,9 @@ const RateDetailsV2 = ({
                                                             Additional rate
                                                             certification
                                                         </h3>
-                                                        <Button
+                                                        <button
                                                             type="button"
-                                                            outline
-                                                            className={
-                                                                styles.addRateBtn
-                                                            }
+                                                            className={`usa-button usa-button--outline ${styles.addRateBtn}`}
                                                             onClick={() => {
                                                                 const newRate =
                                                                     convertGQLRateToRateForm(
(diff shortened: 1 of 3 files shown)


services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
```diff
@@ -549,12 +549,9 @@ const RateDetailsV2 = ({
                                                             Additional rate
                                                             certification
                                                         </h3>
-                                                        <Button
+                                                        <button
```
haworku, line 552: ugh sorry about having to make the change back lol. So annoying

haworku (approved):
LGTM - tested in JAWS as well. It's annoying we aren't on latest react-uswds, usually I would say lets fix this issue in the third party lib. But also maybe this won't matter with forthcoming react 19 and the way refs are on all components

PR #2433, opened 2024-05-08
MCR-4097 Fix frontend monitoring, add back error boundary fallback UI

haworku:
## Summary
- [x] Get proper error component rendering again when JS error thrown 
   - See [4910df5](https://github.com/Enterprise-CMCS/managed-care-review/pull/2433/commits/4910df59beadfb2ede26a047d1bda34bc9a8d2e2) 
   - Fixed this error by bringing back `<PageHeading/>`. This is needed because the`usePage` context is not reliable if the entire app is crashing, it's too far down in the stack and will not be available in a crash.
- [x] Get thrown errors logging again in monitoring (added New Relic snippet to handle this)
- [x] New Relic config set up in higher level environment

**Config changes**
I added 4 new config for each environment for new relic. Several are the same across all environments. I think eventually `REACT_APP_NR_LICENSE_KEY` could totally replace existing `NR_LICENSE_KEY` used to set up OTEL - but decided that could switch on future PR since its a change to existing setup

**New Relic Changes**
- set up three new applications for NR browser monitoring `mc-review-web` (for dev), `mc-review-web-val`, and  `mc-review-web-prod`
- added errors from val and prod to the `app-web errors` alert policies
- added to dashboards

#### Related issues
https://jiraent.cms.gov/browse/MCR-4097

#### Test cases covered
No tests but made a way to manually check monitoring

## QA guidance
- Login in as Admin user
- Navigate to `/settings?test=crash`

You should see 
1. the error boundary which is a react error component NOT a blank screen
2.  an error in the browsering monitoring for that environment alert in slack if this is triggered in a higher environment
Screenshot 2024-05-17 at 10 52 38 AM
3. (if this is set up) an alert in mc-review alerts channel on slack


services/app-web/src/pages/App/App.tsx
```diff
@@ -16,18 +16,28 @@ import TraceProvider from '../../contexts/TraceContext'
...
 }
 
+
 function App({
     authMode,
     apolloClient,
     s3Client,
 }: AppProps): React.ReactElement {
+    const environmentName = process.env.REACT_APP_STAGE_NAME || ''
+    const isHigherEnv = ['prod', 'val', 'dev'].includes(environmentName )
```
mojotalantikite, line 37: `STAGE_NAME` in dev is confusingly `main`, so we'll want that in this array. Really not sure why it was called `main` 🤷🏻‍♂️

mojotalantikite (approved):

PR #2438, opened 2024-05-10
Add column for rate programs to state program JSON

pearl-truss:
## Summary

As part of this ticket the state programs [spreadsheet ](https://docs.google.com/spreadsheets/d/138HHJTs4sGZucyffNFDe2dlSvYxPRSrP/edit#gid=687010743) was updated to include a new `isRateProgram` column. This PR updates the `import-programs` script to account for the new column and generate a UUID when none was specified in the sheet

#### Related issues

https://jiraent.cms.gov/browse/MCR-4042


scripts/import-programs.ts
```diff
@@ -127,15 +130,11 @@ fs.createReadStream(file)
...
             }
 
-            if (!data.id) {
-                console.error('No ID set for program, make sure to set an ID in the spreadsheet', data)
-                process.exit(1)
-            }
-
             states[code]!.programs.push({
-                id: data.id,
+                id: !data.id ? uuidv4() : data.id,
```
JasonLin0991, line 134: I saw on the AC's was to generate a UUID. Do we expect the ID field in the CSV to be undefined? If we did would it be better to throw an error so we don't get data disparity between our JSON and the CSV?

JasonLin0991 (approved):
LGTM.

docs/technical-design/howto-update-state-programs.md
```diff
@@ -12,5 +12,5 @@ The source of truth for that file comes from a CSV maintained by product and des
 1. Download the latest version of csv from google docs when prompted by product/design.
 2. Run the script following the command listed in the  `import-programs.ts`. This will overwrite existing state programs JSON with the new output. Your usage of the script will likely look something like this:  `cd scripts && yarn tsc && node import-programs.js path/to/data.csv > ../services/app-web/src/common-code/data/statePrograms.json`
 3. Double check the diff. It's important not to delete any programs that have already been used for a submission because although programs are not in the database, we still store references to the program ID in postgres as if they are stable. Also, we want to be sure we are only changing programs expected to change.
-4. For any newly created programs, manually populate the `id` field using a UUID generator
-5. Make a PR to update the statePrograms file in the codebase
+4. Make a PR to update the statePrograms file in the codebase
+5. If an ID wasn't present on the spreadsheet and was autogenerated by the script, be sure to add the id to the spreadsheet so that it's reflected there
```
pearl-truss, line 16: @JasonLin0991 does this cover your question? I think the assumption is the ID will still be present on the spreadsheet but if it was autogenerated instead we would need to go back and update the CSV

JasonLin0991: Maybe not throw and error, but this code could be brought back

```
if (!data.id) {
    console.error('No ID set for program, make sure to set an ID in the spreadsheet', data)
    process.exit(1)
}
```

pearl-truss: ^^ I like that idea

PR #2440, opened 2024-05-10
MCR-4043: read stateProgram json data on frontend

pearl-truss:
## Summary

- This PR updates the stateProgram shape/schema so there's a new field `isRateProgram` 
- Adds a flag on the Submission Type/ Contract page so that only contract programs are displayed 
- Filters rate programs from the state dashboard

#### Related issues
https://jiraent.cms.gov/browse/MCR-4043

#### Test cases covered

- PackageSelect filters out rate programs when new `contractProgramsOnly` flag is true
- StateDashboard doesn't display rows from rate programs


services/app-web/src/components/Select/ProgramSelect/ProgramSelect.test.tsx
```diff
@@ -72,12 +72,43 @@ describe('ProgramSelect', () => {
         await selectEvent.openMenu(combobox)
 
         await waitFor(() => {
-            expect(screen.getByText('MSHO')).toBeInTheDocument()
-            expect(screen.getByText('SNBC')).toBeInTheDocument()
-            expect(screen.getByText('PMAP')).toBeInTheDocument()
-            expect(screen.getByText('MSC+')).toBeInTheDocument()
+            expect(screen.getAllByText('MSHO')).toHaveLength(1)
+            expect(screen.getAllByText('SNBC')).toHaveLength(2)
```
haworku, line 76: So if rate programs are visible we would end up displaying certain programs twice even though they have the same name? That seems odd-   I'm confused by how the test has changed here and why `MSHO` has lenght 1 but `SNBC` has length 2.  Can you explain this a bit?

haworku (approved):

services/app-web/src/components/Select/ProgramSelect/ProgramSelect.test.tsx
haworku: I see now- there are mocked inside stateMock that set up two programs with the same nickname. In that case, please convert these tests to check for the distinct full names so we are ensuring the expected and unique program (or rate program) display

PR #2441, opened 2024-05-10
MCR-4120 Fix flickering from page remounting on RevisionSummary page

haworku:
## Summary
 This bug shows up when we have package submissions that have different contract names. This will happen when program names change. 
 
 **Root cause:** The bug appears because we render too early - first using data based on current revision, then as the client determines which package submission should actually be used for a previous submission, data updates. The infinite loop comes in because in cases where the contract name would also change between revisions (if programs had changed) this would trigger the `useEffect` in place that makes sure the proper page names are displaying in a higher level heading outside this component (`usePage` updates something in the header). This side effect would then cause page render going back to using old data, then it would correct itself, then the effect is triggered, and the cycle repeats in a loop.  
 
#### Related issues
https://jiraent.cms.gov/browse/MCR-4120

#### Screenshots
video from buggy behavior is in bug ticket

#### Test cases covered
-  `renders with correct submission name even when previous revisions have different programs` - this is a failing test that will pass after the change. It makes sure the first render is the correct one and uses correct data starting out.
- The fix for this bug also caused an existing test to fail. Would appreciate 👀  here.
   - `renders the error indexed version 3` was the existing test that started failing after my change. Seems like (from looking at the test data) that test shouldn't have passed, we would expect version 3 to exist. There are three contract submissions on the test data. Made that change
   - Related, also wrote new test for Version 4 to 404 in order to still cover 404 case.

## QA guidance
- When a package has been resubmitted three times and each time the program name is changed, the previous submission for each version should load as expected


services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummaryV2.tsx
```diff
@@ -49,18 +49,25 @@ export const SubmissionRevisionSummaryV2 = (): React.ReactElement => {
...
     })
     const contract = fetchContractData?.fetchContract.contract
-    //We offset version by +1 of index, remove offset to find revision in revisions
+
+    // Offset version by +1 of index, remove offset to find target previous submission in the history list
     const revisionIndex = Number(revisionVersion) - 1
-    const name =
-        contract &&
-        contract?.packageSubmissions.length > Number(revisionVersion)
-            ? contract.packageSubmissions.reverse()[revisionIndex]
```
haworku, line 57: This right here was also part of the problem - we were not using the same logic for the `name` and the data calculations - we should be filtering on CONTRACT_SUBMISSION also here. I changed things to calculate the target previous submission once and use that for both the name and the data calculations for the page

services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummaryV2.tsx
```diff
@@ -49,18 +49,25 @@ export const SubmissionRevisionSummaryV2 = (): React.ReactElement => {
...
+    const targetPreviousSubmission = packageSubmissions[revisionIndex] && packageSubmissions[revisionIndex].__typename ? packageSubmissions[revisionIndex] : undefined
+    const name = targetPreviousSubmission?.contractRevision.contractName
+
     useEffect(() => {
-        updateHeading({
+        // make sure you do not update the page heading until we are sure the name for that previous submission exists
+        if (name) {
+            updateHeading({
             customHeading: name,
-        })
```
haworku, line 63: I saw necessarily re-renders happening here, ideally we dont try to update (or clear) the heading in the <header/> at all until we are sure we have the proper submission name

services/app-web/src/pages/SubmissionRevisionSummary/SubmissionRevisionSummaryV2.tsx
```diff
@@ -77,35 +84,28 @@ export const SubmissionRevisionSummaryV2 = (): React.ReactElement => {
     }
 
     if (
-        !contract ||
-        contract.packageSubmissions.length <= Number(revisionVersion)
```
haworku, line 81: similar issue to https://github.com/Enterprise-CMCS/managed-care-review/pull/2441/files#r1597233952

services/app-web/src/testHelpers/apolloMocks/contractPackageDataMock.ts
```diff
@@ -671,6 +671,13 @@ function mockContractPackageSubmittedWithRevisions(
...
+function mockContractPackageWithDifferentProgramsInRevisions (): Contract {
+   return{ "id": "e670adsfdfadsfc", "status": "RESUBMITTED", "createdAt": "2024-05-07T19:44:53.732Z", "updatedAt": "2024-05-07T19:44:53.732Z", "initiallySubmittedAt": "2024-05-07", "stateCode": "FL", "mccrsID": null, "state": { "code": "FL", "name": "Florida", "programs": [{ "id": "712277fb-f43f-4eb5-98c5-6c6a97509201", "name": "NEMT", "fullName": "Non-Emergency Medical Transportation", "__typename": "Program" }, { "id": "037af66b-81eb-4472-8b80-01edf17d12d9", "name": "PCCMe", "fullName": "Healthy Start MomCare Network, Inc.", "__typename": "Program" }, { "id": "5c10fe9f-bec9-416f-a20c-718b152ad633", "name": "MMA", "fullName": "Managed Medical Assistance Program", "__typename": "Program" }, { "id": "3b8d8fa1-1fa6-4504-9c5b-ef522877fe1e", "name": "LTC", "fullName": "Long-term Care Program", "__typename": "Program" }, { "id": "08d114c2-0c01-4a1a-b8ff-e2b79336672d", "name": "Dental", "fullName": "Prepaid Dental Health Program", "__typename": "Program" }], "__typename": "State" }, "stateNumber": 221, "__typename": "Contract", "draftRevision": null, "draftRates": null,
```
JasonLin0991, line 676: nit: It's a bit hard to read. We could probably format this.

JasonLin0991 (approved):
LGTM!

PR #2445, opened 2024-05-14
MCR-4123 Fix contract actions missing on certain submissions

haworku:
## Summary
Resolves an issue where a specific VAL submission (see ticket) is missing some of its related contract actions 

#### Related issues
https://jiraent.cms.gov/browse/MCR-4123

#### Test cases covered
- I tested this by pulling down VAL data locally. I saw that the faulty rate submission summary for  **MCR-FL-0210-PCCME-RATE-20240502-20240509-CERTIFICATION-20240404** was loading correctly.
- Tried for awhile to write a failing test. I wasn't able to achieve that. The test continues to pass with or without the fix added.
   - I have added the test as `skip` for now.
   - I would like a reviewer to take a look at the test. Do you see an option for what I should do to try and get a legitimate failing test for the bug that resolves for the fix?  Otherwise, I will remove the test, just merge this fix as is without a test.

## QA guidance
- QA regression check for Contract Actions behavior on rate summary - add, remove, edit linked rates and check that rate summary still displays as expected


services/app-api/src/resolvers/rate/fetchRate.test.ts
```diff
@@ -352,4 +353,82 @@ describe('fetchRate', () => {
             '2024-02-02'
         )
     })
+
+    // TEST NOT FAILING WHEN EXPECTED, NEED TO FIGURE OUT HOW TO GET RATE REVISIONS OUT OF SYNC
+    it.skip('returns correct linked contractRevisions, even when rate has changed', async()=>{
```
JasonLin0991, line 358: I couldn't think of a way to get revisions out of sync. I wonder if anything outside the workflow happened. Feature flagging, migrations, or maybe something in the API silently errored and continued with the request.

JasonLin0991 (approved):
LGTM.

services/app-api/src/resolvers/rate/fetchRate.test.ts
haworku: Thanks for checking it out.  I will delete this test then... but appreciate the eyes and the QA

PR #2447, opened 2024-05-15
Fix local OTEL. Change OTEL to use tracer and ctx instead of span

mojotalantikite:
## Summary

While investigating how OTEL and the error boundary in React interacted I noticed that our local OTEL setup (i.e. in local dev) wasn't working. It turns out that Jaegar changed how it deals with OTEL traces coming into the system, so this updates the local setup so we can use the most recent Jaegar and collect traces in local dev.

Once that was fixed, I was having issues seeing spans tied properly between `app-web` and `app-api`. I've never felt good about how we passed around a span instead of a tracer to our resolvers, so this updates our usage. Basically, it adds the tracer to our `Context` that is passed around, along with the extracted otel context that we get from the incoming `app-web` http headers, so in each resolver we can properly create a span tied to it's invocation.

Finally, there were some changes and cleanups that I wanted to do to `app-web`. We never used the ZoneContextManager, as is recommended by OTEL for frontends. We also had our TracerContext below our routing in react, so we never got navigation events (these are coming in now). 

I also needed to add a new env variable for local dev, as we need to distinguish between where `app-web` and `app-api` send their traces with the Jaeger changes.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4159
https://jiraent.cms.gov/browse/MCR-4119
https://jiraent.cms.gov/browse/MCR-4160


mojotalantikite:
Also, I verified that we indeed are now recording crashes in OTEL, using Hana's settings page crash idea from #2433:
Screenshot 2024-05-17 at 3 09 26 PM

services/app-web/src/contexts/TraceContext.tsx
```diff
@@ -56,6 +60,7 @@ registerInstrumentations({
                     /(.*)\.adoberesources\.net/g,
                     /(.*)\.google-analytics\.com/g,
                 ],
+                //propagateTraceHeaderCorsUrls: [/.+/g],
```
haworku, line 63: nit: cleanup

haworku (approved):
Looks good - can you demo next RSP what the traces look like too once we get this in production ?  Also please add my docs once my PR lands - lets pull some of the stuff in this PR description into the montoring docs

mojotalantikite:
I threw my name on a topic to talk about OTEL in RSP. I'll also pull in your changes that just merged and add to the docs!

PR #2450, opened 2024-05-17
MCR-4045: Create table for state programs

pearl-truss:
## Summary

- Creates migration for creating a table to represent state programs

#### Related issues

https://jiraent.cms.gov/browse/MCR-4045

### DIFF
--- a/services/app-api/prisma/schema.prisma
+++ b/services/app-api/prisma/schema.prisma
@@ -304,6 +304,7 @@ model State {
   users                       User[]
   contracts                   ContractTable[]
   rates                       RateTable[]
+  statePrograms               StateProgram[]
 }
 
 model StateProgram {
@@ -313,6 +314,7 @@ model StateProgram {
   fullName           String
   name               String
   stateCode          String
+  state              State                @relation(fields: [stateCode], references: [stateCode])
   isRateProgram      Boolean
 }
 
--- a/services/app-api/prisma/schema.prisma
+++ b/services/app-api/prisma/schema.prisma
@@ -306,6 +306,16 @@ model State {
   rates                       RateTable[]
 }
 
+model StateProgram {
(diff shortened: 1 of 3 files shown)


services/app-api/prisma/schema.prisma
```diff
@@ -306,6 +306,16 @@ model State {
   rates                       RateTable[]
```
JasonLin0991, line 306: I wonder if we should add the relationship between `StateProgram` and `State`.

JasonLin0991 (approved):
Looks good with a suggestion.

services/app-api/prisma/schema.prisma
pearl-truss: I wondered this too. Are you thinking `StateProgram` has a `State` and a `State` has a list of `StatePrograms`?

JasonLin0991: Like this

```postgres
model State {
  stateCode                   String                   @id @default(uuid())
  name                        String
  latestStateSubmissionNumber Int                      @default(0)
  latestStateRateCertNumber   Int                      @default(0)
  statePackages               HealthPlanPackageTable[]
  users                       User[]
  contracts                   ContractTable[]
  rates                       RateTable[]
  statePrograms               StateProgram[]
}

model StateProgram {
  id                 String                @id @default(uuid())
  createdAt          DateTime              @default(now())
  updatedAt          DateTime              @default(now()) @updatedAt
  fullName           String
  name               String
  stateCode          String
  isRateProgram      Boolean
  state              State   @relation(fields: [stateCode], references: [stateCode])
}
```

macrael: Yes this is great, I'm trying for us to be able to use the stateCode as the ID for states since it's such a good one. No need to make UUIDs when we have human readable two character identifiers

PR #2475, opened 2024-05-31
Move to Vite from Create React App

mojotalantikite:
## Summary
This swaps out Create React App for Vite. Some things of note:

* protobufjs had to be upgrade to latest along with CLI tools in order to emit a format that Vite could injest (ESM vs CommonJS)
* graphql macro package needed to change in order to import the `.graphql` schema file
* There's a weird `custom.d.ts` that I needed to make in order to get `.svg` icons working properly
* Vite requires that env vars be exported with `VITE_` prefixed. All the `REACT_` instances needed to be moved

Things left to do:

- [x] Move our Jest tests to [vitest](https://vitest.dev/). The API should stay largely the same, but there is a lot of work left to be done to get our tests passing.
- [x] Confirm that we have env vars in the right place since there is a major change to using `VITE_`. Namely, make sure param store is ok.

#### Related issues
https://jiraent.cms.gov/browse/MCR-1797


haworku:
This change looks good to me so far, recognize we are working on tests.

haworku:
Just found two react-uswds things broken from the upgrade running the app locally. These are blocking.
- The calendar date pickers are not working - you click and nothing happens not date picker pop up.
- The icons showing up on file upload component for errors or complete 

This could be form upgrading react-uswds without the required uswds upgrade as well.

JasonLin0991:
> Finding some react-uswds things broken from the upgrade running the app locally. These are blocking.
> 
> * The calendar date pickers are not working - you click and nothing happens not date picker pop up.
> * The icons showing up on file upload component for errors or complete
> * seems like that sidebar nav is not displaying
> 
> This could be from upgrading react-uswds without the required uswds upgrade as well. We will need to investigate and fix before landing this work. I suggest lets get this deploying and confirm if it happens also in deployed environments before trying to fix.

I updated the `vite.config.ts`. All the issues here should be resolved now.

haworku (approved):
Looks ready to me. Just a small merge conflict left

mojotalantikite:
Code coverage reporting hasn't been fully working, so we're going to take care of that as follow-on work (it shouldn't hold up landing this though): https://jiraent.cms.gov/browse/MCR-4287

PR #2476, opened 2024-05-31
Add Pearl's ssh key to authroized keys

mojotalantikite:
## Summary

Pearl was having some issues pulling down data for testing. Looks like her `ssh-key` was missing, so this adds it in to the environment. I placed it on one of the servers manually and she was able to access it, so merging this should take care of it in the future.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4184


haworku (approved):
TY for addressing this yay

PR #2477, opened 2024-05-31
MCR-4196: updating rates with packages with shared rates causes system error

JasonLin0991:
## Summary
[MCR-4196](https://jiraent.cms.gov/browse/MCR-4196)

Fix system error when updating rates with `packagesWithSharedRateCerts` with `linked-rates` flag on.

This bug was caused by the difference in `packagesWithSharedRateCerts` between the GraphQL `RateFormDataInput` and `RateFormData` type.

There were three solutions
1. Map through `packagesWithSharedRateCerts` and return the object that matches the input type
2. Pass in an empty array, because we no longer want to  carry forward this field in new rate revisions
3. Remove `packagesWithSharedRateCerts` from the update rate mutations all together since it is depreciated and we no longer want to carry forward this data.

I choose option 3, removing it from the input of the mutation altogether by removing `packagesWithSharedRateCerts` from `RateFromDataInput` and removed `PackageWithSameRateInput` type. This still kept `packagesWithSharedRateCerts` in `RateFromData` so historical rate revisions will still contain the field. 

With the new Contract and Rate API, removing the `packagesWithSharedRateCerts` input field will clear out this data from new revisions. The HPP API will still be able to set this field, I want to leave that intact and only removed once we remove the HPP API.

## QA guidance

You can test this with the following steps

1. With `link-rates` feature flag off, start a new contract and rate submission.
2. On the rate details page select a rate for the field `Please select the submissions that also contain this rate certification.`. 
4. Continue the submission and submit.
5. Turn the `link-rates` flag on and as a CMS user unlock the submission.
6. Log in as a State user and go to the rate details page of the unlocked submission.
7. Fix any missing data from depreciating old programs and requiring new rate programs to be selected.
8. Continuing the form should no longer cause a system error.

It will be hard to test retaining `packagesWithSharedRateCerts` on historical rate revisions, since we do not have a rate history yet and I don't think we display it anywhere, but you can verify this using the GraphQL explorer tool. 

Please note that the tool no longer works on LOCAL deployment. Here is a bug [ticket](https://jiraent.cms.gov/browse/MCR-4197) for that.

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/prismaContractRateAdaptors.ts
+++ b/services/app-api/src/postgres/contractAndRates/prismaContractRateAdaptors.ts
@@ -65,6 +65,7 @@ function prismaRateCreateFormDataFromDomain(
         },
         actuaryCommunicationPreference:
             rateFormData.actuaryCommunicationPreference,
+        //TODO: Remove this once HPP resolvers are not used anymore
         contractsWithSharedRateRevision: {
             connect: rateFormData.packagesWithSharedRateCerts
                 ? rateFormData.packagesWithSharedRateCerts.map((p) => ({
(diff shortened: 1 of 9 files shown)


haworku (approved):
Thanks for that PR description, really helped me understand this work. I think this option works well we def need to preserve the old data (though agree without rate history its very convulated to confirm that is happening) until we deal with redunant rates but there is no need to edit that data ever going forward.

PR #2480, opened 2024-06-03
Remove Linked Rates flag

haworku:
## Summary
- remove linked rates flag logic
- refactor Cypress to condense all state submission form specs into one test

Out of scope:
- api only cypress test for rate review spec. Right now we are have not fully implemented contract api. Yet, we have stopped using and maintaining the HPP api for rates. This is difficult to emulate in an API only test that chains requests. Rather than rewriting existing test, then have to keep updating it every time another form page migrates which will break the test,  decided to just skip that cypress that for rate reviews. We can re-implemented  api only tests (or move rate review to be a user interaction test)  after Contract API epic. 
- turn back on accessiblity E2E tests. those have been off for awhile, I did turn them back on to check if its an easy fix, didn't get it working and turned theme back off again. Also conslidated into a single test. 

## QA guidance
- Let's check that I didn't cause regressions for linked rates and contract+rate submissions manually. Pick a scenario or two from the [bug bash](https://docs.google.com/spreadsheets/d/1sYow3l7CMiwQoV03oZAultJ9yTP9atApPmieVjkG6Jc/edit#gid=0) and run it please 🙏


haworku:
Seems like this branch has had some bad merges. I see work missing from https://github.com/Enterprise-CMCS/managed-care-review/commit/321b51a30ae321128ed1a510fab9a2ef2b977672 - now going through commits from last several sprints to make sure other code is not missing

haworku:
LGTM to me, this is my PR I can't approve but I did run through some linked rates workflows and saw things working as before

mojotalantikite (approved):

PR #2484, opened 2024-06-04
MCR-4195: CMS user shouldn't see missing field error for populations covered

JasonLin0991:
## Summary
[MCR-4195](https://jiraent.cms.gov/browse/MCR-4195)

This work refactored where we are determining when to show the missing data error notification. Before this was done at each summary section with slightly different conditions. I refactored to have this be determined at the summary page component, then passed into the summary sections. 

This made more sense since we determine the notifications based on the summary page and the user, not at each summary section component.

- `ReviewSubmitV2` is only for state users and should always notify them of any missing data, not just certain sections.
- `SubmissionRevisionSummaryV2` is for viewing historical data where notifications for missing data is not important for any user because it is no longer in review. 
- `SubmissionSummaryV2` only shows missing data notification for State and Helpdesk users only when the submission is unlocked.
   - I added a test in this for State users because there is code to show this page to State users when submission is unlocked, but I'm not sure when or how a state user would get to this page when the submission is unlocked.
- CMS users should never see the missing data notification.

Other things I did:
- Fix 404 page when help desk users click on a unlocked submission on the dashboard.
- Fix some error with `rateRevisionDataMock` test helper.
- Fixed some field not showing missing data text on summary pages.
- Fixed previous submission summary showing missing data error notification.
- Showing missing data error notification to help desk users.
- Added `mockValidHelpDeskUser` helper.
- Added tests for summary pages to catch regressions.

#### Test cases covered

`ReviewSubmitV2.test.ts`
- `'renders missing data error notifications for state user'`
- `'renders missing modified and unmodified provisions data error notifications'`
- `'renders missing effective rating period data error notification'`

`SubmissionRevisionSummaryV2.test.ts`
- `'does not render missing data notification for CMS users'`
- `'does not render missing data notification for Helpdesk users'`
- `'does not render missing data notification for State users'`

`SubmissionSummaryV2.test.ts`
- `'renders missing data error notifications for helpdesk user'`
- `'renders missing data error notifications for State user'`
- `'does not render missing data error notifications for CMS user'`
- `'does not render missing data error notifications on submitted submissions'`

`SingleRateSummarySection.test.ts`
- `'should not display missing field text to CMS users'`
   - Now tests all fields.
- `'should display missing field text to state users'`
   - Now tests all fields.
- `'should display missing field text to helpdesk users'`
- `'should not display missing field text on submitted rates'`

## QA guidance

There is not a good way to manually test all the data fields in the review app. I think the best way is to locally test this by:
- Manually pass in missing data to the components.
- Load in VAL data and go though old submissions.

### DIFF
--- a/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
+++ b/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
@@ -223,8 +223,7 @@ export const HealthPlanPackageTable = ({
 
     const [tableCaption, setTableCaption] = useState<React.ReactNode | null>()
 
-    const isCMSOrAdminUser =
-        user.__typename === 'CMSUser' || user.__typename === 'AdminUser'
+    const isCMSOrAdminUser = user.__typename !== 'StateUser'
     const tableColumns = React.useMemo(
         () => [
             columnHelper.accessor((row) => row, {
--- a/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
+++ b/services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
@@ -77,9 +77,9 @@ const isSubmitted = (status: HealthPlanPackageStatus) =>
 function submissionURL(
     id: PackageInDashboardType['id'],
     status: PackageInDashboardType['status'],
-    isCMSOrAdminUser: boolean
+    isNotStateUser: boolean
 ): string {
-    if (isCMSOrAdminUser) {
+    if (isNotStateUser) {
         return `/submissions/${id}`
     } else if (status === 'DRAFT') {
         return `/submissions/${id}/edit/type`
@@ -223,7 +223,7 @@ export const HealthPlanPackageTable = ({
 
     const [tableCaption, setTableCaption] = useState<React.ReactNode | null>()
 
-    const isCMSOrAdminUser = user.__typename !== 'StateUser'
+    const isNotStateUser = user.__typename !== 'StateUser'
     const tableColumns = React.useMemo(
         () => [
             columnHelper.accessor((row) => row, {
@@ -237,7 +237,7 @@ export const HealthPlanPackageTable = ({
                         to={submissionURL(
                             info.getValue().id,
                             info.getValue().status,
-                            isCMSOrAdminUser
+                            isNotStateUser
                         )}
                         className={`${styles.ID}`}
                     >
@@ -312,7 +312,7 @@ export const HealthPlanPackageTable = ({
                 },
             }),
         ],
-        [isCMSOrAdminUser, tableConfig.rowIDName]
+        [isNotStateUser, tableConfig.rowIDName]
     )
 
     const reactTable = useReactTable({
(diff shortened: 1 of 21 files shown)


services/app-web/src/components/HealthPlanPackageTable/HealthPlanPackageTable.tsx
```diff
@@ -223,8 +223,7 @@ export const HealthPlanPackageTable = ({
 
     const [tableCaption, setTableCaption] = useState<React.ReactNode | null>()
 
-    const isCMSOrAdminUser =
-        user.__typename === 'CMSUser' || user.__typename === 'AdminUser'
+    const isCMSOrAdminUser = user.__typename !== 'StateUser'
```
haworku, line 226: nit:`isNotStateUser` now

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/SingleRateSummarySection.test.tsx
```diff
@@ -525,4 +445,212 @@ describe('SingleRateSummarySection', () => {
             ).toBeNull()
         })
     })
+
+    describe('Missing data error notifications', () => {
```
haworku, line 449: Just curious why did this summary section get its own tests but the others did not and were just tested view the `ReviewSubmit.test.tsx`

haworku (approved):
I like this solution for simplifying the field level logic and just let the parent decide what state we are in

services/app-web/src/components/SubmissionSummarySection/RateDetailsSummarySection/SingleRateSummarySection.test.tsx
JasonLin0991: I'm glad you asked. I think this could be changed at a later time to follow the pattern, but it's this way because it's only used in `RateSummary.tsx` which is hardcoded to pass in true for prop `isSubmitted` in `SingleRateSummary`.

Usuage in `RateSummary.tsx`
```ts
...
<SingleRateSummarySection
    rate={rate}
    isSubmitted // can assume isSubmitted because we redirect for unlocked
    statePrograms={rate.state.programs}
/>
...
```

From the comment, it didn't make sense to me to test from the `RateSummary.test.ts` because:
1. I don't want to add complexity to `RateSummary` just so I can toggle isSubmitted only when testing.
2. The comment insinuates that this page should never show notifications anyways.
3. There is only 1 Summary component here that has all the data.
4. `SingleRateSummarySection` seems to be the best place to test this even though there does not seem to be a case where this component would show notifications, but because it has the functionality we should have tests to incase this functionality is used.

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/ReviewSubmitV2.tsx
```diff
@@ -114,12 +114,14 @@ export const ReviewSubmitV2 = (): React.ReactElement => {
                     editNavigateTo="../type"
                     statePrograms={statePrograms}
                     isStateUser={isStateUser}
+                    explainMissingData
```
macrael, line 117: where is this defined?

macrael (approved):

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/ReviewSubmitV2.tsx
JasonLin0991: In `DataDetails.tsx`

PR #2500, opened 2024-06-10
Upgrade to Node 20

mojotalantikite:
## Summary

This upgrades us to using Node 20 everywhere. It also includes Jason's discovered [fix](https://github.com/jsdom/jsdom/issues/1612) for our failing jest tests that came up in this upgrade.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4158


haworku (approved):
LGTM - I also ran through the app a bit and looked at the react-uswds components a bit, this change looked fine to me. Good work and thanks also to @JasonLin0991 for 🕵️ skills

PR #2510, opened 2024-06-13
MCR-4199 Adjust NEW tag logic for same date submission and previous revisions

haworku:
## Summary
The bugs here are two specific cases 
1. when submit unlock resubmit happen on same day our date added was off
2. or, when viewing previous subission pages our date added was off 

Resolved this by:
- Fix NEW tag display for same date submissions by using DateTime graphql schema type
- Fix NEW tag display for previous submissions by making sure the `lastSubmittedDate` is based on the submission preceeding the current version being displayed 
- Upgrade react-router-dom minor versions to clean up types for router params
- Use timezone aware `isSameOrAfter` date comparison function from dayjs to determine NEW tag

#### Related issues
https://jiraent.cms.gov/browse/MCR-4199

#### Test cases covered
- renders the NEW tag when a document is submitted after the last submission, even if same day

## QA guidance
- Create and resubmit a package and check new package display is correct both on the current page and on the previous summary pages that are viewed from change history 
- Remember design rules around this feature  (also written into the code) - we do not display the new tag for state users, we do not display new tag on the initial submission.

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
+++ b/services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
@@ -164,37 +164,8 @@ function rateFormDataToDomainModel(
         rateID: rateRevision.rateID,
         rateType: rateRevision.rateType ?? undefined,
         rateCapitationType: rateRevision.rateCapitationType ?? undefined,
-        rateDocuments: rateRevision.rateDocuments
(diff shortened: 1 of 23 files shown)


haworku:
Moving this back into in review there are test failures in app-api that will need workarounds.

JasonLin0991 (approved):
Amazing work!

I do wonder if the new tag makes sense for previous submissions though.

PR #2512, opened 2024-06-14
MCR-4024: Fix cypress tests for linked rates

JasonLin0991:
## Summary
[MCR-4024](https://jiraent.cms.gov/browse/MCR-4024)

- Added a new cypress test `'can unlock and resubmit a linked rate and change history updates'` in `unlockResubmit.spec`
- Updated `createAndSubmitContractWithRates` to use the new Contract and Rate API.
   - The function did not work using the HPP API to create rates, not sure how `rateReview.spec` even passed in other branches. 
   - Instead of trying to fix it with the old API, I updated it to use the new rates API.
   - I'm also returning `Contract` instead of `HealthPlanPackage`. 
      - This is because rate names differ between them and instead of updating HPP to match rate name formats today, I just switched to using `Contract` instead. 
      - This made `rateReview.spec` fail because it was looking for HPP rate names, but the rates were made using new Rate API with the new name format.
- Updated `rateReview.spec` to use `Contract` instead of HPP and update the tests now that rate names are less unique using the new rate name format.
- Fixed `fillOutLinkedRate` to look for the first rate option.
- Fixed `minnesotaStatePrograms` to use booleans for `isRateProgram` instead of string
   - Maybe this was causing `createAndSubmitContractWithRates` to fail.

## QA guidance

If it passes CI it should be good, but you could pull down and run both tests in local.

### DIFF
--- a/services/cypress/integration/cmsWorkflow/rateReview.spec.ts
+++ b/services/cypress/integration/cmsWorkflow/rateReview.spec.ts
@@ -35,30 +35,20 @@ describe('CMS user can view rate reviews', () => {
 
                     const rate1 = latestSubmission.rateRevisions[0]
                     const rate2 = latestSubmission.rateRevisions[1]
-                    let rate1Name = rate1.formData.rateCertificationName
-                    let rate2Name = rate2.formData.rateCertificationName
-
-                    if (!rate1Name || !rate2Name) {
-                        throw new Error(`Unexpected error: Rate name(s) did not exist. Rate1Name: ${rate1Name}, Rate2Name: ${rate2Name}`)
-                    }
 
                     // Then check both rates in rate reviews table
                     cy.logInAsCMSUser({
                         initialURL: `/dashboard/rate-reviews`,
                     })
 
-                    // Rate names can be the same, rare in prod, but common in automated tests and manual tests.
-                    // Here were just checking to make sure the first exists in findAll.
                     cy.get('table')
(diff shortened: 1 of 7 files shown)


services/cypress/integration/cmsWorkflow/unlockResubmit.spec.ts
```diff
@@ -454,6 +456,216 @@ describe('CMS user', () => {
     })
 
     // TODO AFTER LINKED RATES AND LINKED RATES CHANGE HISTORY SHIPS
```
haworku, line 458: we can clear out this comment^

haworku (approved):

services/cypress/integration/cmsWorkflow/rateReview.spec.ts
```diff
@@ -29,49 +29,54 @@ describe('CMS user can view rate reviews', () => {
...
-                }).should('exist')
+                    // Then check both rates in rate reviews table
+                    cy.logInAsCMSUser({
+                        initialURL: `/dashboard/rate-reviews`,
                     })
+
+                    // Rate names can be the same, rare in prod, but common in automated tests and manual tests.
+                    // Here were just checking to make sure the first exists in findAll.
+                    cy.get('table')
+                    .findAllByRole('link', { name: rate1Name }).first()
```
haworku, line 53: I would use a data-testid for this with a unique identifier  (maybe the rate number). this feels like it could cause us issues down the road if we selecting by non unique value

PR #2517, opened 2024-06-17
MCR-4023, MCR-3975 delete old rate tables + submissionPackages on fetchRate

haworku:
## Summary

This PR completes a majority of our cleanup work coming out of the linked rates work. 

This PR also adds `packageSubmissions` to the rate type, and adjusts rate and contract submit to be based on the same code filling out new package submissions for both. 

Example tasks:
- [x] remove outdated tables from Prisma https://github.com/Enterprise-CMCS/managed-care-review/pull/2517/commits/64cb494430805671113f6a17c7e0058e77419fd3
- [x] update ContractType domain model for draft revisions so that  draft contract revisions, find their related rates following the  top level `draftRates` https://github.com/Enterprise-CMCS/managed-care-review/pull/2517/commits/78d941c3102ff8eeb4a8813995608f40fedec9f9
- [x] remove prisma helpers for drafts https://github.com/Enterprise-CMCS/managed-care-review/pull/2517/commits/78d941c3102ff8eeb4a8813995608f40fedec9f9
- [x] do not return an `Error` for Rate parse anymore (match behavior from Contract) https://github.com/Enterprise-CMCS/managed-care-review/pull/2517/commits/2ceda13916b753e2a2ce45d647b82121f6f23939
- [x]  fix `submitRate` – get tests passing [2ed4062](https://github.com/Enterprise-CMCS/managed-care-review/pull/2517/commits/2ed4062812302f6847d82b2afeb56c369bbdbb85)
- [x] fix `submitContract` – get tests passing [61b88d2](https://github.com/Enterprise-CMCS/managed-care-review/pull/2517/commits/61b88d2c1e2a8b33d9325f49600ec7bd8751840e)
- [x] update RateType domain model for draft revisions
    - specifically this PR addresses the types infinity loop issue that has been problematic for us - adds new `baseContractType` and `baseRateType` to zod
- [x] add new `RatePackageSubmissions` type to graphql
- [x] add packageSubmissions to fetchRate [1628ef2](https://github.com/Enterprise-CMCS/managed-care-review/pull/2517/commits/1628ef2408b6805b9ff51b6fa67aaa7c9a147de8)
---

#### Related issues
https://jiraent.cms.gov/browse/MCR-4023
https://jiraent.cms.gov/browse/MCR-3975

#### Test cases covered
`rate returns packageSubmissions` 

## QA guidance
Check for regressions around submit (for child rates, linked rates, combination, unlock, resubmit) in the review app.


macrael:
Where are we using contract.draftRevision.rateRevisions anymore? We should be using contract.draftRates and rate.draftContracts exclusively now.

services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
```diff
@@ -157,15 +157,33 @@ function contractWithHistoryToDomainModel(
...
+            draftRevision = {
+                id: contractRev.id,
+                contract: contractRev.contract,
+                createdAt: contractRev.createdAt,
+                updatedAt: contractRev.updatedAt,
+                unlockInfo: convertUpdateInfoToDomainModel(
+                    contractRev.unlockInfo
+                ),
+                formData: contractFormDataToDomainModel(contractRev),
+                rateRevisions: domainRateRevisions,
```
macrael, line 183: I think we should drop rateRevisions from the domain model, we should just use contract.draftRates instead.

services/app-api/src/postgres/contractAndRates/submitContract.ts
```diff
@@ -100,117 +91,6 @@ export async function submitContract(
...
 
-            // OLD SUBMIT STYLE
-            // Update the contract to include the submitInfo ID
-            const updated = await tx.contractRevisionTable.update({
-                where: {
-                    id: currentRev.id,
-                },
-                data: {
-                    submitInfo: {
-                        connect: {
```
macrael, line 111: we probably still need this part of the call, setting the submitInfo, but maybe nothing else

macrael (commented):
This is what I see, lmk if you have more questions

services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
haworku: Okay, I can work on those domain model changes. @macrael I'm assuming make the change symetrically both contracts and rates have this pattern in their domain model

haworku: Update: Can't make this change symetrically because it causes a types infinity loop. So made the change for contracts (drop rate revisions from the draftRevision domain model)  only.

services/app-api/src/domain-models/contractAndRates/rateTypes.ts
```diff
@@ -1,34 +1,31 @@
...
-    updatedAt: z.date(),
-    status: statusSchema,
-    stateCode: z.string(),
-    parentContractID: z.string().uuid(),
-    stateNumber: z.number().min(1),
-    // If this rate is in a DRAFT or UNLOCKED status, there will be a draftRevision
-    draftRevision: rateRevisionWithContractsSchema.optional(),
-    // All revisions are submitted and in reverse chronological order
-    revisions: z.array(rateRevisionWithContractsSchema),
+const baseCategorySchema = z.object({
```
haworku, line 11: what is this stuff?

services/app-api/src/postgres/contractAndRates/parseContractAndRates.test.ts
haworku, line 1: Why are we deleting the parser tests? Are we assuming this is also tested via resolver specific integration tests?

services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
```diff
@@ -280,7 +254,8 @@ function contractWithHistoryToDomainModel(
             contractRevisions[contractRevIndex + 1]
 
         // Reverse rateRevisions so it is in DESC order.
-        const rateRevisions = contractRev.rateRevisions.reverse()
+        const rateRevisions =
+            contractRev.relatedSubmisions[0].submissionPackages
```
haworku, line 258: @macrael did this end up being right? This was my code I think and I remember feeling confused if this was the right way to get the rate revisions on a contract rev

services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
```diff
@@ -404,14 +371,60 @@ function contractWithHistoryToDomainModel(
...
+        rateWithoutDraftContractsToDomainModel(dr.rate)
+    )
+
+    const draftRatesOrError = arrayOrFirstError(draftRatesOrErrors)
+    if (draftRatesOrError instanceof Error) {
+        return draftRatesOrError
+    }
+
+    // FIX PARENT ID IF NEEDED
+    // This is fragile code. Because of the asymmetry required to break the
```
haworku, line 408: seems like then we should write a unit test at least for this - make sure that placeholder is never showing up

services/app-api/src/postgres/contractAndRates/parseContractAndRates.test.ts
macrael: Yes, there were a few tests where it would have been a lot of work to build new mock data when the integration tests are already generating that data correctly so I just removed them.

services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
macrael: I think that's still right, that's more code that will disappear once we drop all the contractRevisions.rateRevisions code which is a follow on PR to this work

haworku: ticket?

services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.ts
```diff
@@ -357,12 +357,53 @@ export function submitHealthPlanPackageResolver(
...
+                    rateUpdates.rateUpdates.delete.push({
+                        rateID: draftRate.id,
+                    })
+                } else {
+                    // this is a previously submitted child-rate. I'm not sure what to do with it
+                    console.warn(
+                        'Attempting to remove a previously submitted child rate in submit',
+                        contractWithHistory.id
+                    )
+                    throw new Error('no nonono bad child rate submit')
```
haworku, line 396: This is not supposed to be an error? This is supposed to just be unlinked from its parent but not deleted from db. If a user does this valid behavior in the app and we error that means we are preventing valid actions in the app I think??

services/app-api/src/postgres/contractAndRates/submitContract.ts
```diff
@@ -77,8 +69,28 @@ async function submitContractInsideTransaction(
...
+    if (submissionResult instanceof Error) {
+        return submissionResult
+    }
+
+    return await findContractWithHistory(tx, contractID)
+}
+
+async function submitContractAndRates(
+    tx: PrismaTransactionType,
+    contractID: string | undefined,
```
haworku, line 88: So before we had an api almost working with editing and submitting a rate independently. Back when we did rate unlock and edit.


Kind of feels like we walked back from that and doubled down on the package idea. Are you thinking we just reimplemented a new function for independent rate editing?

services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.ts
macrael: No, this supports that. The submitRate function calls this passing an undefined contractID, and it sets all the package submission stuff correctly. My goal was for all that bookkeeping to happen in one place, but submitRate works great with it on its own. And yes, I can move this to its own file to make that more clear.

services/app-api/src/domain-models/contractAndRates/rateTypes.ts
macrael: oh whoops that was something from the internet I was using to understand the cycle problem

services/app-api/src/postgres/contractAndRates/submitContract.ts
macrael: No, this supports that. The submitRate function calls this passing an undefined contractID, and it sets all the package submission stuff correctly. My goal was for all that bookkeeping to happen in one place, but submitRate works great with it on its own. And yes, I can move this to its own file to make that more clear.

services/app-api/src/resolvers/healthPlanPackage/submitHealthPlanPackage.ts
macrael: This check is only running if someone has switched a contract to CONTRACT_ONLY from one where they had rates when it was CONTRACT_AND_RATES, but I think you are right that we should treat this the same as unlink.

services/app-api/src/domain-models/index.ts
```diff
@@ -45,7 +45,9 @@ export type {
     RateFormDataType,
     PackageStatusType,
     ContractPackageSubmissionType,
+    RatePackageSubmissionType,
     ContractPackageSubmissionWithCauseType,
+    RatePackageSubmissionWithCauseType,
```
pearl-truss, line 50: Should RatePackageSubmissionWithCauseType be added to codegen like we have for ContractPackageSubmissionWithCauseType?

pearl-truss (approved):
Left a few minor comments but changes look good to me! I also didn't see anything weird in the review app when I went through a submit, unlock, resubmit x2 flow adding and removing linked rates

services/app-api/src/domain-models/index.ts
haworku: I see it in the codegen ✅

haworku:
@macrael when I pulled main down again into this branch  I started to get  app-web errors in about 20 files for packageSubmissions. The cause was that we have a new required field (can't be missing or undefined ) on the  overall `RateType` and that's used across the frontend in graphql.

e.g. its used for draft rates coming down on the contract and its used in index rates specifically. I started to address the errors by expanding the graphql queries, the test helpers etc, but the scope of the change seemed too much for something not fully implemented. I decided to instead just loosen the required field a bit [e3d5532](https://github.com/Enterprise-CMCS/managed-care-review/pull/2517/commits/e3d55327ba2a59ed8757428dab8dc41a7deb08ad) and put the rest of the work into the ticket about removing revisions and using packageSubmissions fully.

PR #2521, opened 2024-06-19
MCR-4202: Generate Download Links for Documents

pearl-truss:
## Summary

Adds a `downloadURL` field to `GenericDocuments` (used by contracts and rates) and `Documents` (used by questions)

#### Related issues
https://jiraent.cms.gov/browse/MCR-4202

#### Test cases covered

- `downloadURL` is populated on fetch/index for GenericDocument and Documents

## QA guidance

NOTE: if testing this locally you'll have to upload new files in order to test this. Files uploaded via local tests will not be valid

GenericDocument (used by rates and contracts)
- Upload new files to a submission
- View a page that calls fetchContract or fetchRate 
- Using the network tab, grab the `downloadURL` from formData
- Inputting the url should trigger the download for the file

Questions
- Upload a new question
- Using the network tab, grab the `downloadURL` from formData
- Inputting the url should trigger the download for the file


services/app-api/src/resolvers/rate/submitRate.test.ts
```diff
@@ -78,8 +82,51 @@ describe('submitRate', () => {
...
         // expect status to be submitted.
         expect(submittedRate.status).toBe('RESUBMITTED')
-        // expect formData to be the same
-        expect(submittedRateFormData).toEqual(draftFormData)
+        // const formDataKeys = Object.keys(submittedRateFormData)
+        // formDataKeys.forEach((key) => {
+        //     if (key !== 'rateDocuments' && key !== 'supportingDocuments') {
+        //         expect(submittedRateFormData[key]).toEqual(draftFormData[key])
+        //     }
+        // })
```
mojotalantikite, line 90: Probably can clean out that commented code?

mojotalantikite (approved):
Wow, that's a lot of test changes 😂 LGTM, thanks for taking care of this.

services/app-api/src/handlers/apollo_gql.ts
```diff
@@ -387,14 +402,26 @@ async function initializeGQLHandler(): Promise<Handler> {
                   dmcoEmails,
                   helpDeskEmail,
               })
+    const S3_BUCKETS_CONFIG: S3BucketConfigType = {
+        HEALTH_PLAN_DOCS: s3DocumentsBucket,
+        QUESTION_ANSWER_DOCS: s3QABucket,
+    }
+
+    const s3Local = 'http://localhost:4569'
+    if (authMode === 'LOCAL') {
```
macrael, line 411: fab!

services/app-api/src/resolvers/configureResolvers.ts
```diff
@@ -142,6 +146,8 @@ export function configureResolvers(
         RateRevision: rateRevisionResolver,
         Contract: contractResolver(),
         ContractRevision: contractRevisionResolver(store),
+        GenericDocument: genericDocumentResolver(s3Client),
+        Document: questionResponseDocumentResolver(s3Client),
```
macrael, line 150: lol why do we have different document types for these two?

services/app-api/src/resolvers/questionResponse/createQuestion.ts
```diff
@@ -104,8 +104,18 @@ export function createQuestionResolver(
             setErrorAttributesOnActiveSpan(errMessage, span)
             throw new Error(errMessage)
         }
-
-        const questionResult = await store.insertQuestion(input, user)
+        const docs = input.documents.map((doc) => {
+            return {
+                name: doc.name,
+                s3URL: doc.s3URL,
+                downloadURL: doc.downloadURL ?? undefined,
```
macrael, line 111: How come we have downloadURL in the document type here? I thought we would only add it to the graphql type, added at the last second with the resolver. It should definitely not end up in the db (haven't gotten that far in the review yet) because it's only temporary, right?

services/app-api/src/resolvers/questionResponse/questionResponseDocumentResolver.ts
```diff
@@ -0,0 +1,32 @@
...
+} from '../../s3'
+import type { S3ClientT } from '../../s3'
+import type { QuestionResponseDocument } from '../../domain-models'
+
+export function questionResponseDocumentResolver(
+    s3Client: S3ClientT
+): Resolvers['QuestionResponseDocument'] {
+    return {
+        downloadURL: async (parent: QuestionResponseDocument) => {
+            const s3URL = parent.s3URL ?? ''
```
macrael, line 15: I'm surprised this is an optional type

services/app-api/src/s3/s3Error.ts
```diff
@@ -0,0 +1,23 @@
+export const S3ErrorCodes = ['NETWORK_ERROR'] as const
+type S3ErrorCode = (typeof S3ErrorCodes)[number] // iterable union type
+
+export type S3Error = {
+    code: S3ErrorCode
+    message: string
+}
+
+// Wow this seems complicated. If there are cleaner ways to do this I'd like to know it.
```
macrael, line 9: We should use a real Error object here, this kind of fake error was something I did wrong early on. Just return actual errors everywhere and you can use `instanceof Error` instead of this function. If we need to get a specific message from the error, we can make an `S3Error` that inherits from Error

macrael: If it's too much hassle that's ok b/c we are sharing this code with the FE so it's reasonable just to keep it the same while we fix the packaging

services/app-api/src/testHelpers/gqlHelpers.ts
```diff
@@ -98,13 +106,19 @@ const constructTestPostgresServer = async (opts?: {
         })
 
     await insertUserToLocalAurora(postgresStore, context.user)
+    const s3LocalClient = await newLocalS3Client(
```
macrael, line 109: why is the default the LocalS3Client when it seems like so many tests are using the mock one instead? Are there any tests that are actually using this local client? Could the default for tests be the mock?

services/app-graphql/src/schema.graphql
```diff
@@ -881,8 +886,15 @@ input GenericDocumentInput {
     s3URL: String!
     "The sha256 is a unique string representing the file, generated on the FE currently in the FileUpload component"
     sha256: String!
+<<<<<<< HEAD
```
macrael, line 889: merge conflict

macrael (approved):
This looks wonderful! Great job with the s3 clients

services/app-api/src/s3/s3Error.ts
pearl-truss: I fixed everything else but am going to leave this as-is for now to revisit later

PR #2523, opened 2024-06-21
MCR-4187: add create contract and update contract to resolvers

JasonLin0991:
## Summary
[MCR-4187](https://jiraent.cms.gov/browse/MCR-4187)

- Two new resolvers and mutations `updateContractDraftRevision` and `createContract`
- One new Postgres function `updateDraftContract`
   - We use a different mutation and Postgres function for updating rates on the form pages which only update the rates.
   -  I think `updateContractDraftRevision` should not update rates so we keep resolvers simple by separating responsibilities.
- Added a function `preprocessNulls`, to wrap Zod schemas to preprocess `nulls`.
   - GraphQL types return `null` and we need to convert fields that are `null` into `undefined`. We are doing this in some parts of the code by manually doing nullish coalescing before the GraphQL inputs are handed to the Postgres functions.
   - This function in the Zod schema will do this transformation for us when we use Zod to validate our data (see next bullet point).
   - Currently I only implemented this for `contractFormDataSchema` and Zod validations are happening in `updateContractDraftRevision` resolver (see next bullet point)`
-  Added Zod data validations for updating contracts `dataValidatorHelpers.ts`
   - `validateContractDraftRevisionInput` this validates the input contract draft revision data.
      - This function takes our schema is is mostly all optional and extends additional validations on specific fields
         -  `statutoryRegulatoryAttestation` & `statutoryRegulatoryAttestationDescription`
            - If `statutoryRegulatoryAttestation` is false and `statutoryRegulatoryAttestationDescription` has no value, then we return validation error.
            - If `statutoryRegulatoryAttestation` is true and `statutoryRegulatoryAttestationDescription` has a value, we clear `statutoryRegulatoryAttestationDescription` out.
         - `programIDS` checks to make sure the IDs are valid for the state.
         - `populationCovered` validate that population coverage type `CHIP` is also not a `CONTRACT_AND_RATES` submission type.
      - Each extended field validation are functions so they are reusable.
      - `safeParse` will return all the invalid fields, we use this in `updateContractDraftRevision` resolver to return a `UserInputError`.
    - This is a first pass at using Zod validations on different contract status data.
      - The reusable functions use two Zod methods, `superRefine` and `transform`.
         - `superRefine` is when we need to do complex validations and return custom errors/
         - `transform` is not specifically used for validation, but it can. It's mainly used to transform data. For our case we need to do both for validating `statutoryRegulatoryAttestation` and transforming `statutoryRegulatoryAttestationDescription`.
- Updated email Zod contact schemas to `z.string().email().optional().or(z.literal(''))`. This will validate non-empty strings for email formats. Empty strings are also valid.
   - The frontend passes in empty string for contact fields when saving as draft.
- Updated `defaultFloridaProgram` and `defaultFloridaRateProgram ` to get programs from JSON file instead of hard coded values.
#### Test cases covered

`dataValidatorHelpers.test.ts`
- `'Validates input form data and removes statutoryRegulatoryAttestationDescription'`
- `'converts fields that are null to undefined'`
- `'Returns error for invalid data'`
   - This tests all the extended validations in `validateContractDraftRevisionInput`

`updateContractDraftRevison.test.ts`
- Copied all contract tests from `updateHealthPlanFormData.test.ts`
- Removed all references to HPP

`updateDraftContract.test.ts`
- Coped all contract tests from `updateDraftContractWithRates`

## QA guidance

You can use the GraphQL explorer tool to hit both new mutations.


services/app-api/src/resolvers/contract/updateContractDraftRevision.ts
```diff
@@ -0,0 +1,142 @@
...
+            draftRevision.updatedAt.getTime()
+        ) {
+            const errMessage = `Concurrent update error: The data you are trying to modify has changed since you last retrieved it. Please refresh the page to continue.`
+            logError('updateContractDraftRevision', errMessage)
+            setErrorAttributesOnActiveSpan(errMessage, span)
+            throw new UserInputError(errMessage)
+        }
+
+        // Using zod to validate and transform graphQL types into domain types.
+        const parsedFormData = validateContractDraftRevisionInput(
```
macrael, line 102: I would have this function return our `DomainModel | Error` just like we do error handling everywhere else. I think it's reasonable to return the ZodError b/c that already has all the error paths in it that didn't validate. We might want to add some of that info into custom error sections for graphql as well, but I think starting with just the message is fine

services/app-api/src/domain-models/contractAndRates/formDataTypes.ts
```diff
@@ -34,40 +35,55 @@ const packagesWithSharedRateCerts = z.object({
     packageStatus: statusSchema.optional(),
 })
 
+// Wraps schema to make it optional and transforms null to undefined. This will take nulls from graphql inputs and turn them to undefined.
+const nullishTransformWrapper = (schema: ZodSchema) =>
```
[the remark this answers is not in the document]
JasonLin0991: I've updated this to use `z.preprocess()`. The type issue we saw was because I defined the function argument types incorrectly.

```ts
function preprocessNulls<T extends ZodTypeAny>(
    schema: T,
    params?: RawCreateParams
): ZodEffects<T, T['_output'], unknown> {
    return z.preprocess((val) => val ?? undefined, schema, params)
}
```
I think this could replace converter functions like `contractFormDataToDomainModel()`. Something like `prismaUpdateContractFormDataFromDomain` I'm not too sure about, the `safeParse` output type would be more of a Prisma schema.

services/app-api/src/postgres/contractAndRates/updateDraftContract.ts
```diff
@@ -0,0 +1,62 @@
...
+// Update the given draft
+async function updateDraftContract(
+    client: PrismaClient,
+    args: UpdateContractArgsType
+): Promise<ContractType | NotFoundError | Error> {
+    const { contractID, formData } = args
+
+    try {
+        return await client.$transaction(async (tx) => {
+            // Given all the Contracts associated with this draft, find the most recent submitted
```
macrael, line 24: this comment doesn't quite seem right. We're looking for an unsubmitted contract rev

services/app-api/src/resolvers/contract/createContract.ts
```diff
@@ -0,0 +1,95 @@
...
+                count
+            )} not exist in state ${stateFromCurrentUser}`
+            logError('createContract', errMessage)
+            setErrorAttributesOnActiveSpan(errMessage, span)
+            throw new UserInputError(errMessage, {
+                argumentName: 'programIDs',
+            })
+        }
+
+        const riskBasedContract =
```
macrael, line 60: why does this need to be pulled apart this way?

services/app-api/src/resolvers/contract/updateContractDraftRevision.ts
```diff
@@ -0,0 +1,143 @@
...
+        const span = tracer?.startSpan('updateContractDraftRevision', {}, ctx)
+        setResolverDetailsOnActiveSpan(
+            'updateContractDraftRevision',
+            user,
+            span
+        )
+        const featureFlags = await launchDarkly.allFlags(context)
+        const { draftRevision, contractID } = input
+
+        if (draftRevision.submitInfo) {
```
macrael, line 31: why does this input have submitInfo? I would expect only the FormData to be passed in as an argument to this mutation

services/app-graphql/src/schema.graphql
```diff
@@ -577,6 +642,15 @@ type UpdateCMSUserPayload {
     user: CMSUser!
 }
 
+input UpdateContractDraftRevisionInput {
+    contractID: ID!
+    draftRevision: ContractDraftRevisionInput!
```
macrael, line 647: I would not let you send a whole draft revision here. State users are only allowed to change the FormData, they can't change updatedAt or submittedAt or the package name or any of that, so the only thing they should be allowed to send is the form data. 

That's how we handle it with updateDraftContractRates and with submitRate. In general it's best to avoid API inputs that are just going to be ignored, or worse, untrusted.

macrael (approved):
This looks good to go except that I really think you should change the inputs for updateDraftContractRevision. Maybe we should rename it updateDraftContractFormData instead

services/app-api/src/resolvers/contract/updateContractDraftRevision.ts
macrael: Another thing I thought of: this is user submitted data, so we cannot rely on submitInfo being honest. We correctly error out lower down based on the fetched contract, that's the right way to validate this. Even checking this from the user input is a bad idea because someone might accidentally think it's trustworthy later on. Better to not let users send data that we can't trust.

services/app-graphql/src/schema.graphql
JasonLin0991: Good call, there is a lot of data we don't use in the revision. I added the entire revision just to get `updatedAt` so we can check for concurrent editing. My other thought was to just send `updatedAt`, but that seemed vague on where this field came from. Another thought is a new graphQL type for meta data for inputs that includes these types of fields, so its better defined. Also, probably all this data should not be in the `input` variable.

```graphql
input UpdateContractDraftRevisionInput {
    contractID: ID!
    updatedAt: DateTime!
    formData: ContractDraftRevisionFormDataInput!
}
```

I took a look at `updateDraftContractRates` and noticed we don't check for concurrent updates there. I was able to overwrite new rate data with stale rate data using two windows of the app.

macrael: I like just passing in updatedAt, I would maybe call it `lastSeenUpdatedAt` to be more explicit?

JasonLin0991: @macrael I'm going to make a new PR to add checking for concurrent update in `updateDraftContractRates`

PR #2533, opened 2024-06-27
MCR-4143 add unlock contract

pearl-truss:
## Summary
- creates `unlockedContract` resolver 
- creates `unlockedContract` mutation
- creates `unlockedContract `types for graphql and domain models
- reimplemented emails for state and CMS for unlocking
- Add additional contract helpers for emails `handleAsCHIPSubmissionForContract`, `generateCMSReviewerEmailsForContract`, `findAllContractProgramIds`

To Do

- [ ] Add tests for emails
- [ ] Resolve merge conflicts

#### Related issues
https://jiraent.cms.gov/browse/MCR-4143 

#### Test cases covered

unlockContract resolver errors for stateUsers, errors if contract is in the wrong status 

## QA guidance

-As a CMS user, unlock a contract. In the network tab you will see the new unlockContract endpoint is used


services/app-graphql/src/schema.graphql
```diff
@@ -1396,6 +1435,23 @@ type ContractPackageSubmission {
     rateRevisions: [RateRevision!]!
 }
 
+"""
+ContractPackageSubmission is a snapshot of a contract and all its related rates in time
+a ContractPackageSubmission can be created by the user submitting a contract or a related rate
+"""
+type UnlockedContractPackageSubmission {
```
pearl-truss, line 1442: Open question, fields that aren't required in the domain modal (cause for packagesubmission and state for contract) were causing a `cannot update non nullable` error when `unlockContract` called on the frontend. I made those fields optional in the graphql schema but I'm not sure if that's the correct way to address this. Any ideas welcome

macrael: can you give more details on the error? We're not sending this data from app-web so I only expect it to really crop up on returning the model in the resolvers

macrael (approved):
Do we have a plan to replace the Contract type with a LockedContract type in the future?

services/app-graphql/src/schema.graphql
pearl-truss: It's an apollo error `Cannot return null for non-nullable field ContractPackageSubmission.cause.`

Screenshot 2024-07-01 at 10 36 42 AM

services/app-api/src/emailer/emails/unlockContractCMSEmail.ts
```diff
@@ -0,0 +1,80 @@
...
+import {
+    stripHTMLFromTemplate,
+    renderTemplate,
+    findContractPrograms,
+    generateCMSReviewerEmailsForContract,
+} from '../templateHelpers'
+import type { EmailData, EmailConfiguration, StateAnalystsEmails } from '../'
+import type { ProgramType, UpdateInfoType } from '../../domain-models'
+
+export const unlockContractCMSEmail = async (
```
JasonLin0991, line 14: Could you add tests for CMS and State unlock emailers. You can copy over the relevant ones from the old emailers.

pearl-truss:
> Do we have a plan to replace the Contract type with a LockedContract type in the future?

I don't think we've laid out a plan yet. But I do see that we have a [spike ticket](https://jiraent.cms.gov/browse/MCR-3151) in the Contract API epic for discovery on locked/unlocked types

services/app-api/src/emailer/emails/unlockContractCMSEmail.ts
pearl-truss: Yup that is one of the todos for this PR

pearl-truss:
> Do we have a plan to replace the Contract type with a LockedContract type in the future?

@macrael I don't think we've laid out a plan yet. But I do see that we have a [spike ticket](https://jiraent.cms.gov/browse/MCR-3151) in the Contract API epic for discovery on locked/unlocked types

services/app-api/src/emailer/emails/unlockContractCMSEmail.ts
JasonLin0991: Whoops! Totally missed that sorry.

services/app-graphql/src/schema.graphql
macrael: Ah, OK I think I have a guess at the problem. We have a custom resolver for packageSubmissions that adds `cause` to the response. It's not something that's in the domain model b/c it's computed. It looks like since you made a new type here for the UnlockedContractPackageSubmission we're not invoking that resolver so we're skipping adding that field. The most straightforward fix I see would be to reuse the other ContractPackageSubmission type, triggering the resolver we wrote for that, but I'm curious if this is the reason you broke it out as its own type and I'm missing something.

pearl-truss: Ah sorry for the confusion. This is the error I see when I use `ContractPackageSubmission`. That's why I created `UnlockedContractPackageSubmission` because it works when the cause field is allowed to be null

services/app-api/src/emailer/emails/unlockContractCMSEmail.ts
pearl-truss: No worries! I just opened the PR early because I was leaving to be out on Friday

services/app-graphql/src/schema.graphql
pearl-truss: @macrael I got this working, I realized I needed to make a `UnlockedContract` resolver for the resolver chain. It does what the `Contract` resolver does in adding the cause and state fields

macrael: ah perfect! yes I was thinking about that as I fell asleep last night but forgot this morning, glad you sorted it out.

macrael (approved):
Looking great!

PR #2534, opened 2024-06-27
Fix updateDraftContractRates resolver to check for concurrent update

JasonLin0991:
## Summary

- Adds a check in `updateDraftContractRates` for optimistic concurrency control. The mutation takes an extra input field `lastSeenUpdatedAt` from `contract.draftRevision.updatedAt` to compare with the  same draft revisions `updatedAt` in the DB to prevent newer changes being overwritten by stale data.
- I also went ahead and updated all the api commands in Cypress to only use the new API's and remove all references to HPP.

#### Test cases covered
`updateDraftContractRates.test.ts`
- `'errors on concurrent updates'`

## QA guidance
- With the `link-rates` feature flag on.
- Open two windows with the same draft contract and rates submission
- On the first the window:
   - Continue with the form until the Rate details page
- On the second window:
   - Continue with the form until the Rate details page
   - Make a change to the rate.
   - Continue to the Contacts page
- Back on the first window, attempt to continue to Contacts page.
- You will encounter a system error banner.

### DIFF
--- a/services/app-graphql/src/schema.graphql
+++ b/services/app-graphql/src/schema.graphql
@@ -1326,6 +1326,8 @@ input UpdateContractRateInput {
 
 input UpdateDraftContractRatesInput {
     contractID: ID!
+    "date the contracts draft revision was last updated at"
+    lastSeenUpdatedAt: DateTime!
     """
     updatedRates contains ALL the existing rates associated with this contract.
     This includes rates that were previously linked or previously created (LINK & UPDATE)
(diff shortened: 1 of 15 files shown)


pearl-truss (approved):
LGTM one small nit is the name `lastSeenUpdatedAt`, is a bit confusing if it just represents `lastUpdatedAt`?

services/app-graphql/src/schema.graphql
```diff
@@ -1326,6 +1326,8 @@ input UpdateContractRateInput {
 
 input UpdateDraftContractRatesInput {
     contractID: ID!
+    "date the contracts draft revision was last updated at"
```
JasonLin0991, line 1329:
```suggestion
    "The updatedAt of the contract draft revision at the time this update. This is used for optimistic concurrency control"
```
@pearl-truss 
I think my description in the graphQL schema is not accurate on what this field is and what is used for. Does something like this make more sense on the naming of the field?

PR #2543, opened 2024-07-02
MCR-4217: button and link event tagging for tealium

JasonLin0991:
## Summary
[MCR-4217](https://jiraent.cms.gov/browse/MCR-4217)

- Refactored Tealium implementation
   - Using a provider to configure and intialize Tealium, then provide functions in context for logging.
   - Provider takes in a Tealium client that has all the tealium related functions as a prop.
      - `tealiumClient` is the client used in `prod` and `val` for actual logging to Tealium
      - `devTealiumClient` is the client used in all lower environments where we do not want to log to Tealium.
      - `testTealiumClient` used in `renderWithProviders` for tests.
- `useTealium` no longer initializes Tealium. It provides functions to log user events.
- Added Tealium provider to `renderWithProviders`
- Added two wrapper components for logging in `components/TealiumLogging`
   - `ButtonWithLogging` wrapper around `Button` that takes in logging data as props and generates required logging data from required button props.
   -  `NavLinkWithLogging, LinkWithLogging, ReactRouterLinkWithLogging` Same as the button, but with links. I couldn't figure out the type definition that would allow all three to be combined into one. Did not want to spend any more time trying to figure that out in this PR.
- Add docs on Tealium implementation and how to use it.

## QA guidance

- Please check all links and buttons to make sure they still work and the styling has not changed.
- Please make sure I am logging the right events and that the event data being sent looks correct.


services/app-web/src/components/ActionButton/ActionButton.tsx
```diff
@@ -81,10 +88,27 @@ export const ActionButton = ({
                 : inheritedProps['aria-label'],
     }
 
+    const onClickHandler = (
+        e: React.MouseEvent<HTMLButtonElement, MouseEvent>
+    ) => {
+        logButtonEvent({
+            text: extractText(children),
+            button_style: variant,
```
haworku, line 96: For saving time at this stage and to reduce back and forth with Blast team, I would like to just have us use the specs Blast has provided when possible. That means here can we map all `outline` variant to be `transparent` please? e.g. something like  `variant === 'outline'? 'transparent' : variant`

services/app-web/src/tealium/tealium.ts
```diff
@@ -0,0 +1,320 @@
...
+    userId?: string // custom attribute
+    packageId?: string // custom attribute
+    tealium_event?: TealiumEvent // this is required by tealium, TBD what allowed values aer here, usually this is supposed to be configured first .
+}
+
+type ButtonEventStyle =
+    | 'default'
+    | 'primary'
+    | 'success'
+    | 'transparent'
```
haworku, line 31: I think the only additional style we would need is `unstyled`. transparent is outline FYI. I can ask about this but for now please add `unstyled` to the types and then don't use any other types besides these events plus unstyled .

docs/technical-design/tealium-logging.md
```diff
@@ -0,0 +1,97 @@
+# Tealium Logging
+## Background
+We use a tool called Tealium to hook into the CMS customer analytics tools. The data is then sent to Adobe Analytics (previously CMS used Google Analytics). Access to analytics tools is set up through another team, Blast Analytics. They are contractors with a focus on analytics and work across all CMS domains.
```
haworku, line 3: Please link to appropriate monitoring.md section here, it explains what/why analytics in more detail

docs/technical-design/tealium-logging.md
```diff
@@ -0,0 +1,97 @@
...
+   const linkData: TealiumButtonEventObject = {
+      ...tealiumData,
+      event_name: 'button_engagement',
+   }
+   logUserEvent(linkData, pathname, loggedInUser, heading)
+}
+```
+These `useTealium` functions are then used in components like `Button` and `Link` to log when they are clicked. Since most of these components are configured and used similarly, wrapper components were made with built-in logging to replace the default components for easy implementation of logging. They can be found in `components/TealiumLogging`.
+
+Not all components are wrapped like this. Some, like `ModalToggleButton` are rarely used, so in those cases we can directly use `logButtonEvent` in the `onClick` prop.
```
haworku, line 50: I like this you give example of the different ways implementation work in codebase

services/app-web/src/contexts/TealiumContext.tsx
haworku, line 1: I see initilaize and logging button events working well. ⭐  Page View is double logging in some cases - maybe that was existing issue not sure but we should fix in future work with Tealium

haworku (approved):
Good work, the docs in particular were very helpful to me. 

- I ran through the app for a submit and unlock and double checked initialize Everything looks good for local dev. 
- After merge and deploy, lets double check this work production. Want to make sure initialize and page views still work (this can be done by checking adobe analytics directly). We can also double check that the script tag still looks proper by inspecting the HTML. Not expecting any issues but that feels like an important validation step before we consider this ticket fully complete.

docs/technical-design/tealium-logging.md
JasonLin0991: I don't see a section for Tealium in `monitoring.md`. I think adding that section can be a follow-up PR for this so we can test this logging work out in production.

PR #2551, opened 2024-07-05
Recenter step indicator

haworku:
## Summary
This change was originally on the  #2475 PR but since it moved up to in review independently moving it to its own PR.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4215

#### Screenshots
Screenshot 2024-07-05 at 1 05 33 PM

#### Test cases covered
 snapshot test can't catch this we need visual regression testing - started to look into this but wanted to get this fix in first


JasonLin0991 (commented):
The styling is still a bit off on the review and submit page as Toph in the review app. Does this branch have the remove link-rates feature flag work? If not, it might be good to merge that in before more fixes.

Screenshot 2024-07-09 at 12 54 50 PM
Screenshot 2024-07-09 at 12 55 03 PM

JasonLin0991 (approved):
Looks great!

PR #2547, opened 2024-07-03
MRC-4244: Bug fix for Showing an error on a submitted rate

pearl-truss:
## Summary
To replicate:

1. Create a new submission

2. Select "contract action and rate certification" and fill out form

3. On the Rate details page link a rate certification that was submitted prior to when the rate name enhancement feature went live.
4. Click continue and fill out the rest of the submission
5. When you arrive at the Review and submit page...

Current behavior:
The Rate details will flag an error validation on the field  "Rates this rate certification covers" and prevent a submission, despite the rate being locked. 

Expected behavior:

We will not flag an error validation on a locked linked rate.

Displays the historic 'Programs this rate covers' if the linked rate is locked and historic

#### Screenshots

**BEFORE**
Screenshot 2024-07-03 at 10 02 45 AM

**AFTER**
Screenshot 2024-07-03 at 10 57 09 AM

#### Test cases covered

- Displays the old field for historic rates
- Renders the new field on unlocked rate

## QA guidance

I pulled down prod data to test this locally

PR #2558, opened 2024-07-09
MCR-4244 User can submit linked rate with deprecated programs, no errors

haworku:
## Summary
Follow up to #2547 when user could not submit and also error was still showing in some places

This PR includes:
- reimplementation of the frontend work - calculating if the rate is linked more directly
- reimplementation of backend validation- account for linked rates in `parseAndSubmit` (have to shoehorn this because actually this legacy API actually isn't supposed to handle linked rates at all and has no concept of linking in its types) 

#### Related issues
https://jiraent.cms.gov/browse/MCR-4244

#### Screenshots
- renders the deprecated rate programs when no rate programs present on the last submitted version of rate shown to state user

## QA guidance

1. Log in with `toph2@example.com` user on review app. This is a user in a state that has historic rate submissions. Create a new submission
2. Select "contract action and rate certification" and fill out form
3. On the Rate details page link a rate certification that was submitted prior to when the rate name enhancement feature went live.
4. Click continue and fill out the rest of the submission
5. When you arrive at the Review and submit page you should see old deprecated rate programs and no cilentside missing data errors
6. You can submit that package with no backend errors


services/app-web/src/pages/StateSubmission/ReviewSubmit/RateDetailsSummarySection.tsx
```diff
@@ -282,9 +280,24 @@ export const RateDetailsSummarySection = ({
...
+                    - there's no new field values present and CMS user is viewing
+                    - theres no new field values present and contract is locked and state user viewing
+                    - theres no new field values present and the contract unlocked and the rate being displayed as a linked rate
+
+                    otherwise use new fields values going forward
+                    */
+                      const showLegacyRatePrograms =
+                          hasDeprecatedRatePrograms &&
+                          hasNoRatePrograms &&
+                          (isSubmittedOrCMSUser || isLinkedRate)
```
JasonLin0991, line 296: Not sure if we care enough, but once a submission is unlocked going to the submission summary page, as a state user, will show the inline error for the new program field. You can get to the submission summary page when it is unlocked by using the path `/submissions/:id`. We don't expose a way to get to this path in the app itself as a State user, but I think some emails have this link and maybe savvy users.

pearl-truss (approved):
I wasn't able to test this in the UI because the deployment is still going through and I have some tealium compile error after pulling in main. However, the code itself in the PR looks good to me. I didn't realize this issue would need so much, thanks for picking this up!

JasonLin0991 (approved):

services/app-web/src/pages/StateSubmission/ReviewSubmit/RateDetailsSummarySection.tsx
haworku: 🐛 Uh, that seems like a bug, you aren't supposed to be able to view the submission summary page for an unlocked rate - you are supposed to get redirected to review/submit to edit

PR #2569, opened 2024-07-16
Fix jest test failures in app-api post Vite merge

mojotalantikite:
## Summary
When we merged in the Vite branch, it turns out there were incompatabilities between `app-api` and `app-web` with Jest tests. Since Vite moved us to using ES6 in `app-web` and `app-api` is still using CommonJS, and because protobuf generation moved to using ES6 in `app-web`, we hit a bunch of bad import errors when running `app-api` unit tests.

I'm not sure why this was working in the branch we merged, but this as a stop-gap copies over the shared code to `app-api`, generates protos in commonjs for `app-api`, and fixes up imports to make this all work. 

I don't love that this is copy-pasta for `common-code`, but I think this means we need to prioritize ripping `common-code` into it's own module instead of living in `app-web`, which has never felt good in the first place.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4296


macrael (approved):

macrael:
agree on this raising the common-code priority

PR #2573, opened 2024-07-17
Move app-api to esbuild

mojotalantikite:
## Summary

This moves our builds for `app-api` off of using `webpack` and onto `esbuild`. 

#### Related issues
https://jiraent.cms.gov/browse/MCR-4237


pearl-truss (approved):
Good work!

services/app-api/serverless.yml
```diff
@@ -249,7 +227,7 @@ functions:
           method: post
           cors: true
           authorizer: aws_iam
-    timeout: 60 # zipping large amount of files hits default time out of 6 seconds
+    timeout: 30 # zipping large amount of files hits default time out of 6 seconds
```
haworku, line 230: assuming you intended to do this change - is it just that esbuild is faster so this time can go down i'm curious

haworku (approved):

services/app-api/serverless.yml
mojotalantikite: Any Lambda attached to an API gateway has a 30s max timeout, so there are warnings in the logs about this. Just bumped it down to silent those.

PR #2590, opened 2024-07-24
MCR-4225 API for withdraw and replace redundant rate 

haworku:
## Summary
- Add `withdrawInfo` to Rate
- New `replaceRateOnContract` postgres level function
- New graphql resolver and mutation - `withdrawAndReplaceRedundantRate` 
#### Related issues
https://jiraent.cms.gov/browse/MCR-4225

#### Test cases covered

-  results in the expected contract package changes
- withdraws rate and adjusts review status
- does not re-validate contract data to replace rat
- returns forbidden error for non-admin users
- returns error if the contract or individual rates are not currently submitted


services/app-api/src/postgres/contractAndRates/replaceRateOnContract.ts
```diff
@@ -0,0 +1,209 @@
...
+        replaceReason,
+        withdrawnRateID,
+        replacementRateID,
+    } = args
+
+    // unlock contract and rates
+    const unlockContractArgs: UnlockContractArgsType = {
+        contractID,
+        unlockedByUserID: replacedByUserID,
+        unlockReason: replaceReason,
```
macrael, line 55: Should we prefix this reason in some way? I'm imagining someone withdrawing a rate and that reason not quite making sense for a reason for an unlock/submit? like `Withdrawing rate: ${withdrawReason}`

services/app-api/src/postgres/contractAndRates/replaceRateOnContract.ts
```diff
@@ -0,0 +1,209 @@
...
+    )
+    if (updateResult instanceof Error) {
+        return updateResult
+    }
+
+    // resubmit contract
+    const resubmitContractArgs: SubmitContractArgsType = {
+        contractID,
+        submittedByUserID: replacedByUserID,
+        submittedReason: replaceReason,
```
macrael, line 149: ibid

services/app-api/src/postgres/contractAndRates/replaceRateOnContract.ts
```diff
@@ -0,0 +1,209 @@
...
+        withdrawReason: replaceReason,
+    }
+    const withdrawResult = await withdrawRateInsideTransaction(
+        tx,
+        withdrawRateArgs
+    )
+    if (withdrawResult instanceof Error) {
+        return withdrawResult
+    }
+    return updateResult
```
pearl-truss, line 172: Great job with the logic in this file! I follow everything except, what's the reasoning for why is `updateResult` returned instead of `resubmitResult`?

services/app-api/src/postgres/contractAndRates/withdrawRate.ts
```diff
@@ -0,0 +1,65 @@
...
+            },
+        })
+
+        const updatedRate = await tx.rateTable.update({
+            where: {
+                id: rateID,
+            },
+            data: {
+                withdrawInfo: {
+                    connect: { id: withdrawInfo.id },
```
macrael, line 48: It doesn't really matter b/c of the transaction but you could create this directly in the .update() call

services/app-api/src/postgres/contractAndRates/withdrawRate.ts
```diff
@@ -0,0 +1,65 @@
...
+                },
+            },
+        })
+
+        if (updatedRate instanceof Error) {
+            throw updatedRate
```
macrael, line 54: return this instead of throwing it, we'll throw where the transaction is created

services/app-api/src/resolvers/contract/withdrawAndReplaceRedundantRate.test.ts
```diff
@@ -0,0 +1,235 @@
...
+                            .rateRevisions[0].rateID,
+                    replacementRateID:
+                        contractWithReplacementRate.packageSubmissions[0]
+                            .rateRevisions[0].rateID,
+                    replaceReason,
+                },
+            },
+        })
+        expect(adminResult.errors).toBeDefined()
+    })
```
macrael, line 234: A couple more tests for some of the error cases would be nice. Like withdrawing a non-child rate or withdrawing an already withdrawn rate.

services/app-api/src/resolvers/contract/withdrawAndReplaceRedundantRate.ts
```diff
@@ -0,0 +1,82 @@
...
+                replacementRateID,
+                withdrawnRateID,
+                replaceReason,
+                replacedByUserID: user.id,
+            })
+        if (withdrawAndReplaceRedundantRateResult instanceof Error) {
+            const errMessage = `Failed to unlock contract revision with ID: ${contractID}; ${withdrawAndReplaceRedundantRateResult.message}`
+            logError('withdrawAndReplaceRedundantRate', errMessage)
+            setErrorAttributesOnActiveSpan(errMessage, span)
+            throw new GraphQLError(errMessage, {
```
macrael, line 69: There's a few errors we can get here that I think are UserInputErrors, but we're returning everything as an INTERNAL_SERVER_ERROR here. 

I think it's worth making a new error type that we can differentiate on. We already have a NotFoundError in postgresErrors, we could have something else for some of these other user input errors.

macrael (approved):
Looks really solid!

services/app-api/src/resolvers/contract/withdrawAndReplaceRedundantRate.ts
haworku: @macrael Yes I was thinking that too but forgot about it - thanks for calling this out! I do want the errors to be more distinct coming out of the API - will look at this today

services/app-api/src/postgres/contractAndRates/replaceRateOnContract.ts
haworku: Good point. This feels out of scope for this PR and more of a design/product question when we ship the UI.  We can raise this thne. My understanding was they are going to have a very clear reason they will type in

services/app-api/src/postgres/contractAndRates/withdrawRate.ts
haworku: see [fe308e9](https://github.com/Enterprise-CMCS/managed-care-review/pull/2590/commits/fe308e936bd83b0cbcc49212714d0a572fc89cc0)

services/app-api/src/postgres/contractAndRates/replaceRateOnContract.ts
haworku: @pearl-truss Added a comment about this `return updateResult because we want to return updated contract - withdrawResult is the updated standalone rate`

services/app-api/src/resolvers/contract/withdrawAndReplaceRedundantRate.test.ts
haworku: See [45c04d0](https://github.com/Enterprise-CMCS/managed-care-review/pull/2590/commits/45c04d02fd5780108f414fca9b6fe1c51eb293e8)

PR #2593, opened 2024-07-25
[MCR-4188] contract API for index contracts and fetch contract with questions

pearl-truss:
## Summary

This PR creates 2 new resolvers, `indexContracts` and `fetchContractWithQuestions`

#### Related issues

https://jiraent.cms.gov/browse/MCR-4188

#### Test cases covered

FetchContractWithQuestions
- Fetches contract with its associated questions 
- Errors on state users fetching contract from different state

IndexContracts
- returns a list of contracts that includes newly created entries
- synthesizes the right statuses as a contract is submitted/unlocked/etc
- a different user from the same state can index contracts

## QA guidance

- Code and test look through 
- Feedback on `fetchContractWithQuestion`, wasn't sure if there were parts of the code for `fetchContract` that could be reused or if it made more sense to make a completely different query/resolver/schema changes


services/app-api/src/domain-models/contractAndRates/baseContractRateTypes.ts
```diff
@@ -11,6 +11,64 @@ import {
...
+})
+const stateUserType = z.object({
+    id: z.string().uuid(),
+    role: z.literal('STATE_USER'),
+    email: z.string(),
+    stateCode: z.string(),
+    givenName: z.string(),
+    familyName: z.string(),
+})
+const questionResponseType = z.object({
```
macrael, line 45: Since we're redefining all these types here in Zod, let's get rid of our extant QuestionType s and export these generated zod types instead

services/app-api/src/resolvers/contract/fetchContractWithQuestions.ts
```diff
@@ -0,0 +1,144 @@
...
+import {
+    setErrorAttributesOnActiveSpan,
+    setResolverDetailsOnActiveSpan,
+    setSuccessAttributesOnActiveSpan,
+} from '../attributeHelper'
+import { isStateUser } from '../../domain-models'
+import type { QuestionIndexType } from '../../domain-models/contractAndRates/baseContractRateTypes'
+export function fetchContractWithQuestionsResolver(
+    store: Store
+): QueryResolvers['fetchContractWithQuestions'] {
```
macrael, line 13: Take a look at healthPlanPackageResolver, I don't think we need a separate resolver for fetchContractWithQuestions, we can just do the same thing we did in HPPResovler in ContractResovler and have it trigger the fetchQuestions when a client asks for them.

services/app-api/src/resolvers/contract/indexContracts.ts
```diff
@@ -0,0 +1,145 @@
...
+            '\n'
+        )}`
+        logError('indexHealthPlanPackagesResolver', errMessage)
+        setErrorAttributesOnActiveSpan(errMessage, span)
+    }
+
+    return parsedContracts
+}
+
+const validateAndReturnContracts = (results: ContractType[], span?: Span) => {
```
macrael, line 52: this function seems misnamed now. I see it came from indexHPP but I don't like that we're calling logSuccess inside of here. Maybe trim it down into something that just returns the node structure? logSuccess in the handler itself?

services/app-api/src/resolvers/contract/indexContracts.ts
```diff
@@ -0,0 +1,145 @@
...
+        if (isStateUser(user)) {
+            const contractsWithHistory =
+                await store.findAllContractsWithHistoryByState(user.stateCode)
+
+            if (contractsWithHistory instanceof Error) {
+                const errMessage = `Issue finding contracts with history by stateCode: ${user.stateCode}. Message: ${contractsWithHistory.message}`
+                logError('indexContracts', errMessage)
+                setErrorAttributesOnActiveSpan(errMessage, span)
+
+                if (contractsWithHistory instanceof NotFoundError) {
```
macrael, line 85: can we return NotFoundError from findAllContracts*? I assume we just return empty array if there's nothing there

services/app-graphql/src/queries/fetchContractWithQuestions.graphql
```diff
@@ -0,0 +1,360 @@
+query fetchContractWithQuestions($input: FetchContractWithQuestionsInput!) {
```
macrael, line 1: we will still need the fetchContractWIthQuestions.graphql, but if we implement the questions section of the ContractResolver we won't need a separate top level resolver for this, it will just be a call to fetchContract that includes Questions in the requested response

macrael (approved):
This is great. I do think we can avoid adding a separate WithQuestions resolver which would be excellent

services/app-api/src/domain-models/contractAndRates/baseContractRateTypes.ts
haworku: I think we need a plan for what to do with the types here generally. Also defined in this file is StateType and UserType CMSUserType,  all types already and now being moved into zod.. 

Maybe we need a ticket in ContractApi epic to clean this up after this ticket? I would expect also each of these to be in their own file not all living within the base contract type since they can be used across domains.

services/app-api/src/resolvers/contract/indexContracts.ts
haworku: Yes if the contract is not an id found in the db wouldnt it would be a NotFoundError

services/app-api/src/domain-models/contractAndRates/baseContractRateTypes.ts
macrael: So long as these zod types match the current explicit types, I think the fix here should be just defining the zod types in the same files where we have the explicit types instead of here, and then exporting the generated-from-zod types, deleting our explicit types. I think that's worth doing in this PR unless it hits some weird recursive type issue.

services/app-api/src/resolvers/contract/indexContracts.ts
pearl-truss: @macrael @haworku what's the requested change here? findAllContractsWithHistoryByState can return either ContractOrErrorArrayType | NotFoundError | Error

macrael: No change! I forgot it could actually return NotFoundError

services/app-api/src/resolvers/contract/contractResolver.ts
```diff
@@ -99,5 +107,103 @@ export function contractResolver(): Resolvers['Contract'] {
...
+                    totalCount: dmcpQuestions.length,
+                    edges: dmcpQuestions,
+                },
+                OACTQuestions: {
+                    totalCount: oactQuestions.length,
+                    edges: oactQuestions,
+                },
+            }
+            // A state user cannot access contracts that don't belong to their state
+            if (isStateUser(user)) {
```
macrael, line 192: this should be impossible, right? We're already checking for this in fetchContract so I'm not sure how we could get to the contract resolver without having checked this.

In my mind, authorization should happen at the root resolver layer and hopefully we can skip it in the leaf resolvers.

macrael (approved):
looks great! One note about authz in the leaf resolver

PR #2600, opened 2024-07-29
MCR-4317 add admin cypress tests for Settings

haworku:
## Summary
Follow on to #2578. Adds new Admin workflow Cypress tests. This helps protect against the various crash/freeze behaviors we saw on settings page this sprint (one was with users table, one was with state analysts table). 

🔍 I also made a change to the use of `navigate` in AppRoutes and AuthContext. It was failing silently - we didn't notice. A test that could have caught it breaking has been skipped since Contract and Rates API was introduced.  I found out when I tried to write an admin test that would attempt to go to /settings, login, then redirect. Using `window.location.href` directly seems to resolve the issue. 

#### Related issues
https://jiraent.cms.gov/browse/MCR-4317

#### Test cases covered
- update user division assignments
- filter down state analysts by state code

## QA guidance
If Cypress tests pass,  we can merge. The bugfix already happened for this issue - this catches up the tests.

### DIFF
--- a/scripts/add_cypress_test_users.ts
+++ b/scripts/add_cypress_test_users.ts
@@ -62,6 +62,7 @@ function IDMRole(role: UserRole): string {
 async function createUser({
     userPoolID,
     name,
+    familyName,
     email,
     role,
     password,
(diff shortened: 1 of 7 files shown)


haworku:
Have to close in favor of #2606 since I changed the cypress users and branch already deployed.

PR #2603, opened 2024-07-30
MCR-4279: prevent oact from getting emails about non risked based contracts

pearl-truss:
## Summary

If a contract is not risk based than OACT shouldn't be notified about it

#### Related issues

https://jiraent.cms.gov/browse/MCR-4279

#### Test cases covered

Updated existing test to use a risked based contract if expecting oact to receive emails

## QA guidance

Submit a risked based contract, in review app. OACT should not be notified

### DIFF
--- a/services/app-api/src/emailer/emails/newPackageCMSEmail.test.ts
+++ b/services/app-api/src/emailer/emails/newPackageCMSEmail.test.ts
@@ -805,6 +805,31 @@ test('does not include oactEmails on contract only submission', async () => {
     })
 })
 
+test('does not include oactEmails on non risked based contract', async () => {
+    const sub = mockContractAndRatesFormData()
+    sub.riskBasedContract = false
+    const statePrograms = mockMNState().programs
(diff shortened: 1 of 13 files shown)


services/app-api/src/emailer/emails/newPackageCMSEmail.test.ts
```diff
@@ -750,8 +752,10 @@ test('does not include state specific analyst on contract and rate submission',
     })
 })
 
-test('includes oactEmails on contract and rate submission', async () => {
```
haworku, line 753: Can we also add the test here that you did for unlock - oact doesn't get the emails if its a non risk based contract.

haworku:
Pleae also edit the EmailSettingsTable for me. Thats the way our team reminds ourselves what the email settings are and eventually that will be what is shared with CMS users if we surface this UI.

 Right now it say OACT  gets all contract and rate submissions; excluding CHIP programs and PR state. Lets add the other exclusion you are adding here to the list. cc: @MegMurphy

PR #2608, opened 2024-07-30
Bump container for cypress in promote

mojotalantikite:
## Summary

When landing the `pnpm` work we got a failure in the cypress promote action due to an outdated container. This just bumps that container to match what we have in review apps.

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -238,11 +238,10 @@ jobs:
     needs: [promote-app-prod]
     runs-on: ubuntu-20.04
     container:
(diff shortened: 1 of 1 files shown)


haworku (approved):

macrael (approved):

haworku:
Hang on do we need to change anything else? A quick find for node 16 I see a few places with that

macrael:
As much as possible if we can read from .nvrc that would be best

mojotalantikite:
Looks like the `get-changed-services` action still references node16. I'd say let's update that in another PR in case it breaks something.

PR #2613, opened 2024-07-31
Drop contractRevision.rateRevisions and rateRevision.contractRevisions

macrael:
## Summary

We don't want to use the old revision -> revision connection anymore. 

This PR removes those connections from the domain models. 

Required changing the unlock emailers, I expect more changes there in the future. 

#### Related issues
https://jiraent.cms.gov/browse/MCR-4304

## QA guidance

No application changes, but it's worth doing a run-through of the app


services/app-api/src/emailer/templateHelpers.ts
```diff
@@ -160,9 +164,12 @@ const generateCMSReviewerEmailsForContract = (
         ]
```
pearl-truss, line 164: Not sure if it'll get flagged in a merge conflict but [logic was added](https://github.com/Enterprise-CMCS/managed-care-review/pull/2603/files#diff-989f1f3bbd590ba8998db651e954e5d0806141f0ebfd80c3d1fe510ac28e5352) to these helper functions to not add oact to emails for non risked based contracts in a recent PR

services/app-api/src/postgres/contractAndRates/findContractWithHistory.test.ts
```diff
@@ -299,76 +299,77 @@ describe.skip('findContractWithHistory with full contract and rate history', ()
...
-        expect(revisionsInTimeOrder[4].rateRevisions).toHaveLength(2)
-        expect(revisionsInTimeOrder[4].unlockInfo?.updatedReason).toBe(
-            'unlock for 2.1 remove'
-        )
-        expect(revisionsInTimeOrder[4].unlockInfo?.updatedBy).toBe(
-            'zuko@example.com'
-        )
-        expect(revisionsInTimeOrder[4].submitInfo?.updatedReason).toBe(
+        expect(submissionsInTimeOrder[4].rateRevisions).toHaveLength(2)
+        // expect(submissionsInTimeOrder[4].unlockInfo?.updatedReason).toBe(
```
pearl-truss, line 338: Can you add a comment for why these assertions are commented out?

macrael: I commeneted it out because subsissionpackages don't have unlock info, but it turns out these tests are all skipped anyway so I'm going to fix them

PR #2614, opened 2024-08-01
Add command to configure and open Cypress to run against review app.

JasonLin0991:

### DIFF
--- a/dev_tool/src/dev.ts
+++ b/dev_tool/src/dev.ts
@@ -24,16 +24,13 @@ import {
     runBrowserTestsInDocker,
     runWebTests,
     runWebTestsWatch,
+    runBrowserTestsAgainstAWS,
 } from './test/index.js'
 
 // Run clean commands from every single lerna package.
 async function runAllClean() {
     const runner = new LabeledProcessRunner()
-    await runner.runCommandAndOutput(
-        'clean',
-        ['pnpm', 'clean'],
-        ''
-    )
+    await runner.runCommandAndOutput('clean', ['pnpm', 'clean'], '')
 }
 
 // Run lint commands from every single lerna package.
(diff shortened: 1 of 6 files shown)


pearl-truss (commented):
Good job, I'm excited for us to have this command! I ran it locally after following the directions to obtain the temp AWS credentials and I got this error. In the terminal out when I ran the command I did get `Missing baseUrl in compilerOptions. tsconfig-paths will be skipped`

Screenshot 2024-08-02 at 11 46 35 AM

pearl-truss:
@JasonLin0991  I got AWS credentials and copied them into envrc and ran `direnv allow`. Then I ran `./dev test browser --in-review-app` and tried to do the `questionResponse` cypress test

JasonLin0991:
> @JasonLin0991 I got AWS credentials and copied them into envrc and ran `direnv allow`. Then I ran `./dev test browser --in-review-app` and tried to do the `questionResponse` cypress test

@pearl-truss I think you are missing `TEST_USERS_PASS` env variable in your `envrc.local`. I was able to reproduce your error by removing that from my mine. I added that to the instructions. Let me know if that works when you add that variable in.

macrael:
@JasonLin0991 Better than docs is if the tool can print a useful error message. If you've got required env vars, you can check they are set and error explicitly

pearl-truss:
@JasonLin0991 I'm not able to get the secret from github actions, could you add more context on how to go about it? This is what I see from settings -> secrets and variables -> actions

Screenshot 2024-08-08 at 8 22 20 AM

JasonLin0991:
> @JasonLin0991 I'm not able to get the secret from github actions, could you add more context on how to go about it? This is what I see from settings -> secrets and variables -> actions
> 
> Screenshot 2024-08-08 at 8 22 20 AM

@pearl I had incorrectly thought there was a way to get it from GitHub secrets, but we can only update the value there. I updated the instructions, you can get the password from `1password`.

pearl-truss (approved):
LGTM thanks for the extra context!

PR #2625, opened 2024-08-05
Remove lerna from monorepo

mojotalantikite:
## Summary

Now that we're using `pnpm`, which has monorepo support built in, we no longer need to depend on `lerna`. We decided it's best to have fewer dependencies if possible.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4312


dev_tool/src/dev.ts
```diff
@@ -26,45 +26,27 @@ import {
...
 
-// Run lint commands from every single lerna package.
 async function runAllLint() {
     const runner = new LabeledProcessRunner()
-    await runner.runCommandAndOutput(
-        'lint',
-        ['npx', 'lerna', 'run', 'lint'],
-        ''
-    )
+    await runner.runCommandAndOutput('lint', ['pnpm', 'lint'], '')
```
haworku, line 36: This isn't working

mojotalantikite: Thanks, fixed!

haworku (approved):
LGTM

PR #2629, opened 2024-08-06
MCR-4323: grant access to CMS approver role

JasonLin0991:
## Summary
[MCR-4323](https://jiraent.cms.gov/browse/MCR-4323)

- Add CMS approver role for Local and Congito login.
- Add new CMS approver role to the Prisma schema.
- Add a `CMSUsersUnion` type in GraphQL and domain types
   - This was the way I came up with to allow multiple roles to share the same queries and mutations that a CMS user had `updateCMSUser`, `createQuestion`, etc...
   - Take a peek in `configureResolvers`, I had to add a new resolver to handle the union type `CMSUsersUnion`
   - **Let me know if you have a better solution**
- Consolidated permissions checks into helper functions in `app-web` and `app-api`
- Added iterable CMS user tests on the front end.
- Add iterable CMS user tests to the backend
- Fixed Prisma migration dev tool command`

## QA guidance

- The new CMS approver role should have the same permissions as the CMS user role. E2E Manual tests of each CMS user function as a CMS approver be great.


services/app-api/src/domain-models/user.ts
```diff
@@ -64,13 +66,35 @@ function toDomainUser(user: PrismaUser): UserType {
...
+}
+
+// Users with admin permissions
+function hasAdminPermissions(
+    user: UserType
+): user is AdminUserType | HelpdeskUserType | BusinessOwnerUserType {
+    const authorizedAdmins: UserRoles[] = [
         'ADMIN_USER',
         'HELPDESK_USER',
         'BUSINESSOWNER_USER',
```
JasonLin0991, line 100: Is `BUSINESSOWNER_USER` supposed to have `Admin` persmissions? Looking at this [IDM roles doc](https://docs.google.com/document/d/1XeVbEimaNLwhVAa9rlPCVVnMDer22Blz/edit#heading=h.tyjcwt) seems like it's more fine-grained than just full admin access.

haworku: I don't *think* so, but I'm not confident. 

 @macrael or @MegMurphy do you know? Admin permissions for the frontend means they can see all the UI that the Admin user can for support or validation reasons.

services/app-graphql/src/schema.graphql
```diff
@@ -797,6 +797,20 @@ type CMSUser {
     divisionAssignment: Division
 }
 
+"CMSUser is a user that works for CMS, approving CMS approver request and able to do everything a CMSUser can do"
```
haworku, line 800:
```suggestion
"CMSApproverUser is a user that works for CMS, approving CMS approver request and able to do everything a CMSUser can do"
```

services/app-api/src/domain-models/user.ts
macrael: businesowner is Junni, I think giving her admin permissions is right

scripts/add_cypress_test_users.ts
```diff
@@ -243,6 +245,12 @@ async function main() {
             email: 'shi-tong@example.com',
             role: 'BUSINESSOWNER_USER' as const,
             state: undefined,
+        },
+        {
+            name: 'Azula',
+            email: 'azula@example.com',
```
haworku, line 251: please match the familyName given elsewhere. Need users setup between cypress and local to match

haworku (approved):
I ran through unlock and Q&A and saw things working!

services/app-api/src/domain-models/user.ts
MegMurphy: yes, BO should have admin permissions. Only difference btwn BO and Admin is what type of role requests they can approve in IDM. Junni is the BO role, and she can have admin rights.

PR #2637, opened 2024-08-08
MCR-4150: Do not display withdrawn rates on rates dashboard.

JasonLin0991:
## Summary
[MCR-4150](https://jiraent.cms.gov/browse/MCR-4150)

- Rates dashboard will skip rates with `withdrawInfo`.

#### Test cases covered

- `RateReviewsDashboard.test.ts`
   - `'renders dashboard without withdrawn rates'`

## QA guidance

The UI to withdraw a rate is not implemented yet, but this is a small task and it can be verified with tests. We can merge this in, but we should wait for acceptance testing until MCR-4152 and the issue found with the withdraw API have been merged. See this Slack [thread](https://cmsgov.slack.com/archives/C014CRU197U/p1723131423175859) for more information.

For reviewing this PR, I think making sure the code and test are good. Please comment if you think that would not be sufficient to approve this.

### DIFF
--- a/services/app-graphql/src/queries/indexRates.graphql
+++ b/services/app-graphql/src/queries/indexRates.graphql
@@ -19,6 +19,11 @@ query indexRates {
                     }
                 }
                 status
+                withdrawInfo {
+                    updatedAt
+                    updatedReason
+                    updatedBy
+                }
                 initiallySubmittedAt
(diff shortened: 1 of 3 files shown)


haworku (approved):
Yes! Looks good

PR #2638, opened 2024-08-08
MCR-4152, MCR-4231 Add form,  LinkRateSelect, and action buttons to replace rate

haworku:
## Summary
This is MCR-4231 and an initial pass of work for MCR-4152. Add on PR will be needed to complete AC.

#### MCR-4152
1. ✅ If I am an admin user on the submission summary page 
Then I see a "replace rate" button next to each rate review
2.  ✅ If I click on "replace rate" 
Then I am navigated to a screen where I see an form field to "Add a rate review to replace this rate" where I can select from a dropdown all the rate reviews for that state
3. ❌  If I am viewing the dropdown of rate reviews to replace the rate with
Then only rates for the the state whose submission I am on should appear in the dropdown, not rates from other states 
4.  ✅ If I type in part of the rate name into the typeahead 
Then the dropdown options narrow to the criteria of my search 
5.  ✅ If I make a selection and click "continue" 
Then I am navigated back to the submission summary page and the new rate I selected appears on the submission summary page.  
6. ❌ We call the new API successfully
7.  ✅ The withdrawn rate is not shown on submission summary or unlocked submissions (for either CMS or State users)

#### MCR-4231
 ✅ If a rate has been withdrawn for in a state 
Then if I am logged in as a user in that state and I start a new submission then I do not see the withdrawn rate appear on the linked rate dropdown on the rate details page

#### Related issues
https://jiraent.cms.gov/browse/MCR-4152
https://jiraent.cms.gov/browse/MCR-4231

#### Screenshots
- Don't bother reviewing for styles - can add polish and adjust margin later if we decide its priority. Here is current state:
Screenshot 2024-08-08 at 12 00 11 PM

#### Test cases covered
- Not React component tests yet
- I honed in some of the API tests a bit, I have plans to add more if I verify multi-rates is broken 

## QA guidance
- Please review for the AC with the green check mark only
- Please help me ensure this PR does not cause regressions for linked rates --  verify in whatever way makes sense. I did touch the component used for the connecting linked rates


services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
```diff
@@ -75,7 +77,7 @@ export const LinkRateSelect = ({
                 revision.formData.rateCertificationName ??
                 'Unknown rate certification',
             rateProgramIDs: programNames(
-                statePrograms,
```
JasonLin0991, line 78: I still see the withdrawn rate in the link rate dropdown selection when filling out the rate details page on a new submission. Somewhere in this component should filter out rates with `withdrawInfo`. 

Also, should withdrawn rates be shown in the rates dropdown on the replace rate page?

services/app-web/src/pages/ReplaceRate/ReplaceRate.tsx
```diff
@@ -1,63 +1,225 @@
...
+
+                                </fieldset>
+                        <PageActionsContainer>
+                    <ButtonGroup type="default">
+                        <ActionButton
+                            type="button"
+                            variant="outline"
+                            data-testid="page-actions-left-secondary"
+                            parent_component_type="page body"
+                            link_url={`/submissions/${id}`}
```
JasonLin0991, line 202: This cancel button doesn't navigate the user.

JasonLin0991 (approved):
Looks good for the 4152 items (minus the ones you pointed out), but for 4231 I still see the withdrawn rates on the rate details page.

services/app-web/src/pages/LinkYourRates/LinkRateSelect.tsx
haworku: >Also, should withdrawn rates be shown in the rates dropdown on the replace rate page?

No they shouldn't. You can't use a withdrawn rate to replace a rate on some other random submission.

😮‍💨  I forgot I need to adjust the props for RateDetails still thank you for pointing out.

haworku: @JasonLin0991 Oh wait your work is going to fix this  - since index rates with filter out withdrawn rates right

haworku: Oh right ...eh,whatever. Yes, it can be done either way. I would probably make the API accept optional filters like I just added for state - where it either can return all states or return just a specific state. Same think for including withdrawn or not.

To speed this up for now, I'll just do it clientside on this PR... follow the existing pattern we have currently.

PR #2639, opened 2024-08-08
MCR-4302: Fix GraphQL explorer tool

JasonLin0991:
## Summary
[MCR-4302](https://jiraent.cms.gov/browse/MCR-4302)

In our vite.config.ts we are using `'vite-plugin-graphql-loader'` to load our GraphQL schema. The output is a `DocumentNode` which I think differred from when we used webpack which outputted a string. The `ApolloExplorer` component only takes an `IntrospectionQuery` or `SDLstring` for the schema. So the fix here is to convert the document node into the `SDLstring` using `print()` from `graphql` library.


mojotalantikite (approved):
I swear I tried stringify-ing it 😂 🎉

PR #2645, opened 2024-08-09
avscan service: check if there is an error value before logging

mojotalantikite:
## Summary

I was looking around our cloudwatch logs for MCR-4354 and noticed a lot of `err undefined` in the logs. Looking at the code this seems normal, we were logging if there was something in `avResult.error` -- which usually is `undefined`. Just added checks here so that we don't see a bunch of `err undefined` in the logs thinking we're missing something.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4354


haworku (approved):
TY

PR #2686, opened 2024-08-20
MCR-4297 Remove CMS_QUESTIONS feature flag

haworku:
## Summary
Remove feature flag. Also addressed any errors under app-web `lint` command

#### Related issues
https://jiraent.cms.gov/browse/MCR-4297

#### Test cases covered
We already had feature flags for the feature ⭐  so we just kept to keep these on and don't have to adjust E2E tests much. 

## QA guidance
Q&A feature should still be visible (I think cypress alone would catch this)


services/app-web/src/contexts/AuthContext.tsx
```diff
@@ -97,7 +97,7 @@ function AuthProvider({
             )
             modalCountdownTimers.current = []
         }
-    }, [sessionIsExpiring]) // full dep array causes a loop, because we're resetting the dep in the useEffect
+    }, [sessionIsExpiring, logoutCountdownDuration]) // full dep array causes a loop, because we're resetting the dep in the useEffect
```
haworku, line 100: ooo just saw this comment seems like I need to lint ignore this one - will go back and do this

JasonLin0991 (approved):
✅

PR #2695, opened 2024-08-23
Pr mcr 4138 submission type contract details pages

pearl-truss:


services/app-web/src/testHelpers/apolloMocks/contractPackageDataMock.ts
```diff
@@ -1866,6 +1866,370 @@ function mockContractPackageUnlocked(
     }
 }
 
+function mockContractPackageUnlockedWithUnlockedType(
```
pearl-truss, line 1869: I'm aware this is very close to an existing mock. However switching out the existing mock to expect and return an unlockedContract type would have been a bit of a refactor since it's being used in multiple places. If folks feel like it is in scope to make that change, lmk and I can include it as a part of this PR

services/app-web/src/hooks/useContractForm.ts
```diff
@@ -0,0 +1,260 @@
...
+            downloadURL: doc.downloadURL,
+            name: doc.name,
+            s3URL: doc.s3URL,
+            sha256: doc.sha256
+        }
+    })
+}
+
+const stateContactsInput = (contacts: StateContact[]): StateContactInput[] => {
+    return contacts.map((contact) => {
```
haworku, line 47: Is the point of these input functions mainly to strip out `__typename `? 

For cases where that is the only difference -  could something like [this ](l https://www.apollographql.com/docs/react/api/link/apollo-link-remove-typename/#remove-__typename-from-all-variables ) be useful. Starting to wonder with moving to graphql types everywhere might it be better to do graphql level config change?

This is not something we have to implement now don't want to derail your work,  raising it more for discussion.

services/app-web/src/hooks/useContractForm.ts
```diff
@@ -0,0 +1,260 @@
...
+                        lastSeenUpdatedAt: contract!.draftRevision!.updatedAt,
+                        formData: input.formData
+                    },
+                },
+            })
+            const updatedSubmission =
+                updateResult?.data?.updateContractDraftRevision.contract
+            if (!updatedSubmission) {
+                setShowPageErrorMessage(true)
+                console.info('Failed to update form data', updateResult)
```
haworku, line 138: FYI everywhere with `console.info` preceeding `recordJSException` we can delete the inline console. The record exception function also does a `console.error` now.

services/app-web/src/hooks/useContractForm.ts
```diff
@@ -0,0 +1,260 @@
...
+                    return {interimState, createDraft, updateDraft,  showPageErrorMessage }
+            }
+        }
+        if (err.name !== 'SKIPPED') {
+            recordJSException(err)
+            interimState = 'GENERIC_ERROR'// api failure or protobuf decode failure
+            return { interimState, createDraft, updateDraft,  showPageErrorMessage}
+        }
+
+        if (!contract || !contract.draftRevision || !contract.draftRevision.formData || contract?.status === 'RESUBMITTED' || contract?.status === 'SUBMITTED') {
```
haworku, line 185: Now that we have UnlockedContract, I think this can be handled with a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html) like `isUnlockedContract`

services/app-web/src/testHelpers/apolloMocks/contractPackageDataMock.ts
haworku: Thanks for calling this out and paying attention to tests. My vote is let's land this work, ship,  but then you immediately follow on with that cleanup refactor PR during this sprint while its in mind.

services/app-web/src/hooks/useContractForm.ts
pearl-truss: Thanks! I meant to call this out as well because I felt like there had to be another way

services/app-web/src/hooks/useContractForm.ts
pearl-truss: I was getting errors without this check iirc, but I will remove and see if the type guard works here

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.test.tsx
```diff
@@ -187,19 +183,13 @@ describe('ContractDetails', () => {
...
-                    updateDraft: mockUpdateDraftFn,
-                    showPageErrorMessage: false,
-                    draftSubmission: {
-                        ...mockContractAndRatesDraft(),
-                        populationCovered: 'CHIP',
-                    },
-                }
+            const draftContract = mockContractPackageUnlockedWithUnlockedType()
+            draftContract.draftRevision!.formData.populationCovered = 'CHIP'
+            vi.spyOn(useContractForm, 'useContractForm').mockReturnValue({
```
macrael, line 188: I'm interested in us discussing this pattern in a broader testing patterns discussion. I don't understand the benefits to mocking out a bunch of the logic in useContractForm vs. the pattern we've been using mostly of mocking the fetchContract response at the ApolloClient layer

macrael: That said, this is obviously adapting a pattern we already had so I think this PR doesn't need to wait on that discussion

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.tsx
```diff
@@ -271,232 +217,359 @@ export const ContractDetails = ({
...
+            booleanAsYesNoFormValue(
+                draftSubmission.draftRevision.formData
+                    .modifiedNonRiskPaymentArrangements === null
+                    ? undefined
+                    : draftSubmission.draftRevision.formData
+                          .modifiedNonRiskPaymentArrangements
+            ) ?? '',
+        statutoryRegulatoryAttestation:
+            booleanAsYesNoFormValue(
+                draftSubmission.draftRevision.formData
```
macrael, line 387: how come we don't just make booleanAsYesNoFormValue work with null instead of doing this check everywhere?

services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
```diff
@@ -193,27 +203,137 @@ export const SubmissionType = ({
...
 
             try {
-                const updatedDraft = await updateDraft(draftSubmission)
+                const updatedContractInput: UpdateContractDraftRevisionInput = {
+                    formData: updatedDraftSubmissionFormData,
+                    contractID: draftSubmission.id,
+                    lastSeenUpdatedAt: draftSubmission.draftRevision.updatedAt,
+                }
+                const updatedDraft = await updateDraft(updatedContractInput)
                 if (updatedDraft instanceof Error) {
```
macrael, line 327: can this throw anymore? I thought we wrapped that in the updateDraft call so we could skip try/catch here?

macrael (approved):
This is great! Comments inline

services/app-web/src/hooks/useContractForm.ts
```diff
@@ -0,0 +1,258 @@
...
+            }
+        },
+        skip: !contractID
+    })
+    const contract = fetchResultData?.fetchContract.contract
+    if (fetchResultLoading) {
+        interimState = 'LOADING'
+        return {interimState, createDraft, updateDraft, showPageErrorMessage }
+    }
+    if (fetchResultError) {
```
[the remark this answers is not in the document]
pearl-truss: I found I had to recreate these fields explicitly because these particular fields that came back from `fetchContract` were set to being possibly undefined and unlockedType doesn't accept undefined for them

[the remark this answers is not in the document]
pearl-truss: I'll look into using the apolloQueryWrapper. I think useHealthPlanForm uses it iirc

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.test.tsx
pearl-truss: I wasn't sure if there were strong opinions on this but like the idea of a wider discussion. For this I used the existing pattern for this file to save time

haworku (approved):
I did manual QA today trying to cover the save as draft back and forth behaviors that caused issues earlier. I saw things working ⭐

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.test.tsx
haworku: FYI there's nothing intentional going on here in the code from before. Down to talk in RSP but I don't consider anything with useHealthPlanPackageForm a pattern to copy, it was hurridedly created trying to de-scope linked rates work, it was not planned for, and it  was always meant to be throwaway. I want us to create patterns and be more intentional now since we have actual sprint time dedicated to Contract API.

pearl-truss: I still need to make a separate cleanup ticket for the unlockedContract mock in test. I can add in refactoring the contract details test to use the updated pattern as well

PR #2699, opened 2024-08-26
Fixes for local and CI linting

mojotalantikite:
## Summary

Some of us have been noticing local pre-commit linting issues. This changes how we call lint-staged and adds a CI step for running our linters across the repo. Running lint against the repo only shows some errors in the `uploads` service, so there are some fixes for that here as well.

#### Related Issues
https://jiraent.cms.gov/browse/MCR-4390

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -94,6 +94,10 @@ jobs:
       - name: Setup env
         uses: ./.github/actions/setup_env
 
+      - name: Generate files
+        shell: bash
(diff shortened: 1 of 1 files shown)


mojotalantikite:
Ok there are more errors than I was expecting. Looks like `pnpm` fails fast, and the `uploads` and `app-proto` lint failures stopped the full lint to stop early. There are a bunch in `app-api` around protos that need fixing too. I'll reopen this with just the pre-commit change and we can focus on the CI stuff some other time.

PR #2709, opened 2024-08-28
MCR-4410 & MCR-3757 fix show more text on banners

JasonLin0991:
## Summary
[MCR-4410](https://jiraent.cms.gov/browse/MCR-4410)
We moved banners outside the grid containers that set a max width. The banners `show more` text needs to have a max width in order for it to know when to clamp long strings.

This fix places both the step indicator and the page banners inside it's own grid container,`FormNotificationContainer`.

A good refactor is to move the step indicator and page banners to the `StateSubmissionForm` component, but for now we need to get this merged in so that users can continue their work.

[MCR-3757](https://jiraent.cms.gov/browse/MCR-3757)
I did not see this happen, but included it anyways to review that it is no longer an issue. This was created last year in November so changes in the project since then may have fixed this.

### DIFF
--- a/services/app-web/src/components/FormContainer/FormContainer.module.scss
+++ b/services/app-web/src/components/FormContainer/FormContainer.module.scss
@@ -6,3 +6,8 @@
     display: flex;
     flex-direction: column;
 }
+
+.notificationContainer {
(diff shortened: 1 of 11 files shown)


pearl-truss (approved):
Tested this locally and it looks good

PR #2715, opened 2024-08-29
MCR-3080: bring back a11y tests in ci

JasonLin0991:
## Summary
[MCR-3080](https://jiraent.cms.gov/browse/MCR-3080)

- Removed `pa11y` plugin
   - Tried unsuccessfully to try to get `pa11y` to work again.
- Added `cypress-axe` and `axe-core` to run our a11y tests.
- Updated documentation.
- Added 2 accessibly tests
   - `'has no a11y violations on submission form with form input errors'`
   - `'has no a11y violations on CMS dashboards'`
- Removed `protobufjs` from `package.json` in `services/cypress` 
- Fixed an issue with local Cypress LD interceptions.
   - We are using some kind of proxy for LD api requests so for local deployment we have to look for a different url.

The a11y logs for violations could be better. Rather than including it into this ticket I have a new one to improve the logs.
https://jiraent.cms.gov/browse/MCR-4416

These are two more tickets that add more a11y testing to other parts of the app that I think could really use it. They are not completely filled out and could use some refinement.
https://jiraent.cms.gov/browse/MCR-4417
https://jiraent.cms.gov/browse/MCR-4418

There are three violations with the accessibility testing. So to not increase the scope of this ticket, I made two tickets to address the violations

https://jiraent.cms.gov/browse/MCR-4420
https://jiraent.cms.gov/browse/MCR-4421

#### Test cases covered

`acccessibility.spec.ts`
   - `'has no a11y violations on submission form with form input errors'`
   - `'has no a11y violations on CMS dashboards'`


pnpm-lock.yaml
```diff
@@ -13740,10 +13756,10 @@ snapshots:
     dependencies:
       '@aws-crypto/sha256-browser': 5.2.0
       '@aws-crypto/sha256-js': 5.2.0
-      '@aws-sdk/client-sso-oidc': 3.609.0(@aws-sdk/client-sts@3.637.0)
+      '@aws-sdk/client-sso-oidc': 3.609.0(@aws-sdk/client-sts@3.609.0)
       '@aws-sdk/client-sts': 3.609.0
       '@aws-sdk/core': 3.609.0
-      '@aws-sdk/credential-provider-node': 3.609.0(@aws-sdk/client-sso-oidc@3.609.0(@aws-sdk/client-sts@3.609.0))(@aws-sdk/client-sts@3.609.0)
+      '@aws-sdk/credential-provider-node': 3.609.0(@aws-sdk/client-sso-oidc@3.609.0(@aws-sdk/client-sts@3.609.0))(@aws-sdk/client-sts@3.637.0)
```
JasonLin0991, line 13762: @mojotalantikite I had a bad merge and reverted. But when I merge main in and install, it changes this part of the lock file. I'm not sure why this differs from main as I did'n't change anything with these libraries. Is this going to be an issue?

package.json
```diff
@@ -71,6 +71,8 @@
     "devDependencies": {
         "c8": "^10.1.2",
         "cypress": "^13.13.2",
+        "axe-core": "^4.10.0",
+        "cypress-axe": "^1.5.0",
```
JasonLin0991, line 75: @mojotalantikite I added these two dependencies here too. Correct me if things changed but, I did this because Cypress in CI runs from root and on local is runs from `services/cypress` without this here CI could not find the dependencies. A lot has changed since so is there a better, but simple, way to not have to install in both places? 

I could do some configuration in each test to point to the location of the install based on deployment, but if installing the deps in both doesn't cost or complicate much would this be ok?

docs/architectural-decision-records/005-frontend-a11y-toolset.md
```diff
@@ -19,7 +19,7 @@ Together these tools include linting, testing, and auditing loaded web pages for
 
 [eslint jsx-a11y][eslint-a11y] Provides React-friendly code linting which allows developers to find accessibility issues during local development.
 
-[pa11y][pa11y] and [pa11y-ci][pa11y-ci] - Accessibility test runner for end to end testing in the command line or Node.js. End to end testing can catch concerns that are not noticeable in linting or unit tests since tests run on browser.
+[cypress-axe][cypress-axe] and [axe-core][axe-copre] - Accessibility test runner for end to end testing in the command line or Node.js. End to end testing can catch concerns that are not noticeable in linting or unit tests since tests run on browser. Our previous choice [pa11y][pa11y] seems to not be maintained anymore and no longer works with our end to end testing.
 
 [jest-axe][jest-axe] - This tool provides additional jest matchers, like `.toHaveNoViolations`, to use in unit tests. This may be redundant when end to end accessibility testing is present.
```
haworku, line 24: Also a request - could you please also clarify that `jest-axe`is removed from our tool set?  Reason is we aren't using it and it was removed from package.json I think awhile ago so its not in use.

services/cypress/integration/stateWorkflow/stateSubmissionForm/accessiblility.spec.ts
```diff
@@ -1,73 +1,114 @@
...
+        }).should('not.have.attr', 'aria-disabled')
+        cy.findByRole('button', {
+            name: 'Continue',
+        }).safeClick()
+        cy.checkA11yWithWcag22aa()
 
-            cy.deprecatedNavigateV1Form('CONTINUE')
-            cy.deprecatedNavigateV1Form('CONTINUE')
+        cy.fillOutContractActionAndRateCertification()
+        cy.deprecatedNavigateV1Form('CONTINUE_FROM_START_NEW')
```
haworku, line 31: I thought Pearl merged her work and start, submission and contract details are are all on v2. I think this could be using the v2 helpers?

JasonLin0991: I don't see a v2 of `fillOutContractActionAndRateCertification`. I do see `startNewContractOnlySubmissionWithBaseContractV2`, but the purpose of the code here is to:
- check a11y violations on the empty form with field errors
- fill out the form and continue so that we have a submission to go to each form page.

Using `startNewContractOnlySubmissionWithBaseContractV2` would fill this out and continue before we could check for a11y violations.

services/cypress/support/accessibilityCommands.ts
```diff
@@ -0,0 +1,32 @@
...
+    cy.task('table', violationData)
+}
+
+Cypress.Commands.add(
+    'checkA11yWithWcag22aa',
+    () => {
+        cy.checkA11y('', {
+            runOnly: {
+                type: 'tag',
+                values: ['wcag22aa']
```
haworku, line 28: I don't think this is working - I added basic accessibility violations (like an image missing alt text, a button with no text content) and there were no errors. When I change this value I start to get errors again. I think it is not working for some version saw a [github issue](https://github.com/component-driven/cypress-axe/issues/123) too

haworku: I think I'm seeing `best-practice` work lets just use that if you agree that 22aa is broken and so is wcagaa right now.

services/cypress/integration/stateWorkflow/stateSubmissionForm/accessiblility.spec.ts
haworku: Look at `navigateCOntractAndRatesForm` and `navigateContractForm`.

package.json
haworku: @mojotalantikite @JasonLin0991 imo this could alkso be good time to also consider if cypress should be moved considering new world we are. Throwing it out there, not trying to suggest moving the entire folder is needed ...just want us to think expansively since we are also doing a lot around modularizing and simplifying our top level architecture right now so maybe we are seeing another opportunity here.

mojotalantikite: This is totally fine. I personally think it's clearer to declare the dependencies in each service's `package.json` that needs them. Back with Yarn 1 everything got hoisted and when we moved to pnpm it started complaining about missing packages, as it was easy to just ignore the dependency if it was already installed elsewhere with Yarn. 

Plus with pnpm everything is symlinked around, so it's not actually going to install it in two places, it's just going to throw in a symlink from the global package store to each service's `node_modules` that declares it.

pnpm-lock.yaml
mojotalantikite: hmm, I'm not sure why in this specific case, but I'd guess that maybe there are some transitive dependencies that might specify `client-sts` somewhere? I'd have to dig in to check, but honestly I don't think it'll be a problem. As long as CI is passing I'd say we're fine!

services/cypress/support/accessibilityCommands.ts
JasonLin0991: Just to make visible our conversation. Rules for each standard are not cumulative. So in order to test all rules, all the standards must be included. If we only include `'wcag22aa'` then only the specific rules added in for that standard will be tested.

The configuration has been updated to include the complete rules 

```
      cy.checkA11y('', {
          runOnly: {
              type: 'tag',
              values: ['wcag2a','wcag2aa', 'wcag21a', 'wcag21aa','wcag22aa']
```

haworku (approved):

services/cypress/support/accessibilityCommands.ts
sethims: Hi @JasonLin0991 and @haworku , Even in this above eg:
cy.checkA11y('', {
    runOnly: {
        type: 'tag',
        values: ['wcag2a','wcag2aa', 'wcag21a', 'wcag21aa','wcag22aa']
        

No violations are getting shown for wcag21a and wcag22aa.

JasonLin0991: @sethims I followed the `axe-core` docs under [axe.run()](https://www.deque.com/axe/core-documentation/api-documentation/#parameters-axerun) to figure out that the tags need to be combined. You can look through the rule of each standard using [axe.getRules()](https://www.deque.com/axe/core-documentation/api-documentation/#api-name-axegetrules). Compare the rules in there with the specific rule violations you are testing for. Good luck!

PR #2719, opened 2024-09-03
MCR-3849 Fix session timeout

haworku:
## Summary
Address issues with session timeout and re-implement with [IdleTimer ](https://idletimer.dev/). 

**Issues addressed include:**
- logout and continue work across tabs
- if one tab has session timeout modal open then user opens another tab and then lets computer go to sleep, both tabs will be logged out of MC-Review after time elapses with no action
- if user tries to add a document with file upload but it fails due to IDM having been logged out in the background for whatever reason, MC-Review will also force logout at this point and display session timeout message so the user knows what happened

Still in progress
- [x] write up acceptance jams steps
- [x] finish adding `AuthenticatedRouteWrapper` tests for idle timer behaviors

#### Related issues
https://jiraent.cms.gov/browse/MCR-3849

#### Screenshots
Screenshot 2024-09-04 at 4 13 22 PM

#### Test cases covered
Session Timeout acceptance jam document forthcoming

## QA guidance
Please re-test[ the steps from this comment in Jira](https://jiraent.cms.gov/browse/MCR-3849?focusedId=2925189&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2925189).

And also pick a couple of these [acceptance jam](https://docs.google.com/spreadsheets/d/1IYXcXkc47ogrjFWKrHUoBGFfNHrhfovXDU5CSf2oF0c/edit?gid=0#gid=0) cases to retest. Keep in mind the timeout for toph in review apps is set to 2min to make things go faster so you won't have to wait as long to verify things.


pearl-truss:
@haworku for "logout and continue work across tabs", is this the desired behavior? When I logged out of one tab I was still able to do work in another tab

haworku:
@pearl-truss No, that's the buggy behavior that we want to test for and hopefully we don't see it. Let me re-check this - I think I tested multi-tabs mainly locally not on the deployed app.

pearl-truss:
> @pearl-truss No, that's the buggy behavior that we want to test for and hopefully we don't see it.
> 
> Can you send me the replication steps you used and which user in the review app you were using (toph currently has shorter session duration)

I was using aang. 

1. I signed into 2 tabs
2. I signed out of one tab (it wasn't a session time out, I explicitly signed out)
3. I used the remaining open tab to view a draft submission

haworku:
@pearl-truss TY for finding this!  I added to [acceptance testing script](https://docs.google.com/spreadsheets/d/1IYXcXkc47ogrjFWKrHUoBGFfNHrhfovXDU5CSf2oF0c/edit?gid=0#gid=0) under a new section called basic login and logout.

FYI from discussion in standup today, looks like this one is an [existing bug ](https://jiraent.cms.gov/browse/MCR-1719) not a new regression, and not in scope for this ticket.  However, I will still look into solving it also on this branch in case I can fix. The bug shouldn't block this work going through, which is focused on session timeout.

 If you see any other issues,  please raise them – this manual testing is very helpful.

pearl-truss (approved):
I saw your comment, the other bullet points I tested looked good

haworku:
I just found a bug 🐛 with the review and submit modal blocking the session expiration modal which causes the app to freeze up. Fixing now.

services/app-web/src/contexts/S3Context.tsx
```diff
@@ -51,9 +53,17 @@ const useS3 = (): S3ContextT => {
         if (isS3Error(s3Key)) {
             const error = new Error(`Error in S3: ${file.name}`)
             recordJSException(error)
+            // s3 file upload failing could be due to IDM session timeout
+            // double check the user still has their session, if not, logout to update the React state with their login status
+            try {
+                await checkAuth()
```
macrael, line 59: why does checkAuth throw instead of return an error?

services/app-web/src/components/Modal/SessionTimeoutModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export const SessionTimeoutModal = ({
+  modalRef,
+}: SessionTimeoutModalProps): React.ReactElement | null => {
+    const idleTimer = useIdleTimerContext()
+    const [countdownSeconds, setCountdownSeconds]= useState(idleTimer.getRemainingTime() / 1000)
+
+    const handleLogoutSession = async () => {
+        idleTimer.message({action:'LOGOUT_SESSION'}, true)
```
macrael, line 19: could these actions be constants?

services/app-web/src/pages/Auth/cognitoAuth.ts
```diff
@@ -129,33 +130,28 @@ export async function resendSignUp(
...
-                throw e
+                recordJSException(`UNEXPECTED SIGNIN ERROR AmplifyError ${e.code} – ${e.message}`)
             }
         } else {
-            console.info('didnt even get an amplify error back from login')
-            throw e
+            recordJSException(`UNEXPECTED SIGNIN ERROR – 'didnt even get an amplify error back from login`)
         }
     }
+    throw Error('Cognito SignIn error')
```
macrael, line 154: It looks like we used to return errors when it was an error we understood and only threw when it was something truly unexpected. I don't love turning this into something that always throws. Honestly I think since writing this we've moved towards just returning un-understood errors as well

macrael (approved):
Love getting to replace custom code with a solid library. I have some questions about us throwing errors we get from Amplify but otherwise this looks great

services/app-web/src/pages/Auth/cognitoAuth.ts
haworku: @macrael Actually looking more at this code - all of cognitoAuth throws for errors.  The oddity was returning both Errors and throwing and thats why I moved it to throw to unify it and unify that interface on the frontend. What do you want to do here? Shall I make a ticket to refactor cognitoAuth? 

Shall I just change this one still?

macrael: If it's not going to change having to do try/catch in our pages then I don't care if you fix this one or not, but I would like to have a ticket tracking moving this to returning errors instead. My goal is that inside any of our pages/components we never have to call try/catch because anything that can throw has been wrapped up as a dependency

haworku: Ended up rewriting `cognitoAuth` anyway - but there are other files that have the same issue that I left alone. Looks like we already have a[ ticket](https://jiraent.cms.gov/browse/MCR-1817) to address this issue - I updated it to also include adding more eslint rules.

pearl-truss:
@haworku on the review app as soon as I signed in I got a session timeout modal

PR #2720, opened 2024-09-04
MCR-2601 Add supporting docs to contract details page

pearl-truss:
## Summary

This PR adds a file upload for supporting docs the contract details page

#### Related issues
https://jiraent.cms.gov/browse/MCR-2601

#### Screenshots
Screenshot 2024-09-04 at 11 55 34 AM

#### Test cases covered

When flag is on file upload works for the supporting docs field on the Contract Details page. Also updated existing Contract Details test to work regardless of if the flag is on or not

## QA guidance

As a state user create a submission. Contract Details page now has a supporting docs field page, add a supporting doc and follow through with submitting the form


.envrc
```diff
@@ -24,14 +24,14 @@ export VITE_APP_S3_DOCUMENTS_BUCKET='local-uploads'
 export VITE_APP_S3_QA_BUCKET='local-qa'
 export DATABASE_URL='postgresql://postgres:shhhsecret@localhost:5432/postgres?schema=public&connection_limit=5' # pragma: allowlist secret
 export EMAILER_MODE='LOCAL'
-export LD_SDK_KEY='this-value-must-be-set-in-local'
+export LD_SDK_KEY='sdk-9f6315a3-ad22-4af8-9e21-b5caf2e12b98'
```
haworku, line 27: @pearl-truss oops - this value shouldn't be sent in envrc, its always read for local envrc or set in CI. lets regenerate new keys also now that this has been published.

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.tsx
haworku, line 1: I think supporting documents also should be added to the contract details formik schema right? I see `supportingDocuments: validateFileItemsList()`_ getting called for rate details page where we make sure documents if included finish loading before we allow submission. I believe we need something similar here.

haworku:
I noticed something I think we missed in the AC for this epic. 

We need to update the edit link display for contract supporting documents. That Edit button shouldn't display at all above the contract supporting documents table when the flag is on
Screenshot 2024-09-05 at 4 34 57 PM

.envrc
pearl-truss: whoops whoops! I reset the values in LD

haworku (approved):
LGTM

PR #2723, opened 2024-09-04
Bump lambda environment's memory in prod

mojotalantikite:
## Summary

This increases the memory for the lambda zip function in prod, in order to speed up the processing of files. This fixes an issue where a user is attempting to view a submission that has a large amount of files and the zip function is hitting the max runtime of 30s in AWS Lambda.


services/app-api/serverless.yml
```diff
@@ -227,7 +230,8 @@ functions:
           method: post
           cors: true
           authorizer: aws_iam
-    timeout: 30 # zipping large amount of files hits default time out of 6 seconds
+    timeout: 30
```
macrael, line 233: is 30s the longest we can get?

macrael (approved):

services/app-api/serverless.yml
mojotalantikite: It's an API gateway limit, not a Lambda one. Since it's attached to the API gateway we're stuck with a max of 30. Lambda itself I think is 15 minutes.

PR #2728, opened 2024-09-05
MCR-4261: cms and admin can navigate settings via sidebar

JasonLin0991:
## Summary
[MCR-4261](https://jiraent.cms.gov/browse/MCR-4261)

- [x] Each email setting has a named route nested under /mc-review-settings
- [x] Theres a sidebar on the settings page
- [x] As a cms user, when I click the the sidebar item for the specific email setting, I go to that named route and see that --specific settings table
- [x] When a user visits  /mc-review-settings from the link, they should land on the "State assignments" table, also the first setting link in the sidebar

Extra goodies:
- Fixed filtering assignments by email
   - Filtering by `No assignments` still doesn't work. I updated [MCR-4335](https://jiraent.cms.gov/browse/MCR-4335) to reflect this.
- Fixed the `/settings` route to redirect to `/mc-review-settings`
- Renamed all the `Settings` page table components to reflect the new side-nav links and titles.
- All API requests are done in `Settings` and pass down in `Outlet` context.
- Consolidated all the tests into `Settings.test.tsx` there where several similar tests.


haworku:
The redirect off settings broke the thing I built to test crashes and error alerts. See `monitoring.md` "How to test monitoring". Can you move this test=crash test=error handling to `/dev` then instead? And update docs. We still need a way to verify when new relic alerts aren't working (as happened again this week).

services/app-web/src/pages/Settings/SettingsTables/DivisionAssignmentTable.tsx
```diff
@@ -152,17 +150,23 @@ function CMSUserTableWithData({
...
+            size: 250,
+        },
         getCoreRowModel: getCoreRowModel(),
     })
 
     return (
-        <Table>
-            <caption className="srOnly">CMS Users</caption>
-            <thead className={styles.header}>
+        <Table bordered fullWidth className={styles.table}>
```
haworku, line 160: Styles nit:
- This table has a `<p>` description like the others looking at the figma. 
- Also this table and the other tables have a heading that is a large font size but is not bold. Let's make sure we clean that up it effects the polish quite a bit I'm noticing.

Current:
Screenshot 2024-09-05 at 10 11 11 AM

Expected:
Screenshot 2024-09-05 at 10 08 07 AM

services/app-web/src/pages/App/AppRoutes.tsx
```diff
@@ -235,8 +240,36 @@ const CMSUserRoutes = ({
...
+                        element={<AutomatedEmailsTable />}
+                    />
+                    <Route
+                        path={RoutesRecord.SUPPORT_EMAILS}
+                        element={<SupportEmailsTable />}
+                    />
+                </Route>
+                <Route
+                    path={RoutesRecord.SETTINGS}
+                    // Until we update the helpdesk documentation for the /mc-review-settings route, we are keeping this
```
haworku, line 269: 👍 thanks for the comment

services/app-web/src/pages/Settings/Settings.tsx
haworku, line 1: For discussion: The users data only relevant for one of the routes, correct? With the current implementation, I see drawbacks to hinging all the settings display on `indexUsers` and I can't see the benefit to this front loading approach  (assuming we use graph cache as you have been so it wouldn't matter if we tab back and forth).  Let me know what you are thinking though

My reasoning is that this is an index API returns much more data and in a different format. It 1. may have slower response times 2. will likely eventually require different error handling if/when parts of data are incomplete 3. will likely need to bring down paginated data in the future.  Would rather see just that cms users table page error if indexUsers has issues.

haworku (approved):
I see things working according to the main AC. Thanks for fixing filters and conslidating tests.  I left some comments I would like to see responded to or fixed

JasonLin0991:
> The redirect off settings broke the thing I built to test crashes and error alerts. See `monitoring.md` "How to test monitoring". Can you move this test=crash test=error handling to `/dev` then instead? And update docs. We still need a way to verify when new relic alerts aren't working (as happened again this week).

@haworku I added this back into the `Settings` page and updated the docs to use `/mc-review-settings`. I did not go with the `/dev` route because this route isn't available in prod, unless this test is not suppose to be available in prod.

PR #2729, opened 2024-09-05
Move shared code into modules

mojotalantikite:
## Summary

In moving to `pnpm`, `vite`, and `esbuild` we had to make a copy of the shared `common-code` that lived in `app-web` over into `app-api`. This directory of `common-code` had gotten a bit out of control, and had been using relative path imports to bring in dependencies from other parts of `app-web`. There were all sorts of build issues in `app-api` moving to `esbuild` because of this, and that's why we had to temporarily copy over the `common-code` directory to `app-api`. This PR finally breaks `common-code` out of `app-web` and into reusable modules, imported via `pnpm` workspaces.

#### Related issues
https://jiraent.cms.gov/browse/MCR-1873


haworku:
@mojotalantikite Great work defining what packages could even be here. I like the way you divided things on a conceptual level.
I'm sorry it took so long to review. I wish I had asked for a code walkthrough to just give you feedback earlier, looking back.

 I'd like to chat more about implementation. A few things seem off to me for creating new packages to be consumed by our services. 
1. I'm seeing new compiled `.js` and `.js.map` files here in the git diff. I would expect to only see ts files even for new packages. In general, I get suspicious when compiled files like this show up in github, it open the door for the ts and js to get out of sync. 
2. I expect each package to have an `exports` field defined in their package.json and there be scripts defined to get each package to build that are then used in our main build process
3. Any third party lib should have their types defined, don't want to be redefining something coming from a third party that exports their own type definitions (e.g. dayjs I saw, not sure if there are others) 

Let me know your thoughts here or on slack.

haworku (commented):
Also I think each package needs a readme, let me know if you need help with that. We should define the boundaries of each package now.

I'm thinking also we try to totally delete `common-code` rather than keeping two misc bucket packages around. Both helpers and common code feel like catch all. What do you think @macrael ? 

Most of what I see in common code feels like it could go in the catch all `helpers` or should go in a new package. called `config` thats feature flags and auth config stuff.  . And then eventually we can organize helpers into subcategories (later work)

haworku:
@mojotalantikite I don't want to block your work with my proposed changes. I'm not leaning towards lets land this, then iterate on it especially if you said the merge conflicts were annoying.

JasonLin0991 (approved):
It deploys locally. No issues running tests and cypress.

PR #2737, opened 2024-09-10
MCR-4409: update state programs

pearl-truss:
## Summary
- Updated state programs. Noticed we now have statePrograms.json in app-api as well as app-web so I updated both
- Some documentation updates 

#### Related issues
https://jiraent.cms.gov/browse/MCR-4409


haworku (approved):
Oof thanks for remembering to update both. That should go back to 1 soon.

PR #2739, opened 2024-09-11
MCR-4451: update tealium script

pearl-truss:
## Summary

I hope I interpreted the instructions correctly to change the domain to tealium-tags.cms.gov in the tealium.ts inline script

#### Related issues
https://jiraent.cms.gov/browse/MCR-4451


haworku (approved):
👍 TY

PR #2743, opened 2024-09-13
Rate Index Speedup and Perf Docs

macrael:
## Summary

I finally just investigated performance manually and found a massive speedup. We were requesting documents in indexRates which hits the S3 APIs to generate download links for every single document. This removes that on the indexRates query (creating a new indexRatesForDashboard query because we still want documents on the indexRates query used by the link-rates chooser)

Should cause a massive speedup on prod. 

#### Related issues
Already closed, I can't find it.


services/app-graphql/src/queries/indexContracts.graphql
```diff
@@ -87,20 +87,7 @@ fragment rateRevisionFragmentForIndexContract on RateRevision {
     formData {
         rateType
         rateCapitationType
-        rateDocuments {
```
pearl-truss, line 90: 👍 this makes sense, and I'll update the mockIndexContract to account for this. I think I'll let this merge first and then push my changes

pearl-truss (approved):
LGTM

PR #2748, opened 2024-09-16
fix: Pull Tealium change off prod, keep on lower env

haworku:
## Summary
Follow on to #2739 - I realized they instructed us **not** to deploy to prod yet, only to lower environments. Makes sense, I see that the new script isn't working in prod when I look in the console. Bringing back the old script for prod.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4458

#### Test cases covered
This will be tested after deploy


pearl-truss (approved):
Good catch!

PR #2753, opened 2024-09-16
Disable Apollo Browser and introspection queries in val 

mojotalantikite:
## Summary

We were asked to disable the Apollo browser and introspection queries on the Val environment for security compliance. This does the thing.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4431


pearl-truss (approved):
confirmed it still works on review app

PR #2118, opened 2023-12-14
MCR-3814 Fix session timeout

haworku:
## Summary

- The true fix the production issue was adjusting Launch Darkly config for `SESSION_EXPIRING_MODAL` to serve `true` again. That is done.
- I removed `MODAL_COUNTDOWN_DURATION` feature flag entirely from code. I deprecated the flag manually in LD. After this work merges through, I will archive that flag. It is not needed.
- I added a new doc documentation session expiration.  I did not see a product brief on this topic but I know there were several design and eng decisions made during implementation. I looked at closed Jira tickets ([MCR-1665](https://qmacbis.atlassian.net/browse/MCR-1665) and [MCR-1666](https://qmacbis.atlassian.net/browse/MCR-1666)) to get the context and did some manual testing as well to surface requirements for the feature. 

#### Related issues
https://qmacbis.atlassian.net/browse/MCR-3814

## QA guidance
- [x] Check that session expiration modal and timeout works as expected as my review app. I have currently turned on the feature for toph@example.com and shortened the session duration to 2 minutes for easy testing. 
- [x] Read my doc and ask any questions or make direct edits.
- [x] Look at Launch Darkly config and the description of flags. If anything is unclear, ask me questions.

PR #2754, opened 2024-09-16
Fix: get session banners displaying again 

haworku:
## Summary
Related to #2118 

One of the code review fixes caused regression, broke the banners. Bringing that back. Yes, we need Cypress for this workflow, need to make time for this.

Screenshot 2024-09-16 at 2 30 49 PM

#### Related issues
https://jiraent.cms.gov/browse/MCR-3849

## QA guidance
When session times out it should log you yes, but it should also display session expired banner


pearl-truss (approved):
LGTM

PR #2766, opened 2024-09-20
MCR-4472: update Tealium staging urls

pearl-truss:
#### Related issues
https://jiraent.cms.gov/browse/MCR-4472

## QA guidance

I'm trying to figure out how to test this, if there's a way lmk


haworku (approved):
I think thats right, TY. manually building the script tag is annoying.

haworku:
@pearl-truss  I think the tags aren't added on dev. So the way I would test before merging would just be running locally but changing `App.tsx` temporarily to load the tealium client and double check in the head tag or by console.logging that things look right.

haworku (approved):

PR #2767, opened 2024-09-20
update state analyst by state

macrael:
## Summary

This PR adds a new API for updating state analysts by the whole state instead of updating the states assigned to a single user. The UI works this way so the API should work this way. 

#### Test cases covered

Wrote new API tests

## QA guidance

API isn't being hit yet so can't really validate it manually

### DIFF
--- a/services/app-api/src/resolvers/user/updateStateAssignmentsByState.ts
+++ b/services/app-api/src/resolvers/user/updateStateAssignmentsByState.ts
@@ -86,7 +86,7 @@ export function updateStateAssignmentsByState(
                 throw new UserInputError(errMsg, {
                     argumentName: 'assignedUsers',
                     argumentValues: assignedUsers,
-                    cause: 'USERS_ARE_NOT_STATE_USERS',
+                    cause: 'BAD_ASSIGNED_USERS',
                 })
             }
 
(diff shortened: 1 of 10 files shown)


services/app-api/src/resolvers/user/updateStateAssignmentsByState.ts
```diff
@@ -0,0 +1,109 @@
...
+            }
+
+            if (result instanceof UserInputPostgresError) {
+                const errMsg = result.message
+                logError('updateStateAssignmentsByState', errMsg)
+                setErrorAttributesOnActiveSpan(errMsg, span)
+                throw new UserInputError(errMsg, {
+                    argumentName: 'assignedUsers',
+                    argumentValues: assignedUsers,
+                    cause: 'USERS_ARE_NOT_STATE_USERS',
```
haworku, line 89: This error confuses me.  Is it supposed to be USERS ARE NOT CMS USERS or USERS ARE NOT VALID STATE ASSIGNMENTS?

 The way its written makes it seem like the users need to be state users are they are not. But assigned users never  state users for this function

haworku (approved):
LGTM  thank you for catching this for the team!

PR #2778, opened 2024-09-24
Revert "Revert "MCR-4140 CMS submissions dashboard uses contract api""

pearl-truss:
Reverts Enterprise-CMCS/managed-care-review#2771


macrael (approved):
Is the indexContracts query actually used anywhere else? If not can we delete it?

I'm curious what all you stripped out of it that sped things up? Glad for this query to only ask for exactly what it needs

pearl-truss:
> Is the indexContracts query actually used anywhere else? If not can we delete it?
> 
> I'm curious what all you stripped out of it that sped things up? Glad for this query to only ask for exactly what it needs

I couldn't get the perf hooks or node prof to work for me. I just ripped out what the query didn't need

PR #2779, opened 2024-09-24
Save Analyst in settings page

macrael:
## Summary

This PR hooks up the app-web settings UI to the app-api assign state users mutation. 

I got looped by the return type of FieldSelect which was different from the overall Option type we were using. I think if we wanted FieldSelect to take and return a simple array of values as its value, and take in a FieldOptionType of options we could but it would require further wrapping and type munging. So as is, the values, initial values, and options for FieldSelect are all FieldOptionType. This means that Formik will have to adopt the FieldOptionType as its internal representation of the fields, I think.

#### Related issues

https://jiraent.cms.gov/browse/MCR-4266

#### Test cases covered

Test the UI sending new user assignments

## QA guidance

I think we should have enough users in the review app to test their assignments


haworku (approved):
Let's merge this and build off it! I will take over ticket

PR #2780, opened 2024-09-24
Send emails to state analysts configured in db when FF is on

macrael:
## Summary

updates Submit, Unlock, CreateQuestion, and CreateQuestionResponse to all send emails to the db-configured state analysts instead of the parameter store state analysts if the `read-write-state-assignments` FF is enabled. 

#### Related issues

https://jiraent.cms.gov/browse/MCR-4277

#### Test cases covered

had to copy over some tests from unlockHPP to unlockContract that weren't covered there anymore.

## QA guidance

Without #2779 you can't really validate this in the deployed app yet


services/app-api/src/emailer/emails/unlockContractCMSEmail.ts
```diff
@@ -51,21 +51,16 @@ export const unlockContractCMSEmail = async (
...
-        contractRev.formData.submissionType === 'CONTRACT_AND_RATES'
-
     const data = {
         packageName,
         unlockedBy: updateInfo.updatedBy.email,
         unlockedOn: formatCalendarDate(updateInfo.updatedAt),
         unlockedReason: updateInfo.updatedReason,
         shouldIncludeRates:
             contractRev.formData.submissionType === 'CONTRACT_AND_RATES',
-        rateInfos:
```
macrael, line 64: I found a bug where unlocking CONTRACT_ONLY contracts would error in the eta templates. Looks like there is an additional barrier in the resubmit templates that prevented it but it seemed simpler just to remove this check since the templates already check for if rateInfos.length > 0

JasonLin0991 (approved):
✅

PR #2787, opened 2024-09-26
Add MIME checks to uploads avScan service

mojotalantikite:
## Summary

From the security audit the team wanted us to implement MIME file checking, ensuring that a file matches it's declared MIME type. This adds that check to the scan and will tag the file with an `INFECTED` if the file types mis-match, like is done for `clamav` findings. I had to implement custom checks for `csv` and `txt` files, as these don't have header declarations of their file types.

This also fixes `uploads` unit tests, as it turns out they were not working previously. It required a migration to `vitest` for the service and a way to declare `local` for unit tests, as our tooling is different between local and deployed environments unfortunately.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4432


mojotalantikite:
Ugh, when I merged main it looks like my local auto-linting changed some unrelated files here?

services/app-web/src/pages/Settings/EditStateAssign/EditStateAssign.test.tsx
```diff
@@ -146,22 +149,25 @@ describe('EditStateAssign', () => {
 
     it('cancel button moves admin user back to state assignments table', async () => {
         let testLocation: Location // set up location to track URL change
-        const { user } = renderWithProviders(wrapInRoutes(<EditStateAssign />), {
```
macrael, line 149: are there any changes in these app-api/app-web files that aren't whitespace?

services/uploads/src/deps/clamAV/clamAV.ts
```diff
@@ -42,19 +44,53 @@ function NewClamAV(config: Partial<ClamAVConfig>, s3Client: S3UploadsClient) {
...
+            scanForInfectedFiles(fullConfig, avScan, path),
         fetchAVDefinitionsWithFreshclam: (workdir: string) =>
             fetchAVDefinitionsWithFreshclam(fullConfig, workdir),
     }
 }
+interface AVScan {
+    scan(pathToScan: string): SpawnSyncReturns<Buffer>
+}
+
+class ClamscanLocal implements AVScan {
```
macrael, line 69: nice dependency injection!

macrael (approved):
looks good

services/app-web/src/pages/Settings/EditStateAssign/EditStateAssign.test.tsx
haworku: @JasonLin0991 saw something like this on my PR too. I think precommit isn't working anymore  -- both linting and prettier don't seem to be totally synced. I keep fixing others lint errors. And these are my prettier errors (makes sense I dont have prettier in my vscode right now I recently had to reset all my extensions and forgot to add it). But thats the point of precomit to smooth this out before merge.

PR #2793, opened 2024-09-30
MCR-4325: Add state assignments cypress test

JasonLin0991:
## Summary
[MCR-4325](https://jiraent.cms.gov/browse/MCR-4325)

- Add cypress test for state assignments.
- Updated user commands to be used with all the CMS user accounts.

#### Test cases covered

- `settings.spect.ts`
   - `'can update state assignment and see in on the state assignment table'`

### DIFF
--- a/services/cypress/integration/adminWorkflow/settings.spec.ts
+++ b/services/cypress/integration/adminWorkflow/settings.spec.ts
@@ -4,58 +4,58 @@ describe('Admin user can view application level settings', () => {
         cy.interceptGraphQL()
     })
 
-    it('and update user division assignments', () => {
-        // make sure a cms user in db first
-        cy.logInAsCMSUser()
-        cy.logOut()
-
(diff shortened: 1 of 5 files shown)


services/cypress/integration/adminWorkflow/settings.spec.ts
```diff
@@ -1,63 +1,148 @@
...
 describe('Admin user can view application level settings', () => {
     beforeEach(() => {
         cy.stubFeatureFlags()
         cy.interceptGraphQL()
     })
 
-    it('and update user division assignments', () => {
-        // make sure a cms user in db first
-        cy.logInAsCMSUser()
+    // it('and update user division assignments', () => {
```
haworku, line 7: I think we can delete this commented out test, seems like the new test addresses both

haworku (approved):
Nice good to test remove as well

PR #2801, opened 2024-10-02
MCR-4448: create api for new rate question and fetch rate question

JasonLin0991:
## Summary
[MCR-4448](https://jiraent.cms.gov/browse/MCR-4448)
[Notion](https://www.notion.so/mc-review/Rates-Q-A-API-Work-dd2dc97057dd44d39ca0b7481bf75d81#25f75d21b51f4cd5915a94e7860f22ae) discovery docs
- `insertRateQuestion` Postgres function
- `findAllQuestionsByRate` Postgres function
- `createRateQuestion` mutation and resolver
   - Only CMS users are authorized to call this resolver
   - CMS users without division assignments return an error
   - Only submitted or unlocked rates can have rate questions.

#### Test cases covered

- `createRateQuestion.tes.ts`
   - `'creates rate question'`
   - `'allows question creation on UNLOCKED and RESUBMITTED rate'`
   - `'returns an error if the rate is in DRAFT'`
   - `'returns an error of a state user attempts to create a rate question'`
   - `'returns an error on invalid rate id'`
   - `'returns an error when CMS user division is unassigned'`

## QA guidance

- You can test the mutation in the app using the GraphQL explorer tool.


services/app-api/src/postgres/postgresStore.ts
```diff
@@ -12,6 +12,8 @@ import type {
     ContractType,
     UnlockedContractType,
     CMSUsersUnionType,
+    RateQuestion,
```
macrael, line 15: I think this type should have the suffix "Type". "RateQuestion" will be the graphql type

services/app-api/src/postgres/questionResponse/insertRateQuestion.ts
```diff
@@ -0,0 +1,51 @@
...
+    const documents = questionInput.documents.map((document) => ({
+        id: uuidv4(),
+        name: document.name,
+        s3URL: document.s3URL,
+    }))
+
+    try {
+        const result = await client.rateQuestion.create({
+            data: {
+                id: uuidv4(),
```
macrael, line 28: this should be auto?

services/app-api/src/resolvers/questionResponse/createRateQuestion.test.ts
```diff
@@ -0,0 +1,213 @@
...
+
+        const rateID = submittedContractAndRate.draftRates?.[0].id
+
+        if (!rateID) {
+            throw new Error(
+                'Unexpected error: Rate not found in test draft contract and rate'
+            )
+        }
+
+        const rateQuestionRes = must(
```
macrael, line 144: There should be no need for must() here, our convention is for createTest* functions to throw on any error

macrael: This is why I've generally invoked the gql.execute stuff in the tests for the given API, and used the createTest* stuff in other tests that need them. I'd be open to changing this convention and adopting must instead but I think that would need to be a broader change

macrael (approved):

PR #2804, opened 2024-10-03
MCR-4450: Update rate resolver to return questions field

JasonLin0991:
## Summary
[MCR-4450](https://jiraent.cms.gov/browse/MCR-4450)

- `rateResolver` returns questions similar to `contractResolver`.
- Refactored the questionHelpers to work with both types of questions
- new `fetchRateWithQuestions` that includes the `questions` field.

NOTE: As a follow on ticket I wanted to rename all the existing `question` for contract questions to `contractQuestion` to better clarify which question type belongs to a rate and a contract. Separating it from this work will make it easier to review.

#### Test cases covered

`fetchRate.test.ts`
- `'returns the questions on for a rate'`

### DIFF
--- a/services/app-api/src/resolvers/rate/fetchRate.test.ts
+++ b/services/app-api/src/resolvers/rate/fetchRate.test.ts
@@ -1,12 +1,18 @@
 import FETCH_RATE from '../../../../app-graphql/src/queries/fetchRate.graphql'
+import FETCH_RATE_WITH_QUESTIONS from '../../../../app-graphql/src/queries/fetchRateWithQuestions.graphql'
 import { testLDService } from '../../testHelpers/launchDarklyHelpers'
 import {
     constructTestPostgresServer,
+    createTestRateQuestion,
     defaultFloridaRateProgram,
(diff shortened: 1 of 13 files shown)


services/app-api/src/resolvers/rate/fetchRate.test.ts
```diff
@@ -455,4 +462,102 @@ describe('fetchRate', () => {
...
+            query: FETCH_RATE_WITH_QUESTIONS,
+            variables: {
+                input: {
+                    rateID,
+                },
+            },
+        })
+        const rateQuestions2 = result2.data?.fetchRate.rate.questions
+
+        // Expect 2 DMCO questions and latest created question at index 1 by dmco2CmsUser
```
JasonLin0991, line 557:
```suggestion
        // Expect 2 DMCO questions and the latest created question at index 0 by dmco2CmsUser
```

haworku (approved):
This looks good to me.  I agree renaming the other existing resource to be contract questions and response will add clarity.

PR #2805, opened 2024-10-03
Fixes for file upload when virus is present

mojotalantikite:
## Summary

This fixes up the test for av scanning to trigger the case I had seen when a good file and a bad file are added to the file upload dialogue causing the application to reload and clearing out even the good file. This cypress test successfully triggered that behavior, and I elaborated on the test by then removing the bad file and re-uploading a known good file.

Some cleanup around the s3 funcs fixed up this behavior.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4500

### DIFF
--- a/services/app-web/src/contexts/AuthContext.tsx
+++ b/services/app-web/src/contexts/AuthContext.tsx
@@ -170,7 +170,7 @@ function AuthProvider({
     const checkAuth: AuthContextType['checkAuth'] = async () => {
         try {
             const refetchRes = await refetch()
-            console.info(`refetch result: ${refetchRes}`)
+            console.info(`refetch result: ${JSON.stringify(refetchRes)}`)
             return refetchRes
         } catch (e) {
             // if we fail auth at a time we expected logged in user, the session may have timed out. Logout fully to reflect that and force React state update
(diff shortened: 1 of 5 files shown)


services/app-web/src/contexts/AuthContext.tsx
```diff
@@ -169,7 +169,9 @@ function AuthProvider({
     */
     const checkAuth: AuthContextType['checkAuth'] = async () => {
         try {
-            return await refetch()
+            const refetchRes = await refetch()
+            console.info(`refetch result: ${JSON.stringify(refetchRes)}`)
```
haworku, line 173: nit: cleanup

services/cypress/integration/stateWorkflow/virusScan/virusScan.spec.ts
```diff
@@ -12,18 +12,50 @@ describe.only('documents', () => {
...
+                    })
+
+                // Verify the file has been removed
+                cy.findByText(/1 complete, 0 errors, 0 pending/).should('exist')
+                cy.findByText('Failed security scan, please remove').should(
+                    'not.exist'
                 )
+
+                // Upload a clean file
+                fileInput.attachFile(['documents/how-to-open-source.pdf'])
```
haworku, line 55: yes this is good - would have caught this issue.

haworku (approved):
Thanks for taking care of this

PR #2806, opened 2024-10-03
Analytics bugfix

haworku:
## Summary
- Get production tealium upgraded to new hostname now that they are ready for us
- Address random issues in [this ticket](https://jira.cms.gov/browse/BLSTANALYT-7663) as much as possible. 
- Turn console logging in dev so we don't flood cypress console. 
- moved types to their own file
- stopped bothering with section subsection and component type since I found out its required asked that this be excluded from our build

#### Related issues
https://jira.cms.gov/browse/BLSTANALYT-7663
https://jiraent.cms.gov/browse/MCR-4301

#### Screenshots
n/a

#### Test cases covered
n/a

## QA guidance
Things I added logging around that should show up locally
- accordion open and close logs something (its accordion_opened always)
-  form submit logs  (just did main form for now, ran out of steam)
- filters pass all the the various filters set with their values
- a bunch of other events have `link_type` set and pass user role and unique identifier (no user data like emails should be in logs)

### DIFF
--- a/services/app-web/src/tealium/tealium.ts
+++ b/services/app-web/src/tealium/tealium.ts
@@ -7,7 +7,7 @@ import {getRouteName} from '../routeHelpers';
 import {createScript} from '../hooks/useScript';
 import { getTealiumPageName } from './tealiumHelpers';
 import {recordJSException} from '../otelHelpers';
-import { TEALIUM_CONTENT_TYPE_BY_ROUTE, TEALIUM_SUBSECTION_BY_ROUTE } from './constants';
+import { TEALIUM_CONTENT_TYPE_BY_ROUTE } from './constants';
 
 // TYPES
 type TealiumEvent =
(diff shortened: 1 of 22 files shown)


haworku:
@JasonLin0991  right now analytics is off on prod it would be good to land this tomorrow  (there are hostname changes) so at least we get data flowing. But please leave comments of what you find I can address in subsequent PR even if we merge this to get basics through

services/app-web/src/tealium/tealium.ts
```diff
@@ -7,203 +7,12 @@ import {getRouteName} from '../routeHelpers';
...
-        loggedInUser?: User,
-        heading?: string | React.ReactElement
-    ) => void
-}
+import { TEALIUM_CONTENT_TYPE_BY_ROUTE } from './constants';
+import {removeNullAndUndefined} from'../common-code/data/dataUtilities'
+import { TealiumClientType, TealiumEnv, TealiumEventObjectTypes, TealiumInteractionEventDataObject, TealiumViewDataObject } from './types';
 
 const tealiumClient = (tealiumEnv: Omit<TealiumEnv, 'dev'>): TealiumClientType => {
-    const isProd = tealiumEnv == 'prod'
```
haworku, line 205: Removing this `isProd` logic is what will turn on analytics again on production. They have been incrementally migrating environments and are ready for our prod code to be turned on.

JasonLin0991 (approved):
I'm going to approve and merge this, but there still is the issue of logging to tealium when toggling filter accordion. I moved that work to this [ticket](https://jiraent.cms.gov/browse/MCR-4523).

PR #2807, opened 2024-10-03
MCR-4450: Follow up question name change to contractQuestion

JasonLin0991:
## Summary
This is a follow-up to [MCR-4450 PR](https://github.com/Enterprise-CMCS/managed-care-review/pull/2804).

The work in this PR renames all `question` to `contractQuestion` which includes table names, mutation names, resolver names, postgres functions, and types. This better clarifies the two different types of questions for a rate and a contract.

Work done
- Alter `Question` tables
- Update Domain types
- Update `GraphQL` schema
- Update existing Questions API tests to no longer use HPP helpers.
- Remove `indexQuestions` GraphQL query and resolver.
   - This query is never used in the app and is not needed since the `contractResolver` already includes the same functionality to return all questions by contract.
- Remove `fetchContractWithQuestions` query type from `schema.graphql`. This is not needed `fetchContract` already includes `questions`. All that is needed is the `query` in `fetchContractWithQuestions.graphql` that includes the questions field.

## QA guidance

- [Submission](https://d2hp5q41ufbnp4.cloudfront.net/submissions/e4a085f7-be12-489b-885d-df1cce5c5ef6) with existing questions before migration.
   - Test if migration worked and can still view the Q&A page
      - Each division should have at least 1 question and response. Except for DMCO, that was the one I used to verify my work on the review app.
   - Test to see if question files still download
   - Test to see if new questions can still be added
   - Test to see if new responses can still be added


services/app-api/src/emailer/emailer.ts
```diff
@@ -124,27 +124,27 @@ type Emailer = {
...
         statePrograms: ProgramType[],
-        question: Question
+        question: ContractQuestionType
     ) => Promise<void | Error>
     sendQuestionsCMSEmail: (
         contract: ContractRevisionType,
         stateAnalystsEmails: StateAnalystsEmails,
         statePrograms: ProgramType[],
-        questions: Question[]
+        questions: ContractQuestionType[]
```
haworku, line 133: FYI @pearl-truss  since you mentioned working on emailer - types changes incoming

services/app-api/src/resolvers/questionResponse/createContractQuestion.test.ts
```diff
@@ -101,10 +81,17 @@ describe('createQuestion', () => {
             ],
         })
 
-        const indexQuestionsPayload = await indexTestQuestions(
-            cmsServer,
-            submittedPkg.id
-        )
+        const contractWithQuestions = await stateServer.executeOperation({
```
haworku, line 84: @JasonLin0991 Can we get a test helper for this 🙏 like the other api requests . Let me know if you want me to make this change since I know you moved on to other work, I have bandwidth.

haworku (approved):

PR #2816, opened 2024-10-07
MCR-4147: submitContract updated for contract API

pearl-truss:
### Summary

This PR
- Updates the submitContract resolver to not rely on and call the submitHealthPlanPackage resolver (removing the submitHPP resolver and test helper functions will happen in a [subsequent ticket](https://jiraent.cms.gov/browse/MCR-4552))
- Updates the CMS and State emails for submit and resubmit to use contract API
- Creates a zod parser for the contract

#### Related issues

https://jiraent.cms.gov/browse/MCR-4529

#### Test cases covered

- submitContract resolver: ported over and updated existing test
- zod parser function: happy path and checked expected errors for invalid data

## QA guidance

- Manual Q&A - check submit flows for submissions with and without linked rates. I also checked that unlock and resubmit were working
- Include unhappy path in QAing for this work. Can use the graphql explorer update endpoint to get a submission in a bad state (check the parseContract and submitContract for logic that will trigger errors)
- On Hana's suggestion I tested this with val, someone else might want to try it with prod data to confirm before merging

Specific Review of the Code
- Double check the logic of the zod parser (parseContract) that lives in `dataValidatorHelpers.ts`
- Double check that the emails still look good (4 emails for submit and resubmit for CMS and state users), snapshots should all have been updated


haworku:
@pearl-truss For team knowledge, has this work already been run against prod-like data? I see a risk if we merge only having manual tested against basic review app data --- those submissions  usually not as complex for the types of unlock/resubmit we see in production. Especially because we have validations at play.

If this testing step hasn't been taken yet, we should do that before merging I think

services/uploads/src/deps/clamAV/testData/badDummy.pdf
pearl-truss, line 1: I'm not sure why this and `badExcel.xls` are appearing in this PR

services/app-api/src/domain-models/contractAndRates/dataValidatorHelpers.test.ts
```diff
@@ -176,3 +181,380 @@ describe('validateContractDraftRevisionInput', () => {
...
+        if (!(parsedContract instanceof Error)) {
+            throw new Error(
+                'Unexpected error: Was expecting validateContractDraftRevisionInput to return and error'
+            )
+        }
+
+        expect(parsedContract.message).toContain(
+            `populationCoveredSchema of CHIP cannot be submissionType of CONTRACT_AND_RATES`
+        )
+        // expect(parsedContract.message).toContain(
```
pearl-truss, line 544: I commented out this assertion because zod is returning the message as 'Required' instead of the custom message created in `validateAttestation`. This will need some looking into

services/uploads/src/deps/clamAV/testData/badDummy.pdf
haworku: @mojotalantikite Are these testData files something we want committed or are they something we gitignore? Not exactly sure whats going on here but they keep getting added back in folks work off main. Any idea?

pearl-truss:
> @pearl-truss For team knowledge, has this work already been run against prod-like data? I see a risk if we merge only having manual tested against basic review app data --- those submissions usually not as complex for the types of unlock/resubmit we see in production. Especially because we have validations at play.
> 
> If this testing step hasn't been taken yet, we should do that before merging I think

@haworku thanks for this idea. I checked this with val data as a FL user and was able to resubmit the unlocked submissions I saw (I tested about 5-8) that were a mix of contract only and with rates

services/app-api/src/common-code/contractAndRates/ContractProvisions.ts
```diff
@@ -0,0 +1,206 @@
...
+    There are currently three distrinct variants of the provisions: 
+    1. For CHIP amendment
+    2. For non CHIP base contract
+    3. For non CHIP contract amendment
+
+    See also ModifiedProvisions.ts
+*/
+
+// Returns the list of provision keys that apply for given submission variant
+const generateApplicableProvisionsList = (
```
JasonLin0991, line 40: nit: These functions look good. I wonder if it would be helpful to add unit tests for these functions. I will be hard to manually test.

services/app-api/src/emailer/templateHelpers.ts
```diff
@@ -181,6 +182,68 @@ const generateCMSReviewerEmailsForUnlockedContract = (
...
+/* 
+    Determine reviewers for a given health plan package and state
+    - devReviewTeamEmails added to all emails by default
+    - dmcpSubmissionEmails added in both CONTRACT_ONLY and CONTRACT_AND_RATES
+    - oactEmails added for CONTRACT_AND_RATES
+    - dmco is added to emails via state analysts
+    
+    Return should be wrapped in pruneDuplicate to ensure even if config is added twice, we get unique list of reviewers
+*/
+const generateCMSReviewerEmailsForSubmittedContract = (
```
JasonLin0991, line 194: nit: Lot's of duplicate code with `generateCMSReviewerEmailsForUnlockedContract`. I think the only difference is getting data from the `draftRevision` vs `packageSubmissions`, we could probably do something to consolidate the functions together.

JasonLin0991 (commented):
I'll re-review after we get the 438 validation fixed.

services/app-api/src/domain-models/contractAndRates/dataValidatorHelpers.ts
```diff
@@ -93,4 +113,82 @@ const validateContractDraftRevisionInput = (
...
-export { validateContractDraftRevisionInput }
+const parseContract = (
+    contract: ContractType,
+    stateCode: string,
+    store: Store,
+    featureFlags?: FeatureFlagSettings
+): ContractType | Error => {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
+    const formData = contract.draftRevision?.formData!
+    const parsedData = contractSchema
```
JasonLin0991, line 124: I think we need to modify this function with a new schema to validate that the contract data is all there before submitting.

`parseContract` is being used to validate contract data before submitting.`contractFormDataSchema` is the schema used for parsing contract `formData` and almost all of the fields are optional is this schema. So invalid data will be able to be submitted. Testing this with the `parseContract` - `'success if valid form data'` test, you can set fields like `contractDateStart`, `contractDateEnd`, or `contractExecutionStatus` as `undefined` and it will pass the test when it should not for submitting a contract.

The function itself is not an issue, but using this function as is in submitContract would let contracts with invalid data be submitted.

services/app-api/src/common-code/contractAndRates/ContractProvisions.ts
haworku: Most of these are already tested in `provisions.test.tsx` already in app-web. This is copypasta because we don't yet have common code fixed.

haworku: Ah, I see why this probably wasn't tested. We lack a lot of the mock functions on the backend that the frontend has for contract and rates. I started make some but pausing because it seems like a pretty big task. Let me see if I can knock out smaller scope issues first.

services/app-api/src/common-code/contractAndRates/ContractProvisions.ts
haworku, line 1: Looks like we still need to delete out these common-code files from this PR. RIP

services/app-api/src/resolvers/contract/submitContract.ts
```diff
@@ -1,45 +1,419 @@
...
+            )
+            setErrorAttributesOnActiveSpan(
+                'user not authorized to fetch state data',
+                span
+            )
+            throw new ForbiddenError('user not authorized to fetch state data')
+        }
+        const stateFromCurrentUser: State['code'] = user.stateCode
+
+        // fetch contract and related reates - convert to HealthPlanPackage and proto-ize to match the pattern for flag off\
```
haworku, line 90: nit: update comments here

services/app-api/src/domain-models/contractAndRates/dataValidatorHelpers.ts
```diff
@@ -90,4 +92,82 @@ const validateContractDraftRevisionInput = (
...
 }
 
-export { validateContractDraftRevisionInput }
+const parseContract = (
+    contract: ContractType,
+    stateCode: string,
+    store: Store,
+    featureFlags?: FeatureFlagSettings
+): ContractType | Error => {
+    const contractParser = featureFlags?.['438-attestation']
```
haworku, line 101: Not blocking: I would like to see us add a pattern here instead of code right in the parser.  We know will have many more feature flag conditional fields on the backend. And that code will be temporary. 

Maybe a function that takes `submittableContractSchema` and the full list of feature flags as params and returns the enhanced schema

haworku (approved):
LGTM! I poked around more on the review app, looking at prod data and tried to replicate some of these incomplete submissions we have. Excited to land this work and rip out this code!!

The only issue I managed to find is not blocking -- I was able to submit with an incomplete certifying actuary contact - some fields filled in but entirely missing the name. The frontend validation catches it and displays warning.  

Thanks for cleaning up the feature flags code 🙏

pearl-truss:
> LGTM! I poked around more on the review app, looking at prod data and tried to replicate some of these incomplete submissions we have. Excited to land this work and rip out this code!!
> 
> The only issue I managed to find is not blocking -- I was able to submit with an incomplete certifying actuary contact - some fields filled in but entirely missing the name. The frontend validation catches it and displays warning.
> 
> Thanks for cleaning up the feature flags code 🙏

@haworku I can add that as AC on a follow on ticket

PR #2821, opened 2024-10-08
MCR-4456: update submission date to et time zone

JasonLin0991:
## Summary
[MCR-4456](https://jiraent.cms.gov/browse/MCR-4456)

- Update GraphQL schema
   - Update `Contract` and `Rate` types field `initiallySubmittedAt` to be `DateTime` so we can convert the UTC timestamp to ET.
- Update display dates to be ET timezone.
   - Submitted dates:
      - Summary pages
   - Document added dates:
      - Summary pages
      - Q&A page
      - Linked rates on the Rate details page
   - Email templates
   - Dashboard pages
- Update display dates format to `MM/DD/YYYY`
   - Banners
      - Submission Unlocked
         
      - Submission Updated
         
      - Rate withdrawn
         
   - Change history
      
   - Previous submission version date time text
      
- Fixed Linked rate document date added not showing up

#### Test cases covered

- Updated a whole slew of tests that verifies UTC is converted to ET.

## QA guidance

- The Unit tests should validate that the conversions work since `vitest` and `jest` uses UTC time when you call `new Date()`. You wanted to test this in the app, make a submission at 10:00 PM ET, in UTC this would 1 day ahead of whatever date it was at 10 PM ET.


services/app-web/src/common-code/dateHelpers/calendarDate.ts
```diff
@@ -1,12 +1,21 @@
 import { dayjs } from './dayjs'
 
-// We store calendar dates (dates with no time info) in UTC for consistency.
-// This formats a date correctly like '12/02/2022' for display.
-function formatCalendarDate(date: Date | undefined): string {
-    if (!date) {
+// We store calendar dates in UTC for consistency. This formats a date like '12/02/2022'.
+// These formatted dates, should only be used for display, never saved to the database.
+function formatCalendarDate(date: Date | undefined | string, timeZone?: 'UTC' | 'America/New_York'): string {
+    const tz = timeZone || 'UTC'
```
haworku, line 6: I would like to see this be a little more developer foolproof.  I'm worried someone will forget to pass a timezone and then we get the same problem.

Let's make the timeZone prop required. You must defined what you are doing to display the date. And then add a comment that **user actions are always displayed in eastern time**, form data like start and end dates of contracts are displayed in UTC.

haworku, line 6: please just require `timeZone` prop -  make developers think about what they want. 

We can't fall back to UTC anymore, there are valid cases where this should be eastern timezone, in fact it looks like from the function usage, eastern is more likely.

services/app-web/src/common-code/dateHelpers/calendarDate.ts
```diff
@@ -1,12 +1,21 @@
...
+    if (!date || !dayjs(date).isValid()) {
+        return ''
+    }
+    return dayjs(date).tz(tz).format('MM/DD/YYYY')
+}
+
+function formatDateTime(date: Date | undefined | string, timeZone?: 'UTC' | 'America/New_York'): string {
```
haworku, line 13: I think date times displayed are always Eastern in the application going forward. |

That means this can be `formatEasternDateTime` and remove timezone handling all together. Also, I suggest adding the appended `ET` string here as well, that's part of the date time display.

haworku (approved):
Thank you for the testing here I see good test cases included.

Only feedback is on `dateHelpers`  - trying to remove defaults when really a developer needs to think each time "which one of these do I need". Wish we could just have everything be in eastern but I'm not sure what's in the database now for things like `rateDateStart` etc so keeping those in UTC seems safest

PR #2822, opened 2024-10-08
Change handleDeleteFile in app-web

mojotalantikite:
## Summary

We ran into an issue where a state submitted, but the contract document that was uploaded was missing from the package. We decided we don't really have a need for doing an actual delete from `app-web`, so this changes the `handleDeleteFile` func to not actually delete files, but tag the metadata instead when a delete operation occurs.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4491

### DIFF
--- a/services/app-web/src/s3/s3Amplify.ts
+++ b/services/app-web/src/s3/s3Amplify.ts
@@ -105,8 +105,8 @@ function newAmplifyS3Client(bucketConfig: S3BucketConfigType): S3ClientT {
                 console.info('Copying file to update metadata')
                 try {
                     await Storage.copy(
-                        { key: fullKey },
-                        { key: fullKey },
+                        { key: filename },
+                        { key: filename },
                         { metadata: updatedMetadata }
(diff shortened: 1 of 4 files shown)


services/app-web/src/s3/s3Amplify.ts
```diff
@@ -77,12 +77,49 @@ function newAmplifyS3Client(bucketConfig: S3BucketConfigType): S3ClientT {
...
+
+                // Update the file's metadata
+                console.info('Copying file to update metadata')
+                try {
+                    await Storage.copy(
+                        { key: fullKey },
+                        { key: fullKey },
+                        { metadata: updatedMetadata }
+                    )
+                    console.info('Successfully updated file metadata')
```
haworku, line 111: Interesting idea, I like it because we still retain something to show that a field may have been removed in the UI.

mojotalantikite:
We've decided to not go down this path, we've already merged a fix.

PR #2824, opened 2024-10-09
MCR-4449: Rate create rate question response api

JasonLin0991:
## Summary
[MCR-4449](https://jiraent.cms.gov/browse/MCR-4449)

- insertRateQuestionResponse Postgres function
- createRateQuestionResponse mutation and resolver

#### Test cases covered

`createRateQuestionResponse.test.ts`
- `'returns question response data'`
- `'returns an error when attempting to create a response for a question that does not exist'`
- `'returns an error if a cms user attempts to create a response for a rate question'`


services/app-graphql/src/schema.graphql
```diff
@@ -701,7 +705,7 @@ type CreateRateQuestionPayload {
 #    question: Question!
 #}
 
-input CreateContractQuestionResponseInput {
+input CreateQuestionResponseInput {
```
haworku, line 708: Can you clarifying this is a generic type it works for both rate and question responses

haworku (approved):
This looks good - lets merge through and start to build some UI to test

PR #2832, opened 2024-10-12
Add Salesforce model docs

macrael:
## Summary

Added docs with all the fields I think we need to send to salesforce for the POC. 

Notably not sending any "revision" info, just straight contract and rate b/c Salesforce does auto-history-tracking

Also left out contacts for now, we use those for emailing which Salesforce doesn't need to do for us now so it reduces the complexity. 

#### Related issues

https://jiraent.cms.gov/browse/MCR-4493

### DIFF
--- a/docs/contract-rate-non-revision-design.md
+++ b/docs/contract-rate-non-revision-design.md
@@ -6,7 +6,7 @@ This is an ER diagram of the data that is planned to be sent to Salesforce on su
 erDiagram
 
 Contract {
-    String id
+    String externalID
     String externalURL
     DateTime initiallySubmittedAt
     DateTime updatedAt
@@ -54,7 +54,7 @@ Contract {
 }
 
 Rate {
-    String id
+    String externalID
     String externalURL
     DateTime initiallySubmittedAt
     DateTime updatedAt
--- a/docs/contract-rate-non-revision-design.md
+++ b/docs/contract-rate-non-revision-design.md
@@ -14,16 +14,17 @@ Contract {
     String name
     String mccrsID
     String submissionReason
+    Rate[] relatedRates
 
     String[] programNicknames
-    PopulationCoveredEnum populationCovered "MEDICAID, CHIP, MEDICAID_AND_CHIP" 
-    SubmissionTypeEnum submissionType "CONTRACT_ONLY, CONTRACT_AND_RATES"
+    Enum populationCovered "MEDICAID, CHIP, MEDICAID_AND_CHIP" 
+    Enum submissionType "CONTRACT_ONLY, CONTRACT_AND_RATES"
     Boolean riskBasedContract
     String submissionDescription
     Document[] contractDocuments
     Document[] supportingDocuments
-    ContractTypeEnum contractType "BASE, AMENDMENT"
-    ContractExecutionStatusEnum contractExecutionStatus "EXECUTED, UNEXECUTED"
+    Enum contractType "BASE, AMENDMENT"
+    Enum contractExecutionStatus "EXECUTED, UNEXECUTED"
     CalendarDate contractDateStart
     CalendarDate contractDateEnd
     String[] managedCareEntities
@@ -51,6 +52,7 @@ Contract {
     Boolean statutoryRegulatoryAttestation
     Boolean statutoryRegulatoryAttestationDescription
     
+    Contact[] stateContacts
 }
 
 Rate {
@@ -61,9 +63,10 @@ Rate {
     StateCode stateCode
     String name
     String submissionReason
+    Contract[] relatedContracts
     
-    RateTypeEnum rateType "NEW, AMENDMENT"
-    RateCapitationTypeEnum rateCapitationType "RATE_CELL, RATE_RANGE"
+    Enum rateType "NEW, AMENDMENT"
+    Enum rateCapitationType "RATE_CELL, RATE_RANGE"
 
     String[] programNicknames
     Document[] rateDocuments
@@ -76,6 +79,9 @@ Rate {
     CalendarDate amendmentEffectiveDateEnd
 
     Enum actuaryCommunicationPreference "OACT_TO_ACTUARY, OACT_TO_STATE"
+
+    Contact[] certifyingActuaries
+    Contact[] additionalActuaries
 }
 
 Document {
--- a/docs/contract-rate-non-revision-design.md
+++ b/docs/contract-rate-non-revision-design.md
@@ -1,0 +1,77 @@
+# Contract and Rate Without Revisions Design
+
+This is an ER diagram of the data that is planned to be sent to Salesforce on submission. It relies on Salesforce's native diff tracking so elides the concept of Revisions. 
+
+```mermaid
+erDiagram
+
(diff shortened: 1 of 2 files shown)


docs/contract-rate-non-revision-design.md
```diff
@@ -0,0 +1,77 @@
...
+    Boolean modifiedNetworkAdequacyStandards
+    Boolean modifiedLengthOfContract
+    Boolean modifiedNonRiskPaymentArrangements
+    Boolean statutoryRegulatoryAttestation
+    Boolean statutoryRegulatoryAttestationDescription
+    
+}
+
+Rate {
+    String id
```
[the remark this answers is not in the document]
macrael: I was thinking that would be the "name" field, standardized on both Contract and Rate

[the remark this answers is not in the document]
JasonLin0991: Ah! I skipped that. Thanks for pointed it out.

MegMurphy:
Some thoughts / question: 
- On the rate table, don't we also need to include the submission ids that belong to each rate in the data that we send to SFDC, so that we can associate those objects properly in SFDC?
- I would argue for including the state contacts and the actuary contacts. Contact objects will be important part of the data model, and in my opinion, part of the scope of the POC should be coming up with a realistic data model that would support CMS's review needs. They are also required for more than just system automated emails, but also for any CMS outreach related to the submission that is not covered by system notifications, since we are not totally replacing the need to email. I think we need to plan for the data model complexity we know we need. (I am also very curious as to how we model this out in Salesforce, since I see contacts as being associated with the state at the account level, but also with submissions at the contract and rate object levels). 
- I would also argue for including certifying actuary, actuarial firm, and actuarial communication preference on each rate.  
- Will we be sending the "rate name"? Is that what is meant by "name" on the rate table? 
- I think having URLs to view the submission and the rate review in MC-Review would be useful, I assume we can populate that in SFDC since we are sending the ids. 
- I know we will be getting auto-history-tracking for free when something is resubmitted and the record is updated in Salesforce. We won't be getting the "summary of changes" note from the state on resubmit for free though. Maybe as a stretch goal (not necessarily v1 of the API) we think about sending that and where it might be incorporated in the SFDC views, so the reviewer has that context as the view a submission and its history in SFDC?

docs/contract-rate-non-revision-design.md
```diff
@@ -0,0 +1,77 @@
...
+
+Contract {
+    String id
+    DateTime initiallySubmittedAt
+    DateTime updatedAt
+    StateCode stateCode
+    String name
+    String mccrsID
+
+    String[] programNicknames
```
haworku, line 16: I'm assuming we will need more robust programs data - the full names and the nicknames. We have different UI we display in our app already depending on context, sometimes we refer to programs by acronym sometimes its the full name written out.

docs/contract-rate-non-revision-design.md
```diff
@@ -0,0 +1,77 @@
...
+    SubmissionTypeEnum submissionType "CONTRACT_ONLY, CONTRACT_AND_RATES"
+    Boolean riskBasedContract
+    String submissionDescription
+    Document[] contractDocuments
+    Document[] supportingDocuments
+    ContractTypeEnum contractType "BASE, AMENDMENT"
+    ContractExecutionStatusEnum contractExecutionStatus "EXECUTED, UNEXECUTED"
+    CalendarDate contractDateStart
+    CalendarDate contractDateEnd
+    String[] managedCareEntities
```
haworku, line 27: Are you assuming for these two we will return the user friendly names as strings rather than the enum values? Trying to understand why these are strings whereas some others are enums

docs/contract-rate-non-revision-design.md
```diff
@@ -0,0 +1,77 @@
...
+    String id
+    DateTime initiallySubmittedAt
+    DateTime updatedAt
+    StateCode stateCode
+    String name
+    
+    RateTypeEnum rateType "NEW, AMENDMENT"
+    RateCapitationTypeEnum rateCapitationType "RATE_CELL, RATE_RANGE"
+
+    String[] programNicknames
```
haworku, line 64: Reminder that rate programs have the logic where sometimes we display historic rate program names sometimes not. See `useHistoricPrograms` to see that logic. Seems like the mc-review API will have to do the logic and return the proper rate programs depending on submission...

haworku:
> I think having URLs to view the submission and the rate review in MC-Review would be useful, I assume we can populate that in SFDC since we are sending the ids.

Good point. I still see an opportunity for MC-Review sending those as full urls to salesforce not just ids ... we have types control on our urls on our side and we can easily determine if a mc-review URL is still correctly formatted even as urls change. If we let salesforce construct them similar to what happened with mccrs ids on our side, if something changes it will break the urls and salesforce won't be able to tell.

docs/contract-rate-non-revision-design.md
macrael: We talked about this in a meeting, I'm not actually sure we surface the long names anywhere in our app right now. I was thinking that just the program short names would be fine for now

docs/contract-rate-non-revision-design.md
macrael: I'm going back and forth on how to do enums, these two I know at some point one of them was a proper enum and one wasn't but I guess they both are now GraphQL. I think I'd like to try one and the other to see how Salesforce handles that so we can decide if we care about putting enums into that system

macrael:
* On the rate table, don't we also need to include the submission ids that belong to each rate in the data that we send to SFDC, so that we can associate those objects properly in SFDC?
    * I think that is going to be handled in Salesforce itself in using the API we are calling to associate records, not something included in the actual objects
* I would argue for including the state contacts and the actuary contacts. Contact objects will be important part of the data model, and in my opinion, part of the scope of the POC should be coming up with a realistic data model that would support CMS's review needs. They are also required for more than just system automated emails, but also for any CMS outreach related to the submission that is not covered by system notifications, since we are not totally replacing the need to email. I think we need to plan for the data model complexity we know we need. (I am also very curious as to how we model this out in Salesforce, since I see contacts as being associated with the state at the account level, but also with submissions at the contract and rate object levels).
    * OK I’ll add contacts. I’ll find out about how durable they can be, probably using email address as a key
* I would also argue for including certifying actuary, actuarial firm, and actuarial communication preference on each rate.
    * roger
* Will we be sending the "rate name"? Is that what is meant by "name" on the rate table?
    * Yes, contract name and rate name will just be “name” in Salesforce
* I think having URLs to view the submission and the rate review in MC-Review would be useful, I assume we can populate that in SFDC since we are sending the ids.
    * I like that idea. Might even use those urls as the ID in Salesforce
* I know we will be getting auto-history-tracking for free when something is resubmitted and the record is updated in Salesforce. We won't be getting the "summary of changes" note from the state on resubmit for free though. Maybe as a stretch goal (not necessarily v1 of the API) we think about sending that and where it might be incorporated in the SFDC views, so the reviewer has that context as the view a submission and its history in SFDC?
    * Ill think about that, maybe a “submission reason” field

PR #2844, opened 2024-10-17
Add audit s3 files lambda

mojotalantikite:
## Summary

This adds a lambda that compares all the documents stored in our database against what exists in s3.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4543

### DIFF
--- a/services/app-api/src/handlers/audit_s3.ts
+++ b/services/app-api/src/handlers/audit_s3.ts
@@ -1,0 +1,76 @@
+import type { APIGatewayProxyResultV2, Handler } from 'aws-lambda'
+import { initTracer, recordException } from '../../../uploads/src/lib/otel'
+import { configurePostgres } from './configuration'
+import { NewPostgresStore } from '../postgres'
+
+const main: Handler = async (): Promise<APIGatewayProxyResultV2> => {
+    // setup otel tracing
+    const otelCollectorURL = process.env.API_APP_OTEL_COLLECTOR_URL
+    if (!otelCollectorURL || otelCollectorURL === '') {
+        const errMsg =
+            'Configuration Error: API_APP_OTEL_COLLECTOR_URL must be set'
+        return fmtAuditError(errMsg)
+    }
+    const serviceName = 'audit-s3'
+    initTracer(serviceName, otelCollectorURL)
+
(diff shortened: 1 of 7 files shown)


services/app-api/src/handlers/audit_s3.ts
```diff
@@ -0,0 +1,168 @@
...
+        throw error
+    }
+}
+
+function parseS3URL(s3URL: string): { bucket: string; key: string } {
+    const url = new URL(s3URL)
+    const fullPath = url.pathname.slice(1) // Remove leading '/'
+    const uuidPart = fullPath.split('/')[0] // Extract the UUID part
+    const key = `allusers/${uuidPart}` // Construct the key
+
```
mojotalantikite, line 127: Question for review: when I get back the document entries from the DB in my review branch, I'm seeing the `s3URL` returned that has a format like `s3://bucket-name/53003860-9999-7777-8888-bc113cc0b59b.pdf/response_to_questions_for_submission.pdf`, for example. Notice that the regular key UUID is there, but then also the original filename after it. Is this really how we're storing s3 URLs in the DB? I sort of figured we'd be storing a valid s3URL here.

mojotalantikite:
Right now I'm just logging out the results of missing files, but do we want to generate this into another sort of file format? I guess I just want to know what output would feel the most actionable to everyone.

services/app-api/src/handlers/audit_s3.ts
macrael: see s3URLEncoding.ts in common-code for tools for this. We store the s3 url as well as the filename together.

services/app-api/src/postgres/documents/findAllDocuments.ts
```diff
@@ -0,0 +1,230 @@
...
+            questionID: doc.questionID,
+        }))
+    } catch (err) {
+        return err instanceof Error
+            ? err
+            : new Error('Failed to fetch rate question documents')
+    }
+}
+
+async function getRateQuestionResponseDocument(
```
macrael, line 216: Really feeling our choice to split out docs into different tables today :/

macrael (approved):
Excited to see the results!

PR #2837, opened 2024-10-15
Upload rate question

haworku:
## Summary
CMS user can upload rate question. 

Other work:
- made a shared upload question form component. I plan to do the same approach for upload response, share the form, make the wrapper and the specific api call and loading of breadcrumbs distinct for rate versus contract.

#### Related issues
https://jiraent.cms.gov/browse/MCR-2552

#### Screenshots
Screenshot 2024-10-16 at 5 31 30 PM

#### Test cases covered
Copied same tests from upload questions for contract question - added new test helpers

## QA guidance
When you directly navigate to the URL with the flag on, the page loads. Other navigation is tied up in sidebar ticket.

PR #2854, opened 2024-10-21
Unit test flake - try await for elements early

haworku:
## Summary
Maybe address flake from [promote](https://github.com/Enterprise-CMCS/managed-care-review/commit/a6c567061826e2a818bb9a088d60f38c234d57e1/checks) for #2837


JasonLin0991 (approved):
This is the same thing I found. I think it should work. Can you include the fix in this [PR](https://github.com/Enterprise-CMCS/managed-care-review/pull/2853) so we can just promote once.

PR #2863, opened 2024-10-23
Fix rate Q&A links not showing depreciated rate names.

JasonLin0991:
## Summary

If you go on VAL or DEV, earlier submissions with `deprecatedRateProgramIDs` and no `rateProgramIDs` would show empty rate Q&A links.

This fix adds code to switch to `deprecatedRateProgramIDs` if:
- `deprecatedRateProgramIDs` has programs
- `rateProgramIDs` has no programs

If both conditions are not met, the link will fallback to the text `Rate questions: Unknown Program(s)`

#### Test cases covered

`SubmissionSideNav.test.tsx`
- `'loads sidebar nav with expected links for state user'`
   - I updated this test to Include rates for both no programs and only `deprecatedRateProgramIDs`

## QA guidance

- You wont be able to test this on the review app, since we do not have submissions from a time where `deprecatedRateProgramIDs` would exist.

### DIFF
--- a/services/app-web/src/pages/SubmissionSideNav/SubmissionSideNav.test.tsx
+++ b/services/app-web/src/pages/SubmissionSideNav/SubmissionSideNav.test.tsx
@@ -115,7 +115,31 @@ describe('SubmissionSideNav', () => {
                 rateProgramIDs: ['ea16a6c0-5fc6-4df8-adac-c627e76660ab'],
             },
         }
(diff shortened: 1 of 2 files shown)


haworku (approved):
Thank you for finding. I am going to put this on RSP agenda this `deprecatedRateProgramIDs` feels like a problem that should be isolated on backend and not coming through the API if we can help it

PR #2882, opened 2024-10-29
MCR-2550: state new rate question email

JasonLin0991:
## Summary
[MCR-2550](https://jiraent.cms.gov/browse/MCR-2550)
- If CMS submits questions for a rate review 
Then ALL state contacts on ALL  submissions that are linked to or parent of that rate cert should receive a notification email that CMS has questions about the rate. 
- If CMS submits questions for a rate review 
AND the state selected a communication preference for that rate review of "OACT can communicate directly with the state's actuaries but should copy the state on all written communication and all appointments for verbal discussions" 
THEN ALL certifying actuary contacts should also receive the notification email that CMS has questions about the rate. 
- If a state selected a communication preference of the following for the rate review "OACT can communicate directly with the state, and the state will relay all written communication to their actuaries and set up time for any potential verbal discussions"
THEN the actuarial contacts do NOT receive an email confirmation when responses are sent.

Other work:
- Replaced test GraphQL requests to import queries and mutations from generated `gqlClient.tsx` rather than the source `.graphql` files.
   - This caused many API tests on local deployment to fail because tests could not find GraphQL fragments we used in the queries and mutations.
   - Because we moved fragments to their own `.graphql` files in `app-graphql` importing the query or mutation in API tests does not include the fragments. 
   - Instead, I opted to import from the generated `gqlClient.tsx` which does include them. This seems more correct to me as we should be using the generated GraphQL client code when making test requests as that is the purpose of it.
 
#### Test cases covered

`sendRateQuestionStateEmail.test.ts`
 - `'to addresses list includes submitter emails'`
 - `'to addresses list includes all state contacts on all contracts submitted with rate'`
 - `'to addresses does not include actuaries when rate communication preference is OACT_TO_STATE'`
 - `'to addresses list does not include duplicate emails'`
 - `'subject line is correct and clearly states submission is complete'`
 - `'includes link to rate Q&A page'`
 - `'includes expected data on the CMS analyst who sent the question'`
 - `'renders overall email for a new rate question as expected'`

`createRateQuestion.test.ts`
- `'send state email to state submitting a question succeeds'`


haworku (approved):
LGTM - I left a comment on the ticket about linked rates, but I think we can ship

services/app-api/src/emailer/emails/sendRateQuestionStateEmail.ts
```diff
@@ -0,0 +1,99 @@
...
+            return contacts.concat(stateContacts)
+        },
+        []
+    )
+
+    const actuaryEmails = shouldIncludeActuaries
+        ? getActuaryContactEmails(rateFormData)
+        : []
+
+    const toAddresses = pruneDuplicateEmails([
```
haworku, line 58: Random though - Seeing this with fresh eyes today, haven't looked at emailer for awhile.

I am seeing copypasta code and thinking how more of it could be abstracted up to the emailer level  rather than handling in each emailer function. Some of this code maybe goes instead into `sendSESEmail`. 

Things like 
- it should never be possible to send to same address twice
- the subject line logic has pre-pended the environment

JasonLin0991: Maybe this could be a RSP topic, but I rather like keeping this kind of logic contained inside the functions themselves.

PR #2886, opened 2024-10-31
Refactor app-web OTEL

mojotalantikite:
## Summary

This refactors OTEL in `app-web`, as we've been having on and off issues around getting traces into New Relic and alerting on them. This refactors how we do things in `app-web`. Previously, we had a wrapper component that initialized tracing outside of the component lifecycle and the TraceProvider didn't provide the context. Here we use `React.createContext` and have hooks for other components to use.

I also fixed up the alert groups in New Relic and tested that we do get a `JSException` when we trigger the error boundary test in my branch. Those should be available in dev/val/prod as this promotes.

Note that there were a few tests that kept failing that I had to clean up with these changes.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4437


services/app-web/src/pages/Errors/ErrorBoundaryRoot.tsx
```diff
@@ -18,7 +18,6 @@ function ErrorBoundaryRoot({
 }): React.ReactElement {
     useEffect(() => {
         const logError = async () => {
-            await recordSpan('React: ErrorBoundary')
             recordJSException(
```
JasonLin0991, line 21: In `App.tsx` this component is outside the `TracingInitializer` context. Should we use `recordJSExceptionWithContext` (assuming the name does what it says) here instead?

JasonLin0991 (approved):
Looks good. Just a question on the ErrorBoundryRoot component.

services/app-web/src/pages/Errors/ErrorBoundaryRoot.tsx
mojotalantikite: Yeah, exactly -- we have the `TracingInitializer` that does the `setGlobalTracingContext`. The `recordJSExceptionWithContext` function is something we've had for a while, to which I just added additional span recordings. I wanted us to not have to rip out all of the existing `record` function calls everywhere in `app-web` for now, which we can clean up in a follow up ticket.

PR #2901, opened 2024-11-05
Tweak Prisma connection pooling

mojotalantikite:
## Summary

The SFDC team was hitting our API and getting some internal service errors back on a few requests, which would succeed on retries. In the logs we found that the errors coming back were from throttling exceptions in Aurora. This tweaks the connection string to fail faster when a connection cannot be made, and limits the connection from a lambda to 1 (since many lambdas will be spun up for dealing with requests and Prisma manages pooling)


haworku (approved):
lets merge

PR #2916, opened 2024-11-08
MCR-4589: Graphql and DB changes for submission approvals

pearl-truss:
## Summary

- Update the contract GQL schema to have a reviewStatus field

- Create new ApprovalInfo field on ContractTable and contract graphql schema

- Create a reusable contractFields fragment for updating graphql queries and mutations to have the new reviewStatus and approvalInfo fields

#### Related issues

[Technical design doc](https://www.notion.so/mc-review/Submission-Approvals-1304f9e180fa8027a43afe08f65e81c4?d=1314f9e180fa803b9992001c1d32a155#1314f9e180fa804ab801ed8b39cf263f)

https://jiraent.cms.gov/browse/MCR-4589


services/app-graphql/src/schema.graphql
```diff
@@ -1739,6 +1770,11 @@ type Contract {
     SUBMITTED and RESUBMITTED packages cannot be modified
     """
     status: HealthPlanPackageStatus!
+    """
+    Where the contract is in the review submission flow
+    Options are UNDER_REVIEW, APPROVED, WITHDRAWN
+    """
+    reviewStatus: ReviewStatus!
```
pearl-truss, line 1777: I don't think we want this field added to `UnlockedContractType`. Thoughts?

JasonLin0991: I could go either way.

An argument for having it in an unlocked contract is that an unlocked contract could have two review statuses `UNDER_REVIEW` or `WITHDRAWN `. There may be a case where we want to know one of those statuses on an unlocked contract.

pearl-truss: Good point, I brought this up when I spoke with Kelsey this morning and it doesn't sound like we'll need those statuses for an unlocked contract. Since it's not really known yet, I would suggest not implementing it for approvals and wait until we have a better idea about withdraws, etc

services/app-api/prisma/schema.prisma
```diff
@@ -166,6 +165,17 @@ model RateRevisionTable {
   contractsWithSharedRateRevision ContractTable[]          @relation(name: "SharedRateRevisions")
 }
 
+model ContractActionTable {
```
pearl-truss, line 168: @macrael @JasonLin0991 how do you feel about this implementation, given that after speaking with Kelsey this morning, it sounds like "unapproval" and the idea of having multiple approval actions is something confined to V1. In V2 when we implement sending the approval letter, approval will be considered final and there won't be multiples

JasonLin0991: > In V2 when we implement sending the approval letter, approval will be considered final and there won't be multiples (with the caveat that final confirmation hasn't been set yet but hopefully will have a final idea on this soon)

This should be okay for v1. Do you expect the schema to change in v2?

pearl-truss: Sorry @JasonLin0991 I didn't see your comment before I posted mine. I don't expect the schema to change in V2.

services/app-web/src/testHelpers/apolloMocks/contractGQLMock.ts
```diff
@@ -417,7 +419,7 @@ const submitContractMockError = ({
 
 const indexContractsMockSuccess = (
     submissions: Contract[] = [
-        { ...mockContractPackageUnlockedWithUnlockedType(), id: 'test-id-123', __typename: 'Contract' },
+        { ...mockContractPackageUnlockedWithUnlockedType(), id: 'test-id-123', __typename: 'Contract', reviewStatus: 'UNDER_REVIEW' },
```
JasonLin0991, line 422: **suggestion**: We could add `reviewStatus: 'UNDER_REVIEW'` to the default data in `mockContractPackageUnlockedWithUnlockedType`

JasonLin0991 (approved):

services/app-api/src/domain-models/contractAndRates/statusType.ts
```diff
@@ -7,9 +7,15 @@ const statusSchema = z.union([
     z.literal('RESUBMITTED'),
 ])
 
+const reviewStatusSchema = z.union([
```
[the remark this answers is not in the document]
pearl-truss: In my mind there was a difference in verbiage but I don't see why it couldn't be consolidated

services/app-web/src/testHelpers/apolloMocks/contractGQLMock.ts
pearl-truss: We can't because `unlockedContractType` doesn't have a `reviewStatus` field. In this example submissions is type `Contract[]` which does have `reviewStatus`

services/app-api/src/domain-models/contractAndRates/statusType.ts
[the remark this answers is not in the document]
pearl-truss: I spoke with @macrael and he had a stronger opinion about keeping them separate because actions are verbs and the status are adjectives and they won't always match up (added a new example to the [miro](https://miro.com/app/board/o9J_lS5oLDk=/?moveToWidget=3458764577353419403&cot=14))

services/app-graphql/src/schema.graphql
pearl-truss: I added reviewStatus back to unlockedContract. I spoke with @macrael and he thought we should have this status on unlocked because we're defaulting all contracts to UNDER_REVIEW.

services/app-web/src/testHelpers/apolloMocks/contractGQLMock.ts
pearl-truss: Since unlockedContractType now has a reviewStatus, I went ahead and consolidated the change the mock

JasonLin0991: Oh, I see what you're saying. That makes sense, though this brings up the issue of our queries return types.

**Issue**: I think the root issue here isn't that the `unlockedContractType` doesn't contain `reviewStatus`, but some of our queries should return `Contract` or `UnlockedContract`. In this `indexContractsMockSuccess` we are extending the unlocked contract with the  `reviewStatus`, which doesn't pass the smell test because `IndexContractsPayload` doesn't return that type, if it did we wouldn't need to add `reviewStatus`. 

**Suggestion:** We should update `IndexContractsPayload` to return either `Contract` or `Unlocked Contract`.

We don't have to address it in this PR, but we should take care of this in one of the other API tickets in this epic. Let me know what you think.

cc: @macrael

pearl-truss: Yeah that makes sense to me. I don't think the ticket belongs in this approvals epic though. Seems more like something for the contract API epic to me

PR #2925, opened 2024-11-13
MCR-4611 Submission Approval Resolver

pearl-truss:
## Summary

New approveSubmission resolver created:
- Should error if: contract isn’t in the correct status (submitted or resubmitted) or is called by a non CMS/non approver user type
- Responsible for: updating the `reviewStatus` field on the contract to `APPROVED` and adds the `approvalInfo` (updatedAt, updatedBy)

New ApproveContract postgres function
- Responsible for: find contract with a given ID and updating the found contract record with the new reviewStatus and approvalInfo

#### Related issues

https://jiraent.cms.gov/browse/MCR-4611

#### Test cases covered

 `approveContractResolver` 
- Happy path
- Errors if called by a non CMS/CMS Approver user
- Errors if submission not in the correct status (Submitted or Resubmitted)
- Errors if submission not in the correct reviewStatus (under review)

`approveContract` postgres func
- Test WIP


JasonLin0991 (approved):
Some suggestion, otherwise looks good.

services/app-api/src/resolvers/contract/approveContract.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        if (
+            !(
+                contractWithHistory.status === 'SUBMITTED' ||
+                contractWithHistory.status === 'RESUBMITTED'
+            ) ||
+            contractWithHistory.reviewStatus !== 'UNDER_REVIEW'
+        ) {
+            const errMessage = `Attempted to approve contract with wrong status`
```
macrael, line 68: I'd add the current status to this error message for our logs

services/app-api/src/resolvers/contract/approveContract.ts
```diff
@@ -0,0 +1,111 @@
...
+        if (approveContractResult instanceof NotFoundError) {
+            throw new GraphQLError(approveContractResult.message, {
+                extensions: {
+                    code: 'NOT_FOUND',
+                    cause: 'DB_ERROR',
+                },
+            })
+        }
+
+        if (approveContractResult instanceof Error) {
```
macrael, line 94: I like putting all the specific error handling inside the top level "instanceof Error" check (NotFoundError inherits from Error) to really ensure that all error handling happens

services/app-api/src/postgres/contractAndRates/approveContract.ts
```diff
@@ -0,0 +1,67 @@
+import { findContractWithHistory } from './findContractWithHistory'
+import { NotFoundError } from '../postgresErrors'
+import type { ContractType } from '../../domain-models/contractAndRates'
+import type { PrismaTransactionType } from '../prismaTypes'
+
+async function approveContract(
+    tx: PrismaTransactionType,
```
macrael, line 7: This is tricky. The way you are calling this from the resolver you are not passing a transaction directly into this function. If you consolidate the db calls into one, there's no need for a transaction so if you change this to `client: PrismaClient` that would work fine, we do that in other single-call db handlers. If you keep the two db calls that need to be in a transaction, I suggest you model it after a more complex handler, you can look at `submitRate.ts` for an example. The model I've used is to have a separate very small function that takes `client: PrismaClient` that then calls a `approveContractInsideTransaction` function that has a tx. You setup the transaction in the calling function to make sure we don't accidentally return an error inside the transaction and commit a transaction that actually has an error. 

LMK if you want to chat about this one I got bit by this bug and breaking it out this way is more defensive.

macrael (approved):
The transaction thing is a bug waiting to happen so I definitely want a change there but otherwise this looks great.

PR #2930, opened 2024-11-14
MCR-4623: Fix cc address to be on only state email for new rate questions

JasonLin0991:
## Summary
[MCR-4623](https://jiraent.cms.gov/browse/MCR-4623)

Fixed the CC of `mmcratesetting` ( `config.dmcpSubmissionEmails` in code ) to be only on new rate question for STATE email only.

Removed the CC from CMS emails for new rate question and responses.

### DIFF
--- a/services/app-api/src/testHelpers/emailerHelpers.ts
+++ b/services/app-api/src/testHelpers/emailerHelpers.ts
@@ -355,6 +355,7 @@ const mockUnlockedContract = (
         status: 'UNLOCKED',
         stateCode: 'MN',
         stateNumber: 4,
+        reviewStatus: 'UNDER_REVIEW',
 
         draftRevision: mockContractRev(),
         draftRates,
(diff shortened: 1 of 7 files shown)


services/app-api/src/testHelpers/emailerHelpers.ts
```diff
@@ -355,6 +355,7 @@ const mockUnlockedContract = (
         status: 'UNLOCKED',
         stateCode: 'MN',
         stateNumber: 4,
+        reviewStatus: 'UNDER_REVIEW',
```
pearl-truss, line 358: I wonder if the CI flakes are emailer specific? Because I ran into this locally this morning and was surprised it wasn't picked up in the submission backend PR

pearl-truss (approved):

services/app-api/src/testHelpers/emailerHelpers.ts
JasonLin0991: Not sure, there is a ticket in the backlog to investigate https://jiraent.cms.gov/browse/MCR-4670

PR #2936, opened 2024-11-15
MCR-4612: Approval Modal Working with API

pearl-truss:
## Summary

Note: should be merged after https://github.com/Enterprise-CMCS/managed-care-review/pull/2925

As a CMS user,
I want to mark a submission as approved
so that I can more easily keep track of which submissions in MC-Review are approved and which are still active.

AC: 

- If I am viewing the approval confirmation modal and I select "approve submission"
then the modal closes and I am still on the submission summary page but the "approve submission" button no longer appears on the submission summary
- If I select "approve submission"
- Then the reviewStatus for the submission in the database is set to "Approved" (the separate status field should still be set to "Submitted")
- If I click "cancel"
Then the modal closes and the "approve submission" button still appears and the status of the submission in the database is still "Submitted"

Other notes
- Did get confirmation from Kelsey that it's okay the optional note box is standard size and there's not character limit
- This PR also implements https://jiraent.cms.gov/browse/MCR-4592 
- Given that the "UnlockSubmitModal" now handles approval too, I changed the name to "ActionModal"

#### Related issues

https://jiraent.cms.gov/browse/MCR-4612

#### Screenshots

Screenshot 2024-11-15 at 11 14 50 AM

Screenshot 2024-11-15 at 11 19 32 AM

#### Test cases covered

Modal displays expected heading 

## QA guidance

Click 'Approve Submission' and check that the reviewStatus changes to 'APPROVED' and the reviewStatusActions field has a new entry. The updatedReason note should also appear if that optional field was used


services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -302,11 +298,16 @@ export const SubmissionSummary = (): React.ReactElement => {
 
                 <ChangeHistory contract={contract} />
 
-                <UnlockSubmitModal
+                <ActionModal
                     modalRef={modalRef}
                     modalType="UNLOCK_CONTRACT"
                     submissionData={contract}
                 />
+                <ActionModal
```
JasonLin0991, line 306: If you look at line 205, I have the approval modal there. The reason why it's up there is because I don't want to render the modal code if the user is not authorized. It's not required to do that, but it would be good to make it unavailable in the dom.

services/app-web/src/gqlHelpers/mutationWrappersForUserFriendlyErrors.ts
```diff
@@ -66,10 +68,18 @@ export const handleApolloErrorsAndAddUserFacingMessages = (
     apolloError: ApolloError,
     mutation: MutationType
 ) => {
-    let message =
-        mutation === 'SUBMIT_HEALTH_PLAN_PACKAGE'
-            ? ERROR_MESSAGES.submit_error_generic
-            : ERROR_MESSAGES.unlock_error_generic
+    let message 
+    switch (mutation) {
```
macrael, line 72: nice, I love a switch

macrael (approved):
Looks great. I'm a little surprised our approval, submission, unlock logic is in the modal itself instead of having those submit functions be on the pages that display the modal but no need to change that, just a surprise to me.

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
JasonLin0991: We can use the `ActionModal` component, but we need to align the designs with what I have for my approval modal on line 205. My suggestion is to not have `ActionModal` handle all these functions.
- Placement of the modal should be toward the top of the page
- The modal submit button should be the default style.
- Modal width is longer than desgins.
- [Figma](https://www.figma.com/design/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?node-id=15083-79446&node-type=frame&t=jXUxzjSnklPJaQIV-0).

My opinion on the `ActionModal` is that it is doing too much. There was a reason for a common modal component to handle submit, unlock, and resubmit because they all shared the same design with different text. The APIs are closely related by updating the same status of the contract, but with approval, we are handling a different status of the contract, and the design is slightly different. With all this functionality put into one modal, it is increasing the complexity of the component and could be prone to bugs.

**Suggestion**: 
- Revert `ActionModal` back to `UnlockSubmitModal` or a better name. 
- Create a new modal component for the approval modal using the common `Modal` component.
- Place the modal above in the conditional rendering so it is not in the dom with unauthorized users.

pearl-truss: Yeah that makes sense to me, currently looking into breaking it out

services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -162,6 +190,31 @@ export const SubmissionSummary = (): React.ReactElement => {
     // Only show for CMS_USER or CMS_APPROVER_USER users
     const showSubmissionApproval = submissionApprovalFlag && hasCMSPermissions
 
+    const approveContractAction = async (actionModalInput?: string) => {
+        setIsSubmitting(true)
```
JasonLin0991, line 194: We should log this submit event to Tealium. You could do this like `UnlockSubmitModal` does with `logFormSubmitEvent`

JasonLin0991 (approved):

PR #2947, opened 2024-11-19
Change all displayed dates to Pacific Time

macrael:
## Summary

We are switching to pacific time everywhere, this means that calendar dates should be formatted to pacific timezone and all full dates we display should be in PT instead of ET. 

#### Related issues

https://jiraent.cms.gov/browse/MCR-4610


services/app-web/src/common-code/dateHelpers/calendarDate.ts
```diff
@@ -17,16 +17,16 @@ function formatCalendarDate(date: Date | undefined | string, timeZone: 'UTC' | '
 
 /**
  * We store calendar dates in UTC for consistency. This formats a date time into 'MM/DD/YYYY h:mma timezone'.
- * "America/New_York" (ET) is usually for displaying app generated timestamps, where timezones are important. This should
- * be only used for display and not saving. We have decided to show these types of dates in "America/New_York" in the app
+ * "America/Los_Angeles" (ET) is usually for displaying app generated timestamps, where timezones are important. This should
```
pearl-truss, line 20: Nit: (ET) should be (PT)

pearl-truss (approved):
LGTM but if this merges after https://github.com/Enterprise-CMCS/managed-care-review/pull/2948, the submission approval banner will need to be updated to use PT as well

PR #2948, opened 2024-11-19
MCR-4590: cms and state sees approval banner

JasonLin0991:


pearl-truss (commented):
This is looking good but I confirmed with Kelsey that it's expected for the optional note (if it's been supplied) to appear in banner: [designs](https://www.figma.com/design/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?node-id=15738-76172&node-type=frame&t=o3kf4qjDTQI2R56u-0)

JasonLin0991:
> This is looking good but I confirmed with Kelsey that it's expected for the optional note (if it's been supplied) to appear in banner: [designs](https://www.figma.com/design/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?node-id=15738-76172&node-type=frame&t=o3kf4qjDTQI2R56u-0)

Here is what the changes look like

When optional note is supplied


When optional note is not supplied (left textbox blank)

pearl-truss (approved):
Thanks for getting that change in, looks good

PR #2955, opened 2024-11-21
MCR-4633: CMS can filter  on submission status

JasonLin0991:
## Summary
[MCR-4633](https://jiraent.cms.gov/browse/MCR-4633)

- CMS sees a `Status` filter
- `Status` filter has four options `SUBMITTED`, `UNLOCKED`, `WITHDRAWN`, `APPROVED`
   - `WITHDRAWN` submissions has not been implemented yet, the filter works but you wont get any data to work with.
   - The AC on the ticket has the words title cased, but it seems like the new design for the status tags are all caps so the filter options are all caps to match.
- Filter is multi-select

#### Test cases covered

`ContractTable.test.tsx`
- `'can filter table by submission status'`

### DIFF
--- a/services/app-web/src/components/ContractTable/ContractTable.test.tsx
+++ b/services/app-web/src/components/ContractTable/ContractTable.test.tsx
@@ -93,6 +93,24 @@ const submissions: ContractInDashboardType[] = [
         submissionType: 'Contract action and rate certification',
         stateName: 'Puerto Rico',
     },
+    {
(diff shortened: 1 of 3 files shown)


services/app-web/src/common-code/formatters/titleCase.ts
```diff
@@ -0,0 +1,6 @@
+export const titleCaseString = (str: string): string => {
```
pearl-truss, line 1: Nice addition!

pearl-truss (approved):
LGTM

PR #2957, opened 2024-11-22
Fix HealthPlanPackageType import

mojotalantikite:
## Summary

`app-api` was not building properly due to a change in the `HealthPlanPackageType`. This fixes that.

I also noticed our snapshots are out of date, so this updates that as well.


pearl-truss (approved):
App api and web builds for me

PR #2958, opened 2024-11-22
MCR-4591: Design fix for badge style

pearl-truss:
## Summary

- Certain status badges should be emphasized (bolded and border added). 
- Which statuses are emphasized depends on user type
- For state users these are draft and unlocked statuses
- For CMS users these are unlocked and submitted statuses

#### Related issues

https://jiraent.cms.gov/browse/MCR-4591

#### Screenshots
Screenshot 2024-11-22 at 1 02 59 PM
Screenshot 2024-11-22 at 1 03 47 PM


JasonLin0991 (approved):
Looks good. Just need to add the new prop into storybook. Also could you add the new styling into the storybook examples?

PR #2959, opened 2024-11-22
Fix spacing in our informational banners

JasonLin0991:
## Summary

Adding more top and bottom spacing for the text in our informational banners: unlocked. This is to address comments in this ticket [MCR-4590](https://jiraent.cms.gov/browse/MCR-4590)

### DIFF
--- a/services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
+++ b/services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
@@ -274,7 +274,7 @@ export const SubmissionSummary = (): React.ReactElement => {
             >
                 {renderStatusAlerts()}
 
-                {documentError && (
+                {!documentError && (
                     <DocumentWarningBanner className={styles.banner} />
                 )}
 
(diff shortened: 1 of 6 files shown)


services/app-web/src/pages/SubmissionSummary/SubmissionSummary.tsx
```diff
@@ -274,7 +274,7 @@ export const SubmissionSummary = (): React.ReactElement => {
             >
                 {renderStatusAlerts()}
 
-                {documentError && (
+                {!documentError && (
```
pearl-truss, line 277: Is this intentional? I think we would want it to show when there's an error

JasonLin0991, line 277:
```suggestion
                {documentError && (
```

JasonLin0991: Oops that was for making the banner show. Good catch

pearl-truss (approved):
Thanks for the change, LGTM

PR #2967, opened 2024-11-26
MCR-4634 default filtering on approvals

pearl-truss:
## Summary

This PR implements default filtering
- navigating to `/dashboard/submissions` with append a filter for only displaying unlocked and submitted submissions
- to show an unfiltered version of the dashboard use `/dashboard/submissions#filters=`

#### Related issues

https://jiraent.cms.gov/browse/MCR-4634

#### Screenshots

Screenshot 2024-12-02 at 3 08 49 PM
Screenshot 2024-12-02 at 3 09 04 PM

#### Test cases covered

- `ContractTable` filters out approved submissions

## QA guidance

For CMS user
- Dashboard is filtered by default 
- Unfiltered version of dashboard is accessible with a hash of #filters=


services/app-web/src/components/ContractTable/ContractTable.tsx
```diff
@@ -431,6 +430,26 @@ export const ContractTable = ({
...
+                    },
+                    {
+                        label: 'Unlocked',
+                        value: 'UNLOCKED',
+                    },
+                ],
+                'status'
+            )
+        }
+    })
```
JasonLin0991, line 451:
```suggestion
    }, [])
```
nit: I think we only want to run this useEffect once on initial render? If so we could give it a empty dependency array so it only tuns once, otherwise anything that causes a re-render, like selecting filters, will cause this useEffect to fire. The conditional would skip `updatingFilters` but we could avoid firing the useEffect entirely.

JasonLin0991 (approved):
LGTM

services/app-web/src/components/ContractTable/ContractTable.tsx
pearl-truss: @JasonLin0991 made this change but added the dependencies, otherwise an empty array kicks up a linting error

JasonLin0991: That's good with me. 

The other way is to add `// eslint-disable-next-line react-hooks/exhaustive-deps`.

PR #2961, opened 2024-11-22
MCR-4600 disable add question, add response and unlock for approvals

pearl-truss:
## Summary

This PR hides the following action buttons after a contract has been approved:

- Add Questions
- Approve Submission
- Unlock Submission
- Upload Response to a question

This PR also updates the resolvers to restrict these actions for approved contracts 

#### Related issues
https://jiraent.cms.gov/browse/MCR-4600

#### Screenshots

Screenshot 2024-11-22 at 4 26 03 PM

#### Test cases covered

Frontend Test 
- `SubmissionSummary`: 'does not render the unlock submission button for an approved contract'
- `ContractQuestionResponse`: For state users -> does not render the Upload response button for an approved contract
For CMS users -> 'does not render the Add questions  button for an approved contract'

Backend Test
- `createQuestionResponse`: 'returns an error when attempting to create response for a contract that has been approved'
- `createContractQuestion`: 'returns an error if contract has been approved'
- `unlockContract`: 'returns status error if contract has been approved'
- existing test covered for `approvedContract`

PR #2978, opened 2024-12-03
MCR-4600 restrict rate actions on FE

pearl-truss:
## Summary

Follow on PR to https://github.com/Enterprise-CMCS/managed-care-review/pull/2961 previous PR hid rate actions for an approved contract. This PR disables the unlock rate button (instead of hiding it completely) 

No changes to the Rate Q&A experience 

#### Related issues
https://jiraent.cms.gov/browse/MCR-4600


services/app-web/src/pages/QuestionResponse/QuestionResponseSummary/RateQuestionResponse.tsx
```diff
@@ -47,18 +51,31 @@ export const RateQuestionResponse = () => {
...
+    const rate = data?.fetchRate.rate
     useEffect(() => {
         updateHeading({ customHeading: rateName })
     }, [rateName, updateHeading])
 
+    const {
+        data: fetchContractData,
+        loading: loadingContract,
+        error: fetchContractError,
+    } = useFetchContractQuery({
```
JasonLin0991, line 63: Suggestion: This works, but I wonder if the change here is to add the `consolidatedStatus` in the response data of `fetchRateQuestions` fragments. It seems like upcoming features we will need to know more about the review status of the related contracts or rates.

Maybe in `contractRevisionSchema` we add in the `consolidatedStatus` so then reach contract revision has the status so we don't need to make another fetch.

JasonLin0991 (approved):
I wonder how we should approach a linked rate on a `APPROVED` submission in regards to rate Q&A. 

Currently, if the viewing a submission that has a approved contract, but the linked rate's parent contract is not, then you would still see the `Upload response` on the rate Q&A section of the approved submission summary page. 

It might be confusing for the state user because I don't believe there is a easy way to know if a rate is linked when viewing a submission.

This isn't blocking this PR, but just putting my thought out there. I'll post this in slack too.

services/app-web/src/pages/QuestionResponse/QuestionResponseSummary/RateQuestionResponse.tsx
pearl-truss: @JasonLin0991  Fair point. That's something I considered. The issue I saw with that is managing the `consolidatedStatus` on each rev. It seemed like a bigger change than I should independently put into a PR. I don't love needing to do 2 API calls here. I was thinking we might want something like  a `fetchRateWithContract` endpoint. That would remove the need to track `consolidatedStatus` on each rev and any cascading schema changes. I can bring this up in an RSP

PR #3004, opened 2024-12-12
Update dev clone command to work with password protected ssh-keys

mojotalantikite:
## Summary

Meghan was having some issues with the db clone commands. However, I've personally never been able to use them, as they assume an ssh-key that does not prompt for a password. This adds a prompt for a password entry if the user has a password protected ssh-key (which they should).

#### Related Issues
https://jiraent.cms.gov/browse/MCR-3904


dev_tool/src/jumpbox.ts
```diff
@@ -191,6 +194,115 @@ async function ensureAllowlistIP(
...
+        try {
+            await ssh.connect({
+                host,
+                username,
+                privateKey,
+            })
+        } catch (err) {
+            if (
+                err instanceof Error &&
+                (err.message.includes('password') ||
```
JasonLin0991, line 279: For my error message the keyword `password` was not included. Mine used `passphrase` and `Encrypted` with capital `E`. Maybe we always prompt for a password so we don't need to trying to catch this?

Screenshot 2024-12-13 at 2 59 59 PM

JasonLin0991 (approved):

dev_tool/src/jumpbox.ts
```diff
@@ -191,6 +194,115 @@ async function ensureAllowlistIP(
...
+                    }
+                    break
+                default:
+                    password += charStr
+                    break
+            }
+        }
+
+        let password = ''
+        output.write(prompt)
```
JasonLin0991, line 236: One more thing, could we hide the input for the password?

dev_tool/src/jumpbox.ts
mojotalantikite: ok thanks, good catch! I'll update this.

dev_tool/src/jumpbox.ts
JasonLin0991: From our 1:1: @mojotalantikite and I thought it would be worthwhile to implement masking the password input in this PR. We also talked about how team members work in public spaces and trying to use the jumpbox in those spaces is bad practice so @mojotalantikite Could you add that caution to the docs?

PR #3022, opened 2024-12-21
MCR-4847: Refactor withdraw rate api to move contract to rate relationship

JasonLin0991:
## Summary
[MCR-4847](https://jiraent.cms.gov/browse/MCR-4847)

Summary:
This work refactors the withdraw API to remove the withdrawn rates from a submitted submission. 

The API will:
- Find all contracts related to the rate that is being withdrawn.
- UNLOCKED and DRAFT contracts will be untouched
- If even one of the related submissions are APPROVED, it will return an error to the client, we do not want to withdraw a rate if it's a child or linked to a approved contract.
- SUBMITTED and RESUBMITTED contracts will be unlocked, rate being withdrawn will be removed, and then resubmitted.
- [🛑**READ THIS PLEASE**] Withdrawn rate is resubmitted, removing itself from an UNLOCKED or DRAFT contracts `draftRates`. This means in the UI any unlocked submission that had this rate, will no longer have it after withdrawing.
   - We should check with design if withdrawn rates should automatically gets removed from an unlocked or draft submission. For unlocked submission, if the rate was linked in a previous submission it will still be in previous submission summary.
      - If we do want to keep withdrawn rates on draft and unlocked submission so that the UI can display them as withdrawn and leaving it up to the state to remove them, then one way to do this is to not resubmit the withdrawn rate in the postgres function.
   - For an unlocked submission, the withdraw will not show up in the change history because we are not submitting this type of submission when withdrawing a rate.
- Finally add review action of `WITHDRAW` and connect to contracts it was withdrawn from on the new join table.

With this work, the withdrawn rate is now separated from the active rates on the submission. New unlocks and resubmits of any contract that was a parent or linked will no longer create a new revision for the withdrawn rate. The withdrawn rate and its former contracts still maintain a relationship on the join table to help with UI or un-withdraw actions.

AC:
- There is a new join table `WithdrawnRatesJoinTable` that stores the relationship between the withdrawn rate and their parent and linked to contracts.
- withdrawRate API updates all contracts, parent and linked, to remove the withdrawn rates from the contract and creates a new relationship between them using the new WithdrawnRatesJoinTable table.
- Both GraphQL return data of the parent and linked contracts no longer contains the withdrawn rate in its packageSubmission and now contains the withdraw rate in withdrawnRates field.
- Unlocking submission does not create new revision for withdrawn rates.

Additional work:
- Fix minor command typo in dev_tool

#### Test cases covered

`resolvers/withdrawnRate.test.ts`
- `'can withdraw a rate without errors'` (UPDATE)
   - Added more assertions on `packageSubmission`
- `'can still unlock and resubmit after a rate has been withdrawn with expected packageSubmissions'`
- `'withdraws rate when linked to other contracts'`
- `'does not update draft contract linked to withdrawn rate'`

### DIFF
--- a/dev_tool/src/local/graphql.ts
+++ b/dev_tool/src/local/graphql.ts
@@ -17,7 +17,7 @@ export const compileGraphQLTypesWatchOnce = once(compileGraphQLTypesWatch)
 async function compileGraphQLTypes(runner: LabeledProcessRunner) {
     await runner.runCommandAndOutput(
         'gql deps',
-        ['pnpm', 'install', '--prefer--offline'],
+        ['pnpm', 'install', '--prefer-offline'],
         ''
     )
 
--- a/packages/mocks/src/apollo/contractPackageDataMock.ts
+++ b/packages/mocks/src/apollo/contractPackageDataMock.ts
@@ -309,6 +309,7 @@ function mockContractPackageDraft(partial?: Partial<Contract>): Contract {
                 },
             },
         ],
+        withdrawnRates: [],
         packageSubmissions: [],
         ...partial,
     }
@@ -336,6 +337,7 @@ function mockContractPackageSubmittedWithQuestions(
         mccrsID: null,
         draftRates: [],
         draftRevision: null,
+        withdrawnRates: [],
         packageSubmissions: [
             {
                 cause: 'CONTRACT_SUBMISSION',
@@ -909,6 +911,7 @@ function mockContractWithLinkedRateDraft(
                 ],
             },
         ],
+        withdrawnRates: [],
         packageSubmissions: [],
         ...partial,
     }
@@ -932,6 +935,7 @@ function mockContractWithLinkedRateSubmitted(
         state: mockMNState(),
         stateNumber: 5,
         mccrsID: null,
+        withdrawnRates: [],
         packageSubmissions: [
             {
                 cause: 'CONTRACT_SUBMISSION',
@@ -1096,6 +1100,7 @@ function mockContractPackageSubmitted(partial?: Partial<Contract>): Contract {
         mccrsID: null,
         draftRevision: null,
         draftRates: [],
+        withdrawnRates: [],
         packageSubmissions: [
             {
                 cause: 'CONTRACT_SUBMISSION',
@@ -1307,6 +1312,7 @@ function mockContractPackageApproved(
         mccrsID: null,
         draftRevision: null,
         draftRates: [],
+        withdrawnRates: [],
         packageSubmissions: [
             {
                 cause: 'CONTRACT_SUBMISSION',
@@ -1697,6 +1703,7 @@ function mockContractPackageSubmittedWithRevisions(
         stateNumber: 5,
         draftRevision: null,
         draftRates: null,
+        withdrawnRates: [],
         packageSubmissions: [
             {
                 __typename: 'ContractPackageSubmission',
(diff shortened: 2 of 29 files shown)


JasonLin0991:
- [🛑**READ THIS PLEASE**] Withdrawn rate is resubmitted, removing itself from an UNLOCKED or DRAFT contracts `draftRates`. This means in the UI any unlocked submission that had this rate, will no longer have it after withdrawing.
   - We should check with design if withdrawn rates should automatically gets removed from an unlocked or draft submission. For unlocked submission, if the rate was linked in a previous submission it will still be in previous submission summary.
      - If we do want to keep withdrawn rates on draft and unlocked submission so that the UI can display them as withdrawn and leaving it up to the state to remove them, then one way to do this is to not resubmit the withdrawn rate in the postgres function.
   - For an unlocked submission, the withdraw will not show up in the change history because we are not submitting this type of submission when withdrawing a rate.
- Finally add review action of `WITHDRAW` and connect to contracts it was withdrawn from on the new join table.

pearl-truss (commented):
[🛑READ THIS PLEASE] Withdrawn rate is resubmitted, removing itself from an UNLOCKED or DRAFT contracts draftRates. This means in the UI any unlocked submission that had this rate, will no longer have it after withdrawing.
We should check with design if withdrawn rates should automatically gets removed from an unlocked or draft submission. For unlocked submission, if the rate was linked in a previous submission it will still be in previous submission summary.
If we do want to keep withdrawn rates on draft and unlocked submission so that the UI can display them as withdrawn and leaving it up to the state to remove them, then one way to do this is to not resubmit the withdrawn rate in the postgres function.
For an unlocked submission, the withdraw will not show up in the change history because we are not submitting this type of submission when withdrawing a rate.

Not sure I understand this part. When withdrawRate api is called on an unlocked submission you just get the expected error (the rate is kicked into Unlocked when the contract is unlocked): Attempted to withdraw rate with wrong status. Rate: UNLOCKED, Parent contract: UNLOCKED

pearl-truss (approved):
I'm seeing this working when pulled int my withdraw rate API UI ticket and in the graphql explorer. Great work! I left some comments on the unlocked issue, just because it wasn't clear to me what the expected behavior was: to be able to withdraw a rate from an unlocked submission (what I took the comment in the PR description to imply) or for the rate withdraw to be restricted and throw an error if submission is unlocked (current behavior and what I see in comments in the code)

haworku:
> If we do want to keep withdrawn rates on draft and unlocked submission so that the UI can display them as withdrawn and leaving it up to the state to remove them...

Ah, this should not be a use case we support. It makes the withdrawn status muddy in its definition and introduces significant code complexity. I wasn't part of initial epic scoping here but I suggest that WITHDRAWN, from a technical standpoint, means this resource removed from all active edit workflows. The rate is made read-only, it is no longer actionable. I assume that means it's removed from any active forms as soon as it is withdrawn. We should not support the ability to keep withdrawn rates on the active submission form and let states remove, that has several trickle effects and makes our review statuses harder to reason about. 

I could see asking design **if** we want to allow withdrawing a rate on an unlocked submission at all. That's an interesting call-out.  Maybe rate can only be withdrawn if all linked submissions are locked and submitted??

haworku:
I'm going to merge this work since the open design question can be resolved later as we ship ongoing work. Also reminder we are testing this work in dev only not on VAL.

PR #3032, opened 2024-12-26
Refactor bulkDownloadUrl lambda

mojotalantikite:
## Summary

A state made a submission with ~100 files in the contract section. As such, the zip file generator was not working. It looks like the zip file lambda is working fine, but the URL is not being generated fast enough for the front end. This uses exponential back-off to wait for the URL a bit longer.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4852

### DIFF
--- a/services/app-api/src/handlers/bulk_download.ts
+++ b/services/app-api/src/handlers/bulk_download.ts
@@ -1,5 +1,9 @@
 import type { PutObjectCommandInput } from '@aws-sdk/client-s3'
-import { S3Client, GetObjectCommand } from '@aws-sdk/client-s3'
+import {
+    S3Client,
+    GetObjectCommand,
+    HeadObjectCommand,
+} from '@aws-sdk/client-s3'
 import { Upload } from '@aws-sdk/lib-storage'
 import { Readable } from 'stream'
 import type { APIGatewayProxyHandler } from 'aws-lambda'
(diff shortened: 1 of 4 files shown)


mojotalantikite:
closing for #3038

services/app-api/src/handlers/bulk_download.ts
```diff
@@ -5,257 +5,288 @@ import {
...
+                    return downloadFile(bulkDlRequest.bucket, key, timeout)
+                })
+            )
+
+            for (const result of batchResults) {
+                totalBytes += result.size
+                if (totalBytes > MAX_TOTAL_SIZE) {
                     throw new Error(
-                        'Programming Error: Unknown return type for s3 Body'
+                        `Total size exceeds maximum allowed size of ${MAX_TOTAL_SIZE} bytes`
```
haworku, line 190: Is this the error we are expecting to see if this issues happens again?  I'd love to leave this work with a clear error thrown instead of things silently exiting if we are going to hit this again

mojotalantikite: I still don't know what the error path of the original failure was actually. Even after I added a bunch of verbosity to what was going on, the lambda would still just shut itself off after zipping a fraction of the files.

This size check is just what the default temp storage size is on a lambda environment. We clean up after ourselves, but basically we don't want to be creating zip archives of > 500MB. We can tweak this value though, I'm up for suggestions.

Oh, I guess this is the closed PR actually 😂

PR #3039, opened 2024-12-31
MCR-4662: state recieves withdrawn rate email notification

JasonLin0991:
## Summary
[MCR-4662](https://jiraent.cms.gov/browse/MCR-4662)
[Figma](https://www.figma.com/design/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?node-id=13787-32931&t=AeBwKPBzntpFMMaz-0)

If a rate is withdrawn then a rate withdrawal email notification is sent to the state contacts on all submissions the rate is attached to.
- If rate is on one submission:
   - Link to submission is listed at the bottom of the email with the text `View the submission in MC-Review`
- If rate is on multiple submissions:
   - Each submission name is listed as a link to the submission
   - There is no link at the bottom of the page
   - All state contacts for each submission receives an withdrawn rate email

#### Screenshots
Withdraw rate

Withdraw rate linked to other submissions

#### Test cases covered

`sendWithdrawnRateStateEmail.test.ts`
- `'includes expected state contacts in toAddress'`
- `'renders withdrawn rate with single withdrawn from contract as expected'`
- `'renders overall email for a withdrawn rate as expected'`
- `'returns an error if rate consolidatedStatus is not WITHDRAWN'`
- `'returns an error if no reviewStatusActions are present'`
- `'returns an error if latest reviewStatusActions action was not withdraw action'`
- `'returns an error when parsing withdrawn from contract data fails'`
- `'returns an error when rate has no withdrawn from contracts'`


services/app-api/src/emailer/emails/sendWithdrawnRateStateEmail.test.ts
```diff
@@ -0,0 +1,384 @@
...
+            testEmailConfig(),
+            testRate,
+            statePrograms
+        )
+
+        if (template instanceof Error) {
+            throw template
+        }
+
+        console.log(template.bodyHTML)
```
pearl-truss, line 190: Nit: I don't think we want this in the test. Same with line 258

services/app-api/src/emailer/etaTemplates/sendWithdrawnRateStateEmail.eta
```diff
@@ -0,0 +1,20 @@
+<h1 style=<%~'"font-weight: normal;"'%>>Rate <%= it.rateName %> was withdrawn by CMS</h1>
+<b>Withdrawn by:</b> <%= it.withdrawnBy %><br />
+<b>Withdrawn on:</b> <%= it.withdrawnDate %><br />
+<b>Reason for withdrawal:</b> <%= it.withdrawnReason %><br />
+<b>Withdrawn from:</b>
+<% if (it.withdrawnFromContractData.length > 1) { %>
+<ul>
+<% it.withdrawnFromContractData.forEach(function(contract){ %>
+    <li>
+        <a href="<%= contract.submissionURL %>"><%= contract.contractName %></a>
```
pearl-truss, line 10: Nice to see us able to use this new contract field instead of needing to derive this

pearl-truss (approved):
Approving this but there is an unrelated issue I'm seeing in this branch and in dev

Screenshot 2024-12-31 at 4 12 52 PM

haworku:
re: discontinuous rate position-  I think the check throwing that error  (ratePosition !== lastPosition + 1`) could be loosened. To my knowledge, we only use the rate position to do relative ordering/sorting, so `ratePosition > lastPosition + 1` would be enough for that

JasonLin0991:
@pearl-truss Fixed the system error, also fixed an issue where the prisma transaction was not rolling back changes inside the transaction when an error happens.

> re: discontinuous rate position- I think the check throwing that error (ratePosition !== lastPosition + 1) could be loosened. To my knowledge, we only use the rate position to do relative ordering/sorting, so ratePosition > lastPosition + 1` would be enough for that

Whoops, didn't see this before I started fixing it. This would have worked too. At least the changes I made we shouldn't have gaps in positions now.

PR #3042, opened 2024-12-31
Upgrade to Typescript 5

haworku:


haworku:
Two major classes of issues so far:
1.  `dayjs` errors related to the constructed types in our dates package (need help with this @mojotalantikite , I wasn't totally sure why we created that external type definition in the first place, looking through dayjs issues I was seeing a lot of different approaches to inferring types wanted to hear from you more about how we got the types defined righ tnow) 
2. UnlockedContract v Contract types - still haven't figured this out, but I'm getting closer.

PR #3051, opened 2025-01-03
Jumpbox command always prompt for password and allow empty password

JasonLin0991:
## Summary
- This always prompts user for the ssh password. The way we checked to see if the error was about a missing password was brittle and could break if the error message changed at all. Since the error does not contain an error code we are limited how to address this. Always prompting is a quick solution for this.


mojotalantikite (approved):
LGTM! I was able to use it to pull down data via the jumpbox as expected.

PR #3083, opened 2025-01-15
MCR-4653: CMS does not see withdrawn rate reviews on dashboard by default

JasonLin0991:
## Summary
[MCR-4653](https://jiraent.cms.gov/browse/MCR-4653)

AC:
- If I hit my rate dashboard, withdrawn rates are filtered out by default

Other work:
- Simplify loading filters from URL when the URL had filters
   - The original implementation was for some UI flickering. Seems like it's no longer an issue without it.

#### Test cases covered

`RatesReviewTable.tsx`
- `'renders rates table correctly'`
   - Updated to assert withdrawn rate was not visible.


pearl-truss (approved):
Nice improvement getting rid of the flickering

PR #3116, opened 2025-01-29
MCR-4893: Adjust Apollo handler to use remove-parameter-store feature flag

haworku:
## Summary
This is prep work to clear out Parameter Store lookups from apollo_gql - the issue that has caused bottleneck for external APIs. Instead, we add that config to the database. 

Slight adjustment to original PR https://github.com/Enterprise-CMCS/managed-care-review/pull/3099 

#### Related issues

[MCR-4893](https://jiraent.cms.gov/browse/MCR-4893)

#### Test cases covered

- `submitContract.test.ts` - uses email settings from database with remove-parameter-store flag on
- `fetchMcReviewSettings.test.ts`- uses email settings from database with remove-parameter-store flag on
- There are several other resolvers that use the emailer dependency for CMS emails but with the same setup

## QA guidance

- Since the defaults in Prisma are the same as `dev` we wont see any changes on the settings page.  Just test that emails still send both flag on and flag off. 

## Code review guidance
- apollo_gql.ts reads and configures email configuration data into a function and returns the configuration for the Apollo server.
- The new function should gather email configuration based on feature flag state. 
   - With "Remove Parameter Store" feature flag ON, the source of the emails are queried from the database.
   - With "Remove Parameter Store" feature flag OFF the source of the emails are queried from parameter store.
- Unit tests for flag on and flag off
   - Use the new function for gathering email configuration in constructTestPostgresServer to configure unit tests.
   - For fetchMcReviewSettings resolver the configuration emails are from the correct source
 

## Additional notes 

-  Updated the default email settings values in Prisma schema to not have a `cms.hhs.gov` email and only use our `truss email.`
- There is a drift in the types `EmailSettings` (values from the database) and `EmailConfiguration` (values for the emailer).
   - 3 fields `cmsReviewHelpEmailAddress`, `cmsRateHelpEmailAddress`, and `helpDeskEmail` differ in types (`string` and `string[]`).  New ticket for cleanup:  [MCR-5018](https://jiraent.cms.gov/browse/MCR-5018): Clean up type drift between EmailSettingsType and EmailConfiguration type


services/app-api/src/handlers/configuration.ts
```diff
@@ -82,4 +90,125 @@ async function getDBClusterID(secretName: string): Promise<string | Error> {
...
+    if (emailSettings instanceof Error) {
+        return emailSettings
+    }
+
+    return {
+        emailSource: emailSettings.emailSource,
+        devReviewTeamEmails: emailSettings.devReviewTeamEmails,
+        helpDeskEmail: emailSettings.helpDeskEmail[0],
+        cmsReviewHelpEmailAddress: emailSettings.cmsReviewHelpEmailAddress[0],
+        cmsRateHelpEmailAddress: emailSettings.cmsRateHelpEmailAddress[0],
```
mojotalantikite, line 106: Is there a reason why the db type on these is `TEXT[]`, but we're only reading out the first value?

services/app-api/src/handlers/configuration.ts
```diff
@@ -82,4 +90,125 @@ async function getDBClusterID(secretName: string): Promise<string | Error> {
...
+}: {
+    emailParameterStore: EmailParameterStore
+    store: Store
+    ldService: LDService
+    stageName: string
+    emailerMode: string
+    applicationEndpoint: string
+}): Promise<Emailer | Error> {
+    const removeParameterStore = await ldService.getFeatureFlag({
+        key: 'throwaway-key-email-configuration',
```
mojotalantikite, line 189: Is this the correct key?

mojotalantikite (approved):

services/app-api/src/handlers/configuration.ts
haworku: Yes - looks like it was to avoid cleanup. The types changed but we didn't have time to fully refactor so made a cleanup ticket -  [MCR-5018](https://jiraent.cms.gov/browse/MCR-5018): Clean up type drift between EmailSettingsType and EmailConfiguration type. We can do that after this ships but eventually this will handle multiple email address.

services/app-api/src/handlers/configuration.ts
haworku: The key in the LD key application should be the user id, which there is none at this point. Whatever value is here is not going to be like our other keys in us in the application. Gave it weird name since this is just getting overidden. We are passing `anonymous` in as well (though not convinced thats working, need to test on next PR).

PR #3121, opened 2025-01-30
Add serverless v4 and license key to CI

mojotalantikite:
## Summary

This adds the serverless v4 license key to github actions and let's us read from it. We're not currently using it at this point, but it sets us up for the rest of the epic.

#### Related issues
https://jiraent.cms.gov/browse/MCR-5068

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -160,7 +160,7 @@ jobs:
 
   promote-infra-dev:
     needs: [build-prisma-client-lambda-layer, build-clamav-layer, unit-tests]
-    uses: Enterprise-CMCS/managed-care-review/.github/workflows/deploy-infra-to-env.yml@main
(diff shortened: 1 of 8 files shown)


.github/workflows/promote.yml
```diff
@@ -160,7 +160,7 @@ jobs:
 
   promote-infra-dev:
     needs: [build-prisma-client-lambda-layer, build-clamav-layer, unit-tests]
-    uses: Enterprise-CMCS/managed-care-review/.github/workflows/deploy-infra-to-env.yml@main
+    uses: Enterprise-CMCS/managed-care-review/.github/workflows/deploy-infra-to-env.yml@mt-serverless-v4
```
haworku, line 163: assuming you want this and then will pull the change back but commenting in case

haworku (approved):
Tag me in github if you want code review i'll see it faster

PR #3122, opened 2025-01-30
MCR-4888 editEmailSettings API

haworku:
## Summary
Add new `editEmailSettings` part of ripping out Parameter Store bottlebeck in `apollo_gql` startup. 

- This API an only be called by Admin user, allows update of the new fields.
- Enforce email validations here   - create a zod schema type EmailSettings to validate with z.string().email().optional() for each item in the string array.
- We can go ahead and error if emails not sent in for updates of any field, no need for partial updates, just make sure errors tells you which field or email to fix.

## Related issues
[MCR-4888](https://jiraent.cms.gov/browse/MCR-4888)

## Test cases covered

`settings.test.ts`
Tests the Zod custom validator for emailSettingsSchema

    'should validate emails with display names'
    'should validate plain emails'
    'should trim white space of emails'
    'should reject empty email settings'
    'should reject invalid email format'
    'should reject empty email arrays'
    'should reject malformed email aliases'

`updateEmailSettings.test.ts`
resolver tests

    'updates email settings'
    'errors when on failed email validations'
    'errors when user is not an admin'


services/app-api/src/handlers/configuration.ts
```diff
@@ -72,7 +72,7 @@ async function configurePostgres(
         return new Error('Failed to create Prisma Client')
     }
 
-    const client: PrismaClient = prismaResult
+    const client: PrismaClient = prismaResult as unknown as PrismaClient
```
mojotalantikite, line 75: I was wondering why we're now casting to `unknown` -- I guess it's from the recent change to make `NewPrismaClient` return an `ExtendedPrismaClient`? I think we should probably try to stay away from giving up some of the type safety here, even though we know it's still a PrismaClient.

What if we change `ExtendedPrismaClient` to an intersection: 
```typescript 
type ExtendedPrismaClient = PrismaClient & ReturnType<typeof extendedPrismaClient>
```

I'd have to play with it but I think something along those lines would give `tsc` enough info to not need `unknown` here?

mojotalantikite (approved):

services/app-api/src/handlers/configuration.ts
haworku: @mojotalantikite It didn't like the intersection because the narrower PrismaClient also used everywhere else, the intersection type clashes with that. I just changed all the consumers and tests to the new type to workaround. Seem alright?

PR #3133, opened 2025-02-04
[MCR-4677] updated dev tool to auto-build packages

ruizajtruss:
## Summary
This PR adds a new script to watch package files and trigger the build command

#### Related issues
[MCR-4677](https://jiraent.cms.gov/browse/MCR-4677)

## QA guidance

Make a change in one of the files found in the packages directory (I used constants) - you should see the change reflected in the corresponding js files

### DIFF
--- a/dev_tool/src/dev.ts
+++ b/dev_tool/src/dev.ts
@@ -47,8 +47,7 @@ async function runAllBuild(runner: LabeledProcessRunner) {
         ['pnpm', 'build:packages'],
         ''
     )
-    //Ensures we watch files in the packages directory for any changes to trigger build:packages
-    runner.runCommandAndOutput('watch packages', ['pnpm', 'packages:watch'], '')
+
     await runner.runCommandAndOutput('build', ['pnpm', 'build'], '')
 }
 
--- a/dev_tool/src/dev.ts
+++ b/dev_tool/src/dev.ts
@@ -47,6 +47,8 @@ async function runAllBuild(runner: LabeledProcessRunner) {
         ['pnpm', 'build:packages'],
         ''
     )
+    //Ensures we watch files in the packages directory for any changes to trigger build:packages
+    runner.runCommandAndOutput('watch packages', ['pnpm', 'packages:watch'], '')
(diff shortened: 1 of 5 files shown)


haworku:
@JasonLin0991  could you review this since i know you were seeing the pain point recently

dev_tool/src/local/web.ts
```diff
@@ -21,7 +21,8 @@ export async function runWebLocally(runner: LabeledProcessRunner) {
         ['pnpm', 'build:packages'],
         ''
     )
-
+    //Ensures we watch files in the packages directory for any changes to trigger build:packages
+    runner.runCommandAndOutput('watch packages', ['pnpm', 'packages:watch'], '')
```
JasonLin0991, line 25: I'm seeing `pnpm packages:watch` ran twice.
 


**Suggestion**: We should set up the `watch packages` like we do with the `gen` folder for generated graphQL schema. Take a peek in `dev_tool/src/local/graphql.ts` where we are using `once()` from `dev_tool/src/deps.ts`.

JasonLin0991 (approved):
This relieves my pain! There is one suggestion I think we should implement though, so that we are not running the watcher twice.

dev_tool/src/dev.ts
```diff
@@ -47,6 +47,8 @@ async function runAllBuild(runner: LabeledProcessRunner) {
         ['pnpm', 'build:packages'],
         ''
     )
+    //Ensures we watch files in the packages directory for any changes to trigger build:packages
+    runner.runCommandAndOutput('watch packages', ['pnpm', 'packages:watch'], '')
```
JasonLin0991, line 51: **Question**: I'm not sure if we need to run the watcher here. Shouldn't the code on line 45 already build the packages module, and `runAllBuild` should exit after line 51 completes?

dev_tool/src/local/web.ts
ruizajtruss: Ah yea that's cool, hadn't seen it! Looks like it's also used for the proto watcher as well, makes sense to apply it here too - thanks!

PR #3140, opened 2025-02-07
MCR-5071: re-enable frontend caching between summary and dashboard

JasonLin0991:
## Summary
[MCR-5071](https://jiraent.cms.gov/browse/MCR-5071)

This work re-enables frontend caching on the CMS and State submission dashboard, CMS rate dashboard, and summary pages. In addition, the dashboard pages have a 5-minute polling interval to refetch data.

AC:
- Reduce use of the loading indicator as the user travels between **cms submissions dashboard** and **cms rates dashboard**
- Reduce use of the loading indicator as the user travels between **dashboard pages and summary pages**
- NOTE: discussing with design and product, we are ok with not notifying the user that the data is being re-fetched while the cached data is on screen
   - This means on the dashboard page, new submissions or submission updates will update without any sort of loading indicator.
   - We also are not showing any type of error message if a re-fetch fails.

Additional work
- Add polling on dashboard pages set at 5-minute intervals.

## QA guidance

- To test this, while on the dashboard page, use a different window as incognito to make changes to a submission. Then, on the original window, navigate to a submission and back. There should be no loading state, and after a few moments, the submissions on the dashboard should update with the changes made.
- To test polling, do the same as above, except on the original window, do not navigate away and wait 5 minutes for the data to refresh.

### DIFF
--- a/packages/mocks/src/apollo/rateDataMock.ts
+++ b/packages/mocks/src/apollo/rateDataMock.ts
@@ -438,7 +438,6 @@ function mockRateSubmittedWithQuestions(
         reviewStatusActions: [],
         createdAt: rate.createdAt ?? new Date(),
         updatedAt: rate.updatedAt ?? new Date(),
-        id: rateID,
         webURL: `https://testmcreview.example/rate/${rateID}`,
         stateCode: rate.stateCode ?? 'MN',
         state: rate.state ?? mockMNState(),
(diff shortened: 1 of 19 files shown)


pearl-truss (approved):
LGTM definite improvement! I am curious about how the 5 minutes for polling was decided?

haworku (approved):
LGTM !

JasonLin0991:
> LGTM definite improvement! I am curious about how the 5 minutes for polling was decided?

Not much thought was behind it. This could be replaced by a feature flag value we set in LD in the future though.

PR #3148, opened 2025-02-10
[MCR-5020] Spike reduce zod parsing time (not meant to merge)

pearl-truss:
## Summary

This spike explored a few different options for how we might reduce zod parsing times. Each option is encapsulated in its own PR for clarity about what code would need be touched and changed.
 
**As a benchmark, duration with using the current setup of `contractSchema.safeParse`**

Screenshot 2025-02-07 at 1 38 20 PM

**And without any zod parsing (removing parsing from submissions dashboard all together)**

- consistently found zod parsing accounts for about 30-35% of load time on the submission dashboard

Screenshot 2025-02-07 at 1 49 04 PM

### The options explored were: (clicking commit link will take you to the code changes specific to each option)

**Use Zod accelerator library** [commit](https://github.com/Enterprise-CMCS/managed-care-review/pull/3148/commits/0a122b3486d66c0504321be7278113e0a4545935)
- Out of the box, this library causes significant 2x-3x delays in load time on  submission dashboard with ~200 submissions and a timeout with 400+ submissions

Screenshot 2025-02-07 at 1 45 00 PM

**Use zod's `safeParseAsync` over `safeParse`** [commit](https://github.com/Enterprise-CMCS/managed-care-review/pull/3148/commits/25432018d75a5fa6c973f8673db9f20bf0b04a06)
- on `parseContractWithHistory` the `z.safeParseAsync` results in a significant _increase_ in dashboard load duration

Screenshot 2025-02-07 at 1 41 12 PM

**Use zod's lazy method for nested/recursive objects** [commit](https://github.com/Enterprise-CMCS/managed-care-review/pull/3148/commits/a18ae2938bdca4ef68b92ac82c22161c49ecdf62)

Screenshot 2025-02-07 at 3 07 40 PM

**Use Valita library** [commit](https://github.com/Enterprise-CMCS/managed-care-review/pull/3148/commits/6eb83f6e3547c32d8c0d2ec50dcbf6c48bb9d5e3)

 - Supposedly about 4-5x as fast as zod, according to a[ benchmarking page ](https://tmcw.github.io/typescript-runtime-type-benchmarks/)that compares Zod and other parsing libraries to each other based on speed
 - Valita only supports validation and parsing. States that data type definitions is a non goal and beyond the scope of the library
- Limited amount of built in validation types (no built in support for date, uuid, adding min or max rules to numbers, etc)

**Create a flag `useZod` that can be used to bypass using zod parsing on indexContract (and wherever else parsing isn't needed** [commit ](https://github.com/Enterprise-CMCS/managed-care-review/pull/3148/commits/d5d773ee433d324f7096bbe14072abe5eb16418e)

- Gives the same 30-35% reduction in page load time as removing `safeParse` altogether 
- Lightweight and easy to implement 
- Will need to ensure `safeParsing` isn't needed on the dashboard/where `useZod` is set to false

Screenshot 2025-02-11 at 10 35 59 AM

##  Takeaway

- I highly favor the idea @haworku had for limiting how often and where we use `safeParse`. The `useZod` option was easy to implement and gives the best performance boost (30-35% reduction in load)

- Some of the other easier to implement options like Zod Accelerator and `safeParseAsync` resulted in longer load times instead of shorter.

- Using [zod's lazy](https://zod.dev/?id=recursive-types) method, did decrease load times by about 10-15% on the contract dashboard page (when tested with about 400 submissions). This was very easy to implement. However, I didn't see additional returns when I played around with adding `lazy` in multiple places beyond the `contractSchema`
- The Valita library looks promising. Would easily be a lot of work. In the scope of this spike ticket, I started down the path of what it could look like to implement parsing with Valita. The biggest drawback is the minimal nature of the library, we would need to create custom types for dates. Even more than that we would need to develop a plan for domain types. I don't think it would make the most amount of sense to keep zod around just for types given that zod isn't a type library. However, if we move parsing away from Zod and domain types away from Zod, that's a lot of change and heavier lift. I wasn't able to get Valita fulling working, I was getting validation errors so I wasn't able to confirm how much fast Valita would be in our specific case  

I did play around with different zod versions, and I did see that some versions of Zod are significantly faster than others. This is ultimately neither here nor there since some of the faster versions were 4+ years old. 

Ultimately, I think if we want to undertake a large/extra large epic it could be worth it to replace Zod entirely. When looking at the speed performance of different parsing libraries, Zod lagged by a lot. And when Zod was not used for parsing, it resulted in a 30-35% reduction in load times, which is significant. 

#### Resources
- Valita library https://github.com/badrap/valita
- MyZod library https://github.com/davidmdm/myzod (another library that is faster than zod, but more limited in functionality and popularity/support)
- Speed up dashboard notion page (I added some of the spike findings to this doc) https://www.notion.so/mc-review/Speed-Up-Dashboard-1834f9e180fa80aa972deb2545a94489
- Runtime benchmark for different parsing libraries https://tmcw.github.io/typescript-runtime-type-benchmarks/


haworku:
Thanks for the commit history that was very helpful.

1. For this epic, we should do the lazy method on contractSchema and rateSchema . They both have nested structures. 

2. Agree that refactoring out Zod for a different paradigm sounds like significant effort. Interesting to see what Valita does. I do I think we could figure out the custom types. It is the prospect of rethinking `ContractType` at a higher level that concerns me. If we cannot figure out another way to reduce the load times, I think I would advocate for paginating the API before reimplementing our parse library near term. It is mainly the index API that we have performance complaints . _Any thoughts about that?_ 

3. Wondering in terms of options explored from the ticket, did you also still look into the "Reduce number of times we zod Parse (refactor prisma handling code)" from the ticket? 
____________
Looking at code a bit this afternoon, I started to wonder about changing `parseRateWithHistory` and `parseContractWithHistory` to take in a boolean `useZod` that defaults to true. Then...
- the parameter could be passed  in as false in the dashboard queries like indexContract/indexRates. 
- and we make a much simpler typeguard that returns ContractType to use when useZod=false.
- we start to open up possibility to useZod  more selectively.

If I'm reading the code right, it seems like we already get very close to the desired Zod type by manually building out the withHistory types _before_ `contractSchema.safeParse(contractWithHistory)` is run in those parse functions. If that's true, I don't think we have a strong need for the `safeParse` usage in those findAll* functions. Could you take a look at this as part of #3 here - reducing need to call zod safeParse, see if it gives us performance boost on indexContracts?

pearl-truss:
@haworku your `useZod` idea was a really good one! I added it to the notes and commit history. It gives the greatest performance boost and was straightforward to implement

haworku:
Nice yes lets do that in but add in the lazy thing too I'm thinking. When we implement to merge lets def add comments around intended use of this parameter - it should be used when we mutate/write data ... but its not needed when we query I think.  I think we can close this PR and move on to implementation

JasonLin0991:
Adding my thoughts here it might be rehashing what everyone else said here.

Working on MCR-4372: reducing indexContract data, there is significant amount of data that we don't need for the dashboard pages. I wonder if refactoring both indexContract and indexRate api's, beyond the GraphQL return data, is a solution here.

The refactor would include
- New graphql type for the query
- New domain type with reduced amount of data
- Refactor postgres to query only data we need, not the full contract data
   - I forget what the time for query data in postgres was, but removing some relationship lookups should reduce time here.
- New parse function for index to format only data we need
   - Really we only need certain formData fields
   - We don't need the full history only the most recent revision and draft revision
- Skip zod parsing altogether.

This seems likes like a lot of work though, maybe a 5 pointer.

I see the need to zod parse individual submissions, but at the index level it doesn't seem like its needed we can also just rely on our own validations.

I think the reason for this spike was the long loading times on the dashboard pages due to index queries. With that in mind, the Zod parsing contributed a large part to it, but replacing Zod for a different library seems like a lot of work to fix two queries when the other long term fix, pagination, might make it not worth replacing. Maybe the incremental fixes so far along with removing Zod parsing in the index queries would be good enough for now and we can spend the time working on pagination.

pearl-truss:
@JasonLin0991 thanks for the comments. MCR-5020 was specifically about reducing zod parsing times since that was a bottleneck that MacRae identified. 

GraphQL and data changes to how we're doing indexContracts and indexRates should give us additional performance improvements. What you laid about switching to parsing formData on indexContract does contradict the proposal here to no longer do zod parsing on indexContract. I would be curious to see what type of performance reduction we get with that option.

JasonLin0991:
> @JasonLin0991 thanks for the comments. MCR-5020 was specifically about reducing zod parsing times since that was a bottleneck that MacRae identified.
> 
> GraphQL and data changes to how we're doing indexContracts and indexRates should give us additional performance improvements. What you laid about switching to parsing formData on indexContract does contradict the proposal here to no longer do zod parsing on indexContract. I would be curious to see what type of performance reduction we get with that option.

You're right this is out of scope for this spike.

>  I would be curious to see what type of performance reduction we get with that option.

I did some rough testing locally and there is a significant reduction in `parseContractWithHistory` runtime. Postgres also takes a significant amount of time. I used my [MCR-4372](https://github.com/Enterprise-CMCS/managed-care-review/pull/3152) and your benchmarking code inside `findAllContractsWithHistoryBySubmitInfo` to record these durations.

Screenshot 2025-02-11 at 11 42 51 PM

haworku:
Pearl approving this work and closing the PR since we will reopen another. These improvements look good - thank you for spurring the discussion and giving us a more comprehensive look at things. 

@JasonLin0991 thanks for opening a separate PR with the postgres changes idea . Lets discuss there

PR #3152, opened 2025-02-11
MCR-5096: Refactor postgres queries and implement indexRatesStripped functions

JasonLin0991:
## Summary
[MCR-5096](https://jiraent.cms.gov/browse/MCR-5096)

- New resolver  and query `indexRatesStripped` that returns a stripped down version of a rate without history, draft contracts, or contracts rate was withdrawn from
   - This replaces the `indexRatesForDashboard` query.
   - The goal here is to reduce the amount of data we return to the frontend when trying to fetch all rates, as the removed data requires a lot of relational lookups in the database.
- New Postgres function `findAllRatesStripped` queries the database for all rates stripping away relational data.
   -  From discovery, we found out that the `Prisma` query we used `includeFullRate` contained a massive amount of relational lookups that was not needed for a index rates query that only needed a select amount of data.
- `parseStrippedRateWithHistory` parses a rate into a stripped down version, mainly removing the `packageSubmissions` that is included in the full rate.
- Update `RateReviewsDashboard` component to use new query and fields

Other
- Manually sorted revisions in `strippedRateToDomainModel` and `rateWithoutDraftContractsToDomainModel`. We are not guaranteed to get rate revisions in ascending order, so we should just do that to not have issues in the future.
- `getContractRateStatus` fix sorting so it does not mutate the argument passed in. 
- `getContractReviewStatus` fix sorting so it does not mutate the argument passed in. 
- `getParentContractID` abstract out this code into a function as its used in two places now.

#### Screenshots
Here are some of the metrics from local deployment. Be aware that these metrics are only for confirmation that there was a reduce in query time, I doubt the reductions would be this large in `PROD` due to the difference in the environment, but it should be significant.

The `Other` column is just the difference between `Total` and `Postgres`. The small reduction in the `Other` column, is likely mostly from GraphQL processing less rate data.

## QA guidance

- Test that the rate dashboard has no regressions in the UI as we changed the query data for this page.
- Test that different types of rates (except draft rates) show up on the dashboard.


haworku (approved):

haworku (changes requested):
Let's still hold on merge -   I'm seeing something else to change

haworku:
From discussion today - our end state goal is two separate queries in the graphql resolver - indexContracts (returns `Contract[]` and `indexContractsForDashboard` returns `ContractInDashboard[]`

JasonLin0991:
Closing this test PR. Not needed as Pearl's work is essentially the same, so I would think the metrics would be almost the same.

JasonLin0991:
Oop, closed the wrong PR.

haworku (commented):
This is looking good and I think its well isolated from other work which I like. Thanks for weaving things through to the frontend as well. Let me look again in the AM and do the QA test steps you wrote-   then I'll approve

haworku (approved):
Looks good, let's wait for ARMS to re-confirm their `indexRates` not `indexRatesForDashboard` API usage before merging in abundance of caution.

haworku:
We spoke more in slack, decided to just revert the graphql renaming for existing queries for now. Bring back indexRatesForDashboard for now (let's comment the top level file and graphql schema that this should not be used, seek to migrate to stripped API)

PR #3157, opened 2025-02-12
Test reduced postgres query data and not zod parsing for indexContracts resolver

JasonLin0991:
## Summary

This PR is for testing indexContracts API for CMS users with reduced postgres query data and no zod parsing of contracts.

- Add a new postgres function `findAllContractsForCMSDashboard` for querying postgres for contracts by submitInfo.
   - Removed `withdrawnRates` and `draftRates` from the prisma query
      - These two where the easiest to remove. There are more relationships that can be removed, but it would most likely require domain and grahpql type changes. Specifically in `contracts.revisions.relatedSubmisions`.
   - Removed zod parshing
- Add feature flag `test-index-contract-refactor` to toggle `indexContracts` resolver to use `findAllContractsForCMSDashboard`
- Add performance logging for both flag off and flag on for both postgres query and parsing.
- This does not test State view of the app.

NOTES:
- Locally this work shows a big improvement in postgres query time. I do have to caveat that there seems to be a large variance in the postgres query time, the middle graph has no changes to the query, but it is 7.8% faster. So I would reduce the third charts postgres improvement by 8% 
- The test data in the graphs are all contract and rates and lots of withdrawn rates. This means there are heavy relational look ups. The benefits might differ on prod, it would probably be better to update this graph with prod data.

Screenshot 2025-02-12 at 1 35 09 PM


JasonLin0991:
Also please note this was branched off of [MCR-4372](https://github.com/Enterprise-CMCS/managed-care-review/pull/3152): reduce indexContract return data. That PR only removed the fields in the graphQL schema.

haworku:
@JasonLin0991 Checking my understanding -- assuming we land an equivalent version of #3148 as well as #3152 -- what's happening here that's different is just not fetching the `withdrawn` and `draft` rates. And that change alone is accounting for 55% + speed up in prisma just that change?

haworku:
From in person discussion - we will land Pearl's work and then retest with the flag, see what the performance bump looks like

JasonLin0991:
Closing this test PR. Not needed as Pearl's work is essentially the same, so I would think the metrics would be almost the same.

PR #3169, opened 2025-02-18
Add secrets manager password rotator in TypeScript

mojotalantikite:
## Summary

We have been using an AWS supplied secrets manager rotation lambda, but at some point it seems like it stopped working. It is written in Python [here](https://github.com/aws-samples/aws-secrets-manager-rotation-lambdas/blob/master/SecretsManagerRDSPostgreSQLRotationSingleUser/lambda_function.py) and we were importing it from one of their S3 buckets.

This PR refactors that lambda into TypeScript and brings the lambda directly into our repository.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4743


services/postgres/src/secrets.ts
```diff
@@ -0,0 +1,116 @@
...
+        secret: SecretDict
+    ): Promise<void> {
+        const command = new PutSecretValueCommand({
+            SecretId: arn,
+            ClientRequestToken: token,
+            SecretString: JSON.stringify(secret),
+            VersionStages: ['AWSPENDING'],
+        })
+
+        await this.client.send(command)
```
haworku, line 90: Notice there's no error checking here coming out of cilent.send - this happens in `putSecret`, `describeSecret` and `updateSecretStage`-  is that fine?

services/postgres/src/rotator.ts
```diff
@@ -0,0 +1,194 @@
+import { DescribeSecretCommandOutput } from '@aws-sdk/client-secrets-manager'
```
haworku, line 1: Assuming this is based on the aws lambda rotator script we removed - can we add that in here in the comments somewhere in case someone wants to see where the approach was developed

haworku (approved):
LGTM.

services/postgres/src/secrets.ts
```diff
@@ -0,0 +1,116 @@
...
+            if (!(field in secretDict)) {
+                throw new Error(`${field} is missing from secret JSON`)
+            }
+        }
+    }
+
+    async generatePassword(): Promise<string> {
+        const command = new GetRandomPasswordCommand({
+            PasswordLength: 32,
+            ExcludeCharacters: ':/@"\'\\',
```
haworku, line 67: the original script also passed   RequireEachIncludedType: true

services/postgres/src/rotator.ts
haworku: This comments also seems helpful to include in our version too for someone wondering whats happening

   _This script enable single-user rotation scheme to rotate an RDS PostgreSQL user credential. This rotation
    scheme logs into the database as the user and rotates the user's own password, immediately invalidating the
    user's previous password._

services/postgres/src/types.ts
```diff
@@ -0,0 +1,15 @@
+export interface SecretDict {
```
haworku, line 1: Maybe we could be more explicit commenting or describing types here to explain whats expected. What I see from the other script docs
  
   {
        'engine': <required: must be set to 'postgres'>,
        'host': <required: instance host name>,
        'username': <required: username>,
        'password': <required: password>,
        'dbname': <optional: database name, default to 'postgres'>,
        'port': <optional: if not specified, default port 5432 will be used>
    }

services/postgres/src/secrets.ts
mojotalantikite: Thanks good catch, I can add some here

services/postgres/src/rotator.ts
mojotalantikite: Yup absolutely, I'll add some documentation like this into the script

mojotalantikite:
Closing for https://github.com/Enterprise-CMCS/managed-care-review/pull/3197 -- this branch is having issues after a merge from `main`.

PR #3185, opened 2025-02-20
[MCR-4871] Adding rate validation to submitContract resolver

ruizajtruss:
## Summary
The submitContract resolver was updated to validate for existing rates on `CONTRACT_AND_RATES` submissions and the frontend temporary validation was removed
#### Related issues
[MCR-4871](https://jiraent.cms.gov/browse/MCR-4871)

#### Test cases covered

Tests were adjusted to include rates where necessary


services/app-api/src/resolvers/contract/approveContract.test.ts
```diff
@@ -24,7 +24,7 @@ describe('approveContract', () => {
             s3Client: mockS3,
         })
 
-        const contract = await createAndSubmitTestContract(stateServer)
```
JasonLin0991, line 27: **Issue**: I think the fix here is to refactor `createAndSubmitTestContract`. The issue is that `createAndSubmitTestContract` is assumed to create and submit a test `CONTRACT_ONLY` submission, but it uses `createAndUpdateTestContractWithoutRates` which creates a `CONTRACT_AND_RATES` draft without rates that causes the error when trying to submit. There are probably some tests that are explicating trying to test a `CONTRACT_ONLY` submission so we could be invlidating a test by using `CONTRACT_AND_RATES` data.

We need to be explicit about what kind of test data we are creating by adding a new helper, something like `createAndUpdateTestContract`, that will actually create a `CONTRACT_ONLY` draft submission.

services/app-api/src/resolvers/contract/submitContract.ts
```diff
@@ -176,7 +176,10 @@ export function submitContract(
                 )
         }
 
-        const contractToParse = contractWithHistory
+        const contractToParse = {
```
JasonLin0991, line 179: **Suggestion**: @haworku and I had a 1:1, she brought up using `Object.assign()` here instead just to be explicit about what we are doing, creating a new object.

services/app-api/src/resolvers/contract/submitContract.ts
```diff
@@ -249,21 +252,37 @@ export function submitContract(
...
-                    setErrorAttributesOnActiveSpan(errMessage, span)
-                    throw new UserInputError(errMessage, {
-                        argumentName: 'rateID',
-                        cause: 'INVALID_PACKAGE_STATUS',
-                    })
+        //Contract and rates rate verification
+        if (parsedSubmissionType === 'CONTRACT_AND_RATES') {
+            //Ensure a contract and rates contract includes rates
+            if (
+                parsedContract.draftRates &&
```
JasonLin0991, line 259: **Question**: I think `parsedContract.draftRates` could be undefined here since we are checking if it exists. If that is the case we also want to throw the same the error if `parsedContract.draftRates` is undefined. We should add it as a `||` condition here. Could you double check and see if that is true?

services/app-api/src/resolvers/contract/submitContract.ts
```diff
@@ -249,21 +252,37 @@ export function submitContract(
...
+            }
+
+            //Ensure included rates are in a valid state
+            if (
+                parsedContract.draftRates &&
+                parsedContract.draftRates.length > 0
+            ) {
+                for (const draftRate of parsedContract.draftRates) {
+                    if (['WITHDRAWN'].includes(draftRate.consolidatedStatus)) {
+                        const errMessage = `Attempted to submit a rate withdrawn rate: ${draftRate.consolidatedStatus}`
```
JasonLin0991, line 278: **Suggestion**: Oops, I think I badly worded this error message.
```suggestion
                        const errMessage = `Attempted to submit contract with a withdrawn rate. Rate id:  ${draftRate.id}`
```

services/app-api/src/resolvers/rate/withdrawRate.test.ts
```diff
@@ -361,7 +373,8 @@ describe('withdrawRate', () => {
                 }),
             ])
         )
-
+        // eslint-disable-next-line no-console
+        console.log('juan:', resubmittedContract)
```
JasonLin0991, line 377: Missed a console.log.

services/app-api/src/resolvers/contract/submitContract.test.ts
```diff
@@ -107,17 +111,28 @@ describe('submitContract', () => {
 
         expect(draftRates).toHaveLength(1)
 
-        const contractWithRate = await submitTestContract(stateServer, contractID)
+        const contractWithRate = await submitTestContract(
```
JasonLin0991, line 114: ** Chore**: We should add one test to verify that contract and rates submissions returns and error if no rates exist.

JasonLin0991 (changes requested):

services/app-api/src/resolvers/contract/approveContract.test.ts
JasonLin0991: Update: I saw the ticket mention something around the helper functions. I think my comment above would dip into that HPP refactor work so lets skip that for this PR.

PR #3190, opened 2025-02-21
Fix undoWithdrawRate to ignore contract only contracts

JasonLin0991:
## Summary

Updating the `undoWithdrawRate` Postgres function to skip contracts from `withdrawnFromContracts` that are now `CONTRACT_ONLY`.

Also split updating the Rate review status and its `withdrawnFromContracts` relationship. Updating the review status before anything else so we are not modifying a rate that is still in the `WITHDRAWN` status. 

We haven't ran into either of these issues and I doubt that we will, but it's good to make sure we account for it anyways.


pearl-truss (approved):
I think these are great proactive changes

PR #3197, opened 2025-02-25
 Add secrets manager password rotator in TypeScript 

mojotalantikite:
## Summary

(This replaces #3169, which had some trouble with old data after a merge from `main`).

We have been using an AWS supplied secrets manager rotation lambda, but at some point it seems like it stopped working. It is written in Python [here](https://github.com/aws-samples/aws-secrets-manager-rotation-lambdas/blob/master/SecretsManagerRDSPostgreSQLRotationSingleUser/lambda_function.py) and we were importing it from one of their S3 buckets.

This PR refactors that lambda into TypeScript and brings the lambda directly into our repository.

#### Related issues
https://jiraent.cms.gov/browse/MCR-4743


haworku (approved):
LGTM - reviewed previous PR, checked out the new error handling

JasonLin0991 (approved):

PR #3198, opened 2025-02-25
Add backup retention check for rds

mojotalantikite:
## Summary

We're getting a lot of security alerts in our `dev` environment (~140) due to RDS snapshots not having encryption enabled. However, we really shouldn't even need snapshots in `dev`, as they're for the most part all review environments with test data from Cypress or manual testing. 

This disables all backups for review environments and sets regular intervals for the main `dev` account, as well as `val` and `prod`.

#### Related issues
https://jiraent.cms.gov/browse/MCR-5048

### DIFF
--- a/services/postgres/serverless.yml
+++ b/services/postgres/serverless.yml
@@ -70,6 +70,11 @@ custom:
     other: 0.5
     val: 1
     prod: 1
+  backupRetentionPeriod:
+    other: 0 # No backups for feature branches/PR review environments
+    main: 7
+    val: 7
+    prod: 30
   scriptable:
     hooks:
       before:package:createDeploymentArtifacts: pnpm build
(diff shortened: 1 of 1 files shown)


haworku (approved):
Yes please

haworku:
@mojotalantikite seeing 
> [#/BackupRetentionPeriod: 0 is not greater or equal to 1].
ServerlessError2: An error occurred: PostgresAuroraV2 - Properties validation failed for resource PostgresAuroraV2 with message:

mojotalantikite:
> @mojotalantikite seeing
> 
> > [#/BackupRetentionPeriod: 0 is not greater or equal to 1].
> > ServerlessError2: An error occurred: PostgresAuroraV2 - Properties validation failed for resource PostgresAuroraV2 with message:

Yeah, unfortunately it looks like we can't just turn this off in our review environments. There is a minimum value of 1. 

I'm going to just set this to the minimum there, so that we can at least clear out some of these unnecessary snapshots. I wrote a ticket to handle the long term fix of specifying the encryption at rest in the CF configs (val/prod may have been manually configured for encryption when we first cut over to RDS): https://jiraent.cms.gov/browse/MCR-5119

PR #3234, opened 2025-03-12
Fix undoWithdraw API

JasonLin0991:
## Summary

There was a bug when a contract had only a single rate and it was withdrawn, then un-withdrawn. I had mistakenly put creating thew new review action type for the rate at the top of `undoWithdrawRateInsideTransaction`, which would not work because `unlockContractInsideTransaction` will see it as a submitted rate and try to unlock it again.

Moving this block code for creating review action to after the rates have been reattached to the contract fixes the code. I initially had this code block at the bottom, but I believe I was thinking it made more sense to un-withdrawal it before reattaching it.

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/undoWithdrawRate.ts
+++ b/services/app-api/src/postgres/contractAndRates/undoWithdrawRate.ts
@@ -35,22 +35,6 @@ const undoWithdrawRateInsideTransaction = async (
 ): Promise<RateType | Error> => {
     const { rateID, updatedByID, updatedReason } = args
 
-    // Update the review status first
-    await tx.rateTable.update({
-        where: {
-            id: rateID,
-        },
-        data: {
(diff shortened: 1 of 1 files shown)


services/app-api/src/postgres/contractAndRates/undoWithdrawRate.ts
```diff
@@ -211,6 +195,22 @@ const undoWithdrawRateInsideTransaction = async (
         }
     }
 
+    // Update the review status first
```
JasonLin0991, line 198:
```suggestion
```

JasonLin0991, line 198:
```suggestion
    // Update the review status
```

ruizajtruss (approved):
Tested this in the deploy - works great!

PR #3237, opened 2025-03-13
MCR-2645 Use logical DB for review environments

mojotalantikite:
## Summary

Up until now we've been deploying a fully unique environment for git branches and PRs, which includes standing up an individual Aurora Postgres instance just like we have in dev/val/prod. However, this approach causes an increase of resources in our `dev` AWS account (Aurora, networking, snapshots, etc.). 

This change will now have us deploy a logical DB that is a part of the main `dev` database and will not stand up a separate Aurora instance for review environments. It uses a lambda in the `postgres` service to add or delete the logical database. When a new branch is created, the `postgres` deploy step will create the infrastructure and then call the lambda to 1) generate a new AWS secrets manager secret 2) log into the `dev` db to create the new user 3) create the new database inside the `dev` Aurora instance. Once a PR is merged or the branch is deleted, that DB will be dropped from inside the main `dev` Aurora instance. 

This reduces resources in AWS and will bring down our costs. It also should make CI faster, particularly on first run, as we won't be deploying a new Aurora DB in the environment.

#### Related issues
https://jiraent.cms.gov/browse/MCR-2645

#### Testing Guidance

Just need to make sure the application is working as expected, which should be covered by Cypress. I've already verified that we are getting data as expected into this logical DB that is in the main `dev` Aurora instance:

```sh
reviewapp_mtlogicaldbs=> select * from "User";
                  id                  | givenName | familyName |       email       |       role        | stateCode | divisionAssignment |        createdAt        |        updatedAt
--------------------------------------+-----------+------------+-------------------+-------------------+-----------+--------------------+-------------------------+-------------------------
 84b84438-e041-708d-8082-211425e3cf68 | Aang      | Avatar     | aang@example.com  | STATE_USER        | MN        |                    | 2025-03-13 17:48:19.733 | 2025-03-13 17:48:19.733
 b4787438-70f1-704f-6a14-b6864da67d40 | Iroh      | Uncle      | iroh@example.com  | ADMIN_USER        |           |                    | 2025-03-13 17:48:24.954 | 2025-03-13 17:48:24.954
 341834d8-b0d1-7013-bcc6-8b2ef9ca4a67 | Azula     | Hotman     | azula@example.com | CMS_APPROVER_USER |           | DMCO               | 2025-03-13 17:48:48.209 | 2025-03-13 17:48:57.001
 1478c4a8-50b1-707d-5462-ffcaec152744 | Zuko      | Hotman     | zuko@example.com  | CMS_USER          |           | DMCO               | 2025-03-13 17:48:23.117 | 2025-03-13 17:49:20.425
(4 rows)
```

And I've verified that the cleanup works in a test run. 

Before the delete part of the lambda was run:
```bash
=> SELECT datname FROM pg_database WHERE datname LIKE 'reviewapp_%';
        datname
------------------------
 reviewapp_mtlogicaldbs
(1 row)

=> SELECT rolname FROM pg_roles WHERE rolname LIKE 'reviewapp_user_%';
           rolname
-----------------------------
 reviewapp_user_mtlogicaldbs
(1 row)
```
after
```bash
=> SELECT rolname FROM pg_roles WHERE rolname LIKE 'reviewapp_user_%';
 rolname
---------
(0 rows)

=> SELECT datname FROM pg_database WHERE datname LIKE 'reviewapp_%';
 datname
---------
(0 rows)
```

### DIFF
--- a/services/postgres/db-manager.sh
+++ b/services/postgres/db-manager.sh
@@ -1,14 +1,11 @@
 #!/usr/bin/env bash
 set -u
-
 # Script to invoke the database manager lambda function
 # Usage: ./db-manager.sh create|delete stageName dbSecretArn
-
 action="$1"        # create or delete
 stage_name="$2"    # The stage/environment name for the database
 secret_arn="$3"    # ARN of the secrets manager secret for the main dev DB
 lambda_stage="${4:-dev}"  # Stage where the lambda is deployed, defaults to dev
-
 function_name="postgres-${lambda_stage}-dbManager"
 lambda_version="\$LATEST"
 cli_read_timeout=240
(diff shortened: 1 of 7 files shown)


services/postgres/db-manager.sh
```diff
@@ -0,0 +1,144 @@
...
+stage_name="$2"    # The stage/environment name for the database
+secret_arn="$3"    # ARN of the secrets manager secret for the main dev DB
+lambda_stage="${4:-dev}"  # Stage where the lambda is deployed, defaults to dev
+function_name="postgres-${lambda_stage}-dbManager"
+lambda_version="\$LATEST"
+cli_read_timeout=240
+
+# Ensure all required parameters are provided
+if [[ -z "$action" || -z "$stage_name" || -z "$secret_arn" ]]; then
+  echo "Usage: $0 create|delete stageName dbSecretArn [lambdaStage]" 1>&2
```
haworku, line 17: Should there be some kind of guard even in this script to avoid deleting higher environments? I get that the way we are calling it right now, it wouldn't be used with non dev environments... more just being cautious

haworku (approved):

haworku:
@mojotalantikite  Is this significant of a change to deserve quick ADR? Might make it easier to demo this also if it does end up saving us $$$

services/postgres/db-manager.sh
mojotalantikite: Yeah I can add a check that bails if stage_name is main/val/prod. I'll probably want to put that protection in the lambda itself, but will also put it here.

mojotalantikite:
> @mojotalantikite Is this significant of a change to deserve quick ADR? Might make it easier to demo this also if it does end up saving us $$$

Yeah that's a good call, I'll add a ticket to this sprint to do that.

PR #3249, opened 2025-03-18
Refactor RDS snapshot lambda

mojotalantikite:
## Summary

We noticed a little while back that some older snapshots for RDS were not being cleaned out properly. It looks like it might be because I wrote the original cleanup script using a `forEach`, which might not finish before the lambda halts. This moves us to use a `Promise.all` approach so that we can make sure the lambda doesn't stop before the operations finish. 

#### Related issues
https://jiraent.cms.gov/browse/MCR-5114


haworku (approved):
LGTM

PR #3250, opened 2025-03-18
Upgrade storybook 

haworku:
## Summary
Noticed storybook upgrades failing with dependabot when they seem trivial. Tried upgrading with storybook latest tooling, it found a few related things we hadn't upgraded that storybook requires. 

## QA guidance
Tests pass and manually check storybook deploys


haworku:
This seems to work. I see there are unrreleated errors in some storybook stories right now `Error: Tracing context not initialized. Ensure TraceProvider is mounted.` likely from move to Vite - I can make a ticket to reoslve

JasonLin0991 (approved):

PR #3255, opened 2025-03-21
MCR-5000: State cannot add rate cert date that is in the future.

JasonLin0991:
## Summary
[MCR-5000](https://jiraent.cms.gov/browse/MCR-5000)

- Add Yup validation for rate cert date so that the date cannot be after the current date..


haworku (approved):
LGTM

PR #3256, opened 2025-03-21
MCR-4652: CMS sees banner alert about rates that are linked will not be withdrawn

ruizajtruss:
## Summary
This PR adds the warning banner for rates that will not be withdrawn with the submission. There is also some work here done by Jason to add the related contract information used to display the banner (see below).

Other work:
- Update `IndexRatesStripped` API to include related contracts and their `consolidatedStatus` when included in query.
   - Add `relatedContracts` to graphql
   - Add `relatedContracts` to `rateStrippedResolver`
   - Add `findRateRelatedContracts` prisma function to query for related contracts and latests `consolidatedStatus`.
   - Add `indexRatesStrippedWithRelatedContracts` query that includes related contracts. Including relatedContracts adds significant time to the Postgres query. Separating this part to only when needed so that it will not negatively impact the rate review dashboard load times. 

#### Related issues
[MCR-4652](https://jiraent.cms.gov/browse/MCR-4652)
#### Screenshots
image

#### Test cases covered
- Banner testing to ensure it displays properly
- `shouldWithdraw` function has multiple test cases to try and account for a multitude of permutations related to contract status and how that affects whether or not it will be withdrawn with the submission
## QA guidance
- Login as a CMS user
- access a submission with multiple contracts
- If any of the rates wont be withdrawn they should display
- You can also navigate to the parent of a displayed rate to change it's status and see it disappear from the banner


haworku:
Please make sure we comment all these similar index resolvers clearly so we document expected use case for future

services/app-api/src/postgres/contractAndRates/findAllRatesStripped.test.ts
```diff
@@ -113,7 +81,7 @@ it('returns all rates with stripped down data', async () => {
 
     await withdrawTestRate(cmsServer, rateBID, 'Withdraw invalid rate')
 
-    const strippedRatesOrErrors = must(await findAllRatesStripped(client))
+    const strippedRatesOrErrors = must(await findAllRatesStripped(client, {}))
```
haworku, line 84: Nit: I avoid `{}`  for representing  empty/no args since it is not explicit. Prefer making the entire type of args optional.  This could be fixed later though

services/app-api/src/postgres/contractAndRates/findAllRatesStripped.test.ts
```diff
@@ -168,25 +138,68 @@ it('returns all rates with stripped down data', async () => {
...
+            rateIDs: [rateBFromArrayAgain.id],
+        })
+    )
+
+    // Expect only rate B to return
+    expect(selectedStrippedRates).toHaveLength(1)
+    expect(selectedStrippedRates[0].rateID).toBe(rateBFromArrayAgain.id)
+
+    const emptySelectedStrippedRatesOrError = must(
+        await findAllRatesStripped(client, { rateIDs: [] })
```
haworku, line 178: question: Feels odd to be able to pass empty array of  `rateIDs` to a query filter that is for _returning only specific rates_ and then expect all them back still. I'd expect none to be returned with this entry. Why are we handling this way, was there a reason to do this? Are we thinking folks may pass in empty arrays accidentally?
 
  Seems like if you want just all the rates, you would call`findAllRatesStripped` without any arguments.

services/app-api/src/postgres/contractAndRates/findRateRelatedContracts.test.tsx
```diff
@@ -0,0 +1,192 @@
...
+                consolidatedStatus: 'UNLOCKED',
+            },
+            {
+                id: contractC.id,
+                consolidatedStatus: 'SUBMITTED',
+            },
+        ])
+    )
+
+    // resubmit contract B should remove it from the related contracts
```
haworku, line 167: Nice this feels like useful test, basically making sure we don't persist the relationship changes in related contract until submit

services/app-api/src/postgres/contractAndRates/findRateRelatedContracts.test.tsx
haworku, line 1: issue: this should not be a .tsx file

services/app-api/src/postgres/contractAndRates/findRateRelatedContracts.tsx
haworku, line 1: issue: another tsx file

services/app-graphql/src/schema.graphql
```diff
@@ -1898,6 +1898,15 @@ type RateRevisionStripped {
     formData: RateFormDataStripped!
 }
 
+"""
+A rate revision represents a single submission
```
haworku, line 1902: comment text seems wrong here

services/app-graphql/src/schema.graphql
```diff
@@ -2020,6 +2029,9 @@ type RateStripped {
 
     "The latest submitted rate revision"
     latestSubmittedRevision: RateRevisionStripped!
+
+    "Contracts this rate is related to"
```
haworku, line 2033: suggestion: be explicit here in comment  -"related" means  what. It means both the parent contract and the additional linked contracts using the rate

services/app-web/src/components/Banner/WithdrawSubmissionBanner/WithdrawSubmissionBanner.tsx
haworku, line 1: issue: rename to SubmissionWithdrawWarningBanner 

This is quite different than our `RateWithdrawBanner`, which sounds similar but is a success banner shown in all cases.  This is a very a specific warning banner, specific use case

haworku (commented):
Still reviewing but pausing for now

services/app-web/src/pages/SubmissionWithdraw/SubmissionWithdraw.tsx
```diff
@@ -31,6 +46,45 @@ const submissionWithdrawSchema = Yup.object().shape({
     ),
 })
 
+export const shouldWithdraw = (
```
haworku, line 49: suggestion: this is actually storing value for `shouldWarnOnWithdraw` it doesnt prevent withdraw

services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
```diff
@@ -213,6 +216,82 @@ const getParentContractID = (
...
+                const submitInfo =
+                    pkg.contractRevision.contract.revisions[0].submitInfo
+                const unlockInfo =
+                    pkg.contractRevision.contract.revisions[0].unlockInfo
+                const latestAction =
+                    pkg.contractRevision.contract.reviewStatusActions[0]
+
+                // Find the current status of all related contracts as some could be in different status than the rate.
+                // If it has both unlock and submit info, then it was resubmitted
+                if (submitInfo && unlockInfo) {
```
haworku, line 256: issue: We have functions capturing this logic `getContractReviewStatus`, `getRateReviewStatus`, `getContractRateStatus`.   Let's lean into or update the existing functions to make this easier to maintain otherwise we are creating another place where status could get out of sync

haworku: 🤔 I would probably make `getRateRevisionStatus`and `getContractRevisionStatus` call those both here and in the existing helpers used for other functions. Unify the logic now.

JasonLin0991: I think it would be some work to make those functions work for both this data structure type and their existing ones. I recommend we split that work up to a follow up PR.

PR #3286, opened 2025-04-04
ADR on using logical DBs in Aurora

mojotalantikite:
## Summary

This just adds the (already decided) ADR for using logical DBs deployed inside a single Postgres Aurora instance in our `dev` AWS account.

#### Related issues
https://jiraent.cms.gov/browse/MCR-5166

### DIFF
--- a/docs/architectural-decision-records/029-review-envs-use-logical-dbs.md
+++ b/docs/architectural-decision-records/029-review-envs-use-logical-dbs.md
@@ -1,0 +1,49 @@
+---
+title: Review Environments Use Logical Postgres Databases 
+---
+## ADR 029 - Review Environments Use Logical Postgres Databases
+
+- Status: Decided
+- Date: 4/4/2025
+
+## Decision Drivers
+- Minimize infrastructure resources deployed in dev (cost savings)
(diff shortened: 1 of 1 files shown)


haworku:
@mojotalantikite `ApiGwWebAclAssociation - Resource handler returned message: "AWS WAF couldn?t perform the operation because your resource doesn't exist.`  did I cause this with closing and reopening 😮‍💨

mojotalantikite:
> @mojotalantikite `ApiGwWebAclAssociation - Resource handler returned message: "AWS WAF couldn?t perform the operation because your resource doesn't exist.` did I cause this with closing and reopening 😮‍💨

Oh no, not your fault! This happens sometimes with the version of Serverless we're pinned to, as there is a bug with finding the `instanceId` of a serverless run (#3229). I had to add a ton of resource to the `dependsOn` and it more often than not fails now on first run 🫠 . I can write another bug ticket, but I'm not sure there's anything we can do about it unless we move to a more recent version of serverless v4, which breaks local offline dev and requires everyone to have AWS credentials in their environment locally.

haworku (approved):
No need for bug ticket, we moving to AWS CDK soon.

PR #3297, opened 2025-04-09
MCR-5207: reassign parent contract on withdrawn contracts child rates

JasonLin0991:
## Summary
[MCR-5207](https://jiraent.cms.gov/browse/MCR-5207)

- Refactor `withdrawContract` postgres function to reassign child rates that cannot be withdrawn to new parent contracts.
   - Child rates linked to other contracts that are `SUBMITTED`, `UNLOCKED` (but had been submitted with this linked rate), and `APPROVED` cannot be withdrawn.
   - These rates will be reassigned a new parent contract. The contract will chosen from the contracts linked to this rate.
   - The priority of the selected of the new parent contract is based on the status. From highest to lowest priority `SUBMITTED` -> `UNLOCKED` -> `APPROVED`.
      - The `UNLOCKED` contract must have been submitted with this rate previously as a linked rate.
      - `SUBMITTED` includes `RESUBMITTED`, but in the code we are just checking if the latest revision has a `submitInfo`. Knowing the difference between `SUBMITTED` and `RESUBMITTED` is irrelevant here.
   - When reassigning rate to a `SUBMITTED` or `APPROVED` contract we unlock this new parent contract to resubmit it with the rate so it becomes its child rate **by sharing the same submitInfo**
   - When reassigning rate to a `UNLOCKED` contract, we resubmit this contract with the rate so it becomes its child rate **by sharing the same submitInfo**. Then we unlock the contract again to restore it to its `UNLOCKED` state like we found it.
- Refactor `getContractID` function to use the latest submitted revision to find the current parent contract.
   - Before we used the initial submitted revision, but now that we can reassign the parent contract we have to look at the latest submitted revision
- Add `getNewParentContract` function to get new parent contracts for a rate.
- Add `reassignParentContractInTransaction` function.
- Set a global test timeout for API tests. This should also help local testing as you hit the timeout often.

AC:
- When a submission is withdrawn, a child rate linked to another submission is not withdrawn but assigned a new parent submission from one of its linked submissions.
- After submission withdrawal, the linked rate on the rate summary page still can be unlocked (now from a new parent submission)

#### Test cases covered

`withdrawContract.test.ts` - resolver test
- (UPDATE) `cant withdraw contract and rates submissions without withdrawing linked rates` -> `can withdraw contract and reassigns parent contract to rates that cannot be withdrawn`
- (NEW) `prioritizes submitted contracts for rate parent contract reassignment`
- (NEW) `prioritizes unlocked contracts for rate parent contract reassignment when there are no submitted contracts`
- (UPDATE) `withdraws rate if when withdrawing linked contract and parent contract is already withdrawn` -> `withdraws rate with reassigned parent`
- (UPDATE) `withdraws rate with contract when rate is linked to unlocked contract` -> `withdraws rate with parent when linked contract is unlocked without previously submitting with linked rate`

## QA guidance

- Test complex linked rates still get reassigned a new parent contract
   - Unlock and resubmits with links, delinks and removals of rates before testing withdraws
- Test the status of the rate on the dashboard and in rate summary match its new parent submission.

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
+++ b/services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
@@ -28,12 +28,11 @@ import type {
     RateRevisionTablePayload,
     RateTableWithoutDraftContractsStrippedPayload,
     RateRevisionTableWithRelatedSubmissionContracts,
+    SelectSubmissionPackageContractRevisionData,
+    RateTableWithRelatedContractsPayload,
 } from './prismaSubmittedRateHelpers'
 import type { ConsolidatedRateStatusType } from '../../domain-models/contractAndRates/statusType'
-import type { RateTableWithRelatedContractsPayload } from './prismaSubmittedRateHelpers'
-import type { HealthPlanPackageStatusType } from '../../domain-models'
 import type { RelatedContractStripped } from '../../gen/gqlClient'
-import type { ConsolidatedContractStatus } from '../../gen/gqlClient'
 
 const subincludeUpdateInfo = {
     updatedBy: true,
@@ -91,32 +90,42 @@ function getContractRateStatus(
         | ContractRevisionTableWithFormData[]
         | RateRevisionTableWithFormData[]
         | StrippedRateRevisionTableWithFormData[]
+        | SelectSubmissionPackageContractRevisionData[]
 ): PackageStatusType {
     // need to order revisions from latest to earliest
     const revs = [...revisions].sort(
         (revA, revB) => revB.createdAt.getTime() - revA.createdAt.getTime()
     )
     const latestRevision = revs[0]
-    // submitted - one revision with submission status
-    if (revs.length === 1 && latestRevision.submitInfo) {
-        return 'SUBMITTED'
-    } else if (revs.length > 1) {
-        // unlocked - multiple revs, latest revision has unlocked status and no submitted status
-        // resubmitted - multiple revs, latest revision has submitted status
-        if (latestRevision.submitInfo) {
-            return 'RESUBMITTED'
-        }
+
+    const submitInfo = latestRevision.submitInfo
+    const unlockInfo = latestRevision.unlockInfo
+
+    // If both info exists then it was resubmitted, you can only get an unlockInfo after the initial submit.
+    if (submitInfo && unlockInfo) {
+        return 'RESUBMITTED'
+    }
+
+    // If only unlock info exists, then it was never resubmitted,
+    if (!submitInfo && unlockInfo) {
         return 'UNLOCKED'
     }
+
+    // If only submitInfo exists, then it is the first submission as a first revision can never have a unlockInfo.
+    if (submitInfo && !unlockInfo) {
+        return 'SUBMITTED'
+    }
+
+    // If no submitInfo or unlockInfo we know this is a newly created contract with an initial revision.
     return 'DRAFT'
 }
 
 // -----
 function getContractReviewStatus(
-    contract: ContractTableWithoutDraftRates
+    reviewStatusActions: ContractTableWithoutDraftRates['reviewStatusActions']
 ): ContractReviewStatusType {
     // need to order actions from latest to earliest
-    const actions = [...contract.reviewStatusActions].sort(
+    const actions = [...reviewStatusActions].sort(
         (actionA, actionB) =>
             actionB.updatedAt.getTime() - actionA.updatedAt.getTime()
     )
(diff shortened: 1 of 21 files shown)


services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
```diff
@@ -198,19 +198,35 @@ const getParentContractID = (
...
+                    )
+                }
+
+                parentContractID = latestParentContract.contractID
             }
-            const firstContract = submission.submittedContracts[0]
-            parentContractID = firstContract.contractID
+        } else {
+            return new Error(
+                'Could not find a submitted revision to get parent contract id.'
```
haworku, line 228: nit: Return this error at the top in a guard clause. Basically the whole point of the `if ...else` is to guard against nonsensical data because we have overly loose database types -- is that right?

 If so, do the data checks at the top and return promptly. Want to avoid the doubly nested if ...else, it's confusing.

services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
```diff
@@ -198,19 +198,35 @@ const getParentContractID = (
...
-                console.error(msg)
-                return new Error(msg)
+        //find the latest submitted revision with submitted contracts
+        const latestSubmittedRevision = revisions.find(
+            (revision) => revision.submitInfo?.submittedContracts[0]
+        )
+
+        if (latestSubmittedRevision?.submitInfo) {
+            if (latestSubmittedRevision.relatedSubmissions.length == 0) {
+                console.info('No related submission. Unmigrated rate.')
```
haworku, line 208: question: what is an unmigrated rate in this context - is this likely

services/app-api/src/postgres/contractAndRates/unlockRate.ts
```diff
@@ -13,8 +13,12 @@ type UnlockRateArgsType = {
 async function unlockRateInDB(
     tx: PrismaTransactionType,
     rateID: string,
-    unlockInfoID: string
+    unlockInfoID: string,
+    // Used to specify a timestamp when calling multiple prisma functions within the same transaction that need
```
haworku, line 17:
```suggestion
    // Used to override createdAt timestamp when calling multiple prisma functions within the same transaction that need
```

services/app-api/src/postgres/contractAndRates/withdrawContract.ts
```diff
@@ -151,20 +156,33 @@ const withdrawContractInsideTransaction = async (
...
-        const firstRateRevision = rate.revisions[rate.revisions.length - 1]
+        if (!latestRateRevision.submitInfo) {
+            throw new Error(
+                `Child rate ${rate.id} of contract to be withdrawn does is not submitted.`
+            )
+        }
+
         const parentContractID =
-            firstRateRevision.submitInfo?.submittedContracts[0].contractID
+            latestRateRevision.submitInfo?.submittedContracts[0].contractID
```
haworku, line 185: suggestion: I think this is third or fourth time I'm seeing `rateRevision.submitInfo.submittedContracts[0]` logic. This feels like good candidate for a helper - I guess its `getRelatedContracts` on rate revision

services/app-api/src/postgres/contractAndRates/withdrawContract.ts
```diff
@@ -176,42 +194,78 @@ const withdrawContractInsideTransaction = async (
...
+                  contractID: string
+                  status: ConsolidatedContractStatus
+              }
+            | undefined = undefined
+
+        // Looping through each rate's latest revision's related submissions to find a valid new parent contract.
+        for (const pkg of submissionPackages) {
+            const contractID = pkg.contractRevision.contract.id
+            const latestRevision = pkg.contractRevision.contract.revisions[0]
+            const isApproved =
```
haworku, line 217: `isApproved` `isWithdrawn` both should be helpers on a contract

services/app-api/src/postgres/contractAndRates/withdrawContract.ts
```diff
@@ -237,18 +291,141 @@ const withdrawContractInsideTransaction = async (
...
-        submittedByUserID: updatedByID,
-        submittedReason: `CMS has withdrawn the submission ${contractName}${rateNamesMessage}. ${updatedReason}`,
-    })
+    if (unlockContract instanceof Error) {
+        throw new Error(
+            `Cannot unlock contract to be withdrawn. ${unlockContract.message}`
+        )
+    }
+
+    //NOTE: Consider a separate function to do the reassignment
```
haworku, line 306: issue: Please break this code up now before merging, make the ` reassignRateParentContractInTransaction`.

services/app-api/src/resolvers/rate/undoWithdrawRate.test.ts
```diff
@@ -379,7 +379,7 @@ describe('undoWithdrawRate', () => {
                 bodyHTML: expect.stringContaining(contractBName),
             })
         )
-    }, 10000)
```
haworku, line 382: observation: we need to bump the overall timeout on backend tests I think. I see we randomly extend to 15000, 20000, 30000 something places its all scattered/random. I would assign a variable to this `EXTENDED_TIMEOUT` thats shared or just bump the overall settings on the test config

haworku (commented):
Code review: The only thing blocking I saw is breaking up that 500 ln+ resolver more.  Comments were super clear, tests look good. 
Want to break up code now - we already struggle with unlock and update, we should seek to merge something a bit more modular please see where we can reduce the maintenance risk.

services/app-api/src/postgres/contractAndRates/withdrawContract.ts
```diff
@@ -176,42 +194,78 @@ const withdrawContractInsideTransaction = async (
...
-        //      submissions for rate in its latest submission package.
-        ratesToWithdraw.push(rate)
+        let reassignToContractID:
+            | {
+                  contractID: string
+                  status: ConsolidatedContractStatus
+              }
+            | undefined = undefined
+
+        // Looping through each rate's latest revision's related submissions to find a valid new parent contract.
```
haworku, line 213: suggestion:  think this long loop is a good candidate also for standalone function. Seems like it has has a type of something  like `prepareToWithdrawRates(rates) => [ratesToWithdraw, ratesToResassignParent]`

services/app-api/src/postgres/contractAndRates/withdrawContract.ts
```diff
@@ -151,20 +156,33 @@ const withdrawContractInsideTransaction = async (
...
+    // Child rates that need to be reassigned a new parent contract
+    const ratesToReAssign: {
+        rateID: string
+        rateName?: string
+        reassignToContractID: string
+        contractStatus: ConsolidatedContractStatus
+    }[] = []
 
-    // loop through all rates to determine which ones can be withdrawn.
+    // Loop through all rates to determine which ones can be withdrawn or reassigned parent contract
```
haworku, line 174: suggestion:  think this long loop is a good candidate also for standalone function. Seems like it has has a type of something  like `prepareToWithdrawRates(rates) => [ratesToWithdraw, ratesToResassignParent]`

services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
JasonLin0991: I am not sure what a unmigrated rate is, but it was added in this [PR](https://github.com/Enterprise-CMCS/managed-care-review/pull/2349) implementing linking, fetching, and submitting rates. I left it in to test some prod data and doesn't look like any of the `PROD` rates hit this `console.info`. 

The `relatedSubmissions` would be records in the `UpdateInfoTable` that is related to this revision. The only way this would empty is on a draft rate which has never been submitted or unlocked so no relationships to the  `UpdateInfoTable`. The code on line 192 checks to see if this is a draft then using a place holder to find the parent contract later. So new submissions I don't think we would ever hit this because we are checking for `submitInfo` which is a record in the `UpdateInfoTable`.

I couldn't find any more information on where the migration related to this happened. Possibly some bad data in `DEV` or `VAL` might trigger this.

So in summary, very unlikely in prod and unlikely on new submissions. Only some kind of bug or bad data migration would cause this.

services/app-api/src/postgres/contractAndRates/withdrawContract.ts
JasonLin0991: I don't think it would be worth it yet. I have 2 instances of this and writing a new function would be way more work to try to get the types out of the query with nested selects. If I encounter this a couple more times then I'll make a function.

services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
```diff
@@ -181,100 +192,164 @@ function getConsolidatedRateStatus(
...
+    if (!latestParentContract) {
+        return new Error(
+            'programming error: its a submitted rate that was not submitted with a contract initially'
+        )
+    }
+
+    return latestParentContract.contractID
+}
+
+// Function to loop through a rate revisions related submissions to find a new suitable parent contract.
```
haworku, line 241:
```suggestion
// Function to loop through a list of revisions on a rate and find a suitable parent contract. Meant to be used in cases where the original parent contract associated with that rate has been withdrawn but the rate is still linked to other contracts.
```

haworku (approved):

haworku:
LGTM all excellent cleanup

PR #3300, opened 2025-04-10
Bump max file size for zip file generation

mojotalantikite:
## Summary

We had a submission come in with a lot of ~25MB pdfs attached, which hit that upper limit of 500MB we have set for zip file generation. This bumps us up to 600MB in order to get this submission's zip file generated properly.

#### Related issues
https://jiraent.cms.gov/browse/MCR-5227


JasonLin0991 (approved):
Looks good to me.

PR #3307, opened 2025-04-15
[MCR-5015] Undo withdraw cypress test

ruizajtruss:
## Summary
Cypress testing for undoing a withdrawn rate
#### Related issues
[MCR-5015](https://jiraent.cms.gov/browse/MCR-5015)
#### Test cases covered

1. CMS user should be able to successfully withdraw a rate

### DIFF
--- a/services/cypress/integration/cmsWorkflow/rateReview.spec.ts
+++ b/services/cypress/integration/cmsWorkflow/rateReview.spec.ts
@@ -6,51 +6,7 @@ describe('CMS user can view rate reviews', () => {
         cy.interceptGraphQL()
     })
 
-    it('and withdraw a rate', () => {
-        cy.interceptFeatureFlags({
-            'withdraw-rate': true,
-        })
-        cy.apiAssignDivisionToCMSUser(cmsUser(), 'DMCO').then(() => {
-            // Create a new contract and rates submission with two attached rates
-            cy.apiCreateAndSubmitContractWithRates(stateUser()).then(
-                (contract) => {
-                    const latestSubmission = contract.packageSubmissions[0]
-                    const rate1 = latestSubmission.rateRevisions[0]
-
-                    // Then check both rates in rate reviews table
-                    cy.logInAsCMSUser({
-                        initialURL: `/dashboard/rate-reviews`,
-                    })
-
-                    cy.get('table')
-                        .findByTestId(`rate-link-${rate1.rateID}`)
-                        .should('exist')
-                    // click the first rate to navigate to rate summary page
-                    cy.get('table')
-                        .findByTestId(`rate-link-${rate1.rateID}`)
-                        .click()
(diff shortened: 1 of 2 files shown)


services/cypress/integration/cmsWorkflow/rateReview.spec.ts
```diff
@@ -50,6 +50,59 @@ describe('CMS user can view rate reviews', () => {
         })
     })
 
+    it('and undo a withdrawn rate', () => {
```
JasonLin0991, line 53: **Suggestion**: This is almost a duplicate of the test above. We could just use that test and change the name to `and withdraw then undo withdraw a rate` so we can keep the Cypress test run times as slim as possible.

JasonLin0991 (approved):

services/cypress/integration/cmsWorkflow/rateReview.spec.ts
```diff
@@ -50,6 +50,59 @@ describe('CMS user can view rate reviews', () => {
...
+                    cy.findByRole('button', { name: 'Withdraw rate' }).click()
+
+                    cy.findByText('Withdraw a rate').should('exist')
+
+                    cy.get('#rateWithdrawReason').type('Withdrawn for testing')
+                    
+                    cy.findByRole('button', { name: 'Withdraw rate' }).click()
+                    
+                    cy.url({ timeout: 20_000 }).should('contain', `rates/${rate1.rateID}`)
+
```
JasonLin0991, line 91: **Note**: Using a timeout on the `cy.url` may cause flakes in the future because it's imprecise in asserting when a mutation or query has successfully finished. If there were any delays of the queries or mutations between clicking the `withdraw rate` button and being redirected to the summary page then `20_000` might not be long enough. So increasing it would help, until you hit the timeout again then you would need to increase it again.

The way we tackle these assertions are to intercept and wait on the requests. That is a more precise and reliable way to make sure responses were successful. We have a established pattern for this. Take a peek in `cypress/support/command.ts` where we create a custom command that intercepts all of our GraphQL queries. Then in the beginning of our test `describe` block we call the custom command `cy.interceptGraphQL()` in a `beforeEach`. From that point each test  can wait on a query we called.

A good example of the use of this pattern is `questionReponse.spec.ts`. We call the `cy.interceptGraphQL()` command in the `beforeEach()` function. So now each test will have requests intercepted. Later on in the tests we call another custom command `cy.addResponse` which is a reusable command to add responses. Going to the definition of that command you can see that once we click the `Submit response` button we then use `cy.wait` on the requests we expect to get a response from.

**Suggestion**: We should use `cy.wait` on intercepted requests instead of using a timeout on the checking the url. Since the withdraw and undo withdraw rate workflow is pretty small we wouldn't need to make a custom command yet. If we were to test this workflow elsewhere and more often it would be worth abstracting it out to a custom command.

If you decided you want to try to do this in this PR, lets pair on it together.

PR #3310, opened 2025-04-17
ADR for pnpm as build tool 

haworku:
## Summary
ADR for previous work moving to pnpm, also cleaned up some references between ADRS overriding eachother.

I _think_ I caught a couple places that still need to use the pnpm script also? Maybe? There is a small code change here please review. 

#### Related issues
https://jiraent.cms.gov/browse/MCR-4307
https://jiraent.cms.gov/browse/MCR-4315

## QA guidance
Read these docs and directly suggest any changes or edits for things I mistated.

### DIFF
--- a/.github/workflows/validate-prs.yml
+++ b/.github/workflows/validate-prs.yml
@@ -19,7 +19,7 @@ jobs:
         uses: ./.github/actions/setup_env
 
       - name: DangerJS
-        run: yarn danger ci
+        run: pnpm danger ci
         env:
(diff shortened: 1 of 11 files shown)


mojotalantikite (approved):

.github/workflows/validate-prs.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         uses: ./.github/actions/setup_env
 
       - name: DangerJS
-        run: yarn danger ci
+        run: pnpm danger ci
```
mojotalantikite, line 22: oh nice thanks for catching this!

PR #3312, opened 2025-04-17
MCR-5146: update withdraw and undo withdraw rate api for submission withdraw

JasonLin0991:
## Summary

This work updates our `withdrawRate` and `undoWithdrawRate` for the submission withdraw feature which adds a new contract status of `WITHDRAWN` and parent contract reassignment.

**Notes**
- `unlockContract` postgres functions
   - This one needed updated for rate withdraw/un-withdrawal feature because of parent contract reassignment. We still query the rate the same, but now we need to look at the latest 2 rate revisions to find the contract it was submitted with.
   - We only want to unlock this child rate if the latest rate submission was submitted with this contract. If it wasn't then we know this rate was reassigned or withdrawn from this contract.
   - `findContractWithHistory.test.ts`
      - Skipping `matches correct rate revisions to contract revision with independent rate unlocks and submits`
      - See the comments above the test:
         ```
            // Test for independent rate unlocks/submits (not currently supported).
            // Skip reason: Recent withdraw/undo features introduced rate parent contract reassignment,
            // changing how parent contracts are determined (now using latest revision's submitInfo
            // instead of initial revision). This breaks the test's expectations about parent calculation.
            // Keeping for future implementation of independent rate actions.
        ```
   - Update the rate summary page so that `Undo withdraw` does not appear if the parent contract is `WITHDRAWN`.
      - This is so that actions users cannot take because of the API should also not be shown.
  
AC:
 - `withdrawRate` API allows a rate to be withdrawn if its parent contract is withdrawn
    - withdrawn rates will be unlock and resubmitted to move the rate relationship to the WithdrawnRateJoinTable
    - **TESTING NOTE**: This AC will be hard to test because now are reassigning linked rates when the parent contract is withdrawn there should never be a rate in this state from the submission withdraw process. I do think we include this AC and the unit test incase a bug or we figure out that there is an edge case.
 - `undoWithdrawRate` API will not allow a rate to be un-withdrawn if the parent contract is withdrawn.
 - `undoWithdrawRate` API allows rate to be un-withdrawn if linked contracts are withdrawn.
 - There are unit test for APU updates. It can be included in an existing test as an assertion.

#### Related issues
[MCR-5146](https://jiraent.cms.gov/browse/MCR-5146)

#### Test cases covered

`withdrawRate.test.ts`
- `'can withdraw a rate when parent contract is withdrawn'`
- `'withdraws rate when linked to withdrawn contract'`

`undoWithdrawRate.test.ts`
- `'can undo withdraw a rate previously linked to withdrawn contract'`
- `'returns an error if parent contract is in an invalid state'`

`RateSummary.test.tsx`
- `'does not render undo withdraw button when parent submission is withdrawn'`

## QA guidance

- A withdrawn rate, using the withdraw rate feature, should not be able to be un-withdrawn using the undo withdraw feature if the parent contract is now withdrawn
   - Create a contract and rates submission and submit it
   - Withdraw the rate using the withdraw rate feature
   - Withdraw the contract using the submission withdraw feature
   - Attempting to undo withdraw the rate should result in an error
- Allows a rate to be withdrawn, using the rate withdraw feature, if the parent contract is WITHDRAWN
   - This AC will be hard to test because now are reassigning linked rates when the parent contract is withdrawn there should never be a rate in this state from the submission withdraw process. I do think we include this AC incase a bug or we figure out that there is an edge case.
   - The unit test can cover the validation for this small piece of work.

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/findContractWithHistory.test.ts
+++ b/services/app-api/src/postgres/contractAndRates/findContractWithHistory.test.ts
@@ -911,8 +911,14 @@ describe('findContractWithHistory with full contract and rate history', () => {
 })
 
 describe('findContractWithHistory with only contract history', () => {
+    // Test for independent rate unlocks/submits (not currently supported).
+    // Skip reason: Recent withdraw/undo features introduced rate parent contract reassignment,
+    // changing how parent contracts are determined (now using latest revision's submitInfo
+    // instead of initial revision). This breaks the test's expectations about parent calculation.
+    // Keeping for future implementation of independent rate actions.
+    //
     // eslint-disable-next-line jest/no-disabled-tests
-    it('matches correct rate revisions to contract revision with independent rate unlocks and submits', async () => {
+    it.skip('matches correct rate revisions to contract revision with independent rate unlocks and submits', async () => {
         const client = await sharedTestPrismaClient()
 
         const stateUser = await client.user.create({
--- a/services/app-api/src/postgres/contractAndRates/unlockContract.ts
+++ b/services/app-api/src/postgres/contractAndRates/unlockContract.ts
@@ -115,7 +115,7 @@ async function unlockContractInsideTransaction(
                     },
                     unlockInfo: true,
                 },
-                //take: 2 // Only two revision states are possible: mutually unlocked or submitted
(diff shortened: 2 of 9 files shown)


ruizajtruss (approved):
Code looks good and I was able to verify everything works in the review app - great work! 🚀

PR #3321, opened 2025-04-22
MCR-5259: withdraw contract system error

JasonLin0991:
## Summary

Withdrawing a submission linked to an unlocked rate results in a system error. The fix was to to move checking for an unlocked child rate further in the code were we know this is a child rate. I also updated `getContractID` function to be more versatile by narrowing the parameter type to only the data we need.

#### Related issues
[MCR-5259](https://jiraent.cms.gov/browse/MCR-5259)

## QA guidance

Steps to validate fix: 
- Create a contract and rate submission and submit it
- Create a second contract and rate submission, linking to the rate from the first submission and submit it.
- As a CMS user unlock the first submission.
- As a CMS user try to withdraw the second submission that is linked to the rate that is the child of the first submission.
- The second submission should successfully withdraw


ruizajtruss (approved):
Validated this works 🚀

PR #3339, opened 2025-05-07
MCR-4833: post redundant rate migration cleanup

JasonLin0991:
## Summary
After migrations we can clean up the rest of the API code for replace redundant rate feature.

AC:
- Remove withdrawInfo from GraphQL rate schema and domain types
- Remove prisma queries with withdrawInfo.
- Remove from prisma to domain types converter functions
- Remove withdrawInfo from prisma schema
- Migration to delete withdrawInfo records from `UpdateInfoTable`. (Done in a previous migration)

Fixes:
- In the previous migration I forgot to make a relationship for the migrated redundant rates. The SQL code below is needed to make these relationships on those rates. This must be done directly on the `PROD` Postgres instance either before or after this PR is merged in.
   - You will not see errors in the app from this issue. The API that will surface this error is the `indexRatesStrippedWithContracts` query, when the common resolver `rateStrippedResolver` when parsing these migrated rates with `getRelatedContracts`. This function expects submitted rate revisions to have a `relatedSubmisions` entry, but the migration for redundant rates did not have it.
   - `indexRatesStrippedWithContracts` is only called when withdrawing a submission, but since these rates are already withdrawn it will never be called on these rates. However we should still fix the rates incase future work does call the endpoint with these rates.
      - The way I found it was by loading prod data locally and running `indexRatesStrippedWithContracts` unit tests.
   - The SQL code below will loop through all the migrated rates and add that relationship.
   - This [spreadsheet](https://docs.google.com/spreadsheets/d/1Itn6bwTgwQRMTfXOW6yj3FufS9dug0oFQ_6XPz2UjRM/edit?gid=0#gid=0) was the result of the first migration. There are 12 rates that we are doing this second migration on.

```sql
BEGIN;

DO $$
DECLARE
  -- List of rate IDs to process
  rate_ids TEXT[] := ARRAY[
    '13cd3be6-226c-4008-96a9-7a8cac1b2031',
    '33a0a3e8-a645-4729-bab3-39e6764d2f0a',
    '3f467b37-7fc6-45fa-9ccd-b29141fd48c1',
    '53bd1a93-6397-458d-8ed0-08faa1e4fef6',
    '6116527e-77ff-4cbe-b1f0-8ff6bc3d9513',
    '64e08f3d-af53-46fc-8d5e-fe3bc63eb4fa',
    '87d07e4c-bd27-4e4c-8bf4-b6836295f6f7',
    '880d460b-8564-4674-be48-dcc2836be059',
    '8d0f51f6-1d2c-4b9d-b01a-68112c3ba4df',
    'a8c4f89a-a24d-40cb-9007-888809704c2a',
    'b0c60b5e-95c3-4c8a-bafa-15a2cb3a5c20',
    'e276ba27-fee9-425e-88a9-438f51b7a016'
  ];

  -- Variables for loop processing
  current_rate TEXT;
  revision_id TEXT;
  submit_info_id TEXT;

BEGIN
  -- Process each rate ID
  FOREACH current_rate IN ARRAY rate_ids LOOP
    -- Find the latest revision for this rate (regardless of submitInfoID)
    SELECT id, "submitInfoID" INTO revision_id, submit_info_id
    FROM "RateRevisionTable" rrt
    WHERE "rateID" = current_rate
    ORDER BY "updatedAt" DESC
    LIMIT 1;

    -- Validate that we found a revision
    IF revision_id IS NULL THEN
      RAISE EXCEPTION 'No revision found for rate ID %', current_rate;
    END IF;

    -- Validate that the latest revision has a submitInfoID
    IF submit_info_id IS NULL THEN
      RAISE EXCEPTION 'Latest revision % for rate ID % does not have a submitInfoID',
        revision_id, current_rate;
    END IF;

    -- Create relationship in _RateRevisionTableToUpdateInfoTable if it doesn't exist
    BEGIN
        INSERT INTO "_RateRevisionTableToUpdateInfoTable" ("A", "B")
        VALUES (revision_id, submit_info_id);
    EXCEPTION
      WHEN unique_violation THEN
        -- If the relationship already exists, this is fine
        RAISE NOTICE 'Relationship between % and % already exists', revision_id, submit_info_id;
      WHEN foreign_key_violation THEN
        -- If there's a foreign key violation, raise an exception
        RAISE EXCEPTION 'Foreign key violation when inserting relationship between % and %',
          revision_id, submit_info_id;
      WHEN OTHERS THEN
        -- Catch any other exceptions and provide details
        RAISE EXCEPTION 'Error inserting relationship between % and %: %',
          revision_id, submit_info_id, SQLSTATE;
    END;
  END LOOP;
END $$;

COMMIT;
```


mojotalantikite (approved):
LGTM

PR #3344, opened 2025-05-09
[MCR-5273] State cannot remove a rate that has already been submitted

ruizajtruss:
## Summary
This PR removes the `Remove rate certification` for multiple unlocked rates
#### Related issues
[MCR-5273](https://jiraent.cms.gov/browse/MCR-5273)
#### Test cases covered
Added a test to ensure the button doesn't render for multiple unlocked rates

## QA guidance

- Submit a contract with 2 or more rates
- Unlock it as a cms user
- Visit the rate details page as a state user
- Expect the remove button to not render at the bottom of the rate
- However if you add a new rate on that same page expect the button to render for that specific rate

### DIFF
--- a/services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
+++ b/services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
@@ -388,10 +388,11 @@ const RateDetails = ({
 
     const displayRemoveRateBtn = (index: number, rateForm: FormikRateForm) => {
         if (index >= 1 && !displayAsStandaloneRate) {
-            //We expect to display the button regardless of status if the rate is a linked rate
+            //We expect to display the button regardless of status if the rate is a linked rate otherwise only display the button if it is a draft rate
             if (
                 rateForm.ratePreviouslySubmitted === 'YES' ||
-                rateForm.status !== 'UNLOCKED'
+                rateForm.status === 'DRAFT' ||
+                rateForm.status === undefined
             ) {
                 return true
             }
--- a/services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.test.tsx
+++ b/services/app-web/src/pages/StateSubmission/RateDetails/RateDetails.test.tsx
@@ -20,7 +20,6 @@ import {
     indexRatesMockSuccess,
     mockWithdrawnRates,
     mockContractPackageUnlockedWithUnlockedType,
-    mockMNState,
 } from '@mc-review/mocks'
 import { Route, Routes, Location } from 'react-router-dom'
 import { RoutesRecord } from '@mc-review/constants'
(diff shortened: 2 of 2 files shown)


services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
```diff
@@ -533,7 +533,9 @@ const RateDetails = ({
                                                                     />
                                                                 )}
                                                                 {index >= 1 &&
-                                                                    !displayAsStandaloneRate && (
+                                                                    !displayAsStandaloneRate &&
```
JasonLin0991, line 536: **Issue:** This removes the `Remove rate certification` button for linked rates that are unlocked. For linked rates we still want to be able to remove them regardless of their submission status (UNLOCKED, SUBMITTED, RESUBMITTED).

**Suggestion:** Also check if this is a child rate, `rateForm.ratePreviouslySubmitted` calculation includes the parentContract comparison, you might be able to use that. Otherwise you can add `parentContractID` to the return data of `convertGQLRateToRateForm`.

JasonLin0991 (changes requested):

services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
ruizajtruss: Refined the logic so it should display the button for linked rates regardless of status - did some fiddling with different statuses and linked/non-linked rates to make extra sure it works as expected, and thanks for the `ratePreviouslySubmitted ` call out! It was all I needed to make it work 🎉

services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
```diff
@@ -386,6 +386,20 @@ const RateDetails = ({
 
     const fieldNamePrefix = (idx: number) => `rateForms.${idx}`
 
+    const displayRemoveRateBtn = (index: number, rateForm: FormikRateForm) => {
+        if (index >= 1 && !displayAsStandaloneRate) {
+            //We expect to display the button regardless of status if the rate is a linked rate
+            if (
+                rateForm.ratePreviouslySubmitted === 'YES' ||
+                rateForm.status !== 'UNLOCKED'
```
JasonLin0991, line 394: **Question**: What is the purpose of checking the status of a child rate?

JasonLin0991 (approved):
🎊 Works from manual testing. I'll approve, but I have a question I left the comments, lets not merge until we discuss.

services/app-web/src/pages/StateSubmission/RateDetails/V2/RateDetailsV2.tsx
ruizajtruss: The main AC is to remove the 'remove this rate' button from unlocked rates -- that's why we check the status. So first we check whether or not it's a linked rate (since we want to display the button regardless of status). If it is a linked rate we display the button while ignoring the status check. However if it is NOT a linked rate then we go ahead and check the rate status to ensure the rate is not `UNLOCKED` in order to display the button.

PR #3346, opened 2025-05-12
Bump TLS version in CloudFront distro

mojotalantikite:
## Summary

This updates the minimum TLS version in our CloudFront distribution to match more modern versions. The quickstart we branched off of had an old setting here.

### Related Issues
https://jiraent.cms.gov/browse/MCR-5303


allly-b (approved):
LGTM (after googling what this does)

PR #3347, opened 2025-05-12
[MCR-5294] Add oauth client store

shamoons:
## Summary

This pull request introduces functionality for managing OAuth clients in the `services/app-api` service. It includes database schema updates, API methods for CRUD operations, and corresponding tests. Additionally, it reorganizes dependencies in the `package.json` file for clarity and consistency.

### OAuth Client Management

- **Database schema changes**:
  - Added a new `oauth_clients` table with fields for `id`, `clientId`, `clientSecret`, `grants`, `createdAt`, `lastUsedAt`, `description`, and `contactEmail`.
  - Created a unique index on `clientId` to enforce uniqueness.
  - Added the corresponding `OAuthClient` model in the Prisma schema.

- **API methods for OAuth client management**:
  - Implemented functions to create, retrieve (by ID and clientId), update, delete, and list OAuth clients.
  - Added logic to verify client credentials and update the `lastUsedAt` timestamp.

- **Unit tests**:
  - Comprehensive test coverage for CRUD operations.
  - Included tests for credential verification and edge cases like duplicates and non-existent client IDs.

### Dependency Reorganization

- **package.json updates**:
  - Reordered `@mc-review` dependencies alphabetically for better readability and consistency.

## Related issues

https://jiraent.cms.gov/browse/MCR-5294

## Screenshots

_N/A_

## Test cases covered

- CRUD operations for `OAuthClient`.
- Credential verification with valid/invalid `clientId` and `clientSecret`.
- Handling duplicates and non-existent client scenarios.
- Timestamp update on successful verification.

## QA guidance

1. Run migrations to ensure the `oauth_clients` table is created successfully.
2. Verify CRUD operations via API calls or DB queries.
3. Run the unit tests to confirm correctness.
4. Confirm dependency changes in `package.json` are non-functional (organizational only).


services/app-api/src/postgres/oauth/oauthClientStore.test.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    it('retrieves an OAuth client by ID', async () => {
+        const createdClient = await createOAuthClient(client, testClientData)
+        if (createdClient instanceof Error) throw createdClient
+
+        const retrievedClient = await getOAuthClientById(
+            client,
+            createdClient.id
+        )
+        expect(retrievedClient).not.toBeInstanceOf(Error)
```
JasonLin0991, line 51: **Nitpick:** I don't think we need the assertion here. The test would fail at this assertion before hitting the if statement. Leaving only the if condition will suffice to fail for errors and type guarding. The other tests have this pattern too.

services/app-api/prisma/schema.prisma
```diff
@@ -575,3 +575,16 @@ enum SubmissionType {
...
   CONTRACT_AND_RATES
 }
+
+model OAuthClient {
+  id            String    @id @default(uuid())
+  clientId      String    @unique
+  clientSecret  String
+  grants        String[]
+  createdAt     DateTime  @default(now())
+  lastUsedAt    DateTime?
```
JasonLin0991, line 585: **Question:** Is this `lastUsedAt` only updated when using the `verifyClientCredentials`? I took it as the last time this client was last updated or read. If `updateOAuthClient` doesn't trigger and update to this field, I wonder if we need a `updatedAt` field to record when OAuthClient records were updated.

JasonLin0991 (approved):
Looks good!

I left some non-blocking comments. I follow [conventional comments](https://conventionalcomments.org/) to add more clarification on the intent of my comments.

services/app-api/src/postgres/oauth/oauthClientStore.test.ts
shamoons: Fair point - I'll update the tests to not cascade on failure

services/app-api/prisma/schema.prisma
shamoons: I think the intent is to track, but we don't have tracking as a requirement. I'll update to a standard `updatedAt`

PR #3351, opened 2025-05-14
MCR-5149: complete undo withdraw api

JasonLin0991:
## Summary
[MCR-5149](https://jiraent.cms.gov/browse/MCR-5149)

Adds `undoWithdrawContract` API. This will allow a `WITHDRAWN` submission (contract and child rates) to be un-withdrawn.

- The Postgres function here is pretty simple
   - Validate contract is withdrawn by querying the contract's review status to check if latest action was `WITHDRAW`
      - The resolver also checks for `WITHDRAWN` status, but since this is a transaction function that could be use elsewhere we want to validate the status again.
   - Unlocked the contract with default withdraw text and append inputted reason.
   - Gather all draft rates with this contracts ID as the `parentContractID`
      - During a submission withdraw, linked child rates (child rate that is linked to another submission) are NOT withdrawn with the contract and are reassigned a new parent contract. 
      - At this point any rate that is still a child rate, was withdrawn with this contract. Our unlock Postgres function will automatically unlock them because of the parent child relationship.
   - Submit contract with default withdraw text and append inputted reason.
   - Add action type of `UNDER_REVIEW` for the contract
   - Loop through the rates we gathered after unlocking and add an action type of `UNDER_REVIEW`.
   - Returns the un-withdrawn contract along with the rates for the emails.
- The resolver work 
 - Performs user auth
 - contract status validations
 - OUT OF SCOPE: emails.
 
 AC:
When a submission is un-withdrawn
- The contract adds a new review action of action type UNDER_REVIEW
   - Revision unlock and submit info should contain a default message about it being withdrawn and include the inputted unwithdraw reason.
   - The default message should probably include rates unwithdrawn with this submission.
      - **NOTE**: We did NOT do this with submission withdraw, so following that pattern here
- Child rates also adds a new review action of action type UNDER_REVIEW
   - Revision unlock and submit info should contain a default message about it being withdrawn and include the inputted unwithdraw reason.
   - The default message should probably include contract unwithdrawn with this submission.
      - **NOTE**: We did NOT do this with submission withdraw, so following that pattern here.
      - Also the pattern we use for unlocking a submission is sharing the unlock and resubmit reason across the contract and child rates. Having a separate message while taking an action on an entire submission greatly deviates with the pattern.
 
#### Test cases covered

`undoWithdrawContract.test.ts` - resolver
- `'can undo a contract-only submission withdrawal'`
- `'can undo a contract and rate submission withdrawal'`
- `'can undo a submission withdrawal that had linked child rates'`
- `'can undo a submission withdrawal that has linked rates'`
- `'returns an error if contract is in incorrect status'`

## QA guidance

You can manually call this API using the GraphQL explorer tool.

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/undoWithdrawContract.ts
+++ b/services/app-api/src/postgres/contractAndRates/undoWithdrawContract.ts
@@ -1,0 +1,179 @@
+import type { ContractType } from '../../domain-models'
+import type { RateForDisplay } from './withdrawContract'
+import type { PrismaTransactionType } from '../prismaTypes'
+import type { ExtendedPrismaClient } from '../prismaClient'
+import { unlockContractInsideTransaction } from './unlockContract'
+import { submitContractInsideTransaction } from './submitContract'
+import { findContractWithHistory } from './findContractWithHistory'
+
+export type UndoWithdrawContractArgsType = {
+    contract: ContractType
+    updatedByID: string
+    updatedReason: string
+}
+
+export type UndoWithdrawContractReturnType = {
+    contract: ContractType
+    ratesForDisplay: RateForDisplay[]
+}
+
+/**
+ * Undo a withdrawn contract and rates withdrawn along with it.
+ *
+ * This function performs the following operations:
+ * 1. Validates the contract is currently withdrawn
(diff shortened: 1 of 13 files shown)


shamoons (approved):
LGTM!

PR #3354, opened 2025-05-16
Add document zip package model

mojotalantikite:
## Summary

This adds a new table to Prisma to hold references to the zip files for our documents. 

#### Related issues
https://jiraent.cms.gov/browse/MCR-5279


services/app-api/prisma/schema.prisma
```diff
@@ -577,15 +581,38 @@ enum SubmissionType {
...
+
+  rateRevisionID String?
+  rateRevision   RateRevisionTable? @relation(fields: [rateRevisionID], references: [id])
+
+  documentType DocumentZipType
+}
+
+enum DocumentZipType {
+  CONTRACT_DOCUMENTS
+  CONTRACT_SUPPORTING_DOCUMENTS
```
JasonLin0991, line 615: Are we planning to separate out the document packages? I know on the frontend, all contract documents are zipped into one package and all rates are zipped into one package. If not, should there be only 2 enum types here for all contract and all rate documents? Or we could add those 2 enums to these.

JasonLin0991 (approved):
✅

services/app-api/prisma/schema.prisma
mojotalantikite: Yeah good call, we can probably keep this the same then and I'll make this enum just `CONTRACT_DOCUMENTS` and `RATE_DOCUMENTS`

PR #3362, opened 2025-05-20
[DRAFT} Zip files on submit

mojotalantikite:
## Summary

This moves us to generating zip files when a submission comes in. It is currently done async on demand, but we were hitting performance limits in the AWS lambda environment as users were starting to make submissions with >700MB and/or hundreds of files. When `submitContract` or `submitRate` resolvers are used, we then generate the zip file and store it in the DB/S3 for later use. 

This doesn't include seeing it in the frontend UI.

#### Related issues
https://jiraent.cms.gov/browse/MCR-5280


services/app-api/src/postgres/documents/zip.test.ts
pearl-truss, line 1: Great test coverage!

services/app-api/src/resolvers/contract/submitContract.ts
```diff
@@ -448,3 +485,309 @@ export function submitContract(
...
+        submitContractResult.packageSubmissions[0]?.contractRevision
+
+    if (!contractRevision) {
+        return
+    }
+
+    // Only attempt to generate zip if there are actually documents to zip
+    if (
+        contractRevision.formData.contractDocuments &&
+        contractRevision.formData.contractDocuments.length > 0
```
pearl-truss, line 591: Should this be increased to > 1? I'm thinking if there's just 1 file, it doesn't need to be zipped either?

(same question for rates and supporting docs)

services/app-api/src/s3/s3Local.ts
```diff
@@ -106,5 +106,17 @@ export function newLocalS3Client(
...
+            documents: Array<{ s3URL: string; name: string; sha256?: string }>,
+            outputPath: string,
+            options = {
+                batchSize: 50,
+                maxTotalSize: 1024 * 1024 * 1024, // 1GB default limit
+                baseTimeout: 120000,
+                timeoutPerMB: 1000,
+            }
+        ): Promise<{ s3URL: string; sha256: string } | Error> => {
+            return new Error('not yet')
```
pearl-truss, line 119: I know this is still in draft, so flagging to see if intentional

pearl-truss (commented):
This is looking great! I did a look through of the code and left a few comments

pearl-truss (approved):
Gotcha on the `sha256` field. None of my other comments were blocking, great work!

services/app-api/src/s3/s3Local.ts
mojotalantikite: I ended up realizing this belonged elsewhere and shouldn't really be methods on our s3 client -- removed, thanks!

services/app-api/src/resolvers/contract/submitContract.ts
mojotalantikite: Yeah that's a good point. I worry that we'd end up having cases where we expect zip files in places in s3, but end up with single doc files. Or we have data in the table for zips that are just regular files. Maybe the move is to turn off the "download zip" button in the UI if there is only one file in the contract/rate/supporting docs section.

I think the easiest move is to just always zip, but I can add an RSP topic for us to talk about it.

PR #3373, opened 2025-05-27
MCR-4920: Submission undo withdrawal banner

pearl-truss:
#### Related issues
https://jiraent.cms.gov/browse/MCR-4920

#### Screenshots

Screenshot 2025-05-29 at 9 39 55 AM
Screenshot 2025-05-29 at 9 40 21 AM

#### Test cases covered
`submissionSummary.tsx`: `undo withdrawal submission tests`
- renders undo withdraw submission button
- render withdraw and other action buttons for a submission thats withdraw has been undone
- renders status update banner on undo submission

## QA guidance

CMS user -> after undoing a submission's withdrawal, the status updated banner should appear and be temporary. Switching routes, logging out, etc when you return to the submission summary page the banner should no longer appear

State user -> after a submission's withdrawal, the status updated banner should appear and be permanent until another action is taken


allly-b (approved):
Only one comment. Approving conditionally, given the ticket comments if changes need to be made, I can re-review. I also assumed the state user would see the same banner, without the additional info and I see your point that the banner with additional info seems like a more consistent UI for the state user especially. (I'll add the last two comments on the ticket too)

PR #3379, opened 2025-05-29
MCR-4963 Apollo Server: Migrate middleware (COR headers, HTTP body parsing)

shamoons:
## Summary

This pull request introduces a new Express-based Apollo Server handler as part of the preparation for migrating to Apollo Server 4. It includes updates to the `package.json` file to add necessary dependencies and implements a new handler file that sets up Apollo Server with middleware for CORS and body parsing. The handler is not yet used in production but provides a foundation for the migration.

### Dependency Updates

- Added new dependencies to `package.json` to support the Express-based Apollo Server setup:
  - `apollo-server-express`
  - `aws-serverless-express`
  - `cors`
  - `express`

### New Apollo Server Handler

- Introduced a new file, `apollo_gql_express.ts`, which sets up an Apollo Server instance using Express:
  - Configures middleware for CORS and JSON body parsing.
  - Integrates with existing services like Postgres, LaunchDarkly, and emailer.
  - Uses `aws-serverless-express` to wrap the Express app and export a Lambda-compatible handler.
  - Includes comments and placeholders for future enhancements (e.g. Apollo Server 4 compatibility, production readiness).

#### Related issues

- [MCR-4963](https://jiraent.cms.gov/browse/MCR-4963)

#### Screenshots

_N/A_

#### Test cases covered

- No test coverage included yet as the handler is not wired into production usage.


services/app-api/src/handlers/apollo_gql_express.ts
```diff
@@ -0,0 +1,144 @@
...
+// --- Apollo Server setup ---
+async function buildApolloServer() {
+    // These env vars and config mirror your current handler
+    const stageName = process.env.stage || 'local'
+    const dbURL = process.env.DATABASE_URL || ''
+    const secretsManagerSecret = process.env.SECRETS_MANAGER_SECRET || ''
+    const applicationEndpoint = process.env.APPLICATION_ENDPOINT || ''
+    const emailerMode = process.env.EMAILER_MODE || 'LOCAL'
+    const parameterStoreMode = process.env.PARAMETER_STORE_MODE || 'LOCAL'
+    const ldSDKKey = process.env.LD_SDK_KEY || ''
```
mojotalantikite, line 46: The majority of these env vars are required for anything to work, so I think we might want to do some validations that they're set rather than falling back to empty strings. We can probably just copy/paste those over from our current handler, or write a `validateEnvVars` function of some sort if we want to encapsulate it a little better in this refactor.

services/app-api/src/handlers/apollo_gql_express.ts
```diff
@@ -0,0 +1,144 @@
...
+        launchDarkly,
+        jwtLib,
+        dummyS3Client, // Satisfy type requirement
+        applicationEndpoint
+    )
+
+    // Context
+    // const userFetcher =
+    //   authMode === 'LOCAL' ? userFromLocalAuthProvider : userFromCognitoAuthProvider;
+    const contextForRequest = () => ({ user: { id: 'placeholder' } }) // Replace with your real context logic if needed
```
mojotalantikite, line 116: Do we have a ticket in this epic to bring over user auth, or do we want to finish that here?

mojotalantikite (approved):
Thanks for taking this on! Just called out a couple of things we might want to take care of before we land it.

services/app-api/src/handlers/apollo_gql_express.ts
shamoons: I'll add a ticket for that

PR #3385, opened 2025-06-02
Fix deploy and promote workflows

mojotalantikite:
## Summary

This fixes up the github actions workflow files for deploy and promote after the #3328 PR merged for aurora data exports.

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -184,6 +184,8 @@ jobs:
       subnet_public_a_id: ${{ secrets.DEV_SUBNET_PUBLIC_A_ID }}
       db_sm_arn: ${{ secrets.DEV_DB_SM_ARN }}
       aurora_arn: ${{ secrets.DEV_AURORA_ARN }}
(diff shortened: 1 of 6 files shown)


.github/workflows/promote.yml
```diff
@@ -184,6 +184,8 @@ jobs:
       subnet_public_a_id: ${{ secrets.DEV_SUBNET_PUBLIC_A_ID }}
       db_sm_arn: ${{ secrets.DEV_DB_SM_ARN }}
       aurora_arn: ${{ secrets.DEV_AURORA_ARN }}
+      prod_aws_account_id: ${{ secrets.PROD_AWS_ACCOUNT_ID }}
```
allly-b, line 187: Why is prod and val data being added to the dev promotion?

mojotalantikite: So we need the account ID of prod in order to make the IAM role for allowing prod to write to the s3 bucket in val. It's actually not really a secret, but we had started putting config values into secrets in github actions years ago -- we probably can refactor this pattern at some point.

PR #3391, opened 2025-06-04
MCR-5300: add logging and alerting for large api payload

pearl-truss:
## Summary

This PR adds logging and alerting for when a API request's payload exceeds 5.5mb

#### Related issues

https://jiraent.cms.gov/browse/MCR-5300

### DIFF
--- a/services/app-api/src/handlers/apollo_gql.ts
+++ b/services/app-api/src/handlers/apollo_gql.ts
@@ -365,7 +365,10 @@ const gqlHandler: Handler = async (event, context, completion) => {
     const initializedHandler = await handlerPromise
 
     const response = await initializedHandler(event, context, completion)
-
+    const payloadSize = Buffer.from(event.body).length
+    if (payloadSize > 5.5 * 1024 * 1024) {
+        console.warn(`Large payload detected: ${payloadSize} bytes`)
+    }
     // Log response size metrics without modifying the response
     if (response && response.body) {
         const bodySize = Buffer.from(response.body).length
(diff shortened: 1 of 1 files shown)


services/app-api/src/handlers/apollo_gql.ts
```diff
@@ -365,7 +367,20 @@ const gqlHandler: Handler = async (event, context, completion) => {
...
+    const payloadSize = Buffer.from(event.body).length
+    const serviceName = 'gql-handler'
+    const otelCollectorUrl = process.env.API_APP_OTEL_COLLECTOR_URL
+    if (otelCollectorUrl === undefined || otelCollectorUrl === '') {
+        throw new Error(
+            'Configuration Error: API_APP_OTEL_COLLECTOR_URL is required to run app-api'
+        )
+    }
+    initTracer(serviceName, otelCollectorUrl)
+    if (payloadSize > 5.5 * 1024 * 1024) {
```
JasonLin0991, line 379: Sorry, I didn't specify that the response payload is what we need to log and alert for sizes > 5.5mb. I think this block of code is for incoming payload size, which there is a limit, but I'm not sure what it is. 

I think we can keep this alerting for incoming payload size, lets keep it at 5.5mb, and add alerting for outgoing payload. There is a if condition on line 385 that sets up just the console logging, we could place the otel recording there. Please update the size to 5.5mb there too.

pearl-truss:
> I didn't see an alert in the `mc-review-alerts` channel for large incoming payload size. I saw the console.warn though.
> 
> Screenshot 2025-06-05 at 2 05 41 PM

@JasonLin0991 Would it warn for the slack channel if the issue is the dev environment? I'm not familiar with this and could be mistaken! I thought it would only do so in prod (and maybe val?)

PR #3397, opened 2025-06-09
MCR-5313: Ally - fix broken aria label errors

pearl-truss:


JasonLin0991 (commented):
Left some comments. I found it it really helpful to pair on these tickets with Anna. She has a wealth of knowledge.

Just just FYI, incase you didn't know. JAWS and Apple VO have noticeable differences when reading web pages. I believe Anna said most of our users use Jaws, so I've been doing most of my testing on it. There is an option to tunnel the JAWS virtual desktop to your local deployment. Here is the [guide](https://assistivlabs.com/support/assistivtunnel) using AsstivLabs. Sorry if this is old news to you, it was new to me last week lol.

PR #3399, opened 2025-06-09
MCR-5117: Text area component has no accessible name

JasonLin0991:
## Summary
[MCR-5117](https://jiraent.cms.gov/browse/MCR-5117)

When focusing on `textarea` of elements of the app, the screen reader does not read out the helper text above the `textarea`. The screen reader should read out the label, require (is applicable), and helper text.

Fix:
- The issue was that our `FieldTextarea` component setup the accessibility attributes incorrectly. Fixing that component fixed most of the `textarea` inputs.
- The withdraw/un-withdraw pages were not using the `FieldTextarea` component, after fixing the component and using it on these pages, the screenreader correctly reads out the helper text.

## QA guidance
- Using JAWS via [AsstivLabs](https://assistivlabs.com/) verify that focusing on `textarea` inputs read out label text -> required (if applicable) -> and the helper text on these pages.
   - Submission type form page
   - Contract details form page, after selecting "no" on the 438 question
   - Withdraw submission
   - Undo withdraw submission
   - Withdraw rate
   - Undo withdraw rate
- Note that the unlock and resubmit modals already read out the helper text. It does not use the `FieldTextarea` as the setup of those modals does not work well with `FieldTextarea`. We should consider refactoring those modals into pages like the withdraw/un-withdraw pages.

### DIFF
--- a/services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
+++ b/services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
@@ -778,9 +778,9 @@ export const SubmissionType = ({
                                             )}
                                             hint={
                                                 <>
-                                                    <p
+                                                    <span
                                                         id="submissionDescriptionHelp"
-                                                        role="note"
+                                                        className="margin-bottom-2"
                                                     >
                                                         Provide a 1-2 paragraph
                                                         summary of your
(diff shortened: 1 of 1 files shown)


pearl-truss (approved):
Gotcha, and I unintentionally had it on NVDA

PR #3402, opened 2025-06-10
[MCR-5206] Save as draft buttons should not redirect

ruizajtruss:
## Summary
This PR makes changes to the save as draft button to not redirect the user after being clicked
#### Related issues
[MCR-5206](https://jiraent.cms.gov/browse/MCR-5206)
#### Screenshots

https://github.com/user-attachments/assets/70ba2ae0-17d0-4d6d-b369-d5b81d8296a4

#### Test cases covered
- Just minor test changes

## QA guidance
- start a submission as a state user
- partially fill out forms page by page
- save as draft each time
- expect to see success banner each time
- exit form then revisit page you just saved to verify data was saved


services/app-web/src/pages/StateSubmission/SubmissionType/SubmissionType.tsx
```diff
@@ -823,16 +832,13 @@ export const SubmissionType = ({
                                             await handleFormSubmit(
                                                 values,
                                                 setSubmitting,
-                                                RoutesRecord.DASHBOARD_SUBMISSIONS
+                                                true
```
JasonLin0991, line 835: **Suggestion**: Consider another way to toggle when it is a save as draft. Keeping the redirection route in the `PageActions` component makes it easier if we had to change any type of routing because they are so close the dev can see them both.

Or do not hardcode the redirect url at all and use the `RoutesRecord.DASHBOARD_SUBMISSIONS` or the route generating functions so changes can be made at one place.

services/app-web/src/pages/StateSubmission/ReviewSubmit/V2/ReviewSubmit/ReviewSubmitV2.tsx
```diff
@@ -150,22 +149,7 @@ export const ReviewSubmit = (): React.ReactElement => {
                     explainMissingData
                 />
 
-                <PageActionsContainer
-                    left={
-                        <ActionButton
```
JasonLin0991, line 155: **Issue**: I know save as draft doesn't do anything here because there is no form inputs here, but we need a way to exit the submission. This was the `Save as draft` button, but I think we need to add a different button so users can exit. I would talk to design about what we should do here.

JasonLin0991 (commented):
Overall good work, there are some changes required before approval. Also 2 questions in the review that needs to be brought to design (see review for details).

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.tsx
```diff
@@ -151,7 +152,9 @@ export type FormError =
 export const ContractDetails = ({
     showValidations = false,
 }: ContractFormPageProps): React.ReactElement => {
-    const [shouldValidate, setShouldValidate] = React.useState(showValidations)
+    const [shouldValidate, setShouldValidate] = useState(showValidations)
+    const [draftSaved, setDraftSaved] = useState(false)
+    useFocusOnRender(draftSaved, '[data-testid="saveAsDraftSuccessBanner"]')
```
JasonLin0991, line 157: **Issue**: This focuses the banner the first time, but it does not refocus when clicking the `Save as draft` button again even after updating form data.

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.tsx
```diff
@@ -1308,8 +1324,7 @@ export const ContractDetails = ({
                                             values,
                                             setSubmitting,
                                             {
-                                                redirectPath:
-                                                    RoutesRecord.DASHBOARD_SUBMISSIONS,
+                                                type: 'SAVE_AS_DRAFT',
```
JasonLin0991, line 1327: **Question**: I forget, are we applying the same pattern for the `Save as draft` button on the Supporting documents page?

services/cypress/integration/stateWorkflow/stateSubmissionForm/rateDetails.spec.ts
```diff
@@ -24,11 +24,9 @@ describe('rate details', () => {
             cy.navigateFormByDirectLink(
                 `/submissions/${draftSubmissionId}/edit/rate-details`
             )
-
-            // Navigate to dashboard page by clicking save as draft
-            cy.navigateContractRatesForm('SAVE_DRAFT')
-            cy.findByRole('heading', { level: 1, name: /Submissions dashboard/ })
         })
+        cy.navigateContractRatesForm('SAVE_DRAFT')
```
JasonLin0991, line 28: **note**: Strange we moved both these assertions outside the `.then` of `cy.location()`. Was being inside the `.then()` causing failures or was it trying to find the dashboard header?

JasonLin0991 (approved):
Looking good, just need to fix the banner refocus and the supporting documents page (maybe).

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.tsx
ruizajtruss: fixed

services/app-web/src/pages/StateSubmission/ContractDetails/ContractDetails.tsx
ruizajtruss: yes

PR #3408, opened 2025-06-12
Fix export syntax in rescanFailedFiles lambda

mojotalantikite:
## Summary

When working through the file av rescanning the lambda threw ```"errorMessage": "Cannot redefine property: main"```. This just fixes the export syntax for the lambda runtime.


allly-b (approved):
LGTM

PR #3410, opened 2025-06-13
Add pagination, permissions and VPC to av re-scanner 

mojotalantikite:
## Summary

The lambda that I wrote to re-scan the files that were imported into our val bucket has been hitting issues around the number of files in the s3 bucket. This expands that handler to stream things in batches so we can step through > 1000 files at a time.
 
This also adds a missing lambda invocation permission and adds the `rescanWorker` to the VPC.  I've tested the permissions changes manually in `val` and everything runs as expected. This just adds the appropriate configs so they don't get deleted in the future.


allly-b (approved):
LGTM

PR #3411, opened 2025-06-13
MCR-5306: Upload response buttons are not unique for accessibility

JasonLin0991:
## Summary
[MCR-5306](https://jiraent.cms.gov/browse/MCR-5306)

`Upload response` button on both Contract and Rate Q&A pages have no unique accessibility labelling, meaning users using the screen reader cannot differentiate the buttons between questions on the screen.

Fix:
- The fix was to give each `Upload response` button a unique `aria-describedby` attribute linked to both the Q&A section and the question round title.
- To further make the buttons more uniquely accessible, we opted to add the question round number to the end of the Q&A table titles.

Frontend Changes
- We had to add rounds to the state Q&A table titles to make sure they are unique.
- State Q&A are now also collected by rounds in each section, exactly like CMS Q&A.
- Modified `QuestionDisplayTable` and `QuestionResponseRound` for accessibility on `Upload response` question button.

API Changes
- To get the round numbers on Rate Q&A, I modified the `questionResolver ` common resolver to be usable on both Contract and Rate questions.
- The GraphQL schema also added `round: Int!` to the `RateQuestion ` type.
- This should not change queries used by the 3rd party API Users, but the schema has changed.
-  @allly-b tagging you here for the API changes.

#### Screenshots

https://github.com/user-attachments/assets/249306f3-325d-455a-b2fb-a61b15cc0fef

#### Test cases covered
- No new test cases added, the refactor work is still validated with existing tests.

## QA guidance
- Test the Contract and the Rate Q&A page `Upload response` buttons.
   - With JAWS screen reader, make sure tabbing to the button announces the section and the question title.
   - With ANDI make sure the ANDI tool shows no errors under the `Links found` section when selecting the `link/buttons` drop down (see screen shot above)

### DIFF
--- a/services/app-web/src/pages/QuestionResponse/QATable/CMSQuestionResponseTable.tsx
+++ b/services/app-web/src/pages/QuestionResponse/QATable/CMSQuestionResponseTable.tsx
@@ -18,6 +18,63 @@ type CMSQuestionResponseTableProps = {
     userDivision?: Division
 }
 
+// Reusable function for sorting rounds by createdAt and reversing
+export const sortRoundsByDate = (rounds: QuestionRounds): QuestionRounds =>
+    rounds
+        .map((round) =>
+            round.sort(
+                (a, b) =>
+                    new Date(b.questionData.createdAt).getTime() -
+                    new Date(a.questionData.createdAt).getTime()
+            )
+        )
+        .reverse()
+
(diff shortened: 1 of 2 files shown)


pearl-truss (approved):
Changes look good to me. Will these API schema changes need to be flagged for SF and ARMS?

PR #3423, opened 2025-06-20
MCR-5390 Add user relationship to OAuth clients and rename table

shamoons:


services/app-graphql/src/schema.graphql
```diff
@@ -2553,6 +2553,7 @@ type OauthClient {
     contactEmail: String
     createdAt: DateTime!
     updatedAt: DateTime!
+    user: User!
```
JasonLin0991, line 2556: **Todo**: I believe we will need to update the queries and mutations for this new field. 

Lets change the `CreateOauthClientInput` to use the user ID instead of email, the ID is unique where as the email is not.
```
input CreateOauthClientInput {
    description: String
    userID: String!
    grants: [String!]
}
```

This will require some changes in the `createOauthClientResolver`

We also need to add the user to the OauthClient type

```
type OauthClient {
    id: ID!
    clientId: String!
    clientSecret: String!
    grants: [String!]!
    description: String
    contactEmail: String
    createdAt: DateTime!
    updatedAt: DateTime!
    user: CMSUsersUnion!
}
```

services/app-api/src/postgres/oauth/oauthClientStore.ts
```diff
@@ -28,6 +32,7 @@ export async function createOAuthClient(
                 grants,
                 description: data.description,
                 contactEmail: data.contactEmail,
+                userID: data.userID,
             },
```
JasonLin0991, line 36: Lets include the user data in the return.

```suggestion
            },
            include: {
               user: true
            }
```

services/app-api/prisma/schema.prisma
```diff
@@ -590,8 +591,8 @@ model OAuthClient {
   lastUsedAt   DateTime?
   description  String?
   contactEmail String
```
JasonLin0991, line 593: **Suggestion**: The email is in the the user data. I wonder if we need this anymore, unless we imagine the email here and the one tied to the user would be different. 

1. If so, this presents challenges to the UI as we would have figure out how to display the email as `user.email` could be different than `oauth.contactEmail` so associations between the client and user would be confusing. 

2. If we the `user.email` to save into `oauth.contactEmail`, then that would be work, there is a possibility of data drift between emails if the `user.email` changes.

3. The best option would be to remove this field and use the `user.email` instead.

My suggestion would be to do option #3 or #2 if it would be less work and faster to implement.

services/app-api/src/resolvers/oauthClient/createOauthClient.ts
```diff
@@ -16,7 +16,8 @@ export function createOauthClientResolver(
         const oauthClient = await store.createOAuthClient({
             grants: input.grants ?? undefined,
             description: input.description ?? undefined,
-            contactEmail: input.contactEmail ?? undefined,
+            contactEmail: input.contactEmail,
```
JasonLin0991, line 19: **Suggestion**: Depending on how we want to approach `oauth.contactEmail` in the schema, this would have to be changed.

If we go with option 2, copy email from the user, then before calling `createOAuthClient` I would fetch the user by the ID to validate the userID is a valid CMS user, then copy the email to the `contactEmail` argument.

JasonLin0991 (commented):
Thanks for the updates. I left some comments, some need addressed before approval.

PR #3432, opened 2025-06-24
MCR-5391 OAuth grants

shamoons:
## Summary
This PR implements a comprehensive OAuth grant-based authorization system that allows external OAuth clients to securely access MC Review data with read-only permissions. OAuth clients authenticate using the `client_credentials` grant type and inherit the permissions of their associated user while being restricted to read operations only.

Key Features:
- OAuth clients with client_credentials grant can read data but cannot perform write operations
- Permission inheritance: OAuth clients inherit state/CMS access permissions from their associated user
- Comprehensive authorization checks across all major GraphQL resolvers
- Enhanced JWT tokens that include user ID and grants information
- Robust error handling with OAuth-specific error messages

####  Related issues
[MCR-5391](https://jiraent.cms.gov/browse/MCR-5391): Implement OAuth grant-based permission checks for read-only API access


JasonLin0991 (commented):
Overall looks good. 

I can re-review again after some change, validate that the OAuth token works, and validate JWT still works.

JasonLin0991 (dismissed):
I'm still getting internal server errors trying to get the token. I checked the logs for `oauth_token` lambda.

Screenshot 2025-06-25 at 11 31 27 PM

I did notice that all the other lambdas that included the `otel` layer exported `main` differently than your `oauth_token.ts`. For example `audit_s3.ts`  exports `main` like 

```
module.exports = { main }
```

Lets try that? Maybe the export is conflicting with otel `main`

services/app-api/src/handlers/third_party_API_authorizer.ts
```diff
@@ -77,7 +81,8 @@ export const main: APIGatewayTokenAuthorizerHandler = async (
 
 const generatePolicy = function (
     userId: string | undefined,
-    event: APIGatewayTokenAuthorizerEvent
+    event: APIGatewayTokenAuthorizerEvent,
+    context?: Record<string, string>
```
pearl-truss, line 85: Could you add a comment explaining what `context` should be here. Also, if I'm understanding correctly, context has the `oathClient` info. What are the implications of having this be an optional field here? Do we need error handling or a fallback if context is missing?

JasonLin0991:
Closing this in favor of this [PR](https://github.com/Enterprise-CMCS/managed-care-review/pull/3451) where we had to start with a fresh DB

PR #3434, opened 2025-06-25
MCR-5340: Admin user can create new oauth client

JasonLin0991:


ruizajtruss:
Small note: I know the ticket doesn't specify whether the description is required but it might be worth asking design to be sure

JasonLin0991:
> Small note: I know the ticket doesn't specify whether the description is required but it might be worth asking design to be sure

There was no design input into this. We wanted to get this through fast using existing designs. The `description` field is not required in the request input or the database so I think it should be left blank.

PR #3437, opened 2025-06-26
Adjust doc download url expiration time in API

JasonLin0991:


pearl-truss (approved):
thanks for double checking!

PR #3446, opened 2025-06-30
Add document migration lambda

mojotalantikite:
## Summary
This adds a lambda that goes through all the revisions in the database and generates document zips for the submission if they haven't been generated yet. We can run it based on state, with a dry run, or with a full run against everything.

#### Related issues
https://jiraent.cms.gov/browse/MCR-5281


JasonLin0991:
For visibility. @Mojo and I synced over zoom about this PR and I had a couple of concerns:
- Prod and Val have a large amount of documents, we may hit the 15 minute timeout. The code looks resumable, but what happens if it times out during a revision?
   - Mojo will add code to account for the timeout and stop the migration before it so we do not have partial migrations.
- Last was about memory concerns, since we are query all the revisions and all of its data is running out of memory a concern?
   - Mojo said the memory allocation was something like 10gb, we should be well under thay.

JasonLin0991 (approved):
Looks good! Thanks for those changes.

PR #3476, opened 2025-07-09
Refactor document zip as resolver dependency

JasonLin0991:
## Summary

While working on dependabot updates, I noticed a lot of our tests were timing out at > `25000` ms. Figured out it was because unit tests and local deployment used the same code as `dev`, `val` and `prod` deployed environments. This meant in unit tests and local deployments it would try to perform actual document zip by calling S3. Errors because it was not configured and increased unit test times.

The work I did here was to refactor this code as a dependency for our resolvers and creating a local version of `generateDocumentZip`. This eliminates the errors in the API and increased unit test time.

## QA guidance

- Make sure zips still generate in the review app.

## PR Reminders

### DIFF
--- a/services/app-api/src/handlers/apollo_gql.ts
+++ b/services/app-api/src/handlers/apollo_gql.ts
@@ -36,6 +36,11 @@ import {
 } from 'apollo-server-core'
 import { newDeployedS3Client, newLocalS3Client } from '../s3'
 import type { S3ClientT, S3BucketConfigType } from '../s3'
+import {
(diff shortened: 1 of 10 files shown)


pearl-truss (approved):
Great work Jason! I really like the idea of having a mocked local doc generator

mojotalantikite:
Oh awesome, great refactor! I had been trying to just intercept the s3 calls in the tests, but also doing the zips is even better.

PR #3483, opened 2025-07-10
Debug OIDC in promote

mojotalantikite:
## Summary
Promotes are blocked right now as all of a sudden the ARN for the OIDC in val is not getting the appropriate AWS ID from the github action config. This is weird, as nothing has changed on our end in regards to that configuration value.

This adds some debug output to promote so we can see a little bit more of what's going on.


mojotalantikite:
well, nevermind, it was a bug in github actions -- deleting the value that wasn't getting passed in correctly and re-creating it fixed it.

PR #3507, opened 2025-07-17
Upgrade for eslint 9.x and typescript 5.x

mojotalantikite:
## Summary

Looks like dependabot upgraded some `eslint` things that were incompatible with our repo's setup. This brings us from the 8.x line of `eslint` to the 9.x line. It required moving us to `typescript` 5.x for the bundler options. In doing this I decided to move us to the 5.1 line of `typescript` as it was the one with the least amount of breaking changes from `4.9.5` while also getting us to a version that `prisma` will require when we upgrade it. We can do a follow on later to get us to the most recent 5.x version.

#### Related Issues
https://jiraent.cms.gov/browse/MCR-4769
https://jiraent.cms.gov/browse/MCR-3148


pearl-truss (commented):
I'm still seeing the module not found locally when trying to commit. I test changes to the `app-web` service. I deleted and reinstalled the packages in node_modules, if there's anything else to try just lmk

PR #3509, opened 2025-07-18
[MCR-5366] view medicaid populations in rate summaries

ruizajtruss:
## Summary
State sees Medicaid populations in rate summary; CMS sees it in submission and rate review summaries

Description:
Update the Rate Details section in the Review submission page and the Submission Summary view. This section also appears in the rate review page. to display the Medicaid populations field when applicable. This should be visible to both state and CMS users on the submission summary page, and also to CMS users on the rate review summary screen.

[Figma Link](https://www.figma.com/design/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?node-id=20556-27635&t=FxrZEEpA111At3WU-4)

Acceptance Criteria:

Section appears in review submit page and submission summary view for all users
Section appears in rate review summary page
Section label: “Medicaid populations included in this rate certification”
Value displayed:
Medicare-Medicaid dually eligible individuals enrolled through a Dual-Eligible Special Needs Plan (D-SNP)
Medicaid-only
Medicare-Medicaid dually eligible individuals not enrolled through a D-SNP
Field only displays if contract was marked “Yes” for D-SNP
Follow the pattern set above on the rate review page and the submission summary view
#### Related issues
[MCR-5366](https://jiraent.cms.gov/browse/MCR-5366)
#### Screenshots
Review and submit
Rate details
State summary
MCR-MN-MSC+-PMAD-20250703-20250710-CERTIFICATION-20250703
Cms summary
Pasted Graphic 4
#### Test cases covered
1. added testing to check for new field on summary pages

## PR Reminders


services/app-web/src/pages/StateSubmission/ReviewSubmit/RateDetailsSummarySection.test.tsx
```diff
@@ -704,6 +710,35 @@ describe('RateDetailsSummarySection', () => {
...
+                submissionName="MN-PMAP-0001"
+                statePrograms={statePrograms}
+            />,
+            {
+                apolloProvider: apolloProviderCMSUser,
+                featureFlags: { dsnp: true },
+            }
+        )
+
+        await waitFor(() => {
```
pearl-truss, line 727: @ruizajtruss I see what you're saying with this test passing locally and failing in CI. From the output here, it looks like the test is running its assertions before the page has had a full chance to load. If you add an await for the page heading to be populated that should hopefully fix this

```
const rateName =
            'MCR-MN-0005-SNBC-RATE-20221014-20221014-CERTIFICATION-20221014'

        expect(screen.getByText(rateName)).toBeInTheDocument()
```

The[ test on line 324](https://github.com/Enterprise-CMCS/managed-care-review/blob/b7aff2374678223eb172ed47482a87d186ca8ded/services/app-web/src/pages/StateSubmission/ReviewSubmit/RateDetailsSummarySection.test.tsx#L324) '' is a good example and shows how to set the rate name

packages/mocks/src/apollo/rateDataMock.ts
```diff
@@ -38,6 +38,11 @@ const rateRevisionDataMock = (data?: Partial<RateRevision>): RateRevision => {
         formData: {
             rateType: 'AMENDMENT',
             rateCapitationType: 'RATE_CELL',
+            rateMedicaidPopulations: [
```
pearl-truss, line 41: This change would apply to a ton of tests (including the RateDetails test that had the failure) since this inside of a reusable test mock. I don't think we want to set `rateMedicaidPopulations` here because this field is would only be populated when the `dsnp` flag is on. The [test branch](https://github.com/Enterprise-CMCS/managed-care-review/compare/test-rate-medicaid-populations?expand=1#diff-1af1a8a6b4ce16a4da87c355122f0f49f518ed20015036035e049daa8971851aR333) I created shows an example of setting `rateMedicaidPopulations` directly in the `rateDetailsSummarySection` test

I still don't see exactly why this caused the test failure we were seeing earlier. I'm curious what @allly-b thinks as far as how much more engineering time to spend on this? My sense is this is

packages/mocks/src/apollo/contractPackageDataMock.ts
```diff
@@ -474,6 +474,11 @@ function mockContractPackageDraft(partial?: Partial<Contract>): Contract {
                     formData: {
                         __typename: 'RateFormData',
                         rateType: 'AMENDMENT',
+                        rateMedicaidPopulations: [
```
pearl-truss, line 477: Same as what I commented on on the rateDataMock file

PR #3605, opened 2025-08-27
MCR-5418: Migrate existing documents to populate "dateAdded"

pearl-truss:
## Summary

This PR updates the contract and rate documents that exists in the db without a dateAdded field. The dateAdded field is set to equal to the `submitInfo.updatedAt` for its associated Contract/Rate revision. This matches the logic in `setDateAddedForContractRevisions` and `setDateAddedForRateRevisions`

AC
- Migrate the following existing documents in DEV, VAL , and PROD:
  - ContractDocument
  - ContractSupportingDocument
  - RateDocument
  - RateSupportingDocument
- Validate that all submitted documents have dateAdded

#### Related issues

https://jiraent.cms.gov/browse/MCR-5418

## QA guidance

There's a few ways to test this I'll just outline my approach:

- Pull data from val or prod locally using[ jumpbox](https://github.com/Enterprise-CMCS/managed-care-review/blob/main/docs/technical-design/howto-migrations.md#how-to-dump-deployed-data-for-local-testing)
- Open a DB viewer (I use Beekeeper) and first select the documents to see that some have null dateAddded fields
- Apply the migration
- Select the documents again and ensure that the dateAdded fields are populated as expected (should match the updatedAt field of the submitInfo for the related contract or rate revision)
- For each table you can run a query like this to ensure that there are no lingering documents that are missing a dateAdded field after submission

```sql
SELECT cd.*
FROM "ContractDocument" cd
JOIN "ContractRevisionTable" cr ON cd."contractRevisionID" = cr."id"
WHERE cr."submitInfoID" IS NOT NULL
  AND cd."dateAdded" IS NULL;
```

## PR Reminders

### DIFF
--- a/services/app-api/prisma/migrations/20250827184558_populate_date_added_column_for_contract_and_rate_documents/migration.sql
+++ b/services/app-api/prisma/migrations/20250827184558_populate_date_added_column_for_contract_and_rate_documents/migration.sql
@@ -1,0 +1,35 @@
+BEGIN;
+
+UPDATE "ContractDocument" cd
+SET "dateAdded" = ui."updatedAt"
+FROM "ContractRevisionTable" cr
+INNER JOIN "UpdateInfoTable" ui ON cr."submitInfoID" = ui."id"
+WHERE cd."dateAdded" IS NULL
+  AND cd."contractRevisionID" = cr."id"
+  AND cr."submitInfoID" IS NOT NULL;
+
+UPDATE "ContractSupportingDocument" csd
+SET "dateAdded" = ui."updatedAt"
+FROM "ContractRevisionTable" cr
+INNER JOIN "UpdateInfoTable" ui ON cr."submitInfoID" = ui."id"
+WHERE csd."dateAdded" IS NULL
+  AND csd."contractRevisionID" = cr."id"
+  AND cr."submitInfoID" IS NOT NULL;
+
+UPDATE "RateDocument" rd
+SET "dateAdded" = ui."updatedAt"
+FROM "RateRevisionTable" rr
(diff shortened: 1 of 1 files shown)


services/app-api/prisma/migrations/20250827184558_populate_date_added_column_for_contract_and_rate_documents/migration.sql
```diff
@@ -0,0 +1,35 @@
+BEGIN;
+
+UPDATE "ContractDocument" cd
```
JasonLin0991, line 3: **Issue**: The migration does not retain the initial date the document was submitted. The example below is from prod data on local deployment. The document is the same for each submission revision, left is initial and right is latest, the date added is the submitted at date for the revision and not when this document was first submitted.

**Suggestion**: The migration will have to be more complex because we will have to look for the earliest sha256 of a document and get its submitInfoID and apply it to the rest of the document records with the same sha256. We also might want this to be a migration we manually apply to the DB's or a lambda we can call for each.

Screenshot 2025-08-28 at 1 07 29 PM

JasonLin0991 (changes requested):

services/app-api/prisma/migrations/20250827184558_populate_date_added_column_for_contract_and_rate_documents/migration.sql
pearl-truss: @JasonLin0991 this has been addressed in a new PR https://github.com/Enterprise-CMCS/managed-care-review/pull/3619

PR #3619, opened 2025-09-02
MCR-5418: Migration for populating dateAdded field for contract and rate documents

pearl-truss:
## Summary

This PR updates the contract and rate documents that exists in the db without a dateAdded field. The dateAdded field is set to equal to the `submitInfo.updatedAt` for its associated Contract/Rate revision. This matches the logic in `setDateAddedForContractRevisions` and `setDateAddedForRateRevisions`

AC
- Migrate the following existing documents in DEV, VAL , and PROD:
  - ContractDocument
  - ContractSupportingDocument
  - RateDocument
  - RateSupportingDocument
- Validate that all submitted documents have dateAdded

#### Related issues

https://jiraent.cms.gov/browse/MCR-5418

## QA guidance

Best to use prod data locally (ids before are pulled from prod data)

```sql
SELECT 
  cd."id", 
  cd."updatedAt" AS contractDocUpdatedAt, 
  cd."dateAdded" AS contractDocDateAdded, 
  ui."updatedAt" AS contractRevisionSubmittedAt, 
  cd."contractRevisionID",
  cr."contractID"
FROM "ContractDocument" cd
JOIN "ContractRevisionTable" cr ON cd."contractRevisionID" = cr."id"
JOIN "UpdateInfoTable" ui ON cr."submitInfoID" = ui."id"
WHERE cd."sha256" = '7d48ae6de8e210d600d0c1ecb93749588894afd120be5986041519c9f46e3a12'
ORDER BY cr."contractID", cd."id";
```
Run this query or similar prior to running the migration. Choose a sha that is tied to multiple documents across different contract submissions

Screenshot 2025-09-02 at 12 59 21 PM

Notice the dateAdded fields are null. 

Run the migration and then rerun the same query. Expect to see 1) the dateAdded field is populated based on the earliest date from the submitInfo of any of the contract revisions associated with the same contract 

Screenshot 2025-09-02 at 2 45 24 PM

There are also cases to check where the document already has a dateAdded value. This entries should be skipped and not affected by the migration

With prod data, document with id `53e98919-35b7-4156-ab16-12bfa56fa794` is an example of a contract document that already has the dateAdded field supplied 

## PR Reminders


JasonLin0991 (approved):
Great work!

I validated with prod data locally.
- Validated that all documents related with a submitInfo has `dateAdded`
- Validated some submissions have correct date added dates for documents through several revisions.
- Validated linked rates show correct date added for its documents.

There is one possible issue. There are unlocked submissions in prod that when viewing as a state user, the unlocked submission documents have `dateAdded` dates. This should not be the case as the `draftRevision` contains copies of existing documents, but the `dateAdded` should not be populated yet. The dates are correct though and any updates to the unlocked submission will remove the `dateAdded`.

Screenshot 2025-09-03 at 1 23 31 PM

It seems like there weren't any issues trying to resubmit this or update it. Also unlocking submissions after the migration does not have this issue. I'm not sure exactly whats going on here. Maybe the migration is populating documents related to draftRevisions with dateAdded when it should not be.

I'll approve and let you make the call to see if you want to fix that or leave it be.

PR #3622, opened 2025-09-03
MCR-5506 add document id to type document

JasonLin0991:
## Summary
[MCR-5506](https://jiraent.cms.gov/browse/MCR-5506)

We have a fetchDocument endpoint that takes a document id and retrieves that document. However, we have two types used for document objects: Document and Generic Document. Document is used for Question and answer docs and doesn't have document id, but Generic Document has document id.

AC:
- The Document type in the graphQL schema, and domain types for the QuestionsType and QuestionResponseType have an ID field
- Q&A graphql fragments return an id for the document so that the FetchContractWithQuestions endpoint return documents with IDs 
   - contractQuestionEdgeFragment
   - rateQuestionEdgeFragment
   - questionResponseFragment

## PR Reminders

### DIFF
--- a/services/app-graphql/src/schema.graphql
+++ b/services/app-graphql/src/schema.graphql
@@ -1314,7 +1314,7 @@ Document represents a document that has been uploaded to S3. It can be retrieved
 by an authenticated user.
 """
 type Document {
-    id: ID!
+    id: ID
     name: String!
     s3URL: String!
     downloadURL: String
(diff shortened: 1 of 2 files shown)


pearl-truss (approved):
Changes look good and I see the ids in the fetchContract response!

services/app-graphql/src/fragments/documentFragment.graphql
pearl-truss, line 1: Love the new fragments!

PR #3643, opened 2025-09-09
MCR-5516: Migration for removing whitespaces from user emails

pearl-truss:
## Summary

This PR makes a small migration that finds and removes any whitespace in user emails. I confirmed that this migration will only affect emails in VAL and that PROD didn't have any emails containing whitespaces

#### Related issues

https://jiraent.cms.gov/browse/MCR-5516

#### Screenshots
Screenshot 2025-09-09 at 10 17 17 AM

## QA guidance

I tested this with local VAL data

`select email from "User" WHERE email LIKE '% %';` returns 0 rows affects after running the migration

## PR Reminders


JasonLin0991 (approved):
✅ LGTM!

PR #3652, opened 2025-09-11
[MCR-3954] Remove avDownload lambdas

ruizajtruss:
## Summary
We no longer need the antivirus definitions download lambda and associated s3 bucket, as the clamavd server runs `freshclam` every hour to update virus definitions. We can delete all the code that is related to keeping AV definitions up to date.

Definition of Done:
- All AV definition download/update code is deleted from the `uploads` service
- All tests are still working

#### Related issues
[MCR-3954](https://jiraent.cms.gov/browse/MCR-3954)
## QA guidance
- Uploads should work as intended and the bucket should no longer be present in AWS

## PR Reminders


mojotalantikite (changes requested):
Awesome, thanks for taking care of this cleanup! I think there are a couple things left in the `serverless.yml` for `uploads` that deploy the av definitions bucket -- I think we can remove the `ClamDefsBucket` and the policy that goes with it too.

mojotalantikite (approved):

PR #3663, opened 2025-09-12
AWS CDK for Review Environments

mojotalantikite:
## Summary

This adds the AWS CDK deploy to review environments. We've been running on Serverless since the start of this project and this refactors all of our Serverless deploys into AWS CDK and stands it up with the `deploy-cdk.yml` GitHub Actions workflow file in ephemeral review environments. This includes a promote to dev, but not val or prod.

We would like to evaluate using this CDK deploy in parallel to the Serverless deploy in the review environments so that our eng team can discover any issues with the deployment. Once we get about a week of usage out of it, we'll do a follow on to remove the parallel Serverless deploy for the review environments.

#### QA Guidance

Please take a look at the deployed application and make sure it looks ok. Cypress tests are all passing, which gives me some confidence, but it'd be good to have some extra eyes on the deployed app.

#### Related Issues
Many issues from this epic: 
https://jiraent.cms.gov/browse/MCR-4352


mojotalantikite:
closing for #3728

PR #3705, opened 2025-09-30
MCR-4193: remove all refs to protobufs in documentation

JasonLin0991:
## Summary
[MCR-4193](https://jiraent.cms.gov/browse/MCR-4193)

AC:
- Go through all Github tech design docs, database diagrams, etc and make sure no outdated references to protobufs remain. Files can be deleted
- For ADRs - we should scan if any remaining ADRs exist that supposed protobufs in use- supercede them with new ADRS describing the new decisions made
   - The only ADR that needs a superseding ADR is `018-chagning-form-over-time.md`. I don't know what the decision was after we removed protobufs so there will need some investigatin, [conversations, and  digging up notes to make this ADR. I think it would be out of scope for this PR, so I made ](https://jiraent.cms.gov/browse/MCR-5582)MCR-5582 to capture this work.

## PR Reminders


pearl-truss (approved):
LGTM and thanks for making the other ticket!

PR #4290, opened 2026-04-03
MCR-6116: Remove generating Apollo client hooks in codegen

JasonLin0991:
## Summary
[MCR-6116](https://jiraent.cms.gov/browse/MCR-6116)

Apollo Client v4 introduces TypeScript improvements to its `useQuery` and `useMutation` hooks — including function overloads for type narrowing and required variables enforcement. These improvements require the hooks to be called directly from Apollo Client, not through generated wrapper hooks. Our             
  codegen generated hooks (from `@graphql-codegen/typescript-react-apollo`) cannot provide these overloads, making them incompatible with v4.

To unblock upgrading to v4, this PR:
- Replaces the typescript-react-apollo codegen plugin with typed-document-node, which generates TypedDocumentNode constants instead of hook wrappers
- Migrates all call sites in app-web to use useQuery(XxxDocument) / useMutation(XxxDocument) directly — types are inferred automatically from the TypedDocumentNode

This change is compatible with our current Apollo Client v3.8.8 and requires no behavioral changes.

## QA guidance

- The unit tests and Cypress tests validate the hooks still work. But use the app and see if there are issues.


adriannekennedy (approved):
Lg. I tested submission, resubmission, withdraw rate, undo withdraw rate, withdraw contract, and release to state across EQRO and Health Plans and saw no issues. Also did a round of contract questions.

PR #4314, opened 2026-04-09
MCR-6040 Design Review/Feedback

adriannekennedy:
## Summary
Anna noticed that the spacing in between Change History accordions was off. This is only noticed on this branch - where we changed the heading to h5 and applied some styles to make sure it had the heading 5 look. This PR addresses that by removing the margin: 0 and letting the existing styles for margin inherit.
The other reported issue was around the padding of the text - but I'm correctly seeing 16px above, 20 px on left and right, and 12px below (2nd screenshot)

#### Related issues
https://jiraent.cms.gov/browse/MCR-6040

#### Screenshots
image

image

#### Test cases covered
N/A - styling update only, no test cases affected

## QA guidance
Review a submission that has multiple change history sections and confirm spacing/padding of the change history accordions 

## PR Reminders

### DIFF
--- a/services/app-web/src/components/ChangeHistory/ChangeHistory.module.scss
+++ b/services/app-web/src/components/ChangeHistory/ChangeHistory.module.scss
@@ -17,7 +17,6 @@
     }
     h5[class*='usa-accordion__heading'] {
         @include custom.mcr-h5-style;
-        margin: 0;
     }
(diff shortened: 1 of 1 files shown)


JasonLin0991 (approved):
LGTM based on her comments.

PR #4315, opened 2026-04-09
MCR-6071 Dashboard Table Styles

adriannekennedy:
## Summary
Anna saw some table issues with the spacing and width of columns. After meeting with her, these are the changes I made. Note, I converted the pixels to rems for the table widths so they'd scale better when user's adjust the browser zoom.

- Anna liked the 16pt padding all around when we paired, so made the change to keep that 
- Programs column should have a max-width of ~143px (Note: adjusted the width for the best responsive view for State dashboard, although it's less responsive for CMS b/c of the extra columns)
- The ID column should have a minimum width of ~90px and max width of ~224px on both dashboards
- Add fixed width of ~130px to contract type column on both CMS and State dashboards

Additionally, I adjusted all the column widths to rem instead of pixels. Also added css to prevent the Not Subject to Review tag from wrapping in desktop view, but remove that nowrap when the screen is sized down. 

Global padding styles are shared with Rate Review, as well as the ID and Programs column widths. The rest of the changes were specifically for submissions dashboards (both state and cms). 

#### Related issues
https://jiraent.cms.gov/browse/MCR-6071
https://jiraent.cms.gov/browse/MCR-6069
https://jiraent.cms.gov/browse/MCR-6070

#### Test cases covered
N/A - only styling changes that did not affect test cases

## PR Reminders

### DIFF
--- a/services/app-web/src/components/ContractTable/ContractTable.module.scss
+++ b/services/app-web/src/components/ContractTable/ContractTable.module.scss
@@ -32,8 +32,8 @@
     tbody td {
         padding-top: uswds.units(2);
         padding-bottom: uswds.units(2);
-        padding-left: uswds.units(1);
-        padding-right: uswds.units(1);
+        padding-left: uswds.units(2);
+        padding-right: uswds.units(2);
         vertical-align: top;
     }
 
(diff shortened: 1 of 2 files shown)


vyakusha (approved):
LGTM Great work!

PR #4317, opened 2026-04-09
MCR-6114: Update state users email address

JasonLin0991:
## Summary
[MCR-6114](https://jiraent.cms.gov/browse/MCR-6114)

A user updated her email in IDM but is still seeing the old email in state portal. We want to update the email for that user in our database.

Instead of making a direct migration in the db instance via SQL commands, we opted to refactor the API to update the user information in our database if it had changed.

Notes:
- The refactor will update the database user record if there is a difference from the Cognito user data. 
- There is a fallback if Cognito returns an error we use the user data from the database if it exists.
- If both fails, when we return an error response.

Questions:
- @mojotalantikite I think this refactor queries Cognito on every request, will that be an issue?.

AC:
- Reactor API to update user information in our database if any user information had changed.

#### Test cases covered

- `fetchCurrentUser.test.ts`
   - updates DB when user info changes

- `cognitoAuthn.test.ts`
  - inserts a new user into the DB from Cognito
  - updates DB when Cognito user info has changed
  - updates DB when only role changes
  - falls back to DB user when Cognito fails for existing user
  - returns error when Cognito fails and no DB user exists
  - returns error when Cognito returns no matching user and no DB user exists

## QA guidance


services/app-api/src/authn/cognitoAuthn.ts
```diff
@@ -212,57 +293,32 @@ export async function userFromCognitoAuthProvider(
...
-        if (auroraUser instanceof Error) {
-            return auroraUser
-        }
-        const endRequest = performance.now()
-        console.info(
-            'User lookup in postgres takes ms:',
-            endRequest - startRequest
+        const cognitoUserResult = await lookupUserCognito(
+            userInfo.userId,
+            userInfo.poolId
```
mojotalantikite: @JasonLin0991 Hmm, yeah, this is probably going to add at least 100ms to every request in our API. We might want to talk about this in RSP along with those two Notion docs I posted about how to refactor Users considering the SF/Mulesoft OBO token work.

JasonLin0991: Another thing we can do for now is only update user data once per day.

services/app-api/src/authn/cognitoAuthn.ts
```diff
@@ -195,6 +195,87 @@ export function userTypeFromAttributes(
     return new Error('Unsupported user role:  ' + roleAttribute)
 }
 
+function hasUserInfoChanged(sourceUser: UserType, dbUser: UserType): boolean {
+    if (
+        sourceUser.email !== dbUser.email ||
+        sourceUser.givenName !== dbUser.givenName ||
+        sourceUser.familyName !== dbUser.familyName ||
+        sourceUser.role !== dbUser.role
+    ) {
```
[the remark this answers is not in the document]
JasonLin0991: divisionAssignment is not in Cognito. This value is set in app.

services/app-api/src/authn/cognitoAuthn.ts
```diff
@@ -212,57 +293,32 @@ export async function userFromCognitoAuthProvider(
...
-                role: cognitoUser.role,
-                givenName: cognitoUser.givenName,
-                familyName: cognitoUser.familyName,
-                email: cognitoUser.email,
-            }
-
-            // if it is a state user, insert the state they are from
-            if (cognitoUser.role === 'STATE_USER') {
-                userToInsert.stateCode = cognitoUser.stateCode
+        if (cognitoUserResult instanceof Error) {
```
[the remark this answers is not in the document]
JasonLin0991: Checking only once per day now.

services/app-api/src/postgres/user/updateUserInfo.ts
```diff
@@ -0,0 +1,44 @@
...
+import type { ExtendedPrismaClient } from '../prismaClient'
+import type { UserRoles } from '../../domain-models/UserType'
+
+export type UpdateUserInfoArgsType = {
+    email: string
+    givenName: string
+    familyName: string
+    role: UserRoles
+    stateCode?: string
+}
```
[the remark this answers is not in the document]
JasonLin0991: divisionAssignment not in Cognito. This value is set in app.

services/app-api/src/authn/cognitoAuthn.test.ts
```diff
@@ -205,3 +211,236 @@ describe('cognitoAuthn', () => {
         })
     })
 })
+
+// Mock the Cognito SDK
+const mockSend = vi.fn()
+vi.mock('@aws-sdk/client-cognito-identity-provider', () => ({
+    CognitoIdentityProviderClient: vi.fn(() => ({ send: mockSend })),
+    ListUsersCommand: vi.fn(),
+}))
```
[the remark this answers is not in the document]
JasonLin0991: Hoisted

services/app-api/src/authn/localAuthn.ts
```diff
@@ -13,51 +17,20 @@ export async function userFromLocalAuthProvider(
...
 
-        if (auroraUser instanceof Error) {
-            return auroraUser
+        if (dbUser instanceof Error) {
+            return localUser
+        }
+
+        // Mimics the logic for syncing cognito user
+        if (!dbUser || !wasUpdatedToday(dbUser.updatedAt)) {
+            return syncUserWithAurora(store, localUser)
```
mojotalantikite: do we want to pass the `dbUser` here too? If it was updated the previous day it looks like we might take the `insert` path in `syncUserWithAurora`.

services/app-api/src/postgres/user/updateUserInfo.ts
```diff
@@ -0,0 +1,44 @@
...
+                    orderBy: {
+                        stateCode: 'asc',
+                    },
+                },
+            },
+        })
+
+        return domainUserFromPrismaUser(updateResult)
+    } catch (err) {
+        return err
```
mojotalantikite, line 42: This could be `any`, and generally I think we tend to check if we're returning an `Error` type here -- I guess we're not forcing the `catch` blocks to do that though with the typescript compiler. Not a huge deal, but I hadn't noticed that we weren't linting on that before.

mojotalantikite (approved):
cool lgtm -- thanks for the tests!

services/app-api/src/postgres/user/updateUserInfo.ts
JasonLin0991: There seems to be a lot of functions in `src/postgres/` that are not type checking this. I created a ticket for making all of them have it https://jiraent.cms.gov/browse/MCR-6189.

PR #4333, opened 2026-04-14
MCR-6100: Block submission of deprecated programs at api level

JasonLin0991:
## Summary
[MCR-6100](https://jiraent.cms.gov/browse/MCR-6100)

This work prevents submitContract from submitting or resubmitting `HEALTH_PLAN` and `EQRO` submissions when contract or child rates (only for `HEALTH_PLAN`) contain a deprecated program.

Acceptance criteria:
- Confirm that we do not need to make changes to createContract related to rejecting requests that include deprecated program IDs
   - **Confirmed**: On new submission, deprecated programs will not be available as a selection.
- Return an error when trying to submit or resubmit a submission with a deprecated program name
   - Includes contract and child rate programs
- Unit tests for each

Possibility unnecessary AC:
- updateContractDraftRevision allows deprecated program IDs only if they were already part of the existing draft's programIDs
   - What is the use case here? I would think we would want them to not use deprecated programs. Also they would still hit an error at submit.
   - Form the UI perspective if we restricted deprecated from being updated at the API level, the UI would not surface a descriptive user error.
- updateDraftContractRates rejects deprecated rateProgramIDs on new rate creation, but allows them on rate updates only if already present on the existing rate
   -  I think this should be handled at the UI level on the rate details page.  New rates would not see the deprecated programs anyways. Trying to figure out the scenario were this would happen.
   - Form the UI perspective if we restricted deprecated from being updated at the API level, the UI would not surface a descriptive user error.

Suggestion: I think we need a new ticket for contract details, rate details and review submit pages that display inline errors and error summaries when users are trying to continue or submit with a deprecated program.
 - Contract details & Rate Details
   - On `Continue` display inline error and error summary when trying to continue with a deprecated program
- Review and Submit page
   - Inline error when a program is deprecated
   - Submit button is disabled?

#### Test cases covered

- `dataValidatorHelpers.test.ts`
    - return error if a submitted programID is deprecated
    - return error if a child rate has a deprecated rateProgramID
    - ignores deprecated rateProgramIDs on linked rates
    - should return an error if a submitted programID is deprecated (EQRO parsing and validation)
- `submitContract.test.ts`
    - returns an error when submitting a contract with a deprecated program
    - returns an error when submitting a contract whose child rate has a deprecated rateProgramID

### DIFF
--- a/services/app-api/src/domain-models/contractAndRates/dataValidatorHelpers.test.ts
+++ b/services/app-api/src/domain-models/contractAndRates/dataValidatorHelpers.test.ts
@@ -7,6 +7,7 @@ import {
 } from './dataValidatorHelpers'
 import {
     mockGqlContractDraftRevisionFormDataInput,
+    mockSubmittableHealthPlanContract,
     must,
 } from '../../testHelpers'
 import type {
@@ -23,6 +24,7 @@ import {
 } from '../../testHelpers/gqlHelpers'
 import type { ContractFormDataType } from './formDataTypes'
 import { eqroValidationAndReviewDetermination } from '@mc-review/submissions'
+import { vi } from 'vitest'
 
 describe('Health plan parsing and validation', () => {
     describe('validateContractDraftRevisionInput', () => {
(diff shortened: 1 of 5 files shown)


ruizajtruss (approved):
LGTM!

I agree we can forgo the AC you mentioned, that was written before we decided to make users replace deprecated programs if they were to unlock a submission that had one.

PR #4339, opened 2026-04-16
MCR-4104: Upgrade Apollo Client

JasonLin0991:
## Summary

Upgrades `@apollo/client` from v3.8.8 → v4.1.7 and `@as-integrations/aws-lambda` from v3.1.0 → v4.0.1 across the whole workspace. MCR-6116 previously unblocked this by swapping codegen's generated hooks for `TypedDocumentNode`, which made the upgrade path tractable.

## What changed

- Bumped `@apollo/client` to `^4.1.7` across all 8 workspace packages 
- Upgrade `@as-integrations/aws-lambda` to `^4.0.1` in app-api
- Added `rxjs` peer dep.
- Ran the Apollo v3→v4 codemod to automate some of the migration work.
- Rewrote error handling in `packages/helpers/src/gql/apolloErrors.ts` for `@apollo/client` V4
- Added `toGQLError` in `@mc-review/helpers` (with unit tests) and converted ~17 call sites that previously read `error.graphQLErrors[0]?.extensions?.code` to use it.
- Widened the `fakeAmplifyFetch` / `localGQLFetch` signatures to `(uri: URL | RequestInfo, options?: RequestInit)` to match v4's stricter `HttpLink` fetch option.
- Set `MockLink.defaultOptions = { delay: 0 }` in `setupTests.ts` so v4's new 20–50 ms default response delay doesn't slow down unit tests; per-mock overrides still work.
- Fixed the cypress harness: migrated `MutateResult.errors` → `.error`, added data guards for v4's stricter nullability, and updated the local `fakeAmplifyFetch` the same way.

## QA guidance

- CI should handle validating most of the issues.

## PR Reminders

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -51,6 +51,7 @@ export IAM_PATH='/'
 export CLOUDFRONT_CERT_ARN='local'
 export CLOUDFRONT_SB_DOMAIN_NAME='local'
 export CLOUDFRONT_DOMAIN_NAME='local'
+export SECRETS_MANAGER_SECRET='local-secrets-manager'
 export stage='local'
 export REGION='us-east-1'
 
--- a/packages/common-code/package.json
+++ b/packages/common-code/package.json
@@ -11,7 +11,8 @@
         "test:coverage": "vitest --coverage"
     },
     "dependencies": {
-        "@apollo/client": "3.8.8",
+        "@apollo/client": "^4.1.7",
+        "rxjs": "^7.8.1",
         "@graphql-typed-document-node/core": "^3.2.0",
         "@mc-review/constants": "workspace:*",
         "@mc-review/helpers": "workspace:*",
--- a/packages/constants/package.json
+++ b/packages/constants/package.json
@@ -8,7 +8,8 @@
     },
     "license": "ISC",
     "dependencies": {
-        "@apollo/client": "3.8.8",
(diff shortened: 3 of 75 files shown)


mojotalantikite (approved):
No notes, LGTM. The error handling refactor looks great!

.envrc
```diff
@@ -51,6 +51,7 @@ export IAM_PATH='/'
 export CLOUDFRONT_CERT_ARN='local'
 export CLOUDFRONT_SB_DOMAIN_NAME='local'
 export CLOUDFRONT_DOMAIN_NAME='local'
+export SECRETS_MANAGER_SECRET='local-secrets-manager'
```
pearl-truss, line 54: Has this variable been set in the higher environments?

JasonLin0991: It is set in higher envs and being used like in `oauth_token.ts`. This should have been added to local env. Without it if you tried to to use the third party API calls on local deploys it would fail. I found this when I ran the thirdParty auth cypress test.

JasonLin0991: For context: SECRETS_MANAGER_SECRET was set in higher envs and added into the codebase from a previous PR.

pearl-truss: Perfect!

pearl-truss (approved):

PR #4340, opened 2026-04-16
Check prisma generated directory before generating

mojotalantikite:
## Summary

Jason hit an issue where the `dev` tool errored because he already had a local `generated` dir for prisma output. This just checks to see if it already exists and clears it if so.

#### Related issues
https://jiraent.cms.gov/browse/MCR-6064


JasonLin0991 (approved):
Thanks for adding this in!

PR #4348, opened 2026-04-17
MCR-6076 Do not display D-SNP question for CHIP only submissions

adriannekennedy:
## Summary
Implemented CHIP-only D-SNP handling behind the existing chip-submission-automation feature flag.

On the web side, we now hide the D-SNP question on Contract Details when it's a CHIP-only submissions, along with its help text and yes/no options. The summary page was also updated to hide the D-SNP row. On submit, logic clears any existing dsnpContract value when the submission is CHIP-only so stale answers don’t persist.

On the API side, I added a server-side backstop that also clears dsnpContract for CHIP-only updates and skips D-SNP-required validation in that scenario. I added focused web and API tests for the hide/clear behavior, and the targeted app-web Vitest files are passing.

#### Related issues
https://jiraent.cms.gov/browse/MCR-6076

#### Screenshots
image

image

#### Test cases covered
ContractDetails.test.tsx & ContractDetailsSummarySection.test.tsx:
Verifies the D-SNP summary row is hidden for CHIP-only submissions when the flag is enabled.

dataValidatorHelpers.test.ts & updateContractDraftRevision.test.ts
Verifies the server side validations are updated and that resubmissions properly clear the dsnpContract variable

## QA guidance
- create CHIP only submission 
- verify dsnp question does not display on Contract Details
- verify dsnp question does not display on Contract Summary
- verify changing a submission that is NOT chip-only to chip-only clears the dsnp value (on resubmit)

## PR Reminders
- [NA] Updated the API Changelog (if this PR changes the API schema)
- [NA] Checked accessibility with ANDI and Wave tools as needed

### DIFF
--- a/services/app-api/src/domain-models/contractAndRates/dataValidatorHelpers.test.ts
+++ b/services/app-api/src/domain-models/contractAndRates/dataValidatorHelpers.test.ts
@@ -3,6 +3,7 @@ import {
     validateEQROContractDraftRevisionInput,
     parseContract,
     parseAndUpdateEqroFields,
+    parseAndUpdateChipOnlyFields,
     parseEQROContract,
 } from './dataValidatorHelpers'
 import {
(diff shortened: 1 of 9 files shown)


JasonLin0991 (approved):
LTGM!

PR #4359, opened 2026-04-20
Add query to support rate arrays

mojotalantikite:
## Summary

ARMS let us know that it'd make their lives easier if they could query rates by a list of rate IDs. We actually already had some of the types, we just didn't have the Prisma piece to support it. This adds it.

#### Related Issues
https://jiraent.cms.gov/browse/MCR-6202


JasonLin0991 (approved):
LGTM!

pearl-truss (commented):
Nice change! Can you update the API changelog as well?

PR #4360, opened 2026-04-20
Add query for rates updated since given date

mojotalantikite:
## Summary

Another major quality of life improvement that would decrease the query load for ARMS is providing a date since last updated. This adds the ability to provide an optional `updatedSince` field that will only return rates that have seen a change since that provided date.

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/findAllRatesStripped.ts
+++ b/services/app-api/src/postgres/contractAndRates/findAllRatesStripped.ts
@@ -15,6 +15,7 @@ type StrippedRateOrErrorArrayType = StrippedRateOrErrorType[]
 type FindAllRatesStrippedType = {
     stateCode?: string
     rateIDs?: string[]
+    updatedSince?: Date
 }
 
 async function findAllRatesStrippedInTransaction(
@@ -23,6 +24,7 @@ async function findAllRatesStrippedInTransaction(
 ): Promise<StrippedRateOrErrorArrayType | Error> {
     const stateCode = args?.stateCode
     const rateIDs = args?.rateIDs
+    const updatedSince = args?.updatedSince
     const rates = await tx.rateTable.findMany({
         where: {
             revisions: {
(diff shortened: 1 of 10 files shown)


services/app-api/src/postgres/contractAndRates/findAllRatesStripped.ts
```diff
@@ -44,6 +46,33 @@ async function findAllRatesStrippedInTransaction(
...
+            AND: updatedSince
+                ? [
+                      {
+                          OR: [
+                              { updatedAt: { gte: updatedSince } },
+                              {
+                                  revisions: {
+                                      some: {
+                                          submitInfoID: { not: null },
+                                          updatedAt: { gte: updatedSince },
```
JasonLin0991: **Issue**: This does not capture changes in statuses or Q&A, I would follow what indexContracts does. In that resolver we query all the data then filter on `lastUpdatedForDisplay`. This date is calculated in `services/app-api/src/resolvers/contract/contractResolver.ts`. Need make sure we calculate it for rates correctly.

JasonLin0991 (changes requested):

services/app-api/src/postgres/contractAndRates/findAllRatesWithHistoryBySubmitInfo.ts
```diff
@@ -39,6 +40,36 @@ async function findAllRatesWithHistoryBySubmitInfo(
...
+                                      submitInfoID: { not: null },
+                                      updatedAt: { gte: args.updatedSince },
+                                  },
+                              },
+                          },
+                          {
+                              revisions: {
+                                  some: {
+                                      submitInfoID: { not: null },
+                                      submitInfo: {
```
JasonLin0991, line 58: We also need to include unlocks in this if it's possible.

JasonLin0991: This is for contracts, but it should calculate the same with rates.

```ts
lastUpdatedForDisplay(parent: ParentType) {
            // These dates are mechanical, draft vs. submit vs. unlock, whatever is latest is latest
            const contractUpdated = parent.updatedAt
            const draftUpdated = parent.draftRevision?.updatedAt

            const lastSubmitted =
                parent.packageSubmissions.length > 0
                    ? parent.packageSubmissions[0].contractRevision.submitInfo
                          ?.updatedAt
                    : undefined
            const lastUnlocked = parent.draftRevision?.unlockInfo?.updatedAt
            const submitStatusDate =
                lastUnlocked || lastSubmitted || draftUpdated || contractUpdated

            // With review actions, we compare if the review action has happened more recently or not than the latest submit action
            if (
                parent.reviewStatusActions &&
                parent.reviewStatusActions.length > 0
            ) {
                const latestAction = parent.reviewStatusActions[0].updatedAt
                if (latestAction && latestAction > submitStatusDate) {
                    return latestAction
                }
            }

            return submitStatusDate
        },
```

I have a suspicion the Prisma OR queries might not get the date we need.

JasonLin0991 (commented):
IMO, we should make a ticket on this and take time to create this functionality. The complexity of the many table relationships and what actions affects the dates is too much for a quick implementation.

mojotalantikite:
> IMO, we should make a ticket on this and take time to create this functionality. The complexity of the many table relationships and what actions affects the dates is too much for a quick implementation.

Yeah I agree, this got complicated pretty quickly and we should spend some time researching this more. I'll close this out and we can take more time to complete this: https://jiraent.cms.gov/browse/MCR-6205

PR #4396, opened 2026-04-28
MCR-6211 Add Graphql Pagination to indexRates

adriannekennedy:
## Summary
MCR-6211 is to add pagination to the graphql layer of the indexRates API. I created a new endpoint for the pagination, so that the work is clearly separated from the existing endpoint and will not break existing customers. 
The page size can be configured - created both a default page size (10) to use if no `first` param is passed, as well as a Max page size. After discussing with Jason, we'll try setting the max page size to 150 since they're currently pulling more than 100 records at a time successfully.
The current logic is using the existing query, `findAllRatesWithHistoryBySubmitInfo` to fetch the rate data and slice it up into pages. 
I added a new helper method `getRateLastUpdatedForDisplay` to order the records by the last time they were actually updated. Need feedback on this method, as I might be missing some logic. I also think I need to add the test case for this to helpers.test.ts. 

#### Related issues
https://jiraent.cms.gov/browse/MCR-6211

#### Screenshots
Default call will give a max of 10 per page:
image

image

#### Test cases covered
- confirms paginated rates, cursor, and pageInfo are returned properly
- confirms an error is generated when max page size is exceeded
- confirms an error is generated if the cursor is invalid
- confirms rates are sorted by lastUpdatedForDisplay
- confirms existing state filter on indexRates still works on indexRatesConnection

## QA guidance
- go to Graphql-explorer and create a request for `indexRatesConnection`
- execute the api with no input params and confirm 10 records are returned
- execute the api with `first` param = 50/100/150 and confirm number of records returned
- execute the api with `first` param and `after` with provided cursor

## PR Reminders


services/app-api/src/resolvers/rate/indexRatesConnection.ts
```diff
@@ -0,0 +1,274 @@
...
+    }
+}
+
+export function indexRatesConnectionResolver(
+    store: Store
+): QueryResolvers['indexRatesConnection'] {
+    return async (_parent, { input, first, after }, context) => {
+        const { user, ctx, tracer } = context
+        const span = tracer?.startSpan('indexRatesConnection', {}, ctx)
+        setResolverDetailsOnActiveSpan('indexRatesConnection', user, span)
```
JasonLin0991: **Todo**: We need to use the new `withResolverSpan` wrapper for tracing. Take a look at fetchContract resolver to see how it us used.

services/app-api/src/resolvers/rate/indexRatesConnection.ts
```diff
@@ -0,0 +1,274 @@
...
+                    extensions: {
+                        code: 'INTERNAL_SERVER_ERROR',
+                        cause: 'DB_ERROR',
+                    },
+                })
+            }
+
+            const rates = validateAndReturnRates(ratesWithHistory, span)
+
+            const filteredRates = stateUser
```
JasonLin0991: **Suggestion**: We don't need to filter rates out for states again 
```ts
await store.findAllRatesWithHistoryBySubmitInfo({
    stateCode: user.stateCode,
    rateIDs: input?.rateIDs ?? undefined,
    useZod: true,
})
```

The state code argument in `findAllRatesWithHistoryBySubmitInfo` filters outs submissions not from the state.

services/app-api/src/resolvers/helpers.ts
```diff
@@ -65,3 +65,29 @@ export function getLastUpdatedForDisplay(
...
+
+    const lastSubmitted = latestDate(
+        (rate.packageSubmissions ?? []).map((ps) =>
+            toDate(ps?.rateRevision?.submitInfo?.updatedAt)
+        )
+    )
+
+    const latestReviewAction = latestDate(
+        (rate.reviewStatusActions ?? []).map((a) => toDate(a?.updatedAt))
+    )
```
JasonLin0991, line 84: **Thought**: We likely want to account for linking/delinking this rate to a contract. I think it might be complex so I wouldn't try to tackle it in this ticket. I can make a ticket for this.

services/app-api/src/resolvers/rate/indexRatesConnection.ts
```diff
@@ -0,0 +1,274 @@
...
+                ? rates.filter((rate) => user.stateCode === rate.stateCode)
+                : rates
+            const sortedRates = [...filteredRates].sort(compareRates)
+            const decodedAfter = normalizeAfterCursor(after ?? undefined)
```
JasonLin0991: **Question**: Wondering what the encoding of `after` is for? Could we not encode this? <strike>Also where is is being encoded at?</strike> Found it at the bottom.

services/app-api/src/resolvers/rate/indexRatesConnection.ts
```diff
@@ -0,0 +1,274 @@
...
+import type { RateOrErrorArrayType } from '../../postgres/contractAndRates'
+import { logError, logSuccess } from '../../logger'
+import { GraphQLError } from 'graphql'
+import type { RateType } from '../../domain-models/contractAndRates'
+import { canRead } from '../../authorization/oauthAuthorization'
+import { NotFoundError } from '../../postgres/postgresErrors'
+import { getRateLastUpdatedForDisplay } from '../helpers'
+
+const DEFAULT_INDEX_RATES_PAGE_SIZE = 10
+const MAX_INDEX_RATES_PAGE_SIZE = 50
```
JasonLin0991: **Thought**: Let me load up val data and see how large indexRates is, we can then see what to set at.

API_Changelog.md
```diff
@@ -1,6 +1,24 @@
 # Managed Care Review - API Changelog
 ## This document highlights API changes that have been introduced since May 2025. See the full [GraphQL schema](services/app-graphql/src/schema.graphql).
 
+### April 28, 2026
+#### Added
+- New endpoint `indexRatesConnection` added to the API for paginated submitted-rate results.
```
JasonLin0991: **Suggestion**: It might be helpful to add some code samples on how to use the inputs. I think GraphQL pagination usage is different from REST pagination patterns.

services/app-graphql/src/schema.graphql
```diff
@@ -47,6 +47,19 @@ type Query {
     """
     indexRates(input: IndexRatesInput): IndexRatesPayload!
     """
+    indexRatesConnection returns full submitted rates using a paginated connection shape
+    indexRatesConnection can be called by CMS and admin users
```
JasonLin0991: **Todo**: Can you update this comment here and for indexRates. Both can be called by State users, but response data will only contain rates for their state.

services/app-graphql/src/schema.graphql
```diff
@@ -47,6 +47,19 @@ type Query {
...
+
+    Errors:
+    - ForbiddenError: User must be a CMS or Admin type user
+    - NotFoundError:  No submitted rates found
+    """
+    indexRatesConnection(
+        input: IndexRatesInput
+        first: Int
```
JasonLin0991: **Suggestion**: I think we weed to break out this input into its own defined GraphQL type. This is so we can add documentation comments above it. Specially need to know what `first` and `after` inputs are and the default if `first` is not provided`

services/app-api/src/resolvers/rate/indexRatesConnection.ts
JasonLin0991: **Suggestion**: Ok, from local deployment using VAL data, looks like 150 max should be good. At 150 records it was Around 4.9 mb. That leaves room for adding more data to the query.

services/app-graphql/src/schema.graphql
```diff
@@ -2292,6 +2310,21 @@ type IndexRatesPayload {
...
 }
 
+type RateConnectionEdge {
+    cursor: String!
+    node: Rate!
+}
+
+type RateConnection {
+    "Total number of submitted rates returned on request"
+    totalCount: Int
```
JasonLin0991, line 2392: **Suggestion**: What do you think about adding `totalPages` to this?

JasonLin0991 (commented):
Pagination works. Left a bunch of comments, none are real deal breakers. I would suggest we change max page count to 150.

The overall consideration I had in mind when review was looking at this API endpoint form a external user perspective. I want it to be somewhat easy for them to use it.

services/app-api/src/resolvers/helpers.ts
adriannekennedy: Ya, I assumed there were some things missing from this logic. I see how in indexContracts we check for several other status actions, but I was not sure how that would translate when looking at just the rates. So separate ticket sounds good.

services/app-api/src/resolvers/rate/indexRatesConnection.ts
adriannekennedy: to give the user an opaque string to use instead of exposing the actual cursor to them. Simplifies it for the user a bit, I guess, was recommended by codex. Just a way to hide the internals of what fields the cursor is using, so the user only ever has to deal with 1 'type' of cursor, but internally we can change what defines the cursor if we need to without disturbing the users.

API_Changelog.md
adriannekennedy: will do -- I forgot to circle back around and add more details to the documentation

services/app-graphql/src/schema.graphql
adriannekennedy: good add on, especially since that will dynamically changed based on how the user fetches data

services/app-api/src/resolvers/rate/indexRatesConnection.ts
adriannekennedy: Note: I define this a bit better in API Changelog for less techy readers

API_Changelog.md
```diff
@@ -1,6 +1,86 @@
...
+            "input": {
+                "first": 50 
+            }
+        }
+        ```
+    - Example next-page variables, with no additional filters - fetches the next 50 records using the `pageInfo.endCursor` from the previous example's response:
+        ```json
+        {
+            "input": {
+                "first": 10,
```
JasonLin0991: **Question**: Is this fetching 50 more since `"first": 10`? Isn't this now fetching only 10?

API_Changelog.md
```diff
@@ -1,6 +1,86 @@
...
+                "after": "string-end-cursor"
+            }
+        }
+        ```
+    - Example first-page variables, with optional filters - fetches less than the default 10 rates per page:
```
JasonLin0991: **Suggestion**: I think we need to group these examples into sections so that the user knows the sequence of queries. For instance, the first 2 examples show how to query the first page, then the subsequent pages.

This example and the following shows how to query with more inputs.

API_Changelog.md
```diff
@@ -1,6 +1,86 @@
...
 
+### April 28, 2026
+#### Added
+- New endpoint `indexRatesConnection` added to the API for paginated submitted-rate results. The api can be called with no input parameters, and a default page size of 10 will be used. 
+    - Accepts the new object `IndexRatesConnectionInput`
+        - The new object contains the same, optional filter paramters that `IndexRatesInput` accepts:
+            - `stateCode`: optional state filter for CMS and admin users
+            - `rateIDs`: optional list of rate IDs to limit the result set
+        - The new object also contains pagination specific arguments:
+            - `first`: optional page size, default is 10, max is 150
```
JasonLin0991: **Thought**: I do wonder if these inputs are confusing for users. I know this is what is specified in the GQL docs, but I don't think there is anything stopping us from changing the names.

JasonLin0991 (approved):
Looks good. Left some suggestions. I do want to get another set of eyes on this though. Let me make a slack post and see if someone can take a look. We can merge this tomorrow or later if no one else reviews it.

API_Changelog.md
adriannekennedy: 10 more - typo in the comment - i'll push the updated text with some more documentation clean up

API_Changelog.md
adriannekennedy: That is true. `pageSize` makes more sense to me than `first`. If it makes it easier for the consumers to understand what they're doing - I'm all for it. 
`after` seems ok, can't think of a better naming convention off the top of my head, but first for sure makes more sense as pageSize.

pearl-truss: Could you also consider renaming `indexRatesConnection`, something like `indexRatesWithPagination` with idea of being more clear on how it differs from the standard indexRates endpoint

API_Changelog.md
```diff
@@ -1,6 +1,88 @@
...
+            - `after`: optional opaque cursor for fetching the next page
+    - Returns `RateConnection`
+        - `totalCount`: total number of matching submitted rates
+        - `totalPages`: total number of pages based on totalCount and requested page size
+        - `edges`: list of `RateConnectionEdge`
+        - `pageInfo`: pagination metadata with `hasNextPage` and `endCursor`
+    - `RateConnectionEdge` includes:
+        - `cursor`: opaque cursor for the current edge
+        - `node`: the `Rate`
+    - Example query:
```
pearl-truss: Praise: thanks for including these example queries! I would just be in favor of having these separated into a pagination.md document file - in case anyone in the future wants more info/examples on how we handle pagination they may not think to come here and having dedicated page will be easier to link external users to

pearl-truss (commented):
Looking good! I left some comments around documentation

API_Changelog.md
adriannekennedy: Sure!!! Honestly I did not like the naming convention to begin with, but supposedly 'connection' == 'pagination' in graphql so I thought this was per the convention but did not like it at all.

API_Changelog.md
adriannekennedy: Can do!

pearl-truss (approved):
Looks great!

PR #4397, opened 2026-04-28
MCR-6203: delete q&a question by implementeding soft delete

JasonLin0991:
## Summary
[MCR-6203](https://jiraent.cms.gov/browse/MCR-6203)

This work adds in the ability to soft delete Questions, Responses and individual documents with the ability to keep a log of the delete action.

#### In scope:
- **New action tables** (one per Q&A table): `ContractQuestionAction`, `ContractQuestionDocumentAction`, `ContractQuestionResponseAction`, `ContractQuestionResponseDocumentAction`, plus the four rate-side equivalents. Each row records the action (`DELETE`, `RESTORE`, `CASCADE_DELETE`, `CASCADE_RESTORE`), who took it, and an optional reason. Indexed on `(parentID, createdAt DESC)` so the latest action is cheap to look up.
- **API**: Added `deleteContractQuestion resolver` only available for admins for now.
   - The store function `softDeleteContractQuestion` for soft deleting the contract question and cascading down to documents and responses.
   - Added code to prevent race conditions in DB updates via Prisma.$transactions isolation level.
      - Second request in sequence will return an error.
- Filtering only active questions.
   - `excludeSoftDeletedQuestionData` function, used for filtering out question data.
   - The two places that are querying the QA data for API responses are in:
      - `services/app-api/src/postgres/questionResponse/findAllQuestionsByContract.ts`
      - `services/app-api/src/postgres/questionResponse/findAllQuestionsByRate.ts`
      - Both places use the `convertNonDeleted*` respective functions to filter out soft deleted questions before converting it to domain types.

#### Out of scope:
- UI work

#### Additional work
- Dev tool: `./dev prisma --` and the test runner now pass `--config=../../prisma.config.ts` so Prisma 7 finds the config from `services/app-api`.

#### Test cases covered
- `softDeleteContractQuestion.test.ts` - store function
   - `'serializes concurrent deletes — no duplicate action rows under a race'`. Attempts to test the race condition. There is work required to figure out how to consistently and accurately make a true concurrent function call in vitest.
- `deleteContractQuestion.test.ts` - resolver
   - `'soft deletes a question and cascades to its responses and documents'`
   - `'rejects non-admin users with FORBIDDEN'`
   - `'returns NOT_FOUND for a missing question id'`

## QA guidance
- Make sure the Q&A workflow still works with this change. 
- Use the `deleteContractQuestion` in the graphql explorer tool to test soft deletions.
   - Make sure contract questions are not displaying in the UI

## PR Reminders


vyakusha:
I run new mutation DeleteContractQuestion in deployment environment and it run successfully, removing Question from UI. Question: will CreateContractQuestionResponse create a new Response (given right inputs), even if Question is soft-deleted? Do we need to cover this eventuality?

JasonLin0991:
> Question: will CreateContractQuestionResponse create a new Response (given right inputs), even if Question is soft-deleted? Do we need to cover this eventuality?

Good catch, we don't check if the question is deleted when we create a response. I can add that

vyakusha:
`findAllDocuments` also does not filter for Deleted questions. Also, "reason" is not included when running "DeleteContractQuestion". Shall we add "reason" as a required filed there?

The audit reason is implemented in storage but cannot be supplied through the GraphQL API, so delete actions created via the new mutation will always store NULL reasons. The store type and action inserts all support reason (
), but DeleteContractQuestionInput only exposes questionID (
) and the resolver only forwards questionID plus user (
). If capturing deletion rationale is part of the requirement, that path is incomplete.

JasonLin0991:
@vyakusha 

>findAllDocuments also does not filter for Deleted questions.

This is intentional, we still want to be able to fetch all documents. Currently this is used in a S3 document audit lambda, which we still want to make sure soft deleted questions are still a part of the audit.

services/app-api/src/domain-models/QuestionsType.ts
```diff
@@ -1,7 +1,26 @@
...
+    'RESTORE',
+    'CASCADE_DELETE',
+    'CASCADE_RESTORE',
+])
+
+const questionActionSchema = z.object({
+    id: z.uuid(),
+    createdAt: z.date(),
+    action: questionActionTypeSchema,
+    reason: z.string().nullable().optional(),
```
pearl-truss, line 20: It looks like this is a mandatory field in other parts of the code

pearl-truss (approved):
LGTM, just left a minor comment

vyakusha (approved):
LGTM. Attempt to delete when no admin: "user not authorized to delete a question". Attempt to delete a question already deleted: "Question with id ... is already deleted". All looks solid.

services/app-api/src/domain-models/QuestionsType.ts
JasonLin0991: Yes it should be. I can make a follow up PR since this review deploy has a deleted question with null reason

PR #4411, opened 2026-05-06
[MCR-6125] Create link to Salesforce from the State Portal

ruizajtruss:
## Summary
As a State user, I want to be able to be routed to the SDP application from the state portal so that I can create State Directed Payment Preprint (SDP) submissions

Acceptance Criteria

- A feature flag is created so that all work associated with surfacing SDP in MC-Review State portal can remain behind a feature flag until the business and CMS stakeholders are ready to expose the functionality in production 
- State Directed Payment Preprint (SDP) will be added as a submission type to the New Submission page in the State Portal
- The State Directed Payment Preprint (SDP) submission type will include a radio button which can be selected
- When the State Directed Payment Preprint (SDP) radio button is selected and the user presses, "Start" then the user will be routed to the submission form within the SDP application (URL: https://cmsapps5--mcrevval.sandbox.my.site.com/s/state-directed-preprint-submission)
- Update the header to reflect "Submission type" instead of "Contract type"
#### Related issues
[MCR-6125](https://jiraent.cms.gov/browse/MCR-6125)
#### Screenshots

https://github.com/user-attachments/assets/f7f31866-7212-493f-8fb9-a57c068fc0a2

## PR Reminders


services/app-web/src/pages/StateSubmission/SharedSubmissionComponents/NewSubmissionForm/NewSubmissionForm.tsx
```diff
@@ -54,11 +61,19 @@ export const NewSubmission = () => {
...
-            generatePath(RoutesRecord.SUBMISSIONS_NEW_SUBMISSION_FORM, {
-                contractSubmissionType: values.contractType,
-            })
-        )
+        if (
+            isSDPEnabled &&
+            values.submissionType === ContractSubmissionTypeRecord['SDP']
+        ) {
+            window.location.href =
+                'https://cmsapps5--mcrevval.sandbox.my.site.com/s/state-directed-preprint-submission'
```
JasonLin0991: Suggestion: We might want to think about how to set this url for different environments. <strike>The testing environment this link is for our VAL environment. DEV and Review app will have no SF environment for this radio and</strike> Prod will be a different link.

services/app-web/src/pages/StateSubmission/SharedSubmissionComponents/NewSubmissionForm/NewSubmissionForm.tsx
```diff
@@ -54,11 +61,19 @@ export const NewSubmission = () => {
...
-        navigate(
-            generatePath(RoutesRecord.SUBMISSIONS_NEW_SUBMISSION_FORM, {
-                contractSubmissionType: values.contractType,
-            })
-        )
+        if (
+            isSDPEnabled &&
+            values.submissionType === ContractSubmissionTypeRecord['SDP']
+        ) {
+            window.location.href =
```
JasonLin0991: **Suggestion**: We will likely want top open the link in a new tab here, in order to track session times in the state portal. We also should see if `'react-router-dom'` has something we can use. Maybe [redirect](https://reactrouter.com/api/utils/redirect#url) could work.

services/app-web/src/pages/StateSubmission/SharedSubmissionComponents/NewSubmissionForm/NewSubmissionForm.tsx
```diff
@@ -95,52 +110,68 @@ export const NewSubmission = () => {
...
-                                        parent_component_heading="Contract type"
+                                        parent_component_heading="Submission type"
                                         radio_button_title="Health plan"
                                         labelDescription="Submit your Medicaid and CHIP managed care plans. This includes base contracts, amendments to base contracts, and rate certifications."
                                         tile
                                     />
                                     <FieldRadio
                                         id="eqro"
-                                        name="contractType"
+                                        name="submissionType"
```
JasonLin0991: **Nit**: This may get confusing since in the form data, there is both a `contractType` and `submissionType` that are different than this value. The name we have been using is `contractSubmissionType`. We might want to rename `contractType` to that instead.

services/app-web/src/pages/StateSubmission/SharedSubmissionComponents/NewSubmissionForm/NewSubmissionForm.tsx
```diff
@@ -95,52 +110,68 @@ export const NewSubmission = () => {
...
                                         label="External Quality Review Organization (EQRO)"
                                         data-testid="eqro"
                                         aria-required
                                         value="eqro"
                                         list_position={2}
                                         list_options={2}
-                                        parent_component_heading="Contract type"
+                                        parent_component_heading="Submission type"
```
JasonLin0991, line 163: **Note**: This can stay as is.

services/app-web/src/pages/StateSubmission/SharedSubmissionComponents/NewSubmissionForm/NewSubmissionForm.tsx
JasonLin0991: We have a ticket for investigating how to track sessions so new tab might not be an issue. Just from off the top of my head, I think without a new tab, we would rely on the IDM session. I just wonder if that session extends for our app when working in SDP.

services/app-web/src/pages/StateSubmission/SharedSubmissionComponents/NewSubmissionForm/NewSubmissionForm.tsx
JasonLin0991: Looks like that link does not require user auth, but likely need to retest this work once user auth is required. We still need to think about how to present different urls for VAL and Prod.

JasonLin0991: **Issue**: Also leaving this url hardcoded in the codebase doesn't seem like a good idea since not auth is needed. I know it's a testing env, but now the link is exposed to the public.

JasonLin0991 (approved):
Looks good.

The issue of hardcoded url should be addressed before merging.

services/app-web/src/pages/StateSubmission/SharedSubmissionComponents/NewSubmissionForm/NewSubmissionForm.tsx
pearl-truss: > We will likely want top open the link in a new tab here

I'm not sure this is true, the keep mentioning wanting a seamless experience with the goal of the user not realizing they've left state portal.

Before making the change, I think it's worth confirming with Ryan and/or Amanda

services/app-web/src/pages/StateSubmission/SharedSubmissionComponents/NewSubmissionForm/NewSubmissionForm.tsx
pearl-truss: Worthing thinking about how we want to do this, my first thought is storing the URL in an environment variable that way we're not hardcoding a URL into the repo and we can cover val and prod (down the line)

pearl-truss (commented):
A few comments but it's looking good

PR #4421, opened 2026-05-07
MCR-6250: Admin only UI for soft deletion of question

JasonLin0991:
## Summary
[MCR-6250](https://jiraent.cms.gov/browse/MCR-6250)

Follow up to [MCR-6203](https://jiraent.cms.gov/browse/MCR-6203).

This work adds in UI for admins to perform a contract question deletion.

Additional work:
- Align some CSS on the Q&A pages
- Fix round numbering [MCR-6037](https://jiraent.cms.gov/browse/MCR-6037)
- Fix some console errors for React 19 upgrade

#### Screenshots

Screenshot 2026-05-07 at 2 51 33 PM

Screenshot 2026-05-07 at 2 51 46 PM

#### Test cases covered

`DeleteContractQuestion.test.tsx`
   - renders the delete question form
  - redirects non-admin users
  - returns 404 on bogus division
  - shows generic error on draft contract
  - shows validation error when reason is missing
  - successfully deletes question and unmounts
  - shows error banner on mutation failure

`questionResponse.spec.ts`
- `can add questions and responses` -> `can add and delete questions and responses`

## QA guidance

- As a CMS user create a contract question
- As a State user create one or many responses to contract question
- As a Admin user, delete this question.

## PR Reminders


pearl-truss (commented):
UI changes look good but the contractOverrides field isn't exposed to graphql, I think the decision to not thread it through the API is a holdover from when we were just doing these changes directly through migrations. If not in this PR, I think we should expose that field to graphql now that this action can be taken in the UI.  @JasonLin0991 what do you think?

JasonLin0991:
> UI changes look good but the contractOverrides field isn't exposed to graphql, I think the decision to not thread it through the API is a holdover from when we were just doing these changes directly through migrations. If not in this PR, I think we should expose that field to graphql now that this action can be taken in the UI. @JasonLin0991 what do you think?

This work isn't using the contractOverrides. That is a different ticket in the sprint [MCR-6198](https://jiraent.cms.gov/browse/MCR-6198). This one is a soft delete for Q&A questions using new tables in the DB to track deleted files without actually removing them, see this [PR](https://github.com/Enterprise-CMCS/managed-care-review/pull/4397).

pearl-truss (approved):

pearl-truss:
> > UI changes look good but the contractOverrides field isn't exposed to graphql, I think the decision to not thread it through the API is a holdover from when we were just doing these changes directly through migrations. If not in this PR, I think we should expose that field to graphql now that this action can be taken in the UI. @JasonLin0991 what do you think?
> 
> This work isn't using the contractOverrides. That is a different ticket in the sprint [MCR-6198](https://jiraent.cms.gov/browse/MCR-6198). This one is a soft delete for Q&A questions using new tables in the DB to track deleted files without actually removing them, see this [PR](https://github.com/Enterprise-CMCS/managed-care-review/pull/4397).

Got it but the main issue is still there. Unless I'm mistaken - is there a way via the API to validate that the question.actions is being populated as expected after a question has been deleted?

PR #4429, opened 2026-05-11
MCR-6036 Display All Question Documents for State User Response page

adriannekennedy:
## Summary
When a state user is responding to a question, only 1 document was showing on the Upload response page. There was already a filter in place `onlyDisplayInitial`, which was filtering to the last document uploaded or showing all documents. To make as minimal changes as possible, I just changed that filter to filter down to all documents added by CMS users or show all documents. Both contract and rates uses this method, so one change fixes both places. There was not test class for this component, so I did not add any test cases - but I can if the team wants.

#### Related issues
https://jiraent.cms.gov/browse/MCR-6036

#### Screenshots
All docs still display on Q&A page:
image

All question docs display on Upload Response:
image

All docs on Rate Review Upload Response:
image

#### Test cases covered
Changes were isolated to QuestionDisplayTable but that component does not have a test file.

## QA guidance
- As CMS user, upload multiple documents on a Question 
- As State User, click Upload Response and verify that all documents display

## PR Reminders

### DIFF
--- a/services/app-web/src/pages/QuestionResponse/QATable/QuestionDisplayTable.tsx
+++ b/services/app-web/src/pages/QuestionResponse/QATable/QuestionDisplayTable.tsx
@@ -14,12 +14,13 @@ type QuestionDisplayTablePropType = {
 export const QuestionDisplayTable = ({
     documents,
     user,
-    onlyDisplayInitial, // used when we only care about initial question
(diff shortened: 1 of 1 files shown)


vyakusha (approved):
LGTM. Multiple uploaded documents are displayed as required.

PR #4434, opened 2026-05-12
[MCR-6198] Add form data override for contract revision

pearl-truss:
## Summary

This PR creates a contract revision override (modeled after the rate revision override).

- It adds support for `contractType` overrides on `ContractRevision`, including:
- unlock preservation so an unlocked draft carries forward the effective overridden `contractType`
- support for overrides on `SUBMITTED`, `RESUBMITTED`, and now `APPROVED` contracts

(NEW since last review): This PR extends the contract revision override to include the ability to add new contract and supporting documents. When a documentID isn't present on the contractDocumentOverride it will add the document, if the documentID is present it will update the document. More details below:

[Notion document for this PR
](https://www.notion.so/mc-review/MCR-6198-3684f9e180fa80829aa8ece2e9036804)

## Contract document overrides

Adds support for adding and updating contract document and supporting document metadata via the existing override pattern 

## What's new

**Schema** — `ContractDocumentOverride` and `ContractSupportingDocumentOverride` models. Nullable `documentID` discriminates add-mode (full doc payload) from update-mode (sparse patch, today just `dateAdded`).

**Write path** — `overrideContractData` accepts `contractDocuments` / `supportingDocuments` arrays. Add rows require `name`, `sha256`, `s3URL`. Update rows reject anything except `dateAdded`.

**Read merge** — `formData.contractDocuments` / `supportingDocuments` now reflect override effects on every read of a submitted revision. Add-mode docs are appended at end (ordered by override-row `createdAt`); update-mode patches apply by `documentID`.

**Unlock materialization** — overrides bake into real `ContractDocument` rows on the new draft. Add docs land with admin-supplied `dateAdded`; update-mode `dateAdded` is intentionally not carried through (restored at resubmit instead, mirroring rate-side).

**Resubmit preservation** — `submitContractAndOrRates` now walks add-override rows by `sha256` during the `prevDocs` trace, so admin-supplied `dateAdded` survives the full submit → override → unlock → resubmit cycle.

## Refactors

- Introduced `mergeContractRevisionOverrides` / `mergeRateRevisionOverrides` to replace scattered `.find()` and `[0]`-index single-latest patterns. All revision-level overrides now route through one merger per side (4 contract + 1 rate call sites).
- Rate-side override merge: single-latest → full-merge. Previously, the read path consulted only the most recent RateRevisionOverrides row on a revision; all older overrides were invisible. So if Override 1 patched docA's dateAdded and Override 2 was later created (touching a different field, or nothing at all), Override 1's patch silently disappeared. Now all override rows on a revision are walked in createdAt-asc order and merged sparsely — each field's effective value is the newest non-null patch across the full set. Matches the skill doc's stated model.

This removes an accidental side-effect of the old code: a newer override row with all-null fields used to function as a tombstone, reverting earlier overrides on the same revision. Under full-merge, null fields mean "no patch" and earlier non-null patches survive. Removing an override now requires an explicit tombstone mechanism (out of scope).
- Contract/supporting doc override merge passes are now kind-separated to prevent cross-contamination on `documentID` collisions (parallel change in `submitContractAndOrRates` for rate-side too).

## Documentation

`skills/skill-api/references/10-revision-overrides.md` updated:
- New "Document Zip Packages And Overrides" section documenting the deferred zip-regeneration gap (tracked under MCR-5928)
- "Things Easy To Get Wrong" gains an entry on `DocumentZipPackage` not reflecting overrides

## Known limitations (deferred)

- **`DocumentZipPackage` staleness**: zip bundles are not regenerated when an add-mode override is created. The merged read view shows the new doc; the zip download does not include it until the next unlock + resubmit cycle. Update-mode overrides are unaffected (no zip bytes change). Tracked under MCR-5928.

#### Related issues

https://jiraent.cms.gov/browse/MCR-6198 

#### Test cases covered

- `overrideContractData.test.ts` — 4 new tests covering add, update, mixed (add+update in one override row), and multi-override accumulation.
- `unlockContract.override.test.ts` — 4 new tests covering add materialization, update-not-carried-through, multi-override accumulation, and the full submit → override → unlock → resubmit lifecycle with `dateAdded` preservation.
- `fetchRate.test.ts` — assertion rewritten for full-merge semantics (newer override wins per documentID; "remove overrides via no-op patch" no longer works).
- Mock data and inferred-type fixtures updated for new fields on Prisma selects.

## QA guidance

**Testing Overriding ContractType With Raw SQL**

Given a `contractID` and `contractRevisionID`, create a contract revision override by inserting a parent `ContractOverrides` row and a child `ContractRevisionOverrides` row.

This does not update `ContractRevisionTable.contractType`; it creates an override record that the app reads as the effective `contractType`.

```sql
WITH new_override AS (
    INSERT INTO "ContractOverrides" (id,
  "contractID", description, "createdAt")
    VALUES (gen_random_uuid()::text,
  '<CONTRACT_ID>', 'Manual 
  contractType override', NOW())
    RETURNING id
  )
  INSERT INTO "ContractRevisionOverrides" (id,
  "contractOverrideID", "contractRevisionID",
  "contractType", "createdAt")
  SELECT gen_random_uuid()::text, new_override.id,
  '<CONTRACT_REVISION_ID>',
  '<AMENDMENT | BASE>'::"ContractType", NOW()
  FROM new_override;
```
After running this migration you should see the updated contract type being used in the app, eg on the submission summary page

**Testing Adding Contract Documents **
```sql
-- Admin override: add one new contract document and one new supporting
-- contract document to a submission via the override pattern.
--
-- Requires:
--   <ADMIN_USER_ID>            -- User.id of the CMS admin creating the override
--   <CONTRACT_ID>              -- ContractTable.id
--   <CONTRACT_REVISION_ID>     -- ContractRevisionTable.id of the LATEST
--                                 submitted revision on that contract
--   For each doc to add:
--     name, sha256, s3URL (required by the add-mode validator)
--     s3BucketName, s3Key, dateAdded (optional but recommended)

WITH new_override AS (
    INSERT INTO "ContractOverrides" (
        id,
        "createdAt",
        "updatedByID",
        "contractID",
        description
    )
    VALUES (
        gen_random_uuid(),
        NOW(),
        '<ADMIN_USER_ID>',
        '<CONTRACT_ID>',
        'Admin add: missing contract + supporting documents (ticket <TICKET>)'
    )
    RETURNING id
),
new_revision_override AS (
    INSERT INTO "ContractRevisionOverrides" (
        id,
        "createdAt",
        "contractOverrideID",
        "contractRevisionID"
    )
    SELECT
        gen_random_uuid(),
        NOW(),
        id,
        '<CONTRACT_REVISION_ID>'
    FROM new_override
    RETURNING id
),
new_contract_doc_override AS (
    INSERT INTO "ContractDocumentOverride" (
        id,
        "createdAt",
        "contractRevisionOverrideID",
        "documentID",        -- NULL = add mode
        name,
        sha256,
        "s3URL",
        "s3BucketName",
        "s3Key",
        "dateAdded"
    )
    SELECT
        gen_random_uuid(),
        NOW(),
        id,
        NULL,
        '<CONTRACT_DOC_FILENAME.pdf>',
        '<CONTRACT_DOC_SHA256>',
        's3://<BUCKET>/<KEY>',
        '<BUCKET>',
        '<KEY>',
        '<YYYY-MM-DD>'::timestamptz
    FROM new_revision_override
    RETURNING id
),
new_supporting_doc_override AS (
    INSERT INTO "ContractSupportingDocumentOverride" (
        id,
        "createdAt",
        "contractRevisionOverrideID",
        "documentID",        -- NULL = add mode
        name,
        sha256,
        "s3URL",
        "s3BucketName",
        "s3Key",
        "dateAdded"
    )
    SELECT
        gen_random_uuid(),
        NOW(),
        id,
        NULL,
        '<SUPPORTING_DOC_FILENAME.pdf>',
        '<SUPPORTING_DOC_SHA256>',
        's3://<BUCKET>/<SUPPORTING_KEY>',
        '<BUCKET>',
        '<SUPPORTING_KEY>',
        '<YYYY-MM-DD>'::timestamptz
    FROM new_revision_override
    RETURNING id
)
SELECT
    (SELECT id FROM new_override)                  AS override_id,
    (SELECT id FROM new_revision_override)         AS revision_override_id,
    (SELECT id FROM new_contract_doc_override)     AS contract_doc_override_id,
    (SELECT id FROM new_supporting_doc_override)   AS supporting_doc_override_id;
```

## PR Reminders


services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
```diff
@@ -725,10 +747,17 @@ type ContractRevisionTableWithFormData =
 function contractFormDataToDomainModel(
     contractRevision: ContractRevisionTableWithFormData
 ): ContractFormDataType {
+    const revisionOverride = contractRevision.revisionOverrides?.find(
```
JasonLin0991: **Suggestion**: The revision overrides should be merged for the same revision. The intention for multiple overrides for the same revision is to track each override so it is isolated from previous overrides. I.E. override 1, overrode contractType by user ABC, but then later override2 by user 123 overrode populationCoverage. This tracks who override what, when we merge we can get the full override picture.

**Suggestion**: Lets make a single merge function to use everywhere for the contract revision overrides.

The are multiple places where we need to merge,  I think we need a single function to handle the merge. 

I can take an AI to add documentation to the technical-design directory.

services/app-api/src/postgres/contractAndRates/parseContractWithHistory.ts
```diff
@@ -314,12 +333,19 @@ function contractWithHistoryToDomainModel(
 function strippedContractFormDataToDomainModel(
     contractRevision: StrippedContractRevisionTableWithFormData
 ) {
+    const revisionOverride = contractRevision.revisionOverrides?.find(
```
JasonLin0991: **Suggestion**: Same here with merging the overrides.

services/app-api/src/postgres/contractAndRates/unlockContract.ts
```diff
@@ -170,6 +179,12 @@ async function unlockContractInsideTransaction(
         relatedRateIDs.push(ratePackage.rateRevision.rateID)
     }
 
+    const contractTypeOverride = currentRev.revisionOverrides.find(
```
JasonLin0991: **Suggestion**: Same here, lets merge the overrides into one.

services/app-api/src/postgres/contractAndRates/prismaSharedContractRateHelpers.ts
JasonLin0991: **Suggestion**: Could you fix my work here too? This should also merge the revisions and not just the first. This one could be complex because of the documents.

JasonLin0991 (approved):
This looks great. 

I highly suggest we just handle merging the revision overrides now. I know we only have 1 field now and merging doesn't make sense, but once we introduce more than 1 field it would be nice to have this merge in place.

JasonLin0991:
Closing this PR in favor of this new [PR](https://github.com/Enterprise-CMCS/managed-care-review/pull/4542)

PR #4437, opened 2026-05-12
MCR-6079: Add rotation notifier Lambda

Btodhunter:
* Added notifier lambda for resetting running lambda environments when db password is rotated
  *  Updates the ROTATION_TIMESTAMP env var on all DB connected lambdas, resetting warm lambda environments
* Added event bridge rule for secret manager secret rotation, used to trigger notifier lambda

### DIFF
--- a/services/infra-cdk/lib/stacks/app-api.ts
+++ b/services/infra-cdk/lib/stacks/app-api.ts
@@ -33,7 +33,7 @@ import { isReviewEnvironment } from '../config/environments'
 import { Architecture, Runtime } from 'aws-cdk-lib/aws-lambda'
 import { CfnWebACL, CfnWebACLAssociation } from 'aws-cdk-lib/aws-wafv2'
 import { SubnetType, Vpc, SecurityGroup } from 'aws-cdk-lib/aws-ec2'
-import { Rule, Schedule } from 'aws-cdk-lib/aws-events'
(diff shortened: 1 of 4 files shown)


services/infra-cdk/lib/stacks/app-api.ts
```diff
@@ -1154,6 +1159,107 @@ export class AppApiStack extends BaseStack {
...
+    const ts = new Date().toISOString();
+    await Promise.all(names.map(async name => {
+        const cfg = await client.send(new GetFunctionConfigurationCommand({ FunctionName: name }));
+        const vars = cfg.Environment?.Variables ?? {};
+        await client.send(new UpdateFunctionConfigurationCommand({
+            FunctionName: name,
+            Environment: { Variables: { ...vars, ROTATION_TIMESTAMP: ts } }
+        }));
+        console.log('Updated ROTATION_TIMESTAMP for', name);
+    }));
```
mojotalantikite: This is a small one so it might not be a big deal, but I worry about having a lambda that is outside of the typical pattern of putting lambdas in `app-api/src/handlers`. Just so someone else will know where to find the lambda, and we're not skipping running linters etc on the code.

mojotalantikite: Another option would be to put it in `postgres` service where most of the db related lambda funcs live

mojotalantikite (approved):

PR #4463, opened 2026-05-15
MCR-6265: fix the submission type page to work if the SDP flag is on independently

pearl-truss:
## Summary

Bryan ran into an issue on Val where he hit a submission error when trying to progress pass the Submission Type page. On Val the SDP flag was on and the EQRO flag is off. Previously the code only called createContract if the new submission form was detected based on the EQRO flag. This PR updates the logic to call createContract (and not updateContract) when the either the EQRO or the SDP flag is on (since these flags should be fully independent of each other)

#### Related issues

https://jiraent.cms.gov/browse/MCR-6265

## QA guidance

1. Turn the SDP flag on and the EQRO flag off. 
2. Select to create a new Health plan submission
3. Fill in the submission type page and click continue
4. you should be able to progress pass the Submission Type page without error

## PR Reminders
- [x ] Updated the API Changelog (if this PR changes the API schema)
- [ x] Checked accessibility with ANDI and Wave tools as needed

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -52,7 +52,7 @@ export IAM_PATH='/'
 export CLOUDFRONT_CERT_ARN='local'
 export CLOUDFRONT_SB_DOMAIN_NAME='local'
 export CLOUDFRONT_DOMAIN_NAME='local'
-export SECRETS_MANAGER_SECRET='local-secrets-manager' // pragma: allowlist secret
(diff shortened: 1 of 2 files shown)


vyakusha (approved):
LGTM. Tested locally. All works as expected

PR #4488, opened 2026-05-18
MCR-6223: Handle race conditions for mutations

JasonLin0991:
## Summary
[MCR-6223](https://jiraent.cms.gov/browse/MCR-6223)

This implements a new way of handling race conditions, [Postgres row-lock](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS). In the ticket, we had mentioned using Prisma's ioslationLevel of Serializable for transactions, but Serializable in Postgres doesn't actually lock rows — it watches which rows each transaction reads and writes, and aborts one with a serialization error if it detects overlap. The catch: that overlap check is approximate. Two transactions updating unrelated contracts can still be flagged as conflicting if they happen to scan the same parts of the underlying table or index, even though they share no actual rows.

I.E. Updating two different contracts and the same time would result in an error, but actually we want that to be allows.

So this new approach locks the row of the record while the transaction is happening and releases it after transaction finishes or rollback.

I only applied this to select updates as not all of them need this row lock. It mostly revolves around contract and rate updates like submit, unlock. withdraw, ...etc.
- approveContract
- overrideRateData
- reverseApproveContract
- submitContract
- submitRate
- undoUnlockContract
- undoWithdrawContract
- undoWithdrawRate
- unlockContract
- unlockRate (unused, but implemented anyways)
- updateDraftContract
- updateDraftContractRates
- updateDraftContractWithRates
- updateMCCRSID
- withdrawContract
- withdrawRate
- softDeleteContractQuestion

I skipped some other resolvers, like create Q&A question and response. This row lock won't prevent those from being able to create duplicates on concurrent API calls. 

#### Test cases covered

`prismaHelpers.test.ts`
- `'runTransactionWithRowLock'`
   - `'prevents concurrent updates on the same locked contract row'`
   - `'does not block concurrent work on different locked contract rows'`

## QA guidance
- Validate that each listed action in the description still works in the app.
- (Difficult) Try to send two action requests at the same time in the app.

## PR Reminders

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/approveContract.ts
+++ b/services/app-api/src/postgres/contractAndRates/approveContract.ts
@@ -21,13 +21,25 @@ async function approveContractInsideTransaction(
             id: contractID,
         },
         include: {
-            reviewStatusActions: true,
+            reviewStatusActions: {
+                orderBy: {
+                    updatedAt: 'desc',
+                },
+                take: 1,
+            },
         },
     })
 
     if (!contract) {
         return new NotFoundError(
             `PRISMA ERROR: Cannot find contract with id: ${contractID}`
+        )
+    }
+
+    const latestReviewStatusAction = contract.reviewStatusActions[0]
+    if (latestReviewStatusAction?.actionType === 'MARK_AS_APPROVED') {
+        return new Error(
+            'Cannot approve contract: contract is already approved'
         )
     }
 
--- a/services/app-api/src/postgres/contractAndRates/approveContract.ts
+++ b/services/app-api/src/postgres/contractAndRates/approveContract.ts
@@ -3,10 +3,7 @@ import { NotFoundError } from '../postgresErrors'
 import type { ContractType } from '../../domain-models'
 import type { PrismaTransactionType } from '../prismaTypes'
 import type { ExtendedPrismaClient } from '../prismaClient'
-import {
-    lockContractRowForUpdate,
-    runTransactionWithRowLock,
-} from '../prismaHelpers'
+import { runTransactionWithRowLock } from '../prismaHelpers'
 
 async function approveContractInsideTransaction(
     tx: PrismaTransactionType,
@@ -73,7 +70,8 @@ async function approveContract(
     return runTransactionWithRowLock({
         client,
         operationName: 'approveContract',
-        lock: async (tx) => await lockContractRowForUpdate(tx, args.contractID),
+        table: 'ContractTable',
+        id: args.contractID,
         transaction: async (tx) =>
             await approveContractInsideTransaction(tx, args),
     })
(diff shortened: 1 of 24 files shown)


services/app-api/src/postgres/contractAndRates/unlockContract.ts
```diff
@@ -319,19 +316,20 @@ async function unlockContract(
...
+    return runTransactionWithRowLock({
+        client,
+        operationName: 'unlockContract',
+        table: 'ContractTable',
+        id: args.contractID,
+        transaction: async (tx) => {
             const result = await unlockContractInsideTransaction(tx, args)
             if (result instanceof Error) {
                 throw result
             }
```
mojotalantikite, line 328: Looks like everywhere else dropped the check here since it's taken care of in the helper?

services/app-api/src/postgres/contractAndRates/approveContract.ts
```diff
@@ -15,49 +15,57 @@ async function approveContractInsideTransaction(
...
-                updatedByID: updatedByID,
-                dateApprovalReleasedToState: dateApprovalReleasedToState,
-                updatedReason,
-                actionType: 'MARK_AS_APPROVED',
-                contractID: contractID,
-            },
-        })
+    const latestReviewStatusAction = contract.reviewStatusActions[0]
+    if (latestReviewStatusAction?.actionType === 'MARK_AS_APPROVED') {
+        return new Error(
```
mojotalantikite: do we want to `return` or `throw` on Error here? the pattern has seemed to be `throw`, which iirc is for the transaction to be triggered for a roll-back

mojotalantikite (approved):

services/app-api/src/postgres/contractAndRates/unlockContract.ts
JasonLin0991: Yea, the `runTransactionWithRowLock` runs all transactions inside a try catch block and handles returning errors from the transaction. I think now it should catch any errors and roll back. Previously, a couple of these functions took the return and passed it back to the resolver without throwing.

services/app-api/src/postgres/prismaHelpers.ts
```diff
@@ -0,0 +1,98 @@
...
+        if (
+            err instanceof NotFoundError ||
+            err instanceof UserInputPostgresError
+        ) {
+            return err
+        }
+
+        console.error(`Prisma error ${args.operationName}`, err)
+        return parseErrorToError(err)
+    }
```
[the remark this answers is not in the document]
JasonLin0991: This is a hard question to answer. While we could use UserInputPostgresError for things like `Cannot withdraw rate: rate is already withdrawn` technically I could be incorrect if the user had request the withdrawal at the same time as another, but was locked out. Then when it finally ran it was withdrawn. I think I will keep this like this for now to revisit later.

services/app-api/src/postgres/contractAndRates/approveContract.ts
JasonLin0991: Good catch.

PR #4499, opened 2026-05-19
MCR-6233: Add additional logging for auth

JasonLin0991:
## Summary
[MCR-6233](https://jiraent.cms.gov/browse/MCR-6233): Add additional logging for Auth requests
[MCR-6254](https://jiraent.cms.gov/browse/MCR-6254): Update UnlockContract resolver to restrict by status 

This work adds additional logging and context into logs for the api.

- Adds more context to error logs in third_party_API_authorizer.ts
- Adds more logging and more context into logs in`apollo_gql.ts` - contextForRequestForFetcher
   - Added throwing and log errors for OAuth clients not permitted for delegated requests. Previously we did not do a check to see if the OAuth user can make a delegated request at this level. We did however gate this using the `canOAuthWrite` function in the resolvers themselves, but this leaves a gap as requests can impersonate a user to read certain endpoints.
   -  Added more context about the auth and user data in logs. This will show things like auth method and token data.
- Refactored resolver logging to add more details into the log
   - `logResolverError`: New function to log resolver errors. This requires context from the dep injection to add more data into the log.
   - `logResolverSuccess: New function to log resolver success. Same as above, this requires context to add more data into the log.
   - Replaced all uses of `logError` and `logSuccess` in resolvers.
- Fixed up gaps in logging in the resolver.
   - Several errors where thrown but not logged
   - Several resolvers throw standard Errors in the resolver instead of GraphQL errors.

Non-logging work
- Adding a check for delegated request permission means our third party request cypress test failed because it was making a delegated request using a client without permission.
   - Fix this by updating createOAuthClient to also take in the `scopes` input to add the write scope of `CMS_SUBMISSION_ACTIONS` scope that we use for delegated write auth.
   - Updated `updateOauthClient` so it also can update the `scopes` of a client.
- UnlockContract resolver restricted by status [MCR-6254](https://jiraent.cms.gov/browse/MCR-6254).

Out of scope:
- There where a few AC's that were originally in the ticket, but I had either removed them completely or move them to a new ticket.
   - Logging inputs: After thinking about it, this seems like a bad practice as some inputs could be PII in the future so we should not log that, but I can be convinced otherwise.
   - Alerting: There are two AC's around alerting for unusual behavior. This is more in the realm of the Otel tracing work so I move this to a new ticket [MCR-6269 ](https://jiraent.cms.gov/browse/MCR-6269)

## PR Reminders


services/app-api/src/logger/logger.ts
```diff
@@ -6,6 +16,27 @@ function logSuccess(operation: string) {
...
+    context: Pick<Context, 'user' | 'oauthClient'>
+) {
+    console.info({
+        message: `${operation} succeeded`,
+        operation: operation,
+        status: 'SUCCESS',
+        requestContext: createRequestContext(context),
+    })
+}
+
```
mojotalantikite, line 39: One enhancement here that I think would be good, since I've been working on OTEL -> datadog stuff, would be to add the trace and span id's to the logger. That way we'll have a way to correlate a trace to a log. Right now we're not shipping the logs off to datadog, but it does support ingesting these cloudwatch logs and having these correlations could get us a lot of extra context on what's going on. Something like:

```typescript
import { trace } from '@opentelemetry/api'

  function getTraceContext() {
      const span = trace.getActiveSpan()
      if (!span) return {}
      const { traceId, spanId } = span.spanContext()
      // Datadog expects decimal, OTEL gives hex — needs conversion
      return {
          'dd.trace_id': BigInt(`0x${traceId.slice(16)}`).toString(),
          'dd.span_id': BigInt(`0x${spanId}`).toString(),
      }
  }
```

And then in the logger func:
```typescript
  function logResolverError(
      operation: string,
      error: Error | string,
      context: Pick<Context, 'user' | 'oauthClient'>
  ) {
      console.error({
          message: `${operation} failed`,
          operation: operation,
          status: 'ERROR',
          error: error,
          requestContext: createRequestContext(context),
          ...getTraceContext(),
      })
  }
```

mojotalantikite: I'll add an RSP topic around this, because we should maybe even think about using a real structured logging lib so we can set log levels (`warn` vs `info` vs `debug`) and whatnot.

services/app-api/src/resolvers/contract/contractRevisionResolver.ts
```diff
@@ -19,15 +20,22 @@ export function contractRevisionResolver(
             const stateCode = parent.contract.stateCode
             const programsForContractState = store.findStatePrograms(stateCode)
             if (programsForContractState instanceof Error) {
-                throw programsForContractState
+                const errMessage = `Failed to fetch state programs. ${programsForContractState.message}`
+                logError('contractRevisionStrippedResolver', errMessage)
```
mojotalantikite: Is there a time/guidance on when to use `logError` vs `logResolverError`?

mojotalantikite (approved):

services/app-api/src/resolvers/contract/unlockContract.ts
```diff
@@ -72,11 +76,12 @@ export function unlockContractResolver(
         }
 
         if (
-            contractResult.draftRevision ||
-            contractResult.consolidatedStatus === 'APPROVED'
+            !['SUBMITTED', 'RESUBMITTED', 'NOT_SUBJECT_TO_REVIEW'].includes(
```
pearl-truss, line 79: nice for handling this here! can you include tracking for https://jiraent.cms.gov/browse/MCR-6254 - I left a comment there

JasonLin0991: I moved it into the sprint and put it in review.

services/app-api/src/resolvers/contract/contractRevisionResolver.ts
JasonLin0991: I missed the field level resolvers. Just added them in a recent commit. The JSDoc comments for `logResolverError` and `logResolverSuccess` is some guidance.

services/app-api/src/logger/logger.ts
JasonLin0991: Yea this would be great thanks!

PR #4535, opened 2026-05-29
MCR-6279 Fix links to submission form guidance page

vyakusha:
## Summary
Links to the submission form guidance page should scroll to the different sections of the page depending on context
#### Related issues
[MCR-6279](https://jiraent.cms.gov/browse/MCR-6279)
## QA guidance
Click on links to submission form guidance page. See if it scrolls to the section noted in the link.

## PR Reminders

### DIFF
--- a/services/app-web/src/pages/App/AppRoutes.tsx
+++ b/services/app-web/src/pages/App/AppRoutes.tsx
@@ -77,24 +77,6 @@ import { CreateOauthClient } from '../Settings/Oauth/CreateOauthClient'
 import { User } from '../../gen/gqlClient'
 import { AddLocalUser } from '../../localAuth/AddLocalUser'
 
-const RedirectPreservingHash = ({
-    pathname,
-}: {
-    pathname: string
-}): React.ReactElement => {
-    const location = useLocation()
-
-    return (
-        <Navigate
-            to={{
-                pathname,
-                hash: location.hash,
-            }}
-            replace
-        />
-    )
-}
-
 const CMSUploadQuestionsRoute = ({
     loggedInUser,
     children,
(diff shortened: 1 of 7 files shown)


services/app-web/src/pages/App/AppRoutes.tsx
```diff
@@ -77,6 +77,24 @@ import { CreateOauthClient } from '../Settings/Oauth/CreateOauthClient'
 import { User } from '../../gen/gqlClient'
 import { AddLocalUser } from '../../localAuth/AddLocalUser'
 
+const RedirectPreservingHash = ({
```
JasonLin0991: **Suggestion**: This works, but I think the core issue here is that the help text links on the form pages are routing to the old url paths.

```ts
<ReactRouterLinkWithLogging
    variant="external"
    to={{
        pathname: '/help',
        hash: '#submission-description',
    }}
    target="_blank"
    className="display-inline-block"
>
    View description examples
</ReactRouterLinkWithLogging>
```

I would update those links so they are pointing to the correct paths instead of adding a wrapper with a hook to fix this. We want the fix to be direct to the core issue so anyone looking at the link knows where it is going instead of having to know that there is a wrapper that resolves the mismatch the link to the actual path.

vyakusha: Ok, I updated relevant links. Agree, it makes for more straightforward fix. Wrapper with a hook was an attempt to localize the fix to one place.

JasonLin0991 (approved):
Looks good.

There is one page that is not used in the app, but has these links `SingleRateCert.tsx`. Lets update that too incase one day we enable it.

PR #4539, opened 2026-06-02
MCR-6279 Fix routing for the help documentation page

vyakusha:
## Summary
In the previous ticket all existing links to help documentation are updated to reflect new routing: "/resources/help". In this ticket the code to silently promote "/help" to "/resources/help" is removed, since it is no longer needed. 

Also, when navigating to 404 page for logged in user (wrong url entered by user), the nav bar was briefly displayed due to user having undefined Auth status for a moment. Fixed that, which also takes care or MCR-6234

#### Related issues
[MCR-6279](https://jiraent.cms.gov/browse/MCR-6279)
[MCR-6234](https://jiraent.cms.gov/browse/MCR-6234)
## QA guidance
Now links: "/help", "/training" are treated as invalid and result in "404" page. Links "/resources/help" and "/resources/training" are recognized.

## PR Reminders

### DIFF
--- a/services/app-web/src/components/Header/Header.test.tsx
+++ b/services/app-web/src/components/Header/Header.test.tsx
@@ -176,6 +176,24 @@ describe('Header', () => {
                 ).not.toBeInTheDocument()
             })
         })
+
+        it('does not show public navigation while auth is still loading on an unknown route', async () => {
+            renderWithProviders(<Header authMode={'AWS_COGNITO'} />, {
+                routerProvider: {
+                    route: '/submissions/new/health-plan/stingray',
+                },
+                apolloProvider: {
+                    mocks: [],
+                },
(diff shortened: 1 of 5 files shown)


ruizajtruss (changes requested):
New links work like a charm but the invalid links just redirect to the homepage with a 403 error in the console instead of a 404 error page. If this was intended then ignore my comment, everything else lgtm 🚀

vyakusha:
> New links work like a charm but the invalid links just redirect to the homepage with a 403 error in the console instead of a 404 error page. If this was intended then ignore my comment, everything else lgtm 🚀 

@ruizajtruss I did fix it. Now for unauthenticated users/pages faulty route returns to the home page.

ruizajtruss (approved):
LGTM!

PR #4554, opened 2026-06-08
MCR-6033 Hide Add question button for Rate if some questions don't have answers

vyakusha:
## Summary
This is a follow-up for vy-mcr-6033-hide-add-questions-button-if-open-round-exists. Fixed logic for Rate Questions page, which was overlooked in the first PR
#### Related issues
[MCR-6033](https://jiraent.cms.gov/browse/MCR-6033)
#### Screenshots
Screenshot 2026-06-08 at 11 04 12 AM
Screenshot 2026-06-08 at 11 05 37 AM

#### Test cases covered
RateQuestionResponse.test
expect last question in round 1 to be DMCP question, and Add question not to be in the document
renders Add questions button when all existing rate questions have responses

## QA guidance
As a State user, create new Rate record. As CMS user, add question for the Rate. "Add questions" button disappears if there are questions which don't have answers, and re-appears once State user creates answers.

## PR Reminders


ruizajtruss (approved):
Looks great! Just a couple small changes I would suggest:

1. You can use `IndexQuestionType` to extract the `canAddQuestions` logic into a reusable function we can export from `questionResponseHelpers.ts` (maybe name it `allQuestionsAnswered`)
2. You can drop the ternary since the logic already returns a boolean

My suggestion:
`  const allQuestionsAnswered = (questions?: IndexQuestionType): boolean =>
      extractQuestions(questions).every(
          (question) => question.responses.length > 0
      )`

PR #4560, opened 2026-06-09
[MCR-6301] Script for generating dsnp submission report

ruizajtruss:
**DSNP submissions report**

  Adds a script to extract all DSNP contracts from the database and generate a `.csv` report.

  **What it does**
  - Pulls every submitted contract that is DSNP as of its latest submitted revision.
  - Output columns: Contract Name, State, contractID, federalAuthorities, rateMedicaidPopulations, dsnpContract, Initial Submission Date, Latest Submission Date.

  **Usage** — run from `scripts/dsnp-report/`:

  | Command | Result |
  |---|---|
  | `./generate-dsnp-report` | All DSNP contracts → `~/Documents/DSNP_Contracts_in_MC-Review.xlsx` |
  | `./generate-dsnp-report 2025-11-04` | Submitted on/after a date |
  | `./generate-dsnp-report 2025-11-04 2025-12-31` | Within a date range |
  | `./generate-dsnp-report 2025-11-04 --by latest` | Filter window on latest (vs initial) submission |


scripts/dsnp-report/dsnp_contracts.sql
```diff
@@ -0,0 +1,84 @@
...
+-- revision. All form data (federalAuthorities, programIDs, rate populations) is
+-- read from that same latest submission, so every field reflects the most
+-- up-to-date version. Contracts that have never been submitted are excluded.
+--
+-- Optional date filtering via the psql vars :from_date / :to_date (an empty
+-- string means unbounded). Pass both for a range, just :from_date for "that date
+-- onward", or neither for everything. :basis selects which submission date the
+-- window applies to: 'initial' (default) or 'latest'.
+
+WITH submitted_revs AS (
```
JasonLin0991: **Suggestion)**: Personally I like just spinning up a unit test to pull the local DB using the resolvers, filtering on what I want, then dumping the csv. This is because using the resolvers we know they are tested for pulling data correctly and the same as what the web app and CMS app uses. With this complex SQL code we don't know how brittle it could be and we don't have CI/CD running tests against it. If it was a simpler SQL code or not other alternatives I would go with SQL.

We could easily just add the unit test into the repo, toggled off using .skip, then run it when we want the data locally without having the scripts (edited). I will defer to @pearl-truss or maybe a RSP topic on this.

services/app-api/src/scripts/generate-dsnp-report
```diff
@@ -0,0 +1,43 @@
...
+
+export NODE_NO_WARNINGS=1
+export DSNP_REPORT=1
+export DSNP_BY="$by"
+[[ ${#positional[@]} -ge 1 ]] && export DSNP_FROM="${positional[0]}"
+[[ ${#positional[@]} -ge 2 ]] && export DSNP_TO="${positional[1]}"
+[[ -n "$out" ]] && export DSNP_OUT="$out"
+
+cd "$APP_API"
+exec ./node_modules/.bin/vitest run --silent=true src/scripts/dsnpReport.test.ts
```
JasonLin0991, line 43: **Question**: If we are keeping this script, idk if my suggestion holds about running it though a test. Can you run the generateDnspReport.ts directly?

JasonLin0991 (commented):
I think this looks good, but I would like a second pair of eyes on this from @pearl-truss and @mojotalantikite on this approach.

services/app-api/src/scripts/generateDsnpReport.ts
```diff
@@ -0,0 +1,165 @@
...
+            formData.programIDs,
+            findStatePrograms(contract.stateCode)
+        ),
+        state: STATE_NAMES.get(contract.stateCode) ?? contract.stateCode,
+        contractID: contract.id,
+        federalAuthorities: (formData.federalAuthorities ?? []).join(','),
+        rateMedicaidPopulations: POP_ORDER.filter((p) => pops.has(p)).join(','),
+        dsnp: formData.dsnpContract === true,
+        initialDate: ymd(initial.submitInfo.updatedAt),
+        latestDate: ymd(latest.submitInfo.updatedAt),
```
JasonLin0991: **Issue**: This doesn't cover review actions. I think you have to calculate this like how `contractResolver` does.

services/app-api/src/scripts/generateDsnpReport.ts
```diff
@@ -0,0 +1,165 @@
...
+            contract.stateNumber,
+            formData.programIDs,
+            findStatePrograms(contract.stateCode)
+        ),
+        state: STATE_NAMES.get(contract.stateCode) ?? contract.stateCode,
+        contractID: contract.id,
+        federalAuthorities: (formData.federalAuthorities ?? []).join(','),
+        rateMedicaidPopulations: POP_ORDER.filter((p) => pops.has(p)).join(','),
+        dsnp: formData.dsnpContract === true,
+        initialDate: ymd(initial.submitInfo.updatedAt),
```
JasonLin0991: **Issue**: This doesn't cover overrides, <strike>there should be a initiallySubmittedAt field at the contract level of the object</strike>. Nvm that is at the resolver level. I think you might have to recalculate this like how `contractResolver` does.

services/app-api/src/scripts/generateDsnpReport.ts
JasonLin0991: Unless these are dates are are a entirely different purpose

services/app-api/src/scripts/generate-dsnp-report
ruizajtruss: To run it directly it would take extra work that's not really worth it for the one use script. Running it with vitest allows the packages to resolve and actually work.

services/app-api/src/scripts/generateDsnpReport.ts
pearl-truss: +1 I think we need to use the more holistic approach for latestDate here and I think that's one benefit to having a script for generating this report - we can encapsulate considerations that are easy to forget to make sure the data is correct

pearl-truss:
@ruizajtruss we just got feedback from Lynn regarding displaying VOLUNTARY for one of the federal authority options (I think the full value is 1915(a) voluntary...". Could you look into making it so that this report uses more user friendly values for the federal authority column?

services/app-api/src/scripts/generateDsnpReport.ts
```diff
@@ -0,0 +1,185 @@
...
+            .map((p) => RateMedicaidPopulationsRecord[p] ?? p)
+            .join(', '),
+        dsnp: formData.dsnpContract === true,
+        initialDate: ymd(
+            resolveInitiallySubmittedAtOverride(
+                initial.submitInfo.updatedAt,
+                contract.contractOverrides
+            )
+        ),
+        latestDate: ymd(
```
JasonLin0991, line 106: TODO: We can use the `lastActionDate` field now and fallback on this logic if it is undefined. `lastActionDate` should not be undefined for submissions that have been submitted at least once.

JasonLin0991 (approved):
Looks good, thanks for the changes. I just left 1 last comment for the latestDate before you merge this.

PR #4561, opened 2026-06-09
MCR-6274: field level resolvers in otel

JasonLin0991:
## Summary
[MCR-6274](https://jiraent.cms.gov/browse/MCR-6274)

This updates our field level resolvers to also use `withResolverSpan` function to set spans for each resolver. In addition I added more data into the span attributes about the request in apollo_gql handler.

Added:
- `http.request_id`: APIGatewayProxyEvent request id
   - This will also be logged in `logResolverError` and `logResolverSuccess`
- `graphql.operation.name`: Extracted GraphQL operation from the request body
- `mcreview.auth_method`: Auth method of the request.
   - `OAuth 2.0` is the external API auth method
   - Cognito is our frontend app

With this we should be able to see what GQL operation the field resolver failed on in traces.

## QA guidance
- Unsure how to test this @mojotalantikite any ideas?

## PR Reminders

### DIFF
--- a/services/app-api/src/handlers/apollo_gql.ts
+++ b/services/app-api/src/handlers/apollo_gql.ts
@@ -138,6 +138,7 @@ function contextForRequestForFetcher(
                     'http.method': event.httpMethod,
                     'http.url': event.path,
                     'http.route': event.requestContext.resourcePath,
+                    'http.request_id': event.requestContext.requestId,
                 },
             },
             parentContext // Use the extracted context as the parent
@@ -164,6 +165,12 @@ function contextForRequestForFetcher(
                 }
 
                 if (fromThirdPartyAuthorizer) {
+                    const authMethod = 'OAuth 2.0'
+                    const operationName = extractOAuthOperationName(event)
+                    requestSpan.setAttributes({
+                        'graphql.operation.name': operationName ?? 'anonymous',
+                        'mcreview.auth_method': authMethod,
+                    })
                     const principalId =
                         event.requestContext.authorizer?.principalId
                     const authorizerContext = event.requestContext.authorizer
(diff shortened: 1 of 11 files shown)


vyakusha (commented):
"The new helper marks swallowed resolver failures as successful spans, so the new field-level tracing will hide the very errors this branch is trying to surface. In both zip-package resolvers, failures are intentionally logged and converted to [] via recordResolverError(...) (contractRevisionResolver.ts (line 83), rateRevisionResolver.ts (line 65)). But withResolverSpan() always sets SpanStatusCode.OK on any non-throwing return (attributeHelper.ts (line 98)). Result: these traces will contain an exception event but still look healthy in span-status-based dashboards and alerts." This brings a question for @mojotalantikite : Is there a reasonable way to test the ground truth in DataDog?

mojotalantikite:
> "The new helper marks swallowed resolver failures as successful spans, so the new field-level tracing will hide the very errors this branch is trying to surface. In both zip-package resolvers, failures are intentionally logged and converted to [] via recordResolverError(...) (contractRevisionResolver.ts (line 83), rateRevisionResolver.ts (line 65)). But withResolverSpan() always sets SpanStatusCode.OK on any non-throwing return (attributeHelper.ts (line 98)). Result: these traces will contain an exception event but still look healthy in span-status-based dashboards and alerts." This brings a question for @mojotalantikite : Is there a reasonable way to test the ground truth in DataDog?

Ok this is a good catch, thanks! It turns out that if you set `OK` on `success` paths in OTEL it can swallow the error set in the resolver. Also, we need to set `code: ERROR` in `recordResolverError`  (not just `recordException`) so the non-fatal/swallowed path actually flags the span. This is a consequence of sometimes `throw`-ing and sometimes `return`-ing for errors in our code. I've made a couple changes for that and pushed it up here.

mojotalantikite (approved):
LGTM

PR #4569, opened 2026-06-11
Add OTEL to impersonated user paths

mojotalantikite:
## Summary

Adds OTEL and monitors to API paths that are acting on behalf of other users. 

#### Related issues

https://jiraent.cms.gov/browse/MCR-6269


JasonLin0991 (approved):
I'll approve, but I think copilot had some good comments in here too.

PR #4576, opened 2026-06-12
[MCR-6296] restrict open question rounds at the API level

pearl-truss:
## Summary

This PR updates the createContractQuestion and createRateQuestion resolvers to reject a new question while a previous question round is still open. When an open round exists, the API returns a BAD_USER_INPUT error.

#### Related issues

https://jiraent.cms.gov/browse/MCR-6296

#### Test cases covered

createContractQuestion.test.ts

'returns an error if there is an open question round (unanswered questions)'

- Asserts a second createContractQuestion on the same contract fails with BAD_USER_INPUT.

'allows a new question once the open round has been answered'

- Asserts a new question succeeds after the prior question is answered.

'allows question creation on UNLOCKED and RESUBMITTED package'

- Updated to answer the first question before creating the second.

'send CMS email to state analysts with correct round number if multiple questions have been asked'

- Updated to answer each question between rounds.

## QA guidance

- Pick a submitted contract to add a question to (easier to do this in the UI)
- Without submitting a response to that question, call createContractQuestion again via the graphql explorer on the same contract. Expect a BAD_USER_INPUT error 
- Submit a response to the first question (createContractQuestionResponse), then call createContractQuestion again.  Expect success

## PR Reminders

### DIFF
--- a/API_Changelog.md
+++ b/API_Changelog.md
@@ -7,6 +7,7 @@
     - A question round is considered open when any existing question on the same contract (for `createContractQuestion`) or rate (for `createRateQuestion`) has not yet received a response.
     - All existing questions must be answered before a new question can be created. 
     - New `BAD_USER_INPUT` error when createContractQuestion or createRateQuestion are called when an open round exists
+    - The open-round check is enforced at the database level with a row lock on the parent contract or rate, so two concurrent requests cannot both create a question round.
 
 ### May 28, 2026
 #### Updated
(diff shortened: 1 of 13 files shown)


vyakusha (approved):
I looked through the code and tested functionality locally. GraphQL shows "BAD_USER_INPUT" when trying to run mutation CreateContractQuestion, if response to the one of the question rounds does not exist.

PR #4582, opened 2026-06-15
MCR-6306 Add filtering to state dashboard

vyakusha:
## Summary
The following filter components are added to the dashboard
Submission type
Programs
Note: we have not filtered on Programs previously
Status
The default filtering should be set to not filtered
Each of the filters will function as a dropdown menu, using the existing functionality that is already in place for the CMS user dashboard filtering
The submissions which are displayed will be updated based upon the filter options which are selected
The user can change the filtering options and the submissions which are displayed will be updated
When a filter is applied, there will be an "X" in the right hand section of the filtering dropdown which will clear all filters 
when selected
The "Contract type" filter should remain behind the EQRO feature flag 

#### Related issues
[MCR-6306](https://jiraent.cms.gov/browse/MCR-6306)

#### Screenshots
Screenshot 2026-06-16 at 11 06 40 AM

Screenshot 2026-06-16 at 11 06 16 AM

Note: State user filter include "Draft" as one of the Status options, and CMS user does not have "Draft" in the Status dropdown.

#### Test cases covered
ContractTable.test.tsx 
'can filter table by programs for state users'
StateDashboard.test.tsx 
'displays state dashboard filters with contract type and status when eqro submissions are enabled'
'displays only status filter on state dashboard when eqro submissions are disabled'

## QA guidance
Log in as the state user and observe filters behavior.

## PR Reminders

### DIFF
--- a/services/app-web/src/components/ContractTable/ContractTable.tsx
+++ b/services/app-web/src/components/ContractTable/ContractTable.tsx
@@ -41,6 +41,7 @@ import {
     featureFlags,
     stateNameToStateCode,
 } from '@mc-review/common-code'
+import { getAvailableContractPrograms } from '../../formHelpers'
 import { formatCalendarDate } from '@mc-review/dates'
 import { RowCellElement } from '..'
 import { useLDClient } from 'launchdarkly-react-client-sdk'
@@ -381,9 +382,9 @@ export const ContractTable = ({
 
     const [tableCaption, setTableCaption] = useState<React.ReactNode | null>()
 
-    const isNotStateUser = user.__typename !== 'StateUser'
-    const isStateUser = !isNotStateUser
-    const statusFilterOptions = isStateUser
+    const isStateUser = user.__typename === 'StateUser'
+    const isNotStateUser = !isStateUser
+    const statusFilterOptions = !isNotStateUser
         ? stateSubmissionStatusOptions
         : submissionStatusOptions
     const tableColumns = React.useMemo(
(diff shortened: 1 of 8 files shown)


services/app-web/src/components/ContractTable/ContractTable.tsx
```diff
@@ -656,95 +695,162 @@ export const ContractTable = ({
...
+                                    <MultiColumnGrid columns={2}>
+                                        <FilterSelect
+                                            value={getSelectedFiltersFromUrl(
+                                                columnFilters,
+                                                'status'
+                                            )}
+                                            name="status"
+                                            label="Status"
+                                            filterOptions={
+                                                submissionStatusOptions
```
pearl-truss: For state users, I think they should be able to filter for draft submissions as well - that's one difference between CMS vs state users. Only state users can see draft submissions 

Image

pearl-truss (commented):
Nice work! I just left one comment

services/app-web/src/components/ContractTable/ContractTable.tsx
vyakusha: Good catch. Thanks. I added "Draft" to the state user Status filter.

services/app-web/src/components/ContractTable/ContractTable.tsx
```diff
@@ -538,6 +564,30 @@ export const ContractTable = ({
...
+    )
+        .sort()
+        .map((contractType) => ({
+            value: contractType,
+            label:
+                contractTypeOptions.find((opt) => opt.value === contractType)
+                    ?.label ?? contractType,
+        }))
+
+    const programFilterOptions = Array.from(
```
JasonLin0991: **Nit**: Maybe we can ask Anna, but I would suggest we show all the programs available to the state for filtering even if some are not currently on any submission. Because status display all status not only ones currently exists in the submissions we should align the same behavior for programs.

services/app-web/src/components/ContractTable/ContractTable.tsx
```diff
@@ -538,6 +564,30 @@ export const ContractTable = ({
             label: state,
         }))
 
+    const contractTypeFilterOptions = Array.from(
```
JasonLin0991: **Nit**: I think this is doing the same as programs filter options? If so, we should also show both contract types at all times.

JasonLin0991 (approved):
LGTM, left a couple of nit picks

services/app-web/src/components/ContractTable/ContractTable.tsx
vyakusha: Let't talk to Anna. State filter dropdown on the CMS user dashboard is also displaying only those options which are present in the submissions table.

pearl-truss:
That's correct for the CMS dashboard - they don't see draft submissions

services/app-web/src/components/ContractTable/ContractTable.tsx
vyakusha: Added logic to the Programs dropdown to populate it with all available program names.
Reverted Contract type dropdown population to use the same logic as CMS user does.

Screenshot 2026-06-18 at 3 48 43 PM

PR #4599, opened 2026-06-23
Admin can delete q&a response

JasonLin0991:
## Summary

This work allows admin to delete contract question responses.

Notes:
- The DB work was already there, this just built the API and UI.
   - Deleting a response adds a`DELETE` action to its action log and attached documents get a `CASCADE_DELETE`
- This is a soft delete, so the responses will still be retained in the DB.
- Deleting a response will be captured in the submission history and the lastActionDate in the database.

#### Screenshots
Screenshot 2026-06-23 at 5 58 20 PM

Screenshot 2026-06-23 at 5 58 31 PM

Screenshot 2026-06-23 at 5 58 57 PM

## PR Reminders


vyakusha (approved):
Tested UI for deleting questions and responses together, single responses and single questions. All actions are resulting in correct views for State, CMS, and Admin users. They are also correctly reflected in fetchSubmissionHistory output, and in db records. LGTM

PR #4600, opened 2026-06-24
MCR-6415: Set last updated at date in db

JasonLin0991:
## Summary
[MCR-6415](https://jiraent.cms.gov/browse/MCR-6415) - Implementation ticket
[MCR-6213](https://jiraent.cms.gov/browse/MCR-6213) - discovery ticket

This work adds a stable DB stored lastActionDate that will be used for the `lastUpdatedForDisplay` field in GraphQL and for filtering in `indexContracts`. The legacy code calculates this date somewhat correctly for `indexContracts` but not fore `lastUpdatedForDisplay`. The issue here is that this is a heavy computed field because it requires querying many related tables for actions that materially update a submission data. For example rate changes and Q&A actions.

Instead of computing this on read, we should just store this into contract and rate tables at contract and rate action writes into the database that create the materially significant change to data. Then we can rely on this date on DB reads and removing the additional queries and computation.

In addition to this work I also added a resolver to get submission history based on these lastActionDate writes. This history is a more complete history than what can be computed using `packageSubmission` like we do for change history. This submission history can also be used to validate the lastAction dates as the latest history should be the `lastActionDate` in the database. This also centralizes the logic to compute the complete history log so if we need to get this data in many places we use the same logic to prevent drifts.

The lastActionDate writes are not feature flagged, so actions will be written at all times. The response is feature flagged so it it will use legacy code to calculate dates on flag off and from the DB with flag on.

- Added lastActionDate column to contract and rate tables
   - The idea is any action that materially changes the contract or rate data should write that action date into the DB. This allows quick access from the DB rather then the heavy prisma queries for data to calculate on on read.
   - Date is set when contract or rate is:
      - submitted
      - review determination
      - unlocked
      - resubmitted
      - overrides
         - Contracts lastActionDate s updated on related rates override actions.
      - Q&A
         - Contracts get lastActionDate writes on both contract and rate Q&A
         - Rate Q&A is not expected to be used, but included just incase
      - approved
      - withdrawn
      - undo withdraw
   - I placed most lastActionDate writes at the lowest level prisma transaction functions, like `submitContractAndOrRate`. This is so when store functions like `withdrawRate`, which uses a collection of transaction functions, can capture lastActionDates reliability without too many one time writes specific for this action.
      - For example, in withdrawRate related contracts are unlocked and resubmitted instead of figuring our where to insert a lastActionDate write in withdrawRate, we rely on the use of `submitContractAndOrRate` in this action to do that.

- Submission history resolver, `fetchSubmissionsHistory`
   - In addition to writing the lastActionDate, I also added functions to build submission history based on the actions that writes to `lastActionDate`.
   - The idea here is that those writes represent dates for each action  that materially change the data and we should base the history on those since it would be also the most relevant.
   - There are instances of duplicate actions that can be labelled as different events, so these have to be handled based on which history is viewed.
      - For example, linking and delinking rates on a contract. 
         - This is a material change to a submission, but the way linking and delinking works, it can only be done on an unlocked contract and recorded as an action once submitted/resubmitted. 
         - At this point this is captured in the contract history as a submit action, but when we combine history the rate side will have this event, so we have to ignore these to prevent dual entries. 
         - When we think about combining the rate history with the contract, we only need to record things not captured in the shared unlock and resubmit logs for contracts and rates. So only actions like rate overrides and rate Q&A are merged in.
         - Rate withdraw actually requires unlock and resubmit of related contracts so those are already captured.

- Other work:
   - Fixed a number of existing flaky tests.
#### Related issues

- Follow up work:
   - Add undo unlock to the history and writing to lastActionDate
      - This is needed so Mulesoft syncs pick up these events.
   - Once this this merged a lambda for migrating existing records to populate the `latestActionDate` will need to be created and ran before we can turn this flag on in lower environments.
      - The logic for submission history can be reused in this lambda to find the `latestActionDate` for any given contract or rate.
      - Likely have to plan out a full sync with Mulesoft team to capture dates as they will change. The legacy code should be roughly the same, but it does not account for override or undo unlock actions.

## QA guidance
NOTE: Turing on the feature flag in DEV for this review may cause issues in dev when rendering last updated date so keep that in mind.
- Unit tests cover a lot of this work, but you can validate some of this work in the app via the fetchSubmissionHistory api.
   - Make a complex submission with actions and makes sure they are captured in the history.
   - The last Updated date in the dashboard matches the latest submission history entry

## PR Reminders


services/app-api/src/postgres/contractAndRates/approveContract.ts
```diff
@@ -52,14 +52,13 @@ async function approveContractInsideTransaction(
         },
     })
 
+    // Update lastActionDate for approval
     await tx.contractTable.update({
         where: {
             id: contractID,
         },
         data: {
-            reviewStatusActions: {
```
pearl-truss, line 60: Why is this being deleted here?

services/app-api/src/postgres/contractAndRates/withdrawRate.ts
```diff
@@ -293,6 +296,24 @@ const withdrawRateInsideTransaction = async (
...
+
+            // Withdrawing a rate can resubmit the affected health plan
+            // contract. Run the same review determination as normal submit so
+            // the contract action history and lastActionDate reflect the final
+            // visible action. Skip the withdrawn-contract edge case because the
+            // review helper intentionally rejects currently withdrawn review
+            // status.
+            if (['SUBMITTED', 'RESUBMITTED'].includes(consolidatedStatus)) {
+                const reviewDetermination =
+                    await healthPlanContractReviewDeterminationAction(
```
pearl-truss, line 308: In submitContract this is behind the chip automation flag

services/app-api/src/postgres/contractAndRates/undoWithdrawRate.ts
```diff
@@ -198,9 +201,33 @@ const undoWithdrawRateInsideTransaction = async (
...
+        // Undoing a rate withdraw can resubmit every submitted contract that
+        // previously contained the rate. Run the same health plan review
+        // determination as normal submit so each contract's action history and
+        // lastActionDate reflect the final visible action. Linked contracts
+        // that are still withdrawn keep their withdrawn status, so skip the
+        // review helper for that edge case.
+        if (
+            ['SUBMITTED', 'RESUBMITTED'].includes(contract.consolidatedStatus)
+        ) {
+            const reviewDetermination =
```
pearl-truss, line 214: Same comment as on withdrawRate

services/app-api/src/postgres/contractAndRates/reassignParentContract.ts
```diff
@@ -189,6 +191,41 @@ const reassignParentContractInTransaction = async (
...
+    // action history and lastActionDate match a normal contract submit.
+    // Reparenting only applies to rates, so EQRO submissions cannot be the new
+    // parent.
+    if (
+        ['SUBMITTED', 'RESUBMITTED'].includes(
+            resubmittedContract.consolidatedStatus
+        )
+    ) {
+        const reviewDetermination =
+            await healthPlanContractReviewDeterminationAction(
```
pearl-truss, line 219: Same comment as on withdrawRate

services/app-api/src/postgres/prismaClient.ts
```diff
@@ -20,7 +20,13 @@ const errorMessages = {
  * @returns Extended PrismaClient instance with custom query behaviors
  */
 function extendedPrismaClient(optionArgs: { adapter: PrismaPg }) {
-    return new PrismaClient(optionArgs).$extends({
+    return new PrismaClient({
+        ...optionArgs,
+        // Set default timeout for all transactions.
+        transactionOptions: {
+            timeout: 10000,
```
pearl-truss, line 27: This is double Prisma's [default](https://www.prisma.io/docs/orm/prisma-client/queries/transactions#interactive-transactions) max timeout. Were you having issues that made this necessary? I'm wondering if we can scope this down and apply to every transaction

services/app-api/src/postgres/contractAndRates/approveContract.ts
JasonLin0991: Lines 45-53 creates the approve action and makes the relationship connection with the contract table, the code I deleted was a redundant connection from the contract side. So I repurposed it to write the lastActionDate from the result on line 45.

services/app-api/src/postgres/contractAndRates/withdrawRate.ts
JasonLin0991: True, but I think the flag is on in prod. We can implement the flag here, but I thought it might not make sense unless we foresee turning the flag off.

services/app-api/src/postgres/prismaClient.ts
JasonLin0991: Yea, now that we are making extra writes in every transaction I ran into some prisma transactions timeouts in unit tests that were slightly over the 5000ms timeout. So this was more a fix for that, but it don' think it would hurt deployed API.

pearl-truss (approved):
Thanks for responses to my earlier comments

vyakusha (commented):
I tested this branch locally, creating, updating and re-submitting updated records in multiple configurations. No visible problems detected. fetchSubmissionHistory results are consistent with the data entered by user.

PR #4601, opened 2026-06-25
Add usePageLoadSpan hook for dashboard telemetry

mojotalantikite:
## Summary

Measures page load time from mount until primary data is ready for the dashboard tables, emitting a span tagged with route and result count for per-page charting and alerting in Datadog.

This was in an old ticket in an old epic that Pearl pointed out and I figured it was worth adding to our Datadog setup.

#### Related issues
https://jiraent.cms.gov/browse/MCR-2670

### DIFF
--- a/services/app-web/src/pages/CMSDashboard/SubmissionsDashboard/SubmissionsDashboard.tsx
+++ b/services/app-web/src/pages/CMSDashboard/SubmissionsDashboard/SubmissionsDashboard.tsx
@@ -7,6 +7,7 @@ import { mostRecentDate } from '@mc-review/dates'
 import styles from '../../StateDashboard/StateDashboard.module.scss'
 import localStyles from './SubmissionsDashboard.module.scss'
 import { recordJSException } from '@mc-review/otel'
+import { usePageLoadSpan } from '../../../hooks'
 import {
     Loading,
     ContractTable,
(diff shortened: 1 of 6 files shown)


services/app-web/src/pages/CMSDashboard/SubmissionsDashboard/SubmissionsDashboard.tsx
```diff
@@ -22,6 +23,17 @@ const SubmissionsDashboard = (): React.ReactElement => {
         pollInterval: 120000,
     })
 
+    // Measure dashboard load time (mount -> data ready) for the Submissions tab.
+    usePageLoadSpan({
```
JasonLin0991, line 27: **Issue**: This one may be tricky as CMS and States use this component and the same route so we likely need a more reliable way to identify which dashboard this trace is for.

JasonLin0991 (approved):
Looks good.

Only 1 issue but I don't think it is a blocker.

services/app-web/src/pages/CMSDashboard/SubmissionsDashboard/SubmissionsDashboard.tsx
JasonLin0991: **Suggestion**: Passing a prop would be the easiest way.

mojotalantikite: yeah good call -- I added a way to pass a `page.name` attribute to the span, so we can differentiate.

PR #4604, opened 2026-06-26
Add email tracing instrumentation

mojotalantikite:
## Summary
In going through some old epics I found this one related to email monitoring. This adds some OTEL tracing to our emailer along with some tests for paths we weren't testing.

#### Related issues
https://jiraent.cms.gov/browse/MCR-2664

### DIFF
--- a/services/app-api/src/emailer/emailTracing.ts
+++ b/services/app-api/src/emailer/emailTracing.ts
@@ -41,7 +41,6 @@ function toSpanAttributes(attributes: EmailSpanAttributes): Attributes {
         // record of so we know who was (or was not) notified.
         'email.to_addresses': attributes.toAddresses,
         'email.cc_addresses': attributes.ccAddresses,
-        'email.bcc_addresses': attributes.bccAddresses,
         'email.recipient_count':
             attributes.toAddresses.length +
             attributes.ccAddresses.length +
--- a/services/app-api/src/emailer/emailTracing.ts
+++ b/services/app-api/src/emailer/emailTracing.ts
@@ -1,0 +1,108 @@
+import {
+    trace,
+    context as otelContext,
+    SpanStatusCode,
+    type Span,
+    type Attributes,
+} from '@opentelemetry/api'
+import { parseErrorToError } from '@mc-review/helpers'
+
+/**
+ * Instrumentation scope for email-send spans.
+ *
+ * The Datadog `service` tag comes from the tracer provider's resource
+ * (service.name = `app-api-<stage>`), NOT from this scope name, so every email
+ * span still lands under the same `app-api-<env>` service as the rest of the API
+ * while sharing a dedicated, queryable span name.
+ */
+export const EMAIL_TRACER_SCOPE = 'mcreview.email'
+
+/**
+ * Stable span name for every transactional email send via SES.
+ *
+ * Search Datadog APM with `operation_name:email.send` (optionally filtered on the
+ * `@email.*` attributes below) to see who notifications were sent to, the SES
+ * message id for correlating with any future bounce/complaint events, and
+ * whether the send succeeded (`@email.outcome`).
+ */
+export const EMAIL_SPAN_NAME = 'email.send'
+
+export type EmailSpanAttributes = {
+    toAddresses: string[]
+    ccAddresses: string[]
+    bccAddresses: string[]
+    sourceEmail: string
+    subject: string
+}
+
+function toSpanAttributes(attributes: EmailSpanAttributes): Attributes {
+    return {
+        // The recipients the email was sent to — the primary thing we want a
+        // record of so we know who was (or was not) notified.
+        'email.to_addresses': attributes.toAddresses,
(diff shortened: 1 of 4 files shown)


services/app-api/src/emailer/emailTracing.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+function toSpanAttributes(attributes: EmailSpanAttributes): Attributes {
+    return {
+        // The recipients the email was sent to — the primary thing we want a
+        // record of so we know who was (or was not) notified.
+        'email.to_addresses': attributes.toAddresses,
+        'email.cc_addresses': attributes.ccAddresses,
+        'email.bcc_addresses': attributes.bccAddresses,
+        'email.recipient_count':
```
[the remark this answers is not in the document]
mojotalantikite: removed

services/app-api/src/emailer/emailTracing.ts
```diff
@@ -0,0 +1,107 @@
...
+    const tracer = trace.getTracer(EMAIL_TRACER_SCOPE)
+    const span = tracer.startSpan(EMAIL_SPAN_NAME, {
+        attributes: toSpanAttributes(attributes),
+    })
+
+    // Run within the span's context so any nested spans parent to it.
+    const spanContext = trace.setSpan(otelContext.active(), span)
+
+    try {
+        const result = await otelContext.with(spanContext, () => fn(span))
```
JasonLin0991, line 94: **Question**: Will cause issues locally with successful emails since we don't have a local otel setup? I see a lot of otel request errors in the local API logs. Is there a way to easily setup handling otel requests locally or can we disable tracing locally all together?

services/app-api/src/emailer/emailer.ts
```diff
@@ -749,18 +750,32 @@ function emailer(
...
+                if (response && 'MessageId' in response && response.MessageId) {
+                    span.setAttribute(
+                        'email.ses_message_id',
+                        response.MessageId
+                    )
+                }
+                return
+            } catch (err) {
+                const message = parseErrorToError(err).message
+                return new Error('SES email send failed. Error: ' + message)
```
JasonLin0991, line 775: **Question**: This catches all thrown errors, would this ever trigger the catch block in withEmailSpan? Also some email failures are meant to not stop resolvers from succeeding, we just log and move on. We should make sure we handle that.

services/app-api/src/emailer/emailTracing.ts
mojotalantikite: Hmm, there's actually a local OTEL container that runs to take those traces in, but I haven't looked at it in a while -- I guess it might be broken? Let me check on that before merge.

services/app-api/src/emailer/emailer.ts
mojotalantikite: that `catch()` block in `withEmailSpan` is just for unexpected throws that happen with the OTEL logic. The inner errors hit the branch with the closer `catch`, which returns the error:

```typescript
async (span) => {
            try {
                const response = await sendSESEmail(emailRequestParams)  // ← AWS SES call
                if (response?.MessageId) span.setAttribute('email.ses_message_id', ...)
                return                                    // success → undefined
            } catch (err) {
                return new Error('SES email send failed. Error: ' + message)  // ← returned, not thrown
            }
```

But in reading through these code paths it does seem like any email failure is going to return a graphql error, which I guess I hadn't realized -- from `submitContract()`

```typescript
     if (
                    cmsContractEmailResult instanceof Error ||
                    stateContractEmailResult instanceof Error
                ) {
                    if (cmsContractEmailResult instanceof Error) {
                        logResolverError(
                            'submitContract - CMS email failed',
                            cmsContractEmailResult,
                            context
                        )
                    }
                    if (stateContractEmailResult instanceof Error) {
                        logResolverError(
                            'submitContract - state email failed',
                            stateContractEmailResult,
                            context
                        )
                    }
                    throw new GraphQLError('Email failed', {
                        extensions: {
                            cause: 'EMAIL_ERROR',
                        },
                    })
                }
```

So I guess the client is getting an error on `submitContract` even though the data was persisted -- how do we end up dealing with that in the React frontend? Obviously that's not a concern with this individual PR, but thought I'd mention it.

services/app-api/src/emailer/emailTracing.ts
mojotalantikite: ok yeah you're right local OTEL got messed up with the migration to datadog, I hadn't noticed. That's from how the inits are working and setting us up to send direct to datadog ingest -- before the URL was local, and the exporter took care of forwarding, so in our local environment since the collector was just the same container nothing needed to change. I'll throw up a follow on to fix that.

services/app-api/src/emailer/emailer.ts
JasonLin0991: Oh I think I remembered incorrectly then. Looks like we handle this on the frontend by swallowing the emails errors. Likely need to think about how to handle this better in the future.

JasonLin0991 (approved):

PR #4608, opened 2026-06-29
Add OpenTelemetry tracing for S3 operations

mojotalantikite:
## Summary

Wrap S3 client calls in dedicated spans to trace GET/PUT operations consistently. Records retry accounting, HTTP status, and S3 error codes so throttling and failures are visible in Datadog.

### DIFF
--- a/services/app-api/src/otel/s3Tracing.ts
+++ b/services/app-api/src/otel/s3Tracing.ts
@@ -1,0 +1,104 @@
+import {
+    trace,
+    context as otelContext,
+    SpanStatusCode,
+    type Span,
+} from '@opentelemetry/api'
+import { parseErrorToError } from '@mc-review/helpers'
+
+/**
+ * Instrumentation scope for S3 data-plane spans.
+ *
+ * The Datadog `service` tag comes from the tracer provider's resource
+ * (service.name = `app-api-<stage>`), not from this scope name, so S3 spans land
+ * under the same service as the rest of the API while sharing a dedicated,
+ * queryable span name (`s3.<operation>`).
+ *
+ */
+export const S3_TRACER_SCOPE = 'mcreview.s3'
+
+export type S3SpanAttributes = {
+    /** S3 operation name, e.g. 'GetObject' | 'PutObject' */
+    operation: string
+    bucket: string
+    key: string
+}
+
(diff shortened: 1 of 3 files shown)


services/app-api/src/otel/s3Tracing.ts
```diff
@@ -0,0 +1,104 @@
...
+            'aws.s3.bucket': attributes.bucket,
+            'aws.s3.key': attributes.key,
+        },
+    })
+
+    // Run within the span's context so it parents to the active resolver/zip span.
+    const spanContext = trace.setSpan(otelContext.active(), span)
+
+    try {
+        const result = await otelContext.with(spanContext, fn)
```
JasonLin0991, line 95: **Thought**: The way we set all these wrappers up to throw on any error, if any part of the tracing services ever went down would using these wrappers also take down our API?

JasonLin0991 (approved):
Looks good, but Copilot had some good comments though.

services/app-api/src/otel/s3Tracing.ts
```diff
@@ -0,0 +1,104 @@
...
+    fn: () => Promise<T>
+): Promise<T> {
+    const tracer = trace.getTracer(S3_TRACER_SCOPE)
+    const span = tracer.startSpan(`s3.${attributes.operation}`, {
+        attributes: {
+            'aws.s3.operation': attributes.operation,
+            'aws.s3.bucket': attributes.bucket,
+            'aws.s3.key': attributes.key,
+        },
+    })
```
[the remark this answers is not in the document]
mojotalantikite: we don't store the filename in the key, so this doesn't leak a filename (or any PII)

services/app-api/src/otel/s3Tracing.ts
mojotalantikite: So the `throw` that we do here is just re-raising the throw that comes from the s3 operation, we eventually end up turning that into a return higher up the call chain. I do sort of wish we had one way of dealing with errors, we have this mix of throwing and returning that makes it a little more messier to reason about than necessary.

In terms of if Datadog or another OTEL backed goes down though, that wouldn't affect us here since all that is async. OTEL uses a span processor to export spans, and if that ever has an error it just drops the spans and fails quietly. The traces/spans are designed not to take down the actual system with it.

PR #4615, opened 2026-07-01
MCR-6220: Fix loading spinner alignment

pearl-truss:
## Summary

This PR resolves this issue: Currently this is occurring on the form pages with an unlocked submission. The issue is that the loading spinner is not centered. 
This is also occurring on pages with a side nav

Notes
- The loading spinner previously was being centered based on the sub component it was a part of + the side nav was being displayed while the page was loading. This PR updates passes a full load prop to handle displaying the load spinner centered on the full page

#### Related issues

https://jiraent.cms.gov/browse/MCR-6220

#### Screenshots
BEFORE
Screenshot 2026-06-30 at 11 42 25 AM

AFTER
Screenshot 2026-06-30 at 1 50 42 PM

Screenshot 2026-07-09 at 9 54 26 AM

## QA guidance

- View an unlocked submission as a state user and switch between the side nav tags, the loading indicator should be centered on the page
- Note: this is hard to see locally because data loads faster - try to review this in the review app 
## PR Reminders


JasonLin0991 (approved):
Refactor looks good. A few places that could use some looking into:
- `StateDashboard.tsx`: Needs to also use `ErrorOrLandingPage`.
- Settings page and tabs: Spinner is not centered
   - The tabs load when clicked so this one is a bit different than the summary page tabs.
   - Image
- There is a general loading UI somewhere higher up in the component tree that is not centered like `ErrorOrLandingPage` is.
   - Image

PR #4616, opened 2026-07-01
Add undo unlock to history and write to last action date

JasonLin0991:
## Summary

This adds the undo unlock action into the submission history and writes to lastActionDate.

The undo unlock information is in the domain model under `undoUnlockPackages`, but is not surfaced to the GraphQL layer in this iteration. This is because right now we just need to be able to get the data for building submission history, we are not displaying it in any UI. When we get around to that we can expose it to GraphQL and likely have to build the `draftRateRevisionSnapShot` portion in the field resolver level.

## QA guidance

- With the `use-stored-contract-action-dates` on, perform an undo unlock and see the `lastUpdatedForDisplay` updated with the undo unlock date.
- Use the graphql-explorer tool to call `fetchSubmissionHistory` query to get a log of the history and validate undo unlock is captured in the log.

## PR Reminders


JasonLin0991:
Closing this in favor of this [PR](https://github.com/Enterprise-CMCS/managed-care-review/pull/4617) with clean git history.

PR #4619, opened 2026-07-06
Fix local OTEL via Jaeger container

mojotalantikite:
## Summary

When I migrated us to Datadog I missed some errors in the local OTEL setup. This PR updates Jaeger (the local OTEL container) and fixes up some of the configs for local init to send traces to the local OTEL setup.

#### Related issues
https://jiraent.cms.gov/browse/MCR-6424

### DIFF
--- a/services/app-api/src/handlers/otel_proxy.ts
+++ b/services/app-api/src/handlers/otel_proxy.ts
@@ -16,16 +16,34 @@ const main: APIGatewayProxyHandler = async (event) => {
         `otel_proxy: received trace payload, body=${bodyBytes} bytes, content-type=${contentType}`
     )
 
-    // Local dev sets OTEL_EXPORTER_OTLP_TRACES_ENDPOINT (in .envrc) to point
-    // at the local Jaeger container's OTLP port. Deployed environments leave
-    // it unset and forward directly to Datadog.
-    const localEndpoint = process.env.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
-    const tracesUrl = localEndpoint ?? DD_TRACES_URL
-
+    // Local dev (stage === 'local') forwards traces to the local Jaeger
+    // container's OTLP port, set via OTEL_EXPORTER_OTLP_TRACES_ENDPOINT in
+    // .envrc. All other environments forward directly to Datadog; the
+    // endpoint override is ignored there so a stray env var can't redirect
+    // traces.
+    let tracesUrl: string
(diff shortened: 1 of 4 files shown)


services/app-api/src/handlers/otel_proxy.ts
```diff
@@ -16,34 +16,42 @@ const main: APIGatewayProxyHandler = async (event) => {
...
+    // Local dev sets OTEL_EXPORTER_OTLP_TRACES_ENDPOINT (in .envrc) to point
+    // at the local Jaeger container's OTLP port. Deployed environments leave
+    // it unset and forward directly to Datadog.
+    const localEndpoint = process.env.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+    const tracesUrl = localEndpoint ?? DD_TRACES_URL
+
+    const headers: Record<string, string> = {
+        'content-type': contentType,
+    }
+    if (!localEndpoint) {
```
JasonLin0991: **Question**: Do we have access to the deployment environment in this handler? Wondering if it would be better to explicitly set the `tracesUrl` to local env if it is local rather than fallback.

services/app-api/src/otel/otel_handler.ts
```diff
@@ -31,6 +31,17 @@ function getDDHeaders() {
     }
 }
 
+// Local dev sets OTEL_EXPORTER_OTLP_TRACES_ENDPOINT (in .envrc) to point at
+// the local Jaeger container's OTLP port. Deployed environments leave it
+// unset and export directly to Datadog.
+function getExporterConfig() {
+    const localEndpoint = process.env.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+    if (localEndpoint) {
```
JasonLin0991: Same down here about explicitly setting the local endpoint for local deploy environment.

JasonLin0991 (approved):
No more errors on local deploy. Thanks!

Couple of non blocking comments, but I do think being more explicit might be better than falling back when on higher environments.

services/app-api/src/handlers/otel_proxy.ts
mojotalantikite: `stage` is here, so yeah good call, changed it to check if we're local first.

services/app-api/src/otel/otel_handler.ts
mojotalantikite: fixed

PR #4629, opened 2026-07-09
MCR-6384 & MCR-6385: Add and update feature flags

JasonLin0991:
## Summary
[MCR-6384](https://jiraent.cms.gov/browse/MCR-6384):
- Add a feature flag to LaunchDarkly
- Add the feature flag into the codebase
- Name the feature flag something like "revision-diff" to ensure that there is no confusion with the actual revision history functionality
   - NOTE: Naming this flag the same as the epic name `revision-history-enhancements`. I think it should suffice since there is a description for the flag.

[MCR-6385](https://jiraent.cms.gov/browse/MCR-6385):
- Add a feature flag to LaunchDarkly
- Add the feature flag into the codebase

Update default feature flag values:
- This updates our default feature flags in the codebase to match production settings. This is a fallback in case LaunchDarkly servers are down we can still preserve the flag settings in prod.
- Unit tests have been updated to test based on new flag values. Some have been deleted if testing with flag off.
- We will be removing flags for features that are permanent, but for now this is a light lift fallback.

## QA guidance
- Mainly the unit tests pass.
- Check for typos in flags.ts

### DIFF
--- a/packages/common-code/src/featureFlags/flags.ts
+++ b/packages/common-code/src/featureFlags/flags.ts
@@ -3,6 +3,8 @@
  * Ensures some type safety around flag names when we're enabling/disabling features in our code.
  *
  * This file also used to generate types for our Jest unit tests, Cypress tests, and for Yup validation logic.
+ *
+ * Default values are set to match PROD.
  */
 
 const featureFlags = {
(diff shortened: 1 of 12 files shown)


vyakusha (approved):
All looks good.

PR #4642, opened 2026-07-16
MCR-6332 Create backend structures to support submission revision history

vyakusha:
## Summary
Add backend support for revision-to-revision diff for contract submissions.
This is the first step, which involves creating the 'first-level' set of Contract-related fields, which were changed between revisions.
#### Related issues
[MCR-6332](https://jiraent.cms.gov/browse/MCR-6332)
#### Screenshots
Screenshot 2026-07-22 at 1 13 09 PM

#### Test cases covered
revisionDiffPrimitives.test.ts
revisionDiffHelpers.test.ts
fetchRevisionDiff.test.ts

## QA guidance
Create Health Plan Contract, submit, unlock, make changes, resubmit.
Run fetchRevisionDiff via GraphQL, observe the changes being reflected in the output.

JSON output for the local fetchRevisionDiff is atatched
[FetchRevisionDiff-07-16-2026.json](https://github.com/user-attachments/files/30096947/FetchRevisionDiff-07-16-2026.json)

## PR Reminders


services/app-api/src/postgres/contractAndRates/revisionDiffHelpers.ts
```diff
@@ -0,0 +1,306 @@
...
+    const names = programIDs.map((programID) => {
+        const program = statePrograms.find((stateProgram) => {
+            return stateProgram.id === programID
+        })
+
+        if (!program) {
+            return new Error(`Could not find state program ${programID}`)
+        }
+
+        return program.fullName ?? program.name
```
pearl-truss: fullName is a required field on programs and will never be empty

services/app-api/src/postgres/revisionDiff/findRevisionDiffByContractID.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    const selectedSubmissions = contractSubmissions.filter((submission) => {
+        return (
+            submission.contractRevision.id === args.olderContractRevisionID ||
+            submission.contractRevision.id === args.newerContractRevisionID
+        )
+    })
+
+    if (selectedSubmissions.length !== 2) {
```
pearl-truss, line 101: I think this will have to be reconsidered given that packageSubmissions can have multiple entries with the same contractRevisionID in the case of rates being added, linked or unlinked

So you could have a valid packageSubmissions array with a length > 2

services/app-api/src/postgres/revisionDiff/findRevisionDiffByContractID.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (
+        (args.olderContractRevisionID && !args.newerContractRevisionID) ||
+        (!args.olderContractRevisionID && args.newerContractRevisionID)
+    ) {
+        return new InvalidRevisionDiffInputError(
+            `olderContractRevisionID and newerContractRevisionID must both be provided when selecting a specific comparison`
+        )
+    }
+
+    if (!args.olderContractRevisionID && !args.newerContractRevisionID) {
```
pearl-truss, line 81: Given that rate submissions, rate links and rate unlinks can result in the same contractRevision appearing in packageSubmissions multiple times - it's not guaranteed that contractSubmissions[0] and contractSubmissions[1] aren't the same contractRevision

Suggestion: Instead of having a default, always require 2 contractRevisionIDs and throw an error if the ids aren't unique  or you can keep the default behavior but check to ensure you're pulling unique revisions

pearl-truss (commented):
This is looking really good Vira! I think you're on the right track and let me know if you want to discuss my comments on a call

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.ts
```diff
@@ -0,0 +1,306 @@
...
+    if (fieldChanges instanceof Error) {
+        return fieldChanges
+    }
+
+    const submissionIDChange =
+        olderSubmissionID === newerSubmissionID
+            ? []
+            : [
+                  {
+                      fieldPath: 'submissionID',
```
JasonLin0991: Suggestion: We probably want this to be contractName to match the graphql schema. I know we commonly use submissionID to mean the contractName, but we should try to stay consistent in the codebase.

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.ts
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    if (unwrappedSchema instanceof z.ZodDate) {
+        return [buildDateFieldConfig(typedFieldPath)]
+    }
+
+    return []
+})
+
+function buildRevisionDiff(
```
JasonLin0991, line 200: Suggestion: Adding JSDoc style comments for these functions would help devs understand the intended usage of many of functions. We can understand from reading the code, but the abstractions makes it a little more diffcult.

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.ts
```diff
@@ -0,0 +1,306 @@
...
+    },
+}
+
+const excludedFieldPaths = new Set<keyof ContractFormData>([
+    'stateContacts',
+    'supportingDocuments',
+    'contractDocuments',
+])
+
+const diffFieldConfigs: DiffFieldConfig[] = Object.entries(
```
JasonLin0991: Question: Will this only be for contract form data diffs? If so, consider renaming this to something like `diffContractFromDataFieldConfigs`

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.ts
```diff
@@ -0,0 +1,306 @@
...
+    fieldPath: keyof ContractFormData & string
+): DiffFieldConfig {
+    return {
+        fieldPath,
+        displayValue: (formData) =>
+            displayDate(formData[fieldPath] as Date | null),
+    }
+}
+
+const fieldConfigOverrides: Partial<
```
JasonLin0991, line 123: Question: What are these overrides used for? Is it just taking the zod schema literals and converting them to the display strings?

services/app-api/src/postgres/contractAndRates/revisionDiffHelpers.ts
```diff
@@ -0,0 +1,306 @@
...
+}
+
+const displayBooleanAsYesNo = (
+    value: boolean | null | undefined
+): string | null => {
+    if (value === undefined || value === null) {
+        return null
+    }
+
+    return value ? 'Yes' : 'No'
```
JasonLin0991: Thought: I'm seeing a common pattern here with the diff values. Seems like the diff result is something different from domain type or graphql type values. The diff seems to be returning actual display type values. For example  'Yes' or 'No' for the field above. Usually this kind of conversion from data values to display values is done where the intended audience is viewing the data. For users of the app this would be frontend or emails. I think we are going down a road where we are deviating from the pattern where domain layer contains domain types and graphql layer is containing graphql types. We are interpreting these data value types into display types in the data layers of the API and muddying the clear separations. My suggestion is to purely return data types and let the frontend app and emails interpret what the display values are. Looking into the future, external API users likely want to work with the same graphql types and not the display values.

JasonLin0991 (commented):
Left some comments. I think there might need to be some more thought about what he return data of the diffs look like and if that aligns with how we structured data types the API.

Nitpick: `postgres/contractAndRates` is getting pretty large, we might want to consider moving all of this revision diff code to a new directory like `postgres/revisionDiff`.

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.ts
vyakusha: reworked to keep output consistent with domain type values

pearl-truss (approved):
LGTM, thanks for making those changes!

JasonLin0991 (approved):

PR #4643, opened 2026-07-17
MCR-6196: create read only user 

pearl-truss:
## Summary

This PR creates a new user role for a Read Only user. This user type isn't able to view any draft data, any oauth client settings data or take any actions

- A read only user account type is created in the system for all environments
- The read only user account type will differ from the Help Desk account type
- The account type should have the ability to review submission records
- The account type should not have any write capabilites
- The account type should not be able to take any actions within the application

#### Related issues

https://jiraent.cms.gov/browse/MCR-6196 

#### Test cases covered

1. Backend — auth entrypoint (cognitoAuthn.test.ts)

Added a case to the table-driven userTypeFromAttributes test asserting that the IDM attribute macmcrrs-ro-user parses into a READONLY_USER domain user (id, email, given/family name). This directly covers the original bug — the entrypoint no longer throws "Unsupported user role."
2. Frontend — local login (Auth.test.tsx)

Updated the "displays local users when logged out" test to expect 10 login buttons (was 9) and assert the Gyatso read-only card renders — covering that the new local sign-in user shows up.

## QA guidance

In the ephemeral review environment sign in as gyatso@example.com (same 1password password as Zuko and Aang) and confirm that the read only user has the correct permissions (no ability to submit any action or see restricted info such as oauth client settings)

## PR Reminders

### DIFF
--- a/packages/helpers/src/gql/userHelpers.ts
+++ b/packages/helpers/src/gql/userHelpers.ts
@@ -5,6 +5,7 @@ import {
     AdminUser,
     BusinessOwnerUser,
     HelpdeskUser,
+    ReadOnlyUser,
     UpdatedBy,
 } from '../gen/gqlClient'
 const hasCMSUserPermissions = (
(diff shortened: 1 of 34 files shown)


ruizajtruss (approved):
Everything looks good to me 🚀

PR #4656, opened 2026-07-27
MCR-6424 Expand revision diff for the full submission coverage

vyakusha:
## Summary
Expand revision DIFF for the full submission coverage (Contacts, Documents, Rates)

#### Related issues
[MCR-6427](https://jiraent.cms.gov/browse/MCR-6427) (continuation of [MCR-6332](https://jiraent.cms.gov/browse/MCR-6332)) 
#### Screenshots
Screenshot 2026-08-02 at 11 34 58 AM

## QA guidance
1. Create fresh Submission: login as State user, create new Health Plan (fill in subsequent forms, click "Submit". Now this record is reflected on the State dashboard as SUBMITTED.
2. Unlock Submission: login as CMS (Approver) user, find submitted record, created in step1, unlock it. Non the record is listed on dashboard as UNLOCKED.
3. As a State user, update Health Plan record. "Edit" buttons are located on Health Plan summary page. Submit changes. Now we have two submissions for one Health Plan record. 
4. Run fetchRevisionDiff via browser graphQL explorer interface (https://mc-review-dev.onemac.cms.gov/dev/graphql-explorer) to observe the structure and values for the return data. 

## PR Reminders


services/app-api/src/domain-models/contractAndRates/revisionDiffTypes.ts
```diff
@@ -29,17 +31,84 @@ type RevisionDiffCollectionItemChange<TItem, TChange> =
     | RevisionDiffCollectionItemRemoved<TItem>
     | RevisionDiffCollectionItemUpdated<TItem, TChange>
 
+type RevisionDiffCollectionItemNewOrModified<TItem> = {
+    kind: 'new_or_modified'
```
ronak-truss: nit: Should this be an enum?

services/app-api/src/domain-models/contractAndRates/revisionDiffTypes.ts
```diff
@@ -29,17 +31,84 @@ type RevisionDiffCollectionItemChange<TItem, TChange> =
...
+    current: TItem
+}
+
+type RevisionDiffRateActuaryContactChange = {
+    kind: 'new_or_modified'
```
ronak-truss: This implies that the object itself cannot be deleted, right?

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
```diff
@@ -608,5 +636,587 @@ describe('revisionDiffHelpers', () => {
...
+        const comparison = buildRevisionDiff(
+            contract.id,
+            previousSubmission,
+            latestSubmissionPackage,
+            statePrograms
+        )
+
+        expect(comparison).not.toBeInstanceOf(Error)
+        expect(
+            comparison instanceof Error ? [] : comparison.stateContactChanges
```
ronak-truss: Why do we need `comparison instanceof Error ? [] `? If that's the case wouldn't it fail the test on the previous line? I think we'd want to avoid coercing the error into a valid []?

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
```diff
@@ -608,5 +636,587 @@ describe('revisionDiffHelpers', () => {
...
+            {
+                kind: 'new_or_modified',
+                current: {
+                    name: 'New Person',
+                    titleRole: 'Analyst',
+                    email: 'new@example.com',
+                },
+            },
+        ])
+    })
```
ronak-truss, line 791: I'm not sure I understand this output. What is this indicating? Are we not showing the before stage?

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
```diff
@@ -608,5 +636,587 @@ describe('revisionDiffHelpers', () => {
...
+        const postgresStore = NewPostgresStore(prismaClient)
+        const cmsUser = testCMSUser()
+        const stateServer = await constructTestPostgresServer()
+        const cmsServer = await constructTestPostgresServer({
+            context: {
+                user: cmsUser,
+            },
+        })
+
+        const contract = await createAndSubmitTestContractWithRate(
```
ronak-truss, line 870: These tests require a lot of setup to create, submit, unlock, and update. It might be worth a helper function that abstracts some of this setup. But maybe the slight nuances between the setups are worth being explicit about? I don't have enough context to have a strong opinion on this.

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
```diff
@@ -0,0 +1,316 @@
...
+}
+
+const excludedRateFieldPaths = new Set<keyof RateFormData>([
+    'id',
+    'rateID',
+    'rateDocuments',
+    'supportingDocuments',
+    'certifyingActuaryContacts',
+    'addtlActuaryContacts',
+    'packagesWithSharedRateCerts',
```
ronak-truss, line 100: I assume these are all intentional? It's not obvious to me why we are doing this.

services/app-api/src/postgres/revisionDiff/revisionDiffRateActuaries.ts
```diff
@@ -0,0 +1,89 @@
...
+): RevisionDiffFieldChange[] | Error {
+    return buildScalarFieldDiffChanges(
+        previous,
+        current,
+        actuaryContactFieldConfigs,
+        undefined
+    )
+}
+
+function buildCertifyingActuaryContactDiffChanges(
```
ronak-truss: Does position matter? Is it possible to have 3 entries and delete the middle one? What would we expect the changes to be in that instance?

ronak-truss (commented):
Left a handful of comments based on an initial review. I'm missing a lot of context and domain knowledge so take them with a grain of salt. Maybe we can pair and walk through this functionality in the app.

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
vyakusha: Will make sure all instances of this pattern are replaced with simplified/optimized one..

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
vyakusha: We don't need to surface packagesWithSharedRateCerts, but we need to make sure addlActuaryContacts are coming AFTER certifyingActuaryContacts (two DIFFS?)

vyakusha: Implemented two separate DIFFS: one for certifying Actuary Contacts, one for Additional Actuary Contacts.

services/app-api/src/postgres/revisionDiff/revisionDiffRateActuaries.ts
vyakusha: Modified comparison logic for revisions Actuary Contacts lists to match Contract's State Contact. Position should not be used for comparison

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
vyakusha: Modified tests to use API-generated data for assertions.

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
```diff
@@ -953,28 +962,122 @@ describe('revisionDiffHelpers', () => {
...
+        const previousRateSupportingDocuments =
+            previousRateRevision.formData.supportingDocuments ?? []
+        const latestRateSupportingDocuments =
+            latestRateRevision.formData.supportingDocuments ?? []
+
+        expect(comparison.documentChanges).toEqual({
             contractDocuments: {
-                added: ['contract-added.pdf'],
-                removed: ['contract-removed.pdf'],
+                added: latestContractDocuments
```
ronak-truss: Afaict, this works, but I find this difficult to reason about. (Mainly the nested !some.) Is there a way we could simplify this and make it easier to parse?

One option might be to use sets and use the difference method.

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
ronak-truss: That seems reasonable to me given the way that we use these, but can you talk me through the thought process behind this decision?

ronak-truss (commented):
This works as expected. Overall this looks good, but I do want to give the tests and the broader code a second pass. I found the tests a little hard to follow initially (that could just be due to the amount of setup that's required).

ronak-truss (commented):
Some other high level architectural decisions that aren't clear to me. What happens if we add a new field X to health plans? Is this auto included in the diff? Does it only do so if its a scaler or non enum/object? If we add an object should there be a failing test or error if we don't add the accompanying diffing functionality?

One other thing to consider is that the default value of a string is considered an empty string "" while we show null for other values. If all of the string values have to be non empty, it could be worth displaying the value as null. (I'd assume this isn't a guarantee though, so likely not worth doing.)

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
vyakusha: I did make an effort to simplify the process, [now we have a function](https://github.com/Enterprise-CMCS/managed-care-review/pull/4656/changes#:~:text=!previousDocKeys.has(%60%24%7Bdocument.sha256%7D%3A%3A%24%7Bdocument.name%7D%60)) which is using Set .has to come back with lists of "added" and "removed" documents.

vyakusha:
> Some other high level architectural decisions that aren't clear to me. What happens if we add a new field X to health plans? Is this auto included in the diff? Does it only do so if its a scaler or non enum/object? If we add an object should there be a failing test or error if we don't add the accompanying diffing functionality?
> 
For Contract fields we are using [contractFormDataSchema.shape](https://github.com/Enterprise-CMCS/managed-care-review/pull/4656/changes#:~:text=158-,contractFormDataSchema.shape,-as%20Record%3C) in revisionDiffHelper.ts. Code iterates through it, applying standard treatment for ZodBoolean, ZodString, ZodDate, ZodArray. If our new field is falling within those categories, then it could be added without breaking the DIFF. 

There are many types of new fields which can break existing DIFF code because of the need for special logic, and without this special handling they will be skipped or handled incorrectly. For example, programID's need to be compared without regard for order, so "MSA, PSBA" is equal to "PSBA, MSA", and contractName does not come from the database but needs to be constructed separately etc. 

I added [two new tests:](https://github.com/Enterprise-CMCS/managed-care-review/pull/4656/changes#diff-e7a33d5e7da23bce8ffee42ade34a0030666cc2bc3b1533d4c98c3839d23a2c1R80:~:text=contract%20form%20field%20in%20revision-,diff,-handling%27%2C%20()) in revisionDiffHelpers.test, which simply go through contractFormDataSchema.shape,  rateFormDataSchema.shape and fail if field path is found which is not covered by standard "Zod" types, special treatment (overrides), or excluded . Tests are: "accounts for every contract form field in revision diff handling" and "accounts for every rate form field in revision diff handling"

> One other thing to consider is that the default value of a string is considered an empty string "" while we show null for other values. If all of the string values have to be non empty, it could be worth displaying the value as null. (I'd assume this isn't a guarantee though, so likely not worth doing.)

I looked through our UI and it looks that all our string inputs are required. At the same time, we may wish to add some non-required strings in the future. @JasonLin0991, what do you think about that? Is it worth the lift?

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
```diff
@@ -37,7 +41,50 @@ const mockStateUser = () => ({
...
+    )
+
+    return {
+        // return the names of documents that are in the latest submission but not in the previous submission, and vice versa.
+        added: latestDocs
+            .filter(
+                (document) =>
+                    !previousDocKeys.has(`${document.sha256}::${document.name}`)
+            )
+            .map((document) => document.name),
```
ronak-truss: I think we can just simplify this using set methods. Eg:
```
latestDocKeys.difference(previousDocKeys)
```

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
```diff
@@ -37,7 +41,50 @@ const mockStateUser = () => ({
     stateCode: 'KY',
 })
 
+type DocumentDiffItem = {
+    name: string
+    sha256: string
+}
+
+//documents are considered to be changed if the name and sha256 do not match between the two revisions. This function returns the added and removed document names.
```
ronak-truss: Do we actually care about the name changing? I'm just asking because I don't think that logic was included in the previous iteration.

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
```diff
@@ -37,7 +41,50 @@ const mockStateUser = () => ({
...
+function diffDocumentsByShaAndName(
```
ronak-truss: Nit: It could be worth it to scope this function inside of the test it's actually used in?

The main reason for this suggestion is so that the code is closer to where it's being used.

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
```diff
@@ -83,6 +83,18 @@ function buildStringArrayFieldConfig(
...
+        getValue: (formData) => formData.rateType,
+    },
+    actuaryCommunicationPreference: {
+        fieldPath: 'actuaryCommunicationPreference',
+        getValue: (formData) => formData.actuaryCommunicationPreference,
+    },
+    rateCapitationType: {
+        fieldPath: 'rateCapitationType',
+        getValue: (formData) => formData.rateCapitationType,
+    },
```
ronak-truss: We're just passing through the values here. I don't think this is what someone would consider an "override". Can we update how we are handling enums in the diffing process so they are just defaulted to actual value?

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
ronak-truss: We can do something similar for the removed documents as well

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
vyakusha: @ronak-truss Yes, we do care about name changing. It was discussed and the decision was that CMS user wants to know if the file name has changed, so we count it as changed, at least for now.

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
vyakusha: You are absolutely right. I added auto-handling enums to the code

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
vyakusha: Done

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
vyakusha: Unfortunately, .difference is not supported by our TS yet

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.ts
```diff
@@ -74,6 +74,25 @@ function unwrapSchema(schema: z.core.$ZodType): z.core.$ZodType {
...
+ * Detects enum-like schemas expressed as string enums, literals, or unions of string literals.
+ */
+function isStringEnumLikeSchema(schema: z.core.$ZodType): boolean {
+    if (schema instanceof z.ZodString || schema instanceof z.ZodEnum) {
+        return true
+    }
+
+    if (schema instanceof z.ZodLiteral) {
+        return true
+    }
```
ronak-truss: We can combine these two if conditions

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
```diff
@@ -41,6 +41,22 @@ function unwrapSchema(schema: z.core.$ZodType): z.core.$ZodType {
     return schema
 }
 
+function isStringEnumLikeSchema(schema: z.core.$ZodType): boolean {
```
ronak-truss: Do we need two instances of this? Can we just have one

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.ts
ronak-truss: Do we also have to check for ZodNativeEnum?

vyakusha: > We can combine these two if conditions

Yes, you are absolutely right. Done

vyakusha: > Do we also have to check for ZodNativeEnum?

We do not use ZodNativeEnum anywhere in the codebase. Since we are not using this in our schemas, probably it does not make sense to handle them here right now. CMIIW

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
vyakusha: Oh, of course you are right. Fixed

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.ts
ronak-truss: Personally I'd probably lean towards having it, but I don't have as much context on this project or zod to know if we should realistically expect ZodNativeEnum to ever be used. This is primarily driven by wanting this code to never have to change. To me the workflow should basically be to ignore this code and only have to interact with this is we add a new complex/object field that needs a diff. 

cc: @JasonLin0991 In case you have thoughts on this

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
```diff
@@ -608,5 +651,729 @@ describe('revisionDiffHelpers', () => {
...
+        ])
+    })
+
+    it('reports contract and rate document add/remove changes with totals', async () => {
+        type DocumentDiffItem = {
+            name: string
+            sha256: string
+        }
+
+        const diffDocumentsByShaAndName = (
```
ronak-truss: I think I might be missing something. 

What's the purpose of buildDocumentNameChanges? If we have that function and unit tests for that do we need this test and this diffDocumentsByShaAndName helper function?

ronak-truss (approved):
Afaict, this looks good to me, barring this one comment. 

@JasonLin0991 / @ruizajtruss do you mind taking a quick pass and approving?

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.test.ts
vyakusha: @ronak-truss Yes, `buildDocumentNameChanges` has its own unit coverage, therefore `diffDocumentsByShaAndName` helper in the test is a duplication of a sort. At the same time I think this higher-level test is also valuable because it verifies that `buidRevisionDiff` correctly writes document changes into to overall diff, including totals and contract - rate grouping. Should we just replace `diffDocumentsByShaAndName` with the explicit expected added/removed document names? We can use buildDocumentNameChanges in the test, but wouldn't it weaken the test purpose?

ronak-truss: Personally I think it's worth the tradeoff to remove the helper code. We won't really expect the document file names to actually change over time from the test.

ronak-truss: cc: @JasonLin0991 if you have strong opinions

JasonLin0991: We probably don't need this `diffDocumentsByShaAndName` function. You are defining what he original document was in `createAndSubmitTestContractWithRate` and the updated document later in `updateTestContractDraftRevision`. You should save those documents into variables as the data to assert against the diff.

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
```diff
@@ -0,0 +1,338 @@
...
+>[] = diffRateFormDataFieldConfigs.map((fieldConfig) => ({
+    fieldPath: fieldConfig.fieldPath,
+    getValue: (formData) => fieldConfig.getValue(formData),
+}))
+
+function buildRateDisplayName(rateRevision: RateRevisionType): string {
+    return rateRevision.formData.rateCertificationName ?? rateRevision.rateID
+}
+
+function isIncludedInAnotherSubmission(
```
JasonLin0991, line 185: **Question**: We shouldn't be surfacing anything with packagesWithSharedRateCerts. It's all been migrated. Is the intent of this to identify linked rates?

services/app-api/src/resolvers/contract/fetchRevisionDiff.test.ts
```diff
@@ -9,12 +9,407 @@ import { mockSubmittableHealthPlanContract } from '../../testHelpers'
...
+    addNewRateToTestContract,
+    addLinkedRateToRateInput,
+    formatRateDataForSending,
+    updateRatesInputFromDraftContract,
+    updateTestDraftRatesOnContract,
+} from '../../testHelpers/gqlRateHelpers'
 import { packageName } from '@mc-review/submissions'
 
 describe('fetchRevisionDiff', () => {
+    async function createResubmittedContractWithStateContactChanges() {
```
JasonLin0991: **Suggestion**: This is a lot of helper functions for setting up redundant submissions for testing diffs. Using the existing `'returns the store-backed diff through the GraphQL resolver'` test, we already have a submission set up, the data can include diffs on every field or collection if you wanted. Splitting the out to separate tests with setup seems redundant.

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
```diff
@@ -0,0 +1,338 @@
...
+type RateFormData = RateFormDataType
+
+type DiffFieldConfig = {
+    fieldPath: string
+    getValue: (formData: RateFormData) => unknown | Error
+}
+
+const normalizeStringArray = (values: string[]): string[] => [...values].sort()
+
+function buildBooleanFieldConfig(
```
JasonLin0991, line 34: **Thought**: There is a set of the same functions for contracts in `revisionDiffHelpers.ts`. Either using generic types or union types I bet we could have used a single set of function for these common field types. No need to make this update in this PR, just something to think about in the future.

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    return []
+})
+
+/**
+ * lists rate form fields that are not already excluded, overridden, or supported by existing auto-diffing.
+ * used to test if new rate form fields are added which need to be specifically handled in the revision diff logic.
+ */
+function getUnhandledRateDiffFieldPaths(): string[] {
```
JasonLin0991, line 142: **Question**: I see this in unit tests but it doesn't seem like it's being used. Is this setting up for future work?

services/app-api/src/domain-models/contractAndRates/revisionDiffTypes.ts
JasonLin0991: **Suggestion**: I think the GraphQL has the same field. We should try to match that casing.

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
```diff
@@ -0,0 +1,338 @@
...
+const scalarRateFormDataFieldConfigs: ScalarDiffFieldConfig<
+    RateFormData,
+    void
+>[] = diffRateFormDataFieldConfigs.map((fieldConfig) => ({
+    fieldPath: fieldConfig.fieldPath,
+    getValue: (formData) => fieldConfig.getValue(formData),
+}))
+
+function buildRateDisplayName(rateRevision: RateRevisionType): string {
+    return rateRevision.formData.rateCertificationName ?? rateRevision.rateID
```
JasonLin0991: **Issue**: I don't think we should fallback to the rateID. In the frontend app, we handle `rateCertificationName` value of `undefined` in multiple ways depending on how and when it is being viewed. In this scenario we are strictly returning a data diff so I think it should be undefined.

In the emailer, we can use display undefined as `Unknown rate` or something similar. This is how it is handled on the frontend.

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
```diff
@@ -0,0 +1,338 @@
...
+
+    for (const [rateID, olderRateRevision] of olderRatesByID) {
+        const newerRateRevision = newerRatesByID.get(rateID)
+
+        if (!newerRateRevision) {
+            removed.push(buildRemovedRate(olderRateRevision))
+            continue
+        }
+
+        if (olderRateRevision.id !== newerRateRevision.id) {
```
JasonLin0991, line 328: **Issue**: When unlocking, a child rate will always unlock and will create a new rate revision. So using a revision id check will not reliably determine if there was an update. Likely have to use the results of the diff to determine if there was changes.

This would work for linked rates though, since those will not create a new revision if the parent contract was not unlocked and resubmitted.

Let me know if you want to pair on this. It gets deeper into the complex data model. You can also ask Codex to use the API AGENT skills to explain more.

services/app-api/src/postgres/revisionDiff/revisionDiffDocuments.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    const contractSupportingDocuments = buildDocumentNameChanges(
+        olderSubmission.contractRevision.formData.supportingDocuments,
+        newerSubmission.contractRevision.formData.supportingDocuments
+    )
+    if (contractSupportingDocuments instanceof Error) {
+        return contractSupportingDocuments
+    }
+
+    const ratesDocuments = buildRateDocumentGroups(
```
JasonLin0991: **Question**: Why is this being built here and not in the `buildRevisedRate` function?

services/app-api/src/domain-models/contractAndRates/revisionDiffTypes.ts
JasonLin0991: **Thought**: Maybe some future work, but we have `kind` meaning two things:
1. Operation
2. Field type

We likely want to think about how to separate the two in the future as it could be confusing to see the `kind` field in the same diff data model that have different meanings. Also it would be much better to align the domain type models with the GraphQL types. IE. caps vs non-cap strings

services/app-api/src/postgres/revisionDiff/revisionDiffDocuments.test.ts
```diff
@@ -0,0 +1,218 @@
...
+    updatedReason: 'submit',
+})
+
+const mockDocument = (name: string, sha256: string): DocumentType => ({
+    name,
+    s3URL: `s3://bucketname/key/${name}`,
+    sha256,
+})
+
+const mockRateRevision = (
```
JasonLin0991, line 30: **Nit**: There are some similar mock functions in the codebase that could be reused here. They also take in overrides if you want to customize the mock data. This is a nit, because those existing functions can be hard to use for this test but would reduce the amount of duplicate code.

JasonLin0991 (changes requested):
Overall good work. There are some complexities with the data model that was missed in this logic. Mainly how to determine linked rates and identifying when a rate was revised.

Let me know if you want to pair on any of these comments. Link to [conventional comments](https://conventionalcomments.org/) if you want to know more about the tags.

services/app-api/src/postgres/revisionDiff/revisionDiffHelpers.ts
vyakusha: When I add ZodNativeEnum to the function `isStringEnumLikeSchema`, I am getting error "Property 'ZodNativeEnum' does not exist on type 'typeof import("/Users/vyakusha/code/managed-care-review/node_modules/.pnpm/[zod@4.3.6](mailto:zod@4.3.6)/node_modules/zod/v4/classic/external", and adding it can be brittle (or so I think). For future-proofing of isStringEnumLikeSchema I added some fallback handling for enum-shaped zod objects to it. Do you think it is sufficient?

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
JasonLin0991: Update: Vira and I paired. We will be moving the work of identifying a linked rate in a separate ticket.

vyakusha: Update. I created a new ticket [MCR-6481](https://jiraent.cms.gov/browse/MCR-6481) to work on identifying linked rate

services/app-api/src/postgres/revisionDiff/revisionDiffDocuments.ts
vyakusha: Yes, I moved Rate documents logic into rate-specific code, to make rate revision the source of truth for the rate document changes.

services/app-api/src/domain-models/contractAndRates/revisionDiffTypes.ts
vyakusha: You are right, use of kind introduces confusion. I did refactor the `kind` as `operation` to `changeType` and `kind` as field type to `valueType`.

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
vyakusha: Added code to compare Rate revisions for actual changes instead of relying on rate revision ID

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
vyakusha: Changed code to allow for `undefined` instead of falling back on `rateID`

services/app-api/src/resolvers/contract/fetchRevisionDiff.test.ts
vyakusha: Consolidated all helpers into `createResubmittedContractWithAllDiffChanges` to use with all tests

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
JasonLin0991: Suggestion:
- delete this function `isIncludedInAnotherSubmission`
- default `includedInAnotherSubmission` on line 210 below to `false`. Add a comment about how this will be done in MCR-6481 on 210.
- Remove any unit tests that assert on this.

I doubt we will be doing this work in here, but even if we do we should remove dead code as to not try to build on top of something that will not work.

services/app-api/src/postgres/revisionDiff/revisionDiffRates.ts
```diff
@@ -0,0 +1,372 @@
...
```
JasonLin0991, line 328: **Nit**: The conditional really doesn't do anything `olderRateRevision.id !== newerRateRevision.id`. The only the revision id's would equal is on a linked rate that was never unlocked and resubmitted from its parent submission.

JasonLin0991 (approved):
Looks good. Thanks for the changes.

PR #4660, opened 2026-07-28
MCR-6455: Contract data refactor feature flag db schema update and api work

JasonLin0991:
## Summary
[MCR-6455](https://jiraent.cms.gov/browse/MCR-6455)

This work expands our state and actuary contact data models to add `givenName` and `familyName` so that we can require the user to input names separately instead of a single input. 

There are other parts of a name that was also added to the model, but are not surfaced in the API. These are added so existing names can have fields to populate and opens the ability to surface inputs to the UI if the business decides to.

Out of scope: API writes to these new fields.

AC:
- Add a feature flag for the epic.
   - Example flag name: contact-model-update.
- Update database to include the following fields in State and Actuary contact tables
   - givenName, familyName, prefix, middleName, suffix
   - These fields are nullable.
   - The existing full name field still must exist in the database.
- Update API code so at least givenName and lastName is surfaced in the response data
   - Update domain type
   - Update parsing functions
      - populate new `givenName` and `familyName`. The existing full name field still must exist and we should construct it using givenName and familyName, then fallback to the database full name.
   - Update GraphQL types and schemas
- Does not block existing API code that writes to the full name field in DB.

#### Test cases covered
Leaving majority of tests to be included in [MCR-6456](https://jiraent.cms.gov/browse/MCR-6456). The unit tests updates here is to include the new fields in mock test data.

## QA guidance
- Make sure the application can still add, update, and remove contacts (both state and actuary) from a submission.
- Make sure submission still works
- Make sure unlock does not lose contact data

## PR Reminders


ruizajtruss (approved):
Code looks good and I was able to repetitively add, remove, and edit contacts 👍

PR #4672, opened 2026-07-31
MCR-6456: update UI and api to write to new contact fields

JasonLin0991:
## Summary
[MCR-6456 ](https://jiraent.cms.gov/browse/MCR-6456)

This work adds first name (givenName), last name (familyName), and suffix inputs to the State contacts and Actuary contacts form. The UI is behind a feature flag.

The API work is mostly pre-existing for updating `givenName` and `familyName`. This PR adds the suffix fields to the update API. This was requested by the business to include contact suffix for both state and actuary contacts.

AC:
- [Not Feature Flagged] Update updateDraftContractRevision mutation to to take in first name, last name, and suffix in formData input for State contacts.
   - Write first names and last names into DB, maybe feature flag is not needed since they are optional fields and non-null validation is on submit
- [Not Feature Flagged] Update updateDraftContractRates mutation to take in first name, last name, and suffix input for Actuary contacts.
   - Write first names and last names into DB, maybe feature flag is not needed since they are optional fields and non-null validation is on submit
- [Feature Flagged] Update State contacts and actuary Contacts UI to split out full names into first name, last name, and suffix inputs.
   - Both name inputs are required in the UI
- Unit tests

#### Test cases covered

Unit Tests:
- Updated all existing mocks and tests to assert on new UI. Old UI will be removed after this is live so this should help with that work.

Cypress Tests:
- Updated existing `rateDetails.spec` and `submissionForm.spec` Cypress tests to run using the new fields.
- Update apiCommands Cypress commands to create submissions using new contact fields.

## QA guidance
- With feature flag on (Note to not mix contacts from feature flag off and on)
   - Validate first name, last name, and suffix can be saved on state contacts and actuary contacts.
   - Validate submissions can be submitted.
   - Validate contacts are retained after unlock and resubmits
   - Validate review and submit, summary, and previous submission summary pages show full names including suffixes.
- With feature flag off (Note to not mix contacts from feature flag off and on)
   - Validate that single full name input still can save and submissions can still be submitted.
   - Validate contacts are retained after unlock and resubmits 
   - Validate review and submit, summary, and previous submission summary pages show full names from the single name input.

## PR Reminders


services/app-api/src/resolvers/contract/updateContractDraftRevision.test.ts
```diff
@@ -77,13 +77,24 @@ describe(`Tests UpdateContractDraftRevision`, () => {
...
-                familyName: null,
-                titleRole: contact.titleRole,
-                email: contact.email,
-            }))
+            ).map((contact) => {
+                // Mirror constructContactName: build name from givenName +
+                // familyName (no suffix), falling back to the stored name.
+                const constructedName = [contact.givenName, contact.familyName]
+                    .filter((part) => part != null && part.trim() !== '')
+                    .join(' ')
```
ronak-truss, line 85: Should we just export constructContactName and use that here?

services/app-api/src/resolvers/contract/updateDraftContractRates.test.ts
```diff
@@ -493,6 +493,110 @@ describe('updateDraftContractRates', () => {
...
+        expect(draftRates).toHaveLength(1)
+
+        const rateFormData = draftRates[0].draftRevision.formData
+
+        // expect the constructed name to include givenName and familyName only
+        // (no suffix), while suffix is still stored.
+        expect(rateFormData.certifyingActuaryContacts).toEqual([
+            expect.objectContaining({
+                name: 'Foo Person',
+                givenName: 'Foo',
```
ronak-truss, line 575: I don't have a strong opinion here, but for the values except name, we could use a shared variable / reference the original object passed into the update above.

services/app-api/src/testHelpers/gqlRateHelpers.ts
```diff
@@ -325,6 +325,12 @@ function formatGQLRateContractForSending(
     return {
         id: contact.id || undefined,
         name: contact.name,
```
ronak-truss, line 327: In some future state we are planning to make name optional and have given and family name be required right?

services/app-web/src/components/DataDetail/DataDetailContactField/DataDetailContactField.test.tsx
```diff
@@ -5,30 +5,67 @@ import { DataDetailContactField } from './DataDetailContactField'
...
+        expect(screen.queryByText(/Structured Contact Jr./)).toBeNull()
         expect(screen.getByText(/Writer/)).toBeInTheDocument()
         expect(
             screen.getByRole('link', { name: 'wedsaddams@example.com' })
         ).toBeInTheDocument()
     })
 
-    it('renders actuarial field when relevant', () => {
+    it('renders the structured name when the contact model update flag is on', async () => {
+        const contact: StateContact = {
```
ronak-truss, line 31: Nit: For this and the previous test I would maybe rewrite this so that the contact info is the same and we just show the differences based on that.

services/app-web/src/components/DataDetail/DataDetailContactField/DataDetailContactField.tsx
```diff
@@ -15,12 +18,26 @@ export const DataDetailContactField = ({
...
+    const ldClient = useLDClient()
+    const useStructuredContactName = ldClient?.variation(
+        featureFlags.CONTACT_MODEL_UPDATE.flag,
+        featureFlags.CONTACT_MODEL_UPDATE.defaultValue
+    )
+
+    // Display full name constructed using name fields when feature flag is on.
+    const displayName = useStructuredContactName
+        ? [contact?.givenName, contact?.familyName, contact?.suffix]
+              .map((namePart) => namePart?.trim())
```
ronak-truss, line 30: Should we be doing processing (trim) here? Should we just be doing this on the save so that we prevent these from entering the DB and then don't have to worry about doing modifications later?

ronak-truss: Looking at the input field changes, it seems like we are preventing this in the future. Do we have any intentions to retroactively update entries so that we can remove this logic or are we perpetually need to account for trimming in any areas that we use name fields?

services/app-web/src/pages/StateSubmission/HealthPlanSubmission/ContractDetails/ContractDetails.tsx
```diff
@@ -579,10 +579,12 @@ export const ContractDetails = ({
                         })
                     }}
                     validationSchema={() =>
-                        ContractDetailsFormSchema(
-                            draftSubmission,
-                            ldClient?.allFlags()
-                        )
+                        ContractDetailsFormSchema(draftSubmission, {
```
ronak-truss, line 582: Can you explain this change?

ronak-truss (commented):
Generally looks good to me. Left some comments/questions based on an initial pass.

services/app-api/src/resolvers/contract/updateContractDraftRevision.test.ts
JasonLin0991: Idk if we want to use `constructContactName`, since the response we are asserting is using the `constructContactName` function. Maybe hardcoding the contacts to assert the response data with  might be better since this code is exactly like that function.

services/app-api/src/resolvers/contract/updateDraftContractRates.test.ts
JasonLin0991: Yea that could prevent errors from updating the mock data. I think I will make this change here and at the  comment above this one.

services/app-api/src/testHelpers/gqlRateHelpers.ts
JasonLin0991: In the future the `name` field will be removed, no idea when, it depends on external app that consumes the API. Currently the GraphQL response has this field as nullable, same as the DB. I think this particular function is used in unit tests and not actual code.

services/app-web/src/components/DataDetail/DataDetailContactField/DataDetailContactField.tsx
JasonLin0991: Good point, we are prevent extra spaces at this point so no need to trim again.

> Looking at the input field changes, it seems like we are preventing this in the future. Do we have any intentions to retroactively update entries so that we can remove this logic or are we perpetually need to account for trimming in any areas that we use name fields?

We have a ticket to run a migration on existing data to put them into this new data model and trim the spaces. So after that we should not need the extra trim.

services/app-web/src/pages/StateSubmission/HealthPlanSubmission/ContractDetails/ContractDetails.tsx
JasonLin0991: Yes! Using `allFlags()` there is a chance it will [return null instead of the default value for a flag](https://launchdarkly.com/docs/sdk/features/evaluating#react-web). We have defined default values in our codebase, so using the `variation()` method we can use those default values as a fallback.

In prod, if LaunchDarkly was not reachable, we want to fallback to defaults and not necessarily undefined.

ronak-truss (approved):
Functionally this looks good to me. I didn't verify if there are missing areas where we needed to make similar changes. 

(Aside) I'm still running into issues locally uploading docs where it reloads the page because the upload fails because of a CORS issue. I assume other's aren't running into this issue and it might be a setup issue.

vyakusha (approved):
All code LGTM. Great job!

PR #4677, opened 2026-08-03
Vy 6326 add diff values to submission email part 1

vyakusha:
## Summary
Surface Contract DIFF values in the CMS Resubmit email. Part 1: Submission Type, Contract Details, Contract Provisions

#### Related issues
[MCR-6326](https://jiraent.cms.gov/browse/MCR-6326)
#### Screenshots
Screenshot 2026-08-03 at 2 45 28 PM
Screenshot 2026-08-03 at 2 45 51 PM

#### Test cases covered
resubmitRevisionChanges.test.ts
resubmitContractCMSEmail.test.ts

## QA guidance
Locally, log in as a State user, submit Health Plan contract. 
As a CMS Approver user, unlock this record. 
As a State user re-open the contract record, make edits and re-submit. 
Use MC-Review Local Dev Console to examine the content of the CMS Re-submit email, compare the result with data changes and [Figma design](https://www.figma.com/design/frKNnm6lkcpdjJ2eVJE8Bj/Managed-Care-Review?node-id=28705-5195&p=f&m=dev).

## PR Reminders

### DIFF
--- a/services/app-api/src/emailer/emails/resubmitRevisionChanges.ts
+++ b/services/app-api/src/emailer/emails/resubmitRevisionChanges.ts
@@ -1,14 +1,20 @@
 import { formatCalendarDate } from '@mc-review/dates'
-import type {
-    ProgramType,
-    RevisionDiff,
-    RevisionDiffFieldChange,
+import {
+    ContractExecutionStatusRecord,
+    FederalAuthorityRecord,
+} from '@mc-review/submissions'
+import {
+    type ContractType,
+    type ProgramType,
+    type RevisionDiff,
+    type RevisionDiffFieldChange,
 } from '../../domain-models'
 
 type ResubmitRevisionChangeRow = {
     label: string
-    oldValue: string
     newValue: string
+    oldValue?: string
+    isNew?: boolean
 }
 
 type ResubmitRevisionChangeSection = {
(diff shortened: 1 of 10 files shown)


services/app-api/src/emailer/etaTemplates/reusableTemplates/resubmitRevisionChanges.eta
```diff
@@ -0,0 +1,33 @@
+<hr />
+<div style="height: 32px; line-height: 32px;">&nbsp;</div>
```
ruizajtruss: Is this intended for spacing? If so we should update these to use `<br />` instead as the current approach doesn't strip cleanly and can cause issues with certain screen readers

services/app-api/src/emailer/emails/resubmitRevisionChanges.ts
```diff
@@ -0,0 +1,563 @@
...
+
+const getProvisionLabel = (
+    fieldPath: ContractProvisionsFieldPath,
+    currentContract: ContractType
+): string => {
+    const baseLabelConfig: Record<ContractProvisionsFieldPath, string> = {
+        inLieuServicesAndSettings: 'In Lieu-of Services and Settings',
+        modifiedBenefitsProvided: 'Benefits provided',
+        modifiedGeoAreaServed: 'Geo area served',
+        modifiedMedicaidBeneficiaries: isChipOnlyContract(currentContract)
```
ruizajtruss, line 396: isChipOnlyContract (resubmitRevisionChanges.ts:102) and the 'CHIP beneficiaries' ternary (:387) end up just falling through. resubmitContractCMSEmail.ts:85 routes populationCovered === 'CHIP' to resubmitChipOnlyCMSEmail.eta, which doesn't include the diff section, so the only case that produces the CHIP label is the same case where the whole object gets discarded. I looked at the design and it says this label would pop up for chip-only but the design also shows that chip only gets a completely separate email. Might have to follow up with Anna on this issue to get clarification, unless I'm misunderstanding this issue

services/app-api/src/emailer/emails/resubmitRevisionChanges.ts
```diff
@@ -0,0 +1,563 @@
...
+    | 'modifiedStateDirectedPayments'
+    | 'modifiedPassThroughPayments'
+    | 'modifiedMedicalLossRatioStandards'
+    | 'modifiedOtherFinancialPaymentIncentive'
+    | 'modifiedGrevienceAndAppeal'
+    | 'modifiedNetworkAdequacyStandards'
+    | 'modifiedLengthOfContract'
+    | 'modifiedNonRiskPaymentArrangements'
+    | 'modifiedPaymentsForMentalDiseaseInstitutions'
+
```
ruizajtruss, line 68: I believe the `modifiedEnrollmentProcess` field is missing

services/app-api/src/emailer/emails/resubmitRevisionChanges.ts
```diff
@@ -0,0 +1,563 @@
...
+type SubmissionTypeFieldPath =
+    | 'contractName'
+    | 'populationCovered'
+    | 'submissionType'
+    | 'contractType'
+    | 'riskBasedContract'
+    | 'programIDs'
+    | 'submissionDescription'
+
+type ContractDetailsFieldPath =
```
ruizajtruss, line 42: I believe that `statutoryRegulatoryAttestationDescription` and `statutoryRegulatoryAttestation` are missing here aswell

services/app-api/src/emailer/emails/resubmitRevisionChanges.ts
```diff
@@ -0,0 +1,563 @@
...
+        return undefined
+    }
+
+    if (oldValue === undefined) {
+        return {
+            label,
+            newValue,
+            isNew: true,
+        }
+    }
```
ruizajtruss: I think this might need minor rework as, according to the AC, we should expect to see `Fields no longer present show the old value → ⎯ (no new value)`. But, with the current setup, I believe this label would just not display at all. 

Example:
Submit    → federal authority: 1115 Waiver
            D-SNP question appears → "Yes"

Unlock    → CMS unlocks

Edit      → swap 1115 Waiver for 1937 Benchmark
            D-SNP question disappears → value cleared to undefined

Resubmit  → diff: federalAuthorities changed, dsnpContract true → undefined

Email shows: Managed Care authorities: 1115 Waiver Authority → 1937 Benchmark Authority

Email omits: Associated with a D-SNP: Yes → ⎯

The reviewer sees the authority change but has no idea the D-SNP designation went away with it.

ruizajtruss (changes requested):
Overall looks great! Just a few comments, I spotted that some fields were not included but if that was intentional then totally ignore me

services/app-api/src/emailer/emails/resubmitRevisionChanges.ts
vyakusha: Talked to Anna about this. She said we should not include statutoryRegulatoryAttestation and related Description

services/app-api/src/emailer/emails/resubmitRevisionChanges.ts
vyakusha: Fixed

services/app-api/src/emailer/emails/resubmitRevisionChanges.ts
vyakusha: Fixed

services/app-api/src/emailer/emails/resubmitRevisionChanges.ts
vyakusha: Added treatment for "CHIP-only" contract, those values are now injected into resubmitChipOnlyCMSEmail

services/app-api/src/emailer/etaTemplates/reusableTemplates/resubmitRevisionChanges.eta
vyakusha: Updated

ruizajtruss (approved):
LGTM!

PR #4678, opened 2026-08-03
[MCR-6347] undo unlock change history

ruizajtruss:
## Summary
As a CMS user or State user, I want the Undo Unlock action to be recorded in the submission's Change History so that there is a complete and accurate audit trail of all actions taken on the submission.

**Acceptance Criteria:**
- When a CMS user performs an Undo Unlock action on a managed care submission, a new entry is added to the submission's Change History
- The Change History section of the Submission Summary is updated to reflect the Undo Unlock action, following existing patterns for the Change history section.
- The Change History event will include the date, time, and the event of "Undo unlock"
- When expanded, the history event will include:
  - Updated by: [email address of the user who performed the undo unlock ation]
  - Status: [Status that the submission was returned to after the unlock was undone]
  - Reason for undoing the unlock: [User entered verbiage from the undo unlock action]
- The Change History entry is visible to both CMS users and State users on the Submission Summary page
- The Change History entry is consistent in format and style with existing Change History entries for other submission actions
- The Undo Unlock action does not break or alter any existing Change History entries or the current-day revision history of the submission
#### Related issues
[MCR-6347](https://jiraent.cms.gov/browse/MCR-6347)
#### Screenshots
image

## PR Reminders

### DIFF
--- a/services/app-api/src/resolvers/contract/contractResolver.ts
+++ b/services/app-api/src/resolvers/contract/contractResolver.ts
@@ -254,7 +254,7 @@ function genericContractResolver<
             )
         },
 
-        undoUnlockSubmissions(parent: ParentType) {
+        undoUnlockPackages(parent: ParentType) {
             return (parent.undoUnlockPackages ?? []).map((undoUnlock) => ({
                 undoUnlockInfo: undoUnlock.undoUnlockInfo,
                 reversedUnlockInfo:
--- a/services/app-api/src/resolvers/contract/contractResolver.ts
+++ b/services/app-api/src/resolvers/contract/contractResolver.ts
@@ -254,6 +254,14 @@ function genericContractResolver<
             )
         },
 
+        undoUnlockSubmissions(parent: ParentType) {
+            return (parent.undoUnlockPackages ?? []).map((undoUnlock) => ({
+                undoUnlockInfo: undoUnlock.undoUnlockInfo,
+                reversedUnlockInfo:
+                    undoUnlock.draftContractRevisionSnapshot.unlockInfo,
+            }))
+        },
+
         questions: async (
             parent: ParentType,
             _args: Record<string, never>,
--- a/services/app-api/src/resolvers/contract/contractResolver.ts
+++ b/services/app-api/src/resolvers/contract/contractResolver.ts
@@ -254,14 +254,6 @@ function genericContractResolver<
             )
         },
 
-        undoUnlockPackages(parent: ParentType) {
-            return (parent.undoUnlockPackages ?? []).map((undoUnlock) => ({
-                undoUnlockInfo: undoUnlock.undoUnlockInfo,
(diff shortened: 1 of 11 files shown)


ronak-truss:
What's the expected behavior of the last updated column for state users?

After unlock:
image

After undo unlock:
image

Is it intentional that the undo unlock reverts the last updated column?

services/app-api/src/resolvers/contract/contractResolver.ts
```diff
@@ -254,6 +254,14 @@ function genericContractResolver<
             )
         },
 
+        undoUnlockSubmissions(parent: ParentType) {
+            return (parent.undoUnlockPackages ?? []).map((undoUnlock) => ({
+                undoUnlockInfo: undoUnlock.undoUnlockInfo,
+                reversedUnlockInfo:
```
JasonLin0991: Question: What is the different between `undoUnlockInfo` and `reversedUnlockInfo`? I think they are the same thing.

Suggestion: It might be a good idea to follow the same data model pattern as `ContractPackageSubmission`.

services/app-web/src/components/ChangeHistory/ChangeHistory.tsx
JasonLin0991, line 319: Suggestion: If we align the models, so they have similar fields, we can put them in this array.

services/app-api/src/resolvers/contract/contractResolver.ts
```diff
@@ -254,6 +254,14 @@ function genericContractResolver<
             )
         },
 
+        undoUnlockSubmissions(parent: ParentType) {
```
JasonLin0991: Nit: Could we align this with the domain model and call it `UndoUnlockPackages`?

services/app-api/src/resolvers/contract/contractResolver.ts
JasonLin0991: Oh I see now one is the `undoUnlockInfo` and the other is the `unlockInfo`. I wonder if we just surface `draftContractRevisionSnapshot` instead of `reversedUnlockInfo`. I think the if we aligned the schema it would look something like:

```ts
type UndoUnlockPackages {
    cause: SubmissionReason!
    undoUnlockInfo: UpdateInformation!
    draftContractRevisionSnapshot: UndoUnlockContractRevision! // <- Does not need to have complete data yet.
}
```

It's more work but it would align the data to be more similar to packageSubmissions and make it easier to work with in any client.

ruizajtruss:
> What's the expected behavior of the last updated column for state users?
> 
> After unlock: image
> 
> After undo unlock: image
> 
> Is it intentional that the undo unlock reverts the last updated column?

This is just a result of the  `use-stored-contract-action-dates` flag being off

JasonLin0991 (approved):
Thanks for those changes! Looks good.

PR #4684, opened 2026-08-10
[MCR-6482] Review decision is not displayed for submissions with populations other than CHIP-only

ruizajtruss:
## Summary
Bug discovered on submission [MCR-FL-0135-MMA](https://mc-review-dev.onemac.cms.gov/submissions/health-plan/fd6ff171-76ff-4a1c-9076-871b7781ce25) 

**Current behavior:**
- Managed Care Plan (MCP) (health plan) submissions that should NOT have a review decision displayed (submissions with populations other than CHIP-only) in the submitted status, display "Review decision: Submitted" on the submission summary
- Review decision submitted is visible on previous submission versions

**Expected behavior:**
- Review decision is not displayed

**Steps to reproduce:**
- Log in as a state user
- Find an unlocked MCP (health plan submission) that was previously submitted with Medicaid and CHIP populations
- While unlocked, change the populations to Medicaid
- Resubmit
- View the resubmitted submission as a state user
- View past submission version

#### Related issues
[MCR-6482](https://jiraent.cms.gov/browse/MCR-6482)
#### Screenshots
Non-CHIP only:
image

CHIP-only:
image

## QA guidance
1. Log in as a state user and open a Medicaid (non-CHIP) health plan submission that's been submitted -> no "Review decision" row in the Submission type section
2. On that submission, open Change history -> View past submission version -> still no "Review decision" row
3. Open a CHIP-only submission -> "Review decision: Not subject to DMCO review and validation" still shows
4. Log in as a CMS user, repeat step 1 -> no "Review decision" row

## PR Reminders

### DIFF
--- a/services/app-web/src/components/SubmissionSummarySection/SubmissionTypeSummarySection/SubmissionTypeSummarySection.tsx
+++ b/services/app-web/src/components/SubmissionSummarySection/SubmissionTypeSummarySection/SubmissionTypeSummarySection.tsx
@@ -26,6 +26,8 @@ import {
 } from '../SummarySectionFields'
 import { formattedProgramNames } from '../../../formHelpers'
 import { getConsolidatedContractStatusText } from '../../ContractTable'
+import { useLDClient } from 'launchdarkly-react-client-sdk'
+import { featureFlags } from '@mc-review/common-code'
 
 export type SubmissionTypeSummarySectionProps = {
     contract: Contract | UnlockedContract
@@ -51,12 +53,18 @@ export const SubmissionTypeSummarySection = ({
     explainMissingData,
 }: SubmissionTypeSummarySectionProps): React.ReactElement => {
     const contractOrRev = contractRev ? contractRev : contract
+    const ldClient = useLDClient()
     const contractFormData = getVisibleLatestContractFormData(
         contractOrRev,
         isStateUser
     )
 
     if (!contractFormData) return <GenericErrorPage />
+
+    const chipSubmissionAutomation = ldClient?.variation(
+        featureFlags.CHIP_SUBMISSION_AUTOMATION.flag,
+        featureFlags.CHIP_SUBMISSION_AUTOMATION.defaultValue
+    )
 
     const programIDs = contractFormData?.programIDs ?? []
     const programNames = formattedProgramNames(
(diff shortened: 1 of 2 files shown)


services/app-web/src/components/SubmissionSummarySection/SubmissionTypeSummarySection/SubmissionTypeSummarySection.tsx
```diff
@@ -86,20 +94,22 @@ export const SubmissionTypeSummarySection = ({
...
+                {contract.consolidatedStatus &&
+                    isSubmitted &&
+                    chipSubmissionAutomation &&
+                    contractFormData.populationCovered === 'CHIP' && (
+                        <ReviewDecisionSummary
+                            reviewDecision={
+                                contract.consolidatedStatus ===
+                                'NOT_SUBJECT_TO_REVIEW'
+                                    ? 'Not subject to DMCO review and validation'
+                                    : getConsolidatedContractStatusText(
```
JasonLin0991: **Question**: I think the review decision section should on be either subject to review or not for CHIP, it would always be not subject to review..  I don't think if supposed to display getConsolidatedContractStatusText(). 

Take a look at how `services/app-web/src/components/SubmissionSummarySection/SubmissionTypeSummarySection/EQROSubmissionTypeSummarySection.tsx` handles this because it handles this section WAY different than health plans. Which I am suprised it is isn't aligned with EQRO since review decision is binary.

Also, I could have sworn, in the past, we made the decision to display Review decision for all submission types including health plans. The bug to me is that it was showing `getConsolidatedContractStatusText` not just the review decision.

I think we should discuss this again with Anna. Because it seems to break the pattern if Health plans don't have a review status unless it is CHIP, but really health plans other than CHIP is subject to review.

JasonLin0991 (commented):
I think this is ok, but I would want to confirm with Anna that we really want to remove review decision for non CHIP health plan submissions. Seems like breaking the pattern since all other submission types have them.

To me the bug is that status is being displayed in the review decision.

services/app-web/src/components/SubmissionSummarySection/SubmissionTypeSummarySection/SubmissionTypeSummarySection.tsx
ruizajtruss: I see what you're saying, touching base with Anna to see what the answer is

JasonLin0991 (approved):

PR #4689, opened 2026-08-11
MCR-6483: provisions display incorrectly on past submissions

JasonLin0991:
## Summary
[MCR-6483](https://jiraent.cms.gov/browse/MCR-6483)

The issue here was that the components that render the provision data took the entire `contract` data as a prop and then parsing out the latest revision's form data to then display the provisions.

The issue comes when viewing a past submission, we cannot use the latest revision we have to use the revision being viewied. So when viewing a past submission the component was being passed the full `contract` data and not the specific past revision's data. This resulted in past submissions always showing the latest submission provisions.

The fix is was to refactor the provision components to take in a `ContractFormData` instead of the full contract, so that it can be targeted with the data we want to render. This refactor touched other parts of the app, but was pretty minimal.

Current behavior:
- When comparing the current and previous versions of a submission in the UI, the previous version's provisions selections match the current version's selections instead of 
- The resubmit email for this same submission generates the correct diff, confirming the underlying data is accurate and that the defect lies in how the comparison UI reads or displays it.

Expected behavior:
- The previous version's provision selections should reflect what was actually selected at that point in time.

#### Test cases covered
- SubmissionRevisionSummary.test.tsx
   - `'renders modified provisions from the revision being viewed, not the latest revision'`

## QA guidance

- This work updates the component that displays provisions, so it's worth while to take a close look to make sure we do no introduce more regressions.
   - Make sure latest submission summary displays correct provisions
   - Make sure Contract Details page, on health plan submissions, display provisions correctly. While it may not be using the provision components it did used a lot of the shared helper functions that use to take `Contract` as a parameter and now it takes `ContractFormData`.

## PR Reminders

### DIFF
--- a/packages/submissions/src/contract/healthPlanFormDataValidtions.ts
+++ b/packages/submissions/src/contract/healthPlanFormDataValidtions.ts
@@ -102,16 +102,20 @@ const isMedicaidAmendmentProvision = (
 
 // Returns the list of provision keys that apply for given submission variant
 const generateApplicableProvisionsList = (
-    draftSubmission: Contract | UnlockedContract
+    formData: ContractFormData
 ):
     | CHIPProvisionType[]
     | MedicaidBaseProvisionType[]
     | MedicaidAmendmentProvisionType[] => {
-    if (isCHIPOnly(draftSubmission)) {
-        return isContractAmendment(draftSubmission)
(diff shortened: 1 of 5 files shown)


ruizajtruss (approved):
No notes, everything LGTM 🚀

PR #4692, opened 2026-08-12
Fix missing fields in tests that caused type errors.

JasonLin0991:
## Summary

Missing fields in test data caused type errors.

## PR Reminders


vyakusha (approved):
Excellent fix. Thank you!

PR #4699, opened 2026-08-13
MCR-6345: cms email undo unlock

JasonLin0991:
## Summary
[MCR-6345](https://jiraent.cms.gov/browse/MCR-6345)

@ronaktruss work adds the CMS email notification for undo undo unlock. It also contains some refactoring work to abstract out code to get state analysts and programs in the resolver.

Acceptance Criteria

- A CMS user email notification is sent when a CMS user performs an Undo Unlock action on a managed care submission
- The email notification is sent to the DMCO inbox (MCGDMCOActions@cms.hhs.gov) and any assigned DMCO users associated with the submission, consistent with the recipient pattern established for other MC-Review State Portal actions
- The email content follows the approved email content design produced in the email content design task for this epic ([MCR-6343](https://jiraent.cms.gov/browse/MCR-6343))
- The email includes the following information:
   - Title which includes the Submission ID and describes the submission status change
   - "Unlock reversed by: [name of the CMS user who performed the Undo Unlock action]"
   - "Status: [restored pre-unlock status of the submission]"
   - "Reason for unlock reversal: [reason as entered by the CMS user on the Undo Unlock page]"
   - A link allowing the recipient to view the submission in MC-Review
- The email is sent at the time the Undo Unlock action is confirmed by the CMS user
- The email follows the established email formatting and styling patterns used for other MC-Review State Portal action notifications

## PR Reminders


services/app-api/src/resolvers/helpers.ts
```diff
@@ -93,3 +98,56 @@ export function getRateLastUpdatedForDisplay(
         latestReviewAction,
     ])
 }
+
+export async function getStateAnalystsEmails(
```
JasonLin0991, line 109: **Thought**: These abstraction looks good to me. One thought is that our usual pattern is trying to handle errors in the resolvers to be explicit, but this should be ok. Consider adding JSDoc comment above the function to give devs more awareness that this is a resolver helper function only. I could see it being missed even though it lives in `resolvers/helpers.ts`

services/app-api/src/resolvers/helpers.ts
```diff
@@ -93,3 +98,56 @@ export function getRateLastUpdatedForDisplay(
...
+    const stateAnalystsEmailsResult = await store.findStateAssignedUsers(
+        contractRes.stateCode as StateCodeType
+    )
+
+    if (stateAnalystsEmailsResult instanceof Error) {
+        logResolverError(
+            'getStateAnalystsEmails',
+            stateAnalystsEmailsResult.message,
+            context
+        )
```
ronak-truss, line 124: This removes the usage of `recordResolverError` in some locations. I haven't found a good reason to keep it unless the previous iterations were just missing it. We have examples where emails are being sent eg. `createContractQuestion.ts` where we weren't calling this.

JasonLin0991: I think the intended usages of `recordResolverError` here is to record issues to Otel that do not block the operation so we can be alerted. Otel is set up to alert one of our slack channels. For example in submitContract, if getting analysts emails fails we want to record to otel so we can be altered if it happens frequently. That being said, this example scenario was implemented when we use to store the analyst emails in AWS parameter store. Might not apply today, but the gist is that we want to know about frequent issues even if they don't block an operation.

That being said, the usages removed in this PR was around getting the state analysts, so I think it should be ok removing.

ronak-truss:
I added three follow-up commits [here](https://github.com/ronaktruss/managed-care-review/commits/follow-up) if someone could cherry-pick them and add them here OR just manually update.

vyakusha (approved):
Code changes LGTM. Tested manually locally, all looks good. Great job!

PR #4700, opened 2026-08-13
MCR-6465: Add QA environment to Val

mojotalantikite:
## Summary

The Salesforce/Mulesoft teams have asked for a QA environment since they point a few of their lower environments to our Val. This adds one inside the Val AWS account. It is deployed in parallel to Val.

Currently in this PR we're letting promotes continue even with a failure in deploying QA. This is because in order to deploy QA we'll need promote to run, which means we don't have a way to validate that QA will succeed until we run the first promote. We don't want to block promotes to Prod while we figure out QA. Once we've worked out any bugs with the QA deployment we can change this behavior with a follow on PR.

#### Related issues
https://jiraent.cms.gov/browse/MCR-6465


Btodhunter (approved):
lgtm

PR #4706, opened 2026-08-14
[MCR-6480] Undo unlock twice in succession drops rates on third unlock

ruizajtruss:
## Summary
When a submission is unlocked then undo unlock twice in succession, rates are dropped on the submission summary page.

**Current Behavior:**
Unlock then undo unlock twice in succession drops rates from submission when unlocked a third time.

**Expected Behavior:**
Unlock then undo unlock twice or more in succession retains existing rates on the submission.

**Reproduction Steps:**
- Make a contract and rate submission and submit it
- Unlock and undo unlock twice in succession.
- Unlock a third time and view submission as the state user.
- Rates are no longer existing on the submission

**Eng Notes:**
Console errors suggest an issue with unlocking the rate at the third unlock, without undoing unlock. `Programming Error: A child rate is not a draft`. Likely missing some logic in the unlock API code that needs to unlock when two undo unlocks happen in a row.

#### Related issues
[MCR-6480](https://jiraent.cms.gov/browse/MCR-6480)
#### Test cases covered
  - After the third unlock: the child rate is still attached to the contract and gets a fresh draft revision from that same unlock event.
  - After the resubmit: the new package stamps a new child rate revision rather than reusing the one from the initial submission.

## QA guidance
- As a state user, create and submit a contract with at least one rate.
- Cycle twice as a CMS user, unlock then undo unlock; repeat a second time.
- Third unlock, unlock again as CMS.
- Check the state view, open the unlocked submission as the state user and go to review & submit; rates should still be listed, with no "A child rate is not a draft" console error.
- submit as the state user; the rate should appear in the new submission with its data intact.

## PR Reminders

### DIFF
--- a/services/app-api/src/postgres/contractAndRates/unlockContract.ts
+++ b/services/app-api/src/postgres/contractAndRates/unlockContract.ts
@@ -114,6 +114,17 @@ async function unlockContractInsideTransaction(
             // return data only queries data we need.
             id: true,
             revisions: {
+                // Only the revision that was last submitted with a contract
+                // tells us who the current parent is. Undo unlock preserves
+                // reversed unlocked revisions for audit history. Matches how
+                // getParentContractID selects the parent-defining revision.
+                where: {
+                    submitInfo: {
+                        submittedContracts: {
+                            some: {},
+                        },
+                    },
+                },
                 orderBy: {
                     createdAt: 'desc',
                 },
@@ -129,7 +140,7 @@ async function unlockContractInsideTransaction(
                     },
                     unlockInfo: true,
                 },
-                take: 2, // Only two revision states are possible: mutually unlocked or submitted
+                take: 1,
             },
             reviewStatusActions: {
                 orderBy: {
--- a/services/app-api/src/postgres/contractAndRates/submitContract.ts
+++ b/services/app-api/src/postgres/contractAndRates/submitContract.ts
@@ -49,16 +49,9 @@ async function submitContractInsideTransaction(
             if (rate.draftRevision) {
                 unsubmittedChildRevs.push(rate.draftRevision)
             } else {
-                console.info(
-                    'Strange, a child rate is not in a draft state. Shouldnt be true while we are unlocking child rates with contracts.'
-                )
-                const latestSubmittedRate = rate.revisions[0]
-                if (!latestSubmittedRate) {
-                    return new Error(
-                        `Attempted to submit a contract connected to an unsubmitted child-rate. ContractID: ${contractID}`
-                    )
-                }
-                linkedRateRevs.push(latestSubmittedRate)
+                const msg = `Cannot submit contract ${contractID}: child rate ${rate.id} has no active draft revision`
+                console.error(msg)
+                return new Error(msg)
             }
         } else {
             // non-child rate
--- a/services/app-api/src/resolvers/contract/undoUnlockContract.test.ts
+++ b/services/app-api/src/resolvers/contract/undoUnlockContract.test.ts
@@ -632,6 +632,104 @@ describe('undoUnlockContract', () => {
         ).toBe('Initial submission')
     })
 
+    it('unlocks child rates on a third unlock after two reverse-unlock cycles', async () => {
+        const cmsServer = await constructTestPostgresServer({
+            s3Client: mockS3,
+            emailer: mockEmailer,
+            context: {
+                user: testCMSUser(),
+            },
+        })
+
+        const stateServer = await constructTestPostgresServer({
+            s3Client: mockS3,
+            emailer: mockEmailer,
+        })
+
+        const submittedContract =
+            await createAndSubmitTestContractWithRate(stateServer)
+        const childRateID =
+            submittedContract.packageSubmissions[0].rateRevisions[0].rateID
+        const initialRateRevisionID =
+            submittedContract.packageSubmissions[0].rateRevisions[0].id
+
+        await unlockTestContract(
+            cmsServer,
+            submittedContract.id,
+            'First unlock'
+        )
+        await undoUnlockTestContract(
+            cmsServer,
+            submittedContract.id,
+            'First reverse'
+        )
+        await unlockTestContract(
+            cmsServer,
+            submittedContract.id,
+            'Second unlock'
+        )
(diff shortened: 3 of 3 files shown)


services/app-api/src/postgres/contractAndRates/unlockContract.ts
```diff
@@ -114,6 +114,17 @@ async function unlockContractInsideTransaction(
             // return data only queries data we need.
             id: true,
             revisions: {
+                // Only the revision that was last submitted with a contract
```
JasonLin0991, line 117: **Question**: What was the core cause of this bug? Was there something in undo unlock that broke the previous query here?

vyakusha (approved):
I tested triple unlock - undo unlock with multiple rates, with and without saving draft changes in between unlock and 'undo unlock'. All works fine now. I think the reason behind the bug is the fact that code was looking for the last two revisions, and not reaching out 'original' revision which links Contract to Rate. This is fixed.

JasonLin0991:
>  I think the reason behind the bug is the fact that code was looking for the last two revisions, and not reaching out 'original' revision which links Contract to Rate.

@vyakusha What you mean by original revision? Could you explain more on your take?

@ruizajtruss What do you think?

Lets not merge this yet. I want to understand the root cause and also want to test if there are side effects of this fix for other actions.

ruizajtruss:
> > I think the reason behind the bug is the fact that code was looking for the last two revisions, and not reaching out 'original' revision which links Contract to Rate.
> 
> @vyakusha What you mean by original revision? Could you explain more on your take?
> 
> @ruizajtruss What do you think?
> 
> Lets not merge this yet. I want to understand the root cause and also want to test if there are side effects of this fix for other actions.

By having the code: `take: 2`, we were fetching the newest 2 rows of the rate history. This worked when they only existed in two possible states. Introducing undo unlock is what caused this particular line to not work anymore; because after two undo cycles these two rows are cancelled unlocks so it doesn't find any rates.

So the code:
`where: {
                    submitInfo: {
                        submittedContracts: {
                            some: {},
                        },
                    },
                },
 Take: 1`

filters for submitted rows only on the rate's history, keeping the rows submitted with a contract then returning that top row. The original code basically did the inverse, taking the two newest rows of any type then searched for the submitted one (now not present with undo unlock) which is why it was coming up empty.

JasonLin0991:
> > > I think the reason behind the bug is the fact that code was looking for the last two revisions, and not reaching out 'original' revision which links Contract to Rate.
> > 
> > 
> > @vyakusha What you mean by original revision? Could you explain more on your take?
> > @ruizajtruss What do you think?
> > Lets not merge this yet. I want to understand the root cause and also want to test if there are side effects of this fix for other actions.
> 
> By having the code: `take: 2`, we were fetching the newest 2 rows of the rate history. This worked when they only existed in two possible states. Introducing undo unlock is what caused this particular line to not work anymore; because after two undo cycles these two rows are cancelled unlocks so it doesn't find any rates.
> 
> So the code: `where: { submitInfo: { submittedContracts: { some: {}, }, }, }, Take: 1`
> 
> filters for submitted rows only on the rate's history, keeping the rows submitted with a contract then returning that top row. The original code basically did the inverse, taking the two newest rows of any type then searched for the submitted one (now not present with undo unlock) which is why it was coming up empty.

Ah I see. So when doing 2 undo unlocks in a row, you end up with 2 revisions with `unlockInfo` and `undoUnlockIndo`, but no `submitInfo` (This could also produce `n` number of `undoUnlockInfo` revisions). Filtering it to just revisions with submissions fixes this. I do wonder what kind of side effects this would have for other operations likes withdraw and undo withdrawal. Likely none, but wold be worth testing those action sequences.

**Suggestion**: I also wonder if the alternative here is to filter OUT revisions with `undoUnlockInfo` as those are now only for record keeping and no longer revisions to be considered in the unlock resubmit flow. Filtering to only submitted revisions gives you only the submit history and not the unlocks too. If two unlocks somehow happen we want to make sure this revision will properly error here and not add a third unlock. We can also chat about this.

ruizajtruss:
> Ah I see. So when doing 2 undo unlocks in a row, you end up with 2 revisions with `unlockInfo` and `undoUnlockIndo`, but no `submitInfo` (This could also produce `n` number of `undoUnlockInfo` revisions). Filtering it to just revisions with submissions fixes this. I do wonder what kind of side effects this would have for other operations likes withdraw and undo withdrawal. Likely none, but wold be worth testing those action sequences.
> 
> **Suggestion**: I also wonder if the alternative here is to filter OUT revisions with `undoUnlockInfo` as those are now only for record keeping and no longer revisions to be considered in the unlock resubmit flow. Filtering to only submitted revisions gives you only the submit history and not the unlocks too. If two unlocks somehow happen we want to make sure this revision will properly error here and not add a third unlock

I'm pushing up some new changes and to address some of your points:

1. Good call on withdraw, after looking at it, it looks like withdraw would experience the same bug after a couple of cycles which is now taken care of with this fix.

2. I added test cases to cover both to ensure this sort of bug doesn't happen again.

3.  We'd probably want to keep the current filter I applied for consistency's sake and its the safer approach since its direct and finds specifically what we need. If we try to filter out by `undoUnlockInfo` alone we run the risk of ending up in the same spot since there are other rows we would want to avoid. With the current filter we're filtering for rows that can give us the answer we need vs filtering out rows that can't

4. I believe we already have guards against a double unlock in `undoWithdrawRate.ts` so I think we're set in that regard

JasonLin0991 (approved):
> I'm pushing up some new changes and to address some of your points:
>
>Good call on withdraw, after looking at it, it looks like withdraw would experience the >same bug after a couple of cycles which is now taken care of with this fix.
>
>I added test cases to cover both to ensure this sort of bug doesn't happen again.
>
>We'd probably want to keep the current filter I applied for consistency's sake and its >the safer approach since its direct and finds specifically what we need. If we try to >filter out by undoUnlockInfo alone we run the risk of ending up in the same spot since >there are other rows we would want to avoid. With the current filter we're filtering for >rows that can give us the answer we need vs filtering out rows that can't
>
>I believe we already have guards against a double unlock in undoWithdrawRate.ts so I >think we're set in that regard

Thanks for the tests and explanation. Great work!

PR #4709, opened 2026-08-18
MCR-6344: state undo unlock email

JasonLin0991:
## Summary
When a submission is unlocked then undone state emails should be sent. There are 3 variations of the state email. Two for EQROs (one for when it's subject to review and one when its not) and then a third variant for everything else.

Some minor refactors made to make code reusable and styling more consistent. (Snapshots were updated because of format changes.)

#### Related issues
[MCR-6344](https://jiraent.cms.gov/browse/MCR-6344)

#### Screenshots
Screenshot 2026-08-18 at 11 30 22 AM
Screenshot 2026-08-18 at 11 34 45 AM
Screenshot 2026-08-18 at 11 42 12 AM

#### Test cases covered
  - After the third unlock: the child rate is still attached to the contract and gets a fresh draft revision from that same unlock event.
  - After the resubmit: the new package stamps a new child rate revision rather than reusing the one from the initial submission.

## QA guidance
Note: undo `cms-user-undo-unlock` and `eqro-submissions` flags should be turned on
- As a state user, create and submit a contract with at least one rate.
- As a CMS user unlock the submission and then undo the unlock. This should trigger the new state email

- Now as a State user submit an EQRO (reviewable)
- As a CMS user unlock the submission
- Login as an Admin and undo the unlock for the EQRO

- Now as a State user submit an EQRO (not subject to review)
- As a CMS user unlock the submission
- Login as an Admin and undo the unlock for the EQRO

## PR Reminders


vyakusha (approved):
Code changes LGTM. Tested all three variants, look and behave as expected.

PR #4715, opened 2026-08-19
MCR-6214: Adds an automated end-to-end test confirming that Aurora backups can successfully be restored & validated

Btodhunter:
- Runs monthly for dev, val, prod
- Restores Aurora to its latest restorable snapshot.
- Validates table counts, row counts, State data, and Prisma migrations using a VPC connected Lambda.
- Compares results with the previous successful run stored in SSM.
- Reports recovery-point age and restore/validation timing.
- Cleans up temporary database resources, including resources abandoned by interrupted runs.
- Sends Slack notifications on success and failure.


scripts/test-aurora-automated-backup-restore.ts
```diff
@@ -0,0 +1,729 @@
...
+
+        if (baseline) {
+            compareToBaseline(baseline, validatorBody.tableCounts)
+        } else {
+            console.info(
+                'No previous run to compare against; establishing baseline'
+            )
+        }
+
+        await writeBaseline(envName, validatorBody.tableCounts)
```
mojotalantikite: If `compareToBaseline()` is kept throwing (see comment above), then this `writeBaseline()` could potentially never get hit in that scenario --  every later run then fails against the stale value until someone manually deletes the SSM parameter. I think ultimately just warning in `compareToBaseline()` might be the best path, but if not then we'd want to move `writeBaseline()` up so it's reachable.

scripts/test-aurora-automated-backup-restore.ts
```diff
@@ -0,0 +1,729 @@
...
+                    `${tableName}: shrank from ${baselineRows} to ${currentRows}`,
+                ]
+            }
+
+            return []
+        }
+    )
+
+    if (problems.length > 0) {
+        throw new Error(
```
mojotalantikite: we might want this to just be a warning, as I think the numbers that we're getting particularly around drafts could trigger these comparisons that happen -- as in I don't want to assume that numbers are all going up within 10% thresholds like seem to be happening with the comparisons (if I'm understanding this right). The backup shrinking on some tables doesn't necessarily mean the backups are bad, and I think it's good to surface it, but I don't know if it should be a thrown error. Maybe just:  

```typescript
if (problems.length > 0) console.warn(`Restored backup tables shrank since the previous run: ...`)
```

scripts/test-aurora-automated-backup-restore.ts
```diff
@@ -0,0 +1,729 @@
...
+
+const AWS_REGION = 'us-east-1'
+const RESTORED_CLUSTER_PREFIX = 'mcr-backup-restore-'
+
+// A production point-in-time restore can take well over an hour. Timing out mid
+// restore leaves a running cluster behind, so allow generous headroom.
+const CLUSTER_AVAILABLE_TIMEOUT_SECONDS = 7200
+const INSTANCE_AVAILABLE_TIMEOUT_SECONDS = 3600
+const DELETE_TIMEOUT_SECONDS = 1800
+
```
mojotalantikite, line 68: So the action's timeout is set at 180m, but all this sums to more than that (210m plus the lambda timeout of 15m, plus the teardown). If the github action timeout hits first (I don't think it will, but it could) then it kills the process tree, node exits without running finally, so the cluster could survive until the next run. Maybe we need to fit all of this inside, say, 170m? Just making that number up.

mojotalantikite (approved):
Left a couple comments inline but LGTM overall, thanks for doing this!

PR #4723, opened 2026-08-21
MCR-6517: Update email for chip only submissions

ronaktruss:
## Summary
- Updates chip only submissions for undo-unlock (applies to both CMS and state emails)
- Generally updates how we determine status for undo unlock emails. Using actual consolidatedStatus instead of hard coding. 

#### Related issues
[MCR-6517](https://jiraent.cms.gov/browse/MCR-6517)

#### Screenshots
image

#### Test cases covered

Ensures CHIP only email has review decision info. 

## QA guidance

- Ensure CHIP only contracts have the review decision.
- Update status across all emails to use consolidatedStatus. Make sure this is a reasonable thing to do. (This status should match the status on the dashboard.)
- Validate existing emails don't have unintended changes.

## PR Reminders

### DIFF
--- a/services/app-api/src/emailer/templateHelpers.ts
+++ b/services/app-api/src/emailer/templateHelpers.ts
@@ -27,6 +27,7 @@ import type {
     UpdateInfoType,
 } from '../domain-models/contractAndRates'
 import type { ReviewActionTypes } from '../domain-models/contractAndRates/contractReviewActionType'
+import { ConsolidatedContractStatusRecord } from '@mc-review/constants'
 
 // ETA SETUP
 Eta.configure({
@@ -532,6 +533,7 @@ export type UndoUnlockContractEtaData = {
     reason: string
     status: string
     isEQRO: boolean
+    isCHIPOnly: boolean
     isNotSubjectToReview: boolean
     reviewDecisionText: string
     shouldIncludeRates: boolean
@@ -571,11 +573,14 @@ const parseEmailDataUndoUnlockContract = (
         config.baseUrl
     )
 
-    const isEQRO = contract.contractSubmissionType === `EQRO`
-
+    const isEQRO = contract.contractSubmissionType === 'EQRO'
+    const isCHIPOnly = formData.populationCovered === 'CHIP'
+    contract.reviewStatusActions
     const isNotSubjectToReview =
-        isEQRO &&
-        eqroValidationAndReviewDetermination(contract.id, formData) === false
+        (isEQRO &&
+            eqroValidationAndReviewDetermination(contract.id, formData) ===
+                false) ||
+        isCHIPOnly
 
     return {
         packageName: contractPackageName,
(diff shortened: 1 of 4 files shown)


services/app-api/src/emailer/templateHelpers.ts
```diff
@@ -571,11 +573,14 @@ const parseEmailDataUndoUnlockContract = (
         config.baseUrl
     )
 
-    const isEQRO = contract.contractSubmissionType === `EQRO`
-
+    const isEQRO = contract.contractSubmissionType === 'EQRO'
+    const isCHIPOnly = formData.populationCovered === 'CHIP'
+    contract.reviewStatusActions
```
ruizajtruss: Issue: no-op statement here

services/app-api/src/emailer/templateHelpers.ts
```diff
@@ -585,8 +590,9 @@ const parseEmailDataUndoUnlockContract = (
         ),
         updatedBy: updateInfo.updatedBy.email,
         reason: updateInfo.updatedReason,
-        status: isNotSubjectToReview ? `Not subject to review` : `Submitted`,
+        status: ConsolidatedContractStatusRecord[contract.consolidatedStatus],
```
ruizajtruss: Issue: In your PR description you mention that we want to match what's displayed in the dashboard. This new logic displays `Resubmitted` but the dashboard collapses `SUBMITTED` and `RESUBMITTED` to only display `Submitted`. I assume we want to collapse this in the emails in the same way.

services/app-api/src/emailer/templateHelpers.ts
```diff
@@ -571,11 +573,14 @@ const parseEmailDataUndoUnlockContract = (
         config.baseUrl
     )
 
-    const isEQRO = contract.contractSubmissionType === `EQRO`
-
+    const isEQRO = contract.contractSubmissionType === 'EQRO'
+    const isCHIPOnly = formData.populationCovered === 'CHIP'
```
ruizajtruss: The CHIP check isn't scoped to health plans, so an EQRO submission that's CHIP-only gets flagged "not subject to review" even when its own EQRO provisions say it is

```suggestion
    const isCHIPOnly = !isEQRO && formData.populationCovered === 'CHIP'
```

services/app-api/src/emailer/templateHelpers.ts
ronaktruss: Good catch. Updated.

services/app-api/src/emailer/templateHelpers.ts
```diff
@@ -585,8 +589,11 @@ const parseEmailDataUndoUnlockContract = (
         ),
         updatedBy: updateInfo.updatedBy.email,
         reason: updateInfo.updatedReason,
-        status: isNotSubjectToReview ? `Not subject to review` : `Submitted`,
+        status: isNotSubjectToReview
+            ? ConsolidatedContractStatusRecord['NOT_SUBJECT_TO_REVIEW']
+            : ConsolidatedContractStatusRecord['SUBMITTED'],
```
ronaktruss, line 594: There's an argument to just have the string values here, but it might be worth eventually centralizing these to this helper array.

ruizajtruss (approved):

PR #4728, opened 2026-08-25
Add graphql query to get state user by email

ronaktruss:
## Summary
- Adds graphql query to fetch state user based on an email (gets the latest record if there are multiple)
Note: emails aren't unique, but atm they only are duplicated within the same state.

#### Related issues
[MCR-6476](https://jiraent.cms.gov/browse/MCR-6476)

#### Screenshots
Screenshot 2026-08-24 at 6 58 17 PM

#### Test cases covered

- It only fetches state users
- it fetches the latest record if there are multiple matches

## QA guidance

Use graphql explorer to test this new query

## PR Reminders

### DIFF
--- a/services/app-api/src/postgres/user/findStateUser.ts
+++ b/services/app-api/src/postgres/user/findStateUser.ts
@@ -4,16 +4,26 @@ import { includeUsersWithBaseData } from '../contractAndRates/prismaUserHelpers'
 import type { ExtendedPrismaClient } from '../prismaClient'
 import { domainUserFromPrismaUser } from './prismaDomainUser'
 
-export async function findStateUserByEmail(
-    client: ExtendedPrismaClient,
+export type FindStateUserArgsType = {
     email: string
+    givenName: string
+    familyName: string
+}
+
+export async function findStateUser(
+    client: ExtendedPrismaClient,
+    args: FindStateUserArgsType
 ): Promise<StateUserType | Error | undefined> {
+    const { email, givenName, familyName } = args
     try {
-        // email is not unique on the User table
+        // email is not unique on the User table, neither is givenName or familyName
+        // Use all 3 pieces of information to reduce duplicate records from being found
         // If there are duplicates, prefer the newest account
         const result = await client.user.findFirst({
             where: {
-                email: email,
+                email: { equals: email, mode: 'insensitive' },
+                givenName: { equals: givenName, mode: 'insensitive' },
+                familyName: { equals: familyName, mode: 'insensitive' },
                 role: 'STATE_USER',
             },
             include: includeUsersWithBaseData,
--- a/services/app-api/src/postgres/postgresStore.ts
+++ b/services/app-api/src/postgres/postgresStore.ts
@@ -30,10 +30,14 @@ import type {
     SubmissionHistory,
 } from '../domain-models'
 import { findPrograms, findStatePrograms } from './'
-import type { InsertUserArgsType, UpdateUserInfoArgsType } from './user'
+import type {
+    InsertUserArgsType,
+    UpdateUserInfoArgsType,
+    FindStateUserArgsType,
+} from './user'
 import {
     findUser,
-    findStateUserByEmail,
+    findStateUser,
     insertUser,
     updateCmsUserProperties,
     updateUserInfo,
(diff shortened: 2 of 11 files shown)


services/app-api/src/postgres/user/findStateUser.ts
```diff
@@ -0,0 +1,40 @@
...
+import { domainUserFromPrismaUser } from './prismaDomainUser'
+
+export async function findStateUserByEmail(
+    client: ExtendedPrismaClient,
+    email: string
+): Promise<StateUserType | Error | undefined> {
+    try {
+        // email is not unique on the User table
+        // If there are duplicates, prefer the newest account
+        const result = await client.user.findFirst({
```
JasonLin0991, line 22: **Suggestion**: I took a peek into VAL, the env that SDP will be working in, there seems to be a few dozen emails with the same email addresses. 2 large groups of 11 and 14 each sharing 1 emails within the group. 

Most of this came from our Prod data dump + sanitization into the VAL env, but the 11 and 14 are state portal test users that share the same team email. I think Dev likely has the same duplicate emails for test users. These test users have different givenName, familyName, and states so you could legitimately find the wrong state for a user because the emails are not unique enough.

Prod likely has duplicates, but to a smaller extent and it would likely be fine with the code above. But for SDP testing this code will cause issues in testing.

My suggestion is to also require givenName and familyName when querying for the state user. We likely also want to normalize the names before we try to match.

ronaktruss: >But for SDP testing this code will cause issues in testing.

Can you elaborate? Are they unable to create new users? 

Since there is a known and expected functionality from our side, shouldn't they adjust testing accordingly?

JasonLin0991: They can create new users, but if they decide to reuse emails than we likely run into issues getting the correct state for the user. This isn't only a SDP issue, our Dev and Val accounts used for our team also share the same email, so many state users point to the same email address. I would like to account for this for both our testing and SDP app.

ronaktruss: Ultimately we can be flexible, but if they just want to filter by emails we can't guarantee uniqueness.

I don't know what information they already have. If they do already have givenName and familyName, that still doesn't solve the problem as these can also be duplicated, but reduces the likelihood of matches. This likely won't helps as much in prod.

ronaktruss: To me the work around would be for them to create a new user with the same email in the state that they want, they get the latest, so they will have control over state of the user. Or they could create a unique email using the + modifier, or they could just do the testing in a different state.

To me this solution doesn't inherently solve the root cause and is just a workaround. Atm it's not clear that adding this functionality actually makes things easier. As a general note, I'd prefer for workarounds to exist on the consumer side, rather than at the API side where making changes is more difficult. (For clarity, I'm open to making this change, I just don't have enough info to justify it.)

JasonLin0991: > I don't know what information they already have

They will have first names, last names, emails, and other data that wont help identify the user in our db. All user data comes from IDM, same with our app. The only thing SDP does not get is the state assigned to the user. So the only purpose of this endpoint is to get the users State

> If they do already have givenName and familyName, that still doesn't solve the problem as these can also be duplicated, but reduces the likelihood of incorrect matches.

My reason to suggest include 3 criteria (givenName, familyName, email) is to:
- Target the same user, regardless of duplicate contacts. But if the names and emails are the same then we are likely working with the same actual person but somehow ended up with 2 IDM accounts (maybe 1 was deprecated in the IDM system)
- Find the State associated with this user. We don't care if there a 100 of the same user with the same data as long as we get the state from the latest created one. There is still a issue of same user data and only different or no state, but that is something that has to be handled outside of this work.

> This likely won't helps as much in prod.

I think there are no occurrences of same email with different names in prod, so yes it does not help in prod, but in the lower environments we have users with the same email but different names and state (this came from IDM accounts). So we should try to handle this even though it does not happen in prod, it will in lower environments.

Let me know if you want to chat about this.

ronaktruss: Updated in https://github.com/Enterprise-CMCS/managed-care-review/pull/4728/changes/4579f7657d29aa4e833c4ac9502f02065406e312. 

Also makes email, given, and family name all case insensitive.

Also updates postgres function name to not include by email since that's no longer accurate.

ronaktruss: image

JasonLin0991 (approved):

PR #4741, opened 2026-08-28
MCR-6459: create migration lambda to parse and migrate contacts

JasonLin0991:
## Summary
[MCR-6459](https://jiraent.cms.gov/browse/MCR-6459)

This work adds a lambda to migrate the easy to parse State and Actuary contacts for submitted revisions only (revisions with a submitInfo). It will ignore submissions without the the submitInfo.

There is a manual step for manually migrating the more complex contacts. That can be done before or after apply the lambda migration.

Lambda migrator:
- The lambda migrator has a `"dryRun": true` option that will only log out results and not apply the migration.
   - Both `dryRun` true or false will log out the results of the migration.
- The lambda migrator will log our State and Actuary contacts that would not be parsed and migrated using the lambda. These are the records that will need to have the fill name manually split into distinct name fields and migrated.
   - The manual migration process is to run SQL code on the DB instances.
- The lambda code will have a lot of inline and JSDoc comments to describe what exactly everything is doing.
- **IMPORTANT**: This work is a best attempt at migrating the full name to the distinct name fields. The deprecated `name` field will still retain the user inputted full name incase we need to go back and fix anything.

## QA guidance
- Validate the migration runs in Review APP
   - Log into AWS console
   - Navigate to `Lambda -> Functions -> app-api-jlmcr6459createmig16a65-cdk-migrate-contacts -> Test tab`
   - Input any event name
   - Event JSON can be with dryRun true or false
      - `{ "entity": "both", "dryRun": true }`
   - Hit `Test`
   - Expand the details section on `Executing function` banner to see the log out.
      Screenshot 2026-08-28 at 4 29 05 PM.
   - Validate the migration ran and review the output log records for manual migrations.
   - You do not have to perform the manual migration step for this review. The data migration will be performed in other tickets.

## PR Reminders

### DIFF
--- a/services/app-api/src/handlers/migrate_contacts.ts
+++ b/services/app-api/src/handlers/migrate_contacts.ts
@@ -1558,7 +1558,7 @@ export const main: Handler = async (
         dryRun,
     })
     // The response contains only contact ids and counts, never names or emails.
-    console.info('Contacts migration complete', response)
+    console.info('Contacts migration complete', JSON.stringify(response))
 
     return response
 }
--- a/services/app-api/src/handlers/migrate_contacts.ts
+++ b/services/app-api/src/handlers/migrate_contacts.ts
@@ -28,7 +28,6 @@ import {
 const NO_GIVEN_NAME = 'NO_GIVEN_NAME'
 const NO_FAMILY_NAME = 'NO_FAMILY_NAME'
 const NO_TITLE_ROLE = 'NO_TITLE_ROLE'
-const NO_EMAIL = 'no-email@example.com'
 
 // How many updates run at once. The connection pool holds 10 (a node-postgres
 // pool, built in prismaClient.ts), so 10 is the most that can be in flight and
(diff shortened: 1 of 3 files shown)


services/app-api/src/handlers/migrate_contacts.ts
```diff
@@ -0,0 +1,1564 @@
...
+ * contacts-manual-migration-runbook.md at the repository root.
+ */
+
+// Placeholders written when a name part genuinely does not exist in the source
+// data. These are deliberately a single shared constant rather than a numbered
+// series -- see the comment on countPlaceholders() for why that matters.
+const NO_GIVEN_NAME = 'NO_GIVEN_NAME'
+const NO_FAMILY_NAME = 'NO_FAMILY_NAME'
+const NO_TITLE_ROLE = 'NO_TITLE_ROLE'
+const NO_EMAIL = 'no-email@example.com'
```
mojotalantikite: Do we want to touch email and title role during the migration? This could get passed in during an unlock, and then if no one saw it and re-submitted, we'd get email bounce errors trying to send to this address. I'd say we'd maybe want to leave these fields alone.

mojotalantikite (approved):
Just raised a question around the email replacement when not found, but otherwise LGTM

services/app-api/src/handlers/migrate_contacts.ts
JasonLin0991: Removed backfilling empty emails. It now leaves it that field alone.

PR #4742, opened 2026-08-28
MCR-6465: Transition QA to its own GitHub environment and AWS account

mojotalantikite:
## Summary
QA previously shared Val's AWS account and CI credentials during OIDC bootstrap. Now QA has its own GitHub environment (`qa`) and deploys using `QA_AWS_ACCOUNT_ID` instead of `VAL_AWS_ACCOUNT_ID`.

This also extracts account-wide API Gateway CloudWatch role management into a separate `api-gateway-account` CDK stack. Dev/Val/Prod deploy this baseline once per account; QA uses Val's baseline since it shares that account.

Also cleans up some docs.

#### Related issues
https://jiraent.cms.gov/browse/MCR-6465


ruizajtruss (approved):
LGTM

PR #4760, opened 2026-09-03
Refactor Cypress feature flag command and remove deprecated code

JasonLin0991:
## Summary

This work refactors how the Cypress launch darkly commands store and set feature flags for tests.

- Before:
   - cy.stubFeatureFlags(), sets up the test file to intercept actual feature flag calls from the app and return
      - 1. flags from our flag store
      - 2. apply Cypress test defaults
   - cy.interceptFeatureFlags(), used in the tests to update the feature flags for the specific test
      - This would clear our any defaults for Cypress set in  stubFeatureFlags.
- After:
   - cy.stubFeatureFlags(), sets up the test file to intercept actual feature flag calls from the app and return
      - 1. flags from our flag store
      - 2. apply Cypress test defaults
   - cy.interceptFeatureFlags(), used in the tests to update the feature flags for the specific test
      - [NEW]: Now only updates the flag passed into the argument for command to update, all others remain whatever stubFeatureFlags had set for defaults.
    
The previous behavior of the commands blocked turning on the `contact-model-update` flag in dev. The API uses real feature flags in dev, but Cypress can toggle them dynamically in tests and cannot sync them with the API. This causes API validation issues when the flag is on, but Cypress runs a test with `contact-model-update` off. This in turn causes any PR to fail Cypress tests in CI because of API validations on distinct contact name fields.

So we would have to turn this flag on for all tests. This is when I found the issue with the previous behavior clearing out default Cypress flag values and not retaining default values if they were not passing into interceptFeatureFlags arguments.

It would be good to think about a way to sync the Cypress test flags with the API during these CI runs in the review apps. Out of scope for this PR though.

Other work:
- Fixed a small flaky test on the rate reviews dashboard page
   - Before it was looking for `rate reviews` text on the page, this correlates with the filter text, but what we really want to test is if we are looking at the `Rate reviews` tab.
   - Updated the assertion to look for a H2 element with `Rate reviews` text.
- Removed all `438-attestation` Cypress testing code. This was a field that was implemented but never completed and is no longer required. Future work will remove this field from the codebase

## QA guidance

- Cypress tests pass in CI.

### DIFF
--- a/docs/technical-design/launch-darkly-testing-approach.md
+++ b/docs/technical-design/launch-darkly-testing-approach.md
@@ -116,7 +116,7 @@ renderWithProviders(
         },
         featureFlags: {
             'rate-edit-unlock': false,
-            '438-attestation': true,
+            dsnp: true,
         },
     }
 )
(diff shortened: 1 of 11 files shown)


vyakusha (approved):
LGTM.

PR #4774, opened 2026-09-09
MCR-3145: Upgrade yup to v1 and update form validators for compatibility

mojotalantikite:
## Summary

Update yup dependency from v0.32.9 to v1.7.1 in app-web and refactor form validators to work with the new API. The main changes include:

- Adding explicit type parameters to `Yup.mixed()` calls
- Updating `.when()` method signatures to use array destructuring for field values
- Converting `.then` conditions from inline schemas to callback functions
- Adding null coalescing operators for optional values
- Refactoring conditional logic to use ternary operators

Also update pnpm lock file with resolved dependency versions.

#### Related issues

https://jiraent.cms.gov/browse/MCR-3145


services/app-web/src/pages/StateSubmission/HealthPlanSubmission/RateDetails/RateDetailsSchema.ts
JasonLin0991, line 165: **Issue**: yup 1.7.1 loosened email checking regex. This now allows emails like `foo@bar` to be accepted in the form validation. Our API does a second email validation using Zod so it would fail validation on `foo@bar`, but we want to surface the error in the error summary before that.

To do this I think we need to add our own strict email validation regex for Yup since I couldn't find options for yup.email(). Something like `/^(?!\.)(?!.*\.\.)([A-Za-z0-9_'+\-\.]*)[A-Za-z0-9_+-]@([A-Za-z0-9][A-Za-z0-9\-]*\.)+[A-Za-z]{2,}$/` used in yup match() would work. Something like

```ts
const EMAIL_REGEX =                                                                                                                                                                                                               
      /^(?!\.)(?!.*\.\.)([A-Za-z0-9_'+\-\.]*)[A-Za-z0-9_+-]@([A-Za-z0-9][A-Za-z0-9\-]*\.)+[A-Za-z]{2,}$/
...
email: Yup.string()                                                                                                                                                                                                       
    .matches(EMAIL_REGEX, {                                                                                                                                                                                               
        message: 'You must enter a valid email address',                                                                                                                                                                  
        excludeEmptyString: true,                                                                                                                                                                                         
    })                                                                                                                                                                                                                    
    .required('You must provide an email address')
```

JasonLin0991: Image

JasonLin0991, line 169: **Nit**: This could be an issue later, but essentially with 1.7.1 this allows null to become accepted for inputs when it shouldn't.

In .032.9 required() rejects null and could not be overridden when `nullable()` is chained after. In 1.7.1 that changed so `nullable()` now does override the previous required().

Right now, nothing has a null value so it isn't reachable because initial values are empty string or `undefined`.

I think here the way to go is removing `nullable()` from the chain, also doing the same in `addtlActuaryContacts` field schema.  Like so:

```ts
actuarialFirm: Yup.string().required('You must select an actuarial firm'),                                                                                                                                                        
actuarialFirmOther: Yup.string()                                                                                                                                                                                                  
    .nullable()                                                                                                                                                                                                                   
    .when('actuarialFirm', {                                                                                                                                                                                                      
        is: 'OTHER',                                                                                                                                                                                                              
        then: (schema) => schema.required('You must enter a description'),                                                                                                                                                        
    })
```

JasonLin0991 (changes requested):
Looks mostly good.

The only real issue is regression of email validation in the UI.

The other one could be a potential issue in the future,

services/app-web/src/pages/StateSubmission/HealthPlanSubmission/RateDetails/RateDetailsSchema.ts
mojotalantikite: Cool, I added an email regex to check against here.

mojotalantikite: Good catch, I removed the `.nullable()`  from the required `actuarialFirm` field for both `actuaryContacts` and `addtlActuaryContacts`. Added some regression coverage as well.

JasonLin0991 (approved):
Looks good. Thanks!

PR #4777, opened 2026-09-11
[ MCR-6527] Add revision diff data to Submission Summary page

ruizajtruss:
## Summary
Adds and surfaces revision diff data to the summary pages via a new nullable `revisionDiff` field on `Contract`, resolved only when requested; one endpoint instead of two queries.

**AC:**
- SubmissionSummary and EQROSubmissionSummary page has revision diff data for the current submission and latest past submission.
- Either by making 2 queries `fetchContract` and `fetchRevisionDiff` or include the revision diff in the `fetchContract` payload via a field resolver.
  - This PR takes the field resolver route

**Main Changes:**
- **`Contract.revisionDiff` field resolver**: compares the latest two submissions. Returns `null` for single-submission contracts or on diff errors (logged), so the diff can never break the summary page
- **New `fetchContractRevisionDiff` query** — fired by SubmissionSummary/EQROSubmissionSummary only for resubmissions (skip-gated). Kept out of the shared `fetchContractWithQuestions` query so its ~11 other consumers pay nothing. Result is intentionally unconsumed.
-  **Contact diffing now distinguishes New vs. Updated** — `NEW_OR_MODIFIED` replaced with `ADDED | UPDATED` (+ `index`) so we can easier identify what contacts were newly added vs updated. Manually confirmed diff emails remain unaffected by these changes.

> [!NOTE]
> Since I implemented the field resolver approach, `fetchRevisionDiff`  is essentially dead code. We should decide whether to keep or delete.

> [!NOTE]
> This PR lays the ground work for [MCR-6525](https://jiraent.cms.gov/browse/MCR-6525)

#### Related issues
[MCR-6527](https://jiraent.cms.gov/browse/MCR-6527)
#### Screenshots
image

## QA guidance
- Access an unlocked submission
- Update fields
- visit the submission summary page
- As a CMS user, check your network tab for the revision diff data

## PR Reminders


services/app-web/src/pages/SubmissionSummary/EQROSubmissionSummary/EQROSubmissionSummary.tsx
```diff
@@ -85,6 +86,17 @@ export const EQROSubmissionSummary = (): React.ReactElement => {
...
     const contract = data?.fetchContract.contract
+
+    useQuery(FetchContractRevisionDiffDocument, {
+        variables: {
+            input: {
+                contractID: id ?? 'unknown-contract',
+            },
+        },
+        // Only resubmissions have a previous submission to diff against
+        skip: (contract?.packageSubmissions.length ?? 0) < 2,
```
vyakusha, line 97: I like it.

services/app-api/src/postgres/revisionDiff/revisionDiffPrimitives.ts
```diff
@@ -150,36 +150,70 @@ function diffCollectionByKey<TItem, TChange>({
     return changes
 }
 
-function buildNewAndModifiedCollectionChanges<TItem>(
+function buildContactCollectionChanges<TItem>(
```
vyakusha, line 153: I very much like the way the new code compares previous and new state contact values, using name and email as 'anchors'. If at least one of them matches one of the previous records, we are marking the record as "UPDATED". While technically not strict, it gives user the logic which covers their business needs. A couple of naming "nits" is suggested to make this code more readable.

services/app-api/src/postgres/revisionDiff/revisionDiffPrimitives.ts
```diff
@@ -150,36 +150,70 @@ function diffCollectionByKey<TItem, TChange>({
...
-            (previousRemainingCounts.get(key) ?? 0) + 1
-        )
+        previousCounts.set(key, (previousCounts.get(key) ?? 0) + 1)
     }
 
-    const changes: RevisionDiffCollectionItemNewOrModified<TItem>[] = []
+    const leftoverCurrent: { item: TItem; index: number }[] = []
+    current.forEach((item, index) => {
+        const key = getComparisonKey(item)
+        const remaining = previousCounts.get(key) ?? 0
```
vyakusha, line 171: nit: could use unmatchedPreviousCount vs remaining

services/app-api/src/postgres/revisionDiff/revisionDiffPrimitives.ts
```diff
@@ -150,36 +150,70 @@ function diffCollectionByKey<TItem, TChange>({
...
 
-    for (const item of current) {
+        if (remaining > 0) {
+            previousCounts.set(key, remaining - 1)
+        } else {
+            leftoverCurrent.push({ item, index })
```
vyakusha: nit: could use unmatchedCurrentContacts vs leftoverCurrent

services/app-api/src/postgres/revisionDiff/revisionDiffPrimitives.ts
```diff
@@ -150,36 +150,70 @@ function diffCollectionByKey<TItem, TChange>({
...
+        }
+    })
+
+    const leftoverPrevious: TItem[] = []
```
vyakusha: nit: could use unmatchedPreviousContacts vs leftoverPrevious

services/app-graphql/src/queries/fetchRevisionDiff.graphql
```diff
@@ -2,105 +2,7 @@ query FetchRevisionDiff($input: FetchRevisionDiffInput!) {
...
-            current {
-              name
-              titleRole
-              email
-              actuarialFirm
-            }
-          }
-        }
-      }
+      ...revisionDiffFragment
```
vyakusha, line 5: IMHO I would not remove fetchRevisionDiff, in case we would need to call it without bringing up the whole Contract data. @JasonLin0991 What do you think?

services/app-api/src/postgres/revisionDiff/revisionDiffPrimitives.ts
ruizajtruss: Done!

services/app-api/src/postgres/revisionDiff/revisionDiffPrimitives.ts
ruizajtruss: Done!

services/app-graphql/src/queries/fetchRevisionDiff.graphql
ruizajtruss: I'm okay with leaving it personally

vyakusha: I talked with @JasonLin0991 and he is OK with leaving it as is for now.

vyakusha (approved):
All code LGTM, tested locally and found no issues. Great job!

PR #4798, opened 2026-09-17
[MCR-6467] zip generation failure

ruizajtruss:
## Summary
Fixes prod zip generation failures for submissions whose documents span the legacy uploads bucket and the current CDK bucket. generateDocumentZip used the first document's s3BucketName for every download in the zip, so revisions mixing legacy-URL documents with newer uploads failed the whole zip. No documents were actually missing; they were being requested from the wrong bucket.

**AC:**
- Identify and fix the root cause
- Submissions don't have missing zip files

> [!NOTE]
> After this deploys, the affected revisions get their zips back via the existing regenerate-zips lambda

**Main Changes:**
- Per-document download bucket: each document is now downloaded from its own s3BucketName instead of documents[0]'s. A revision can legitimately reference both buckets, so one shared bucket 404s on some files. Validation errors now name the offending document.
- Zip uploads always target the configured documents bucket (VITE_APP_S3_DOCUMENTS_BUCKET): Previously the zip landed in whatever bucket the first document pointed at. Missing env var is now a hard error, checked before any downloads.
- Tests: mixed-bucket downloads hit each document's own bucket, uploads go to the configured bucket, and both missing-s3BucketName and missing-env-var cases return errors.

#### Related issues
[MCR-6467](https://jiraent.cms.gov/browse/MCR-6467)

## PR Reminders


mojotalantikite (commented):
Thanks for taking this bug on! So you're right, it looks like the errors are from mixed buckets on these revisions. I had done a major refactor and migration (across PRs #3911, #4012, #4051, #4139) earlier this year, but I missed some of the documents for a few reasons:

* Assumed the bucket being non-null in the new column meant we successfully migrate
* Didn't block writes to the s3URL field so the migration could be undone so to speak
* There was a gap between migration and follow on work

This PR as it is should work to get zip generating in those edge cases where they still span two buckets (serverless and CDK), but I think we're going to need a more comprehensive fix to the migration, which I've started sketching out in another branch. Essentially we'll have to re-run the migration to fix the stale DB records, copy files that are persisted only on the legacy bucket, and then prevent the old bucket from being written to. Zip generation can then consistently use the configured bucket instead of permanently supporting both old and new buckets and at that point I'd like to remove the migrate lambda and make sure the old bucket is archived.

PR #4803, opened 2026-09-18
MCR-6467: Refactor S3 document migration to use canonical bucket sources

mojotalantikite:
## Summary

Juan was working on a bug ticket (#4798) where some zip files had not been properly generated, where he noticed that in these cases files were being referenced incorrectly across the old bucket (serverless) and the new one (cdk). We had run a migration earlier in this year (across PRs https://github.com/Enterprise-CMCS/managed-care-review/pull/3911, https://github.com/Enterprise-CMCS/managed-care-review/pull/4012, https://github.com/Enterprise-CMCS/managed-care-review/pull/4051, https://github.com/Enterprise-CMCS/managed-care-review/pull/4139), but it seems like there were some cases where this didn't work.

It looks like the original migration left the `s3URL` column stale and skipped records with a non-null but incorrect bucket, allowing frontend roundtrips to restore legacy references.

 This PR:

 - Canonicalizes document bucket metadata on every write.
 - Copies legacy-only objects into the configured CDK bucket before updating database pointers.
 - Makes zip generation use the configured bucket.
 - Adds regression coverage for partial migrations, roundtrips, and copy failures.

After the migration runs successfully, we can remove the one-time Lambda and legacy bucket permissions, as well as make sure we're not ever referencing `s3URL` anywhere, in a follow on.

#### Related issues
https://jiraent.cms.gov/browse/MCR-6467


services/app-api/src/zip/generateZip.ts
```diff
@@ -133,7 +132,7 @@ export const generateDocumentZip: GenerateDocumentZipFunctionType = async (
                     mergedOptions.timeoutPerMB * 1000
                 const downloadResult = await downloadFile(
                     s3Client,
-                    bucket,
+                    documentsBucket,
                     docInfo.key,
```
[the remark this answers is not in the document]
mojotalantikite: This is intentional. The configured CDK bucket is now the canonical source, and we do not want to preserve mixed-bucket download behavior. Most documents have already been migrated anyway. 

I think we can accept that an affected zip may fail during that short interval rather than retaining a permanent legacy-bucket fallback, which is happening on a few submissions right now already.

services/app-api/src/handlers/migrate_s3_urls.ts
```diff
@@ -348,36 +366,272 @@ export const main: Handler = async (
...
+        return false
+    }
+    const s3Error = error as {
+        name?: string
+        $metadata?: { httpStatusCode?: number }
+    }
+    return (
+        s3Error.name === 'NotFound' ||
+        s3Error.name === 'NoSuchKey' ||
+        s3Error.$metadata?.httpStatusCode === 404
```
ruizajtruss, line 534: **Question:** Would we have to include a branch here to account for 403? If the newly copied object in the new bucket lacks a tag for virus scanning for a period of time in which this check takes place then it would return a 403 instead of a 404 due to the `StringNotEquals` check because a non-existent tag would mean it doesn't equal to the expected tag its checking for in `uploads.ts`  (at least that's what I understand from the docs).

ruizajtruss (approved):
Looks good! Just one question in regards to a potential bug

services/app-api/src/handlers/migrate_s3_urls.ts
mojotalantikite: Hmm, good question. I don't think we want to treat every `403` as not found, as it could genuinely be a block from intentional permissions (e.g. it's actually a virus). I think the move is to switch the migration’s existence check to `GetObjectTagging`, which isn’t covered by that deny policy, while continuing to treat only `404/NoSuchKey` as absent.

